PR #4, opened 2016-04-25
Implement FOV camera model

puzzlepaint:
Implements the FOV camera model from Frederic Devernay, Olivier Faugeras. Straight lines have to be straight: Automatic calibration and removal of distortion from scenes of structured environments. Machine vision and applications, 2001.

Potential further work:
* Determine the threshold for small numbers in a sensible way.
* The term ```T(2) * ceres::tan(omega / T(2))``` appearing in the distortion and undistortion could theoretically be cached in case ```omega``` is constant.


ahojnnes:
Looks great! Thanks, Thomas.

PR #19, opened 2016-07-18
Explicit namespace for std::isfinite(x)

omigeot:
Should fix compile problems on Ubuntu Xenial.


ahojnnes:
Thanks very much for the fix. Before merging your fix, could you share on which platform this causes problems?

Thanks!

ahojnnes:
Okay, I should have read the commit message in full.

ahojnnes:
@omigeot Actually, now Clang/MSVC was complaining about a missing include. See my follow-up commit https://github.com/colmap/colmap/commit/a2b5644e87d23f184435af723eec4aa1960b3991

Thanks again for the fix!

PR #21, opened 2016-08-09
Enable BOOST in shared mode (optional).

cbalint13:
This little patch enables to build [COLMAP](https://colmap.github.io/) with shared [**Boost**](http://boost.org) library too without affecting current behaviour. Vast majority of distros ship shared boost library only, perhaps in future this shared mode **may** be considered as default.

 * To invoke this mode one can run:

  ```cmake . -DBOOST_STATIC=OFF```

 * Please notice the mandatory extra [-DBOOST_TEST_DYN_LINK](http://stackoverflow.com/questions/7781374/boost-test-linking) flag in this shared mode.


ahojnnes:
@cbalint13 Thanks very much for the patch!

PR #37, opened 2016-10-04
MSVC fixes

bjornpiltz:


src/base/reconstruction.cc
```diff
@@ -567,7 +567,6 @@ std::vector<image_t> Reconstruction::FilterImages(
     const class Image& image = Image(image_id);
     const class Camera& camera = Camera(image.CameraId());
     if (image.NumPoints3D() == 0) {
-      DeRegisterImage(image_id);
```
ahojnnes: Why is this?

src/ext/FLANN/algorithms/kmeans_index.h
```diff
@@ -373,7 +373,7 @@ class KMeansIndex : public NNIndex<Distance>
         /**
          * The cluster center.
          */
-        DistanceType* pivot;
+        DistanceType* pivot=nullptr;
```
ahojnnes: One of the many uninitialized variables in FLANN causing problems :-)

src/mvs/CMakeLists.txt
```diff
@@ -20,6 +20,10 @@ if(CUDA_ENABLED)
         # Workaround for Visual Studio compiler error,
         # due problems with compiling too complex code.
         string(REPLACE "/Zo" "" CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS}")
+
+        # Avoid pulling in too many header files through <windows.h>
+        add_definitions(-DWIN32_LEAN_AND_MEAN)
```
ahojnnes: Under which MSVC version is this a problem? Just out of curiosity.

ahojnnes:
Thanks very much for the fixes!

src/mvs/CMakeLists.txt
bjornpiltz: I have only got CUDA to work with MSVC 2013, so it only shows up there. It probably also depends on which Win SDK you have installed.

PR #61, opened 2017-01-08
added an option to build cmake without building the tests

cebe:
building colmap for just using it does not need the test binaries,
so here is an option to turn them off.

Not sure exactly about the default value, maybe off is better than on?


ahojnnes:
This is a good option, thanks for the contribution!

cebe:
You're welcome!

PR #127, opened 2017-05-28
Snap for Colmap

pablospe:
Currently working version (even with GPU). Compilation details:
- Ubuntu 16.04.2 (almost vanilla ubuntu iso, in a virtualbox)
- snap 2.24
- snapcraft 2.29

Some steps are still needed in order to properly publish it to the snap store:
[https://snapcraft.io/docs/build-snaps/publish]( https://snapcraft.io/docs/build-snaps/publish) 

Explanation of snap, from Mark Shuttleworth:
[Why we need a different container purely for apps (youtube)](https://www.youtube.com/watch?v=0z3yusiCOCk)


snap/snapcraft.yaml
```diff
@@ -0,0 +1,204 @@
...
+      - libqt5opengl5-dev
+      #- nvidia-cuda-dev       # for CUDA support
+      #- nvidia-cuda-toolkit
+    plugin: cmake
+    build: |
+      mkdir build
+      cd build
+      cmake -DCMAKE_BUILD_TYPE=Release \
+        -DCMAKE_INSTALL_PREFIX=$SNAPCRAFT_PART_INSTALL \
+        -DCeres_DIR=$SNAPCRAFT_STAGE/lib/cmake/Ceres/ \
```
ahojnnes: I guess tests are not need, so you could add a ``-DTESTS_ENABLED=OFF``

snap/snapcraft.yaml
```diff
@@ -0,0 +1,204 @@
...
+  rig-bundle-adjuster:
+    command: rig_bundle_adjuster
+    plugs: [home]
+  vocab-tree-builder:
+    command: vocab_tree_builder
+    plugs: [home]
+
+parts:
+  eigen:
+    source: http://bitbucket.org/eigen/eigen/get/3.2.10.tar.bz2
```
ahojnnes: Can we upgrade to the latest 3.3.3?

snap/snapcraft.yaml
```diff
@@ -0,0 +1,204 @@
...
+
+#grade: devel
+#confinement: devmode
+
+grade: stable
+confinement: strict
+
+#icon: project/src/images/icon.png
+
+apps:
```
ahojnnes: It seems like not all binaries are included here. One missing file would be the ``automatic_reconstructor``.

snap/gui/colmap.desktop
```diff
@@ -0,0 +1,7 @@
+[Desktop Entry]
+Type=Application
+Name=Colmap
```
ahojnnes: COLMAP

snap/snapcraft.yaml
ahojnnes: You should also specify ``-DCMAKE_BUILD_TYPE=Release``

pablospe:
Automatic buildings can be done using Travis CI, that would be awesome. Anyway it should not be too much to compile only the stable versions. Here the documentation:
   https://snapcraft.io/docs/build-snaps/ci-integration

In order to specific a branch, you need to change "**source-tag: '3.0'**" to "**source-branch: _<branch_name>_**". But I think it make senses to leave the _source-tag_ in the _snapcraft.yaml_ configuration file, and to figure it out how to configure in the Travis CI side which branch should be used.

Btw, you will need to register the colmap name in the snap store, so it is reserved by you.

pablospe:
Regarding the non-cuda version I would suggest to have to branches: snap and snap-non-cuda (or snap-cpu), and non-cuda branch only having the following two changes:

1. the name of the package "colmap-non-cuda" (or colmap-cpu);
2. -DCUDA_ENABLE=OFF in the cmake of the colmap parts.

In this way, you just update one branch (the 'snap' branch) and then for the non-cuda branch it is just a fast-forward merge.

pablospe:
Forgot to say, once everything is done and working properly I would try to modify the colmap documentation in order to explain how to install it using snap. And even add a short note that snapcraft could be used for easy compilations with the command: "snapcraft prime", which greatly simplifies the compilation process.

In addition, I think it would be interesting to add in the compilation: gflags, google-glog and suitesparse, in the 'snapcraft.yaml' file in order to have control over the exact versions which colmap is compiled against (probably you will have less random bugs reported as issues, if you have a more controlled environment). But this complicates a bit the compilation. Anyway, it is almost done in the end of the 'snapcraft.yaml' file you have it but it has been commented out (needs more testing).

ahojnnes:
@pablospe Thanks, with the new ``scripts/python/build.py`` it should be much easier to compile COLMAP and all its dependencies. Since snap does not support CUDA, I will close this for now. Thanks in any case :-)

PR #138, opened 2017-06-06
Feature/Serialization

fedeDev:
This PR implements serialization of several of Colmap's data types:
- Point2D
- Point3D
- TrackElement
- Track
- Camera
- Image
- VisibilityPyramid
- SceneGraph::Correspondence
- SceneGraph::Image
- SceneGraph
- Reconstruction

They're all unit tested.

The backend itself, [YAS](https://github.com/niXman/yas), is also unit tested and I wrote a wrapper to make those unit tests available to Colmap's framework. YAS is aimed to be as fast as possible for i/o (as you can see on its GitHub), but it is quite inefficient regarding memory consumption (a binarized Colmap reconstruction is about 1.5x larger than a text-based one). The serialization is implemented such that we can easily swap YAS for something else such as [Cereal](https://github.com/USCiLab/cereal) or Boost Serialize. Take a look at Cereal, I can easily swap the backend to that one if you feel its better.

I also include two executables (textify and binarize models) that work on reconstructions.


src/base/reconstruction.cc
```diff
@@ -767,9 +767,9 @@ void Reconstruction::WriteCameras(const std::string& path) const {
   std::ofstream file(path, std::ios::trunc);
   CHECK(file.is_open()) << path;
 
-  file << "# Camera list with one line of data per camera:" << std::endl;
-  file << "#   CAMERA_ID, MODEL, WIDTH, HEIGHT, PARAMS[]" << std::endl;
-  file << "# Number of cameras: " << cameras_.size() << std::endl;
+  file << "# Camera list with one line of data per camera:\n";
```
fedeDev: @ahojnnes I swapped most of ```std::endl```s for ```\n``` since ```endl``` does a flush, which can be an expensive system call. Instead I do a flush at the end of the method.

fedeDev:
FYI, YAS is included into Colmap as a ```git subtree```.

src/base/reconstruction.cc
ahojnnes: Great, didn't know that. Thanks.

ahojnnes:
Thanks Fede, I didn't want to add another huge dependency for serialization and I also compared the performance of YAS vs. my own binary file reader/writer. COLMAP now uses binary files by default for reading/writing. It should be fully backwards compatible and you can convert between the two formats using the already existing ``model_converter`` executable. Documentation about this can be found in https://colmap.github.io/format.html

Thanks again and sorry for not merging this.

ahojnnes:
Ah, the code is only in the dev branch for now and will be merged sometime soon into master.

fedeDev:
@ahojnnes awesome. All that matters is that I can load/write reconstruction faster now :)

PR #140, opened 2017-06-06
Fix cuda version

fedeDev:
The bug: If having CUDA < 7.0, Colmap's build files will try to build CUDA-enabled libriaries but Colmap uses std=c++11 with nvcc, which CUDA < 7.0 doesn't handle. This results in a compilation error if the system compiling has CUDA < 7.0.

This PR adds a bit of logic to the main CMakeLists file to detect that we have the minimum CUDA version required for colmap to build with Cuda enabled.


ahojnnes:
Looks great, thanks!

PR #142, opened 2017-06-07
Add missing directory change to Mac build instructions.

botBoss:
Mac build instructions were missing

``` cd build ```

after

``` mkdir build ```

before running CMAKE


ahojnnes:
Great, thanks!

PR #147, opened 2017-06-12
Minor correction of input argument for *_matcher's

heimsi:
The documentation advises to use the option "--use_gpu true" or "--use_gpu false", but the *_matcher's have instead as input argument "--SiftMatching.use_gpu".


ahojnnes:
Great, thanks for the fix!

PR #155, opened 2017-06-15
Indicate possibility of dense output visualization

heimsi:
It was not clear to me whether COLMAP could visualize a dense reconstruction obtained via the CLI. I think it does no harm to indicate clearly that COLMAP has this feature. 

Regarding the third point about ``mesh.ply``: There I'm actually not sure if it's not possible. In the doc there is written that the surface normals cannot be visualized, but these are just a "part" of the ``dense_mesher`` output, right? Anyway, COLMAP in my case crashes when I try to load ``mesh.ply``, that's why I wrote it like this.


ahojnnes:
Thanks!

PR #240, opened 2017-10-17
Added support for range specification given by option timages in pmvs

SBCV:
Hi Johannes,

according to http://www.di.ens.fr/pmvs/documentation.html there are two possible value options for timages
- Enumeration (already implemented)
- Range specification (not implemented)
 For instance, the latter is used by OpenMVG when exporting PMVS files.

This pull requests contains the functionality to parse range specifications.

Cheers,
Sebastian


ahojnnes:
Hi Sebastion,

Thanks for the contribution. I had actually implemented this feature recently as well but did not have time to commit the changes yet due to a project deadline. The same functionality should be in the dev branch now.

Thanks again,
Johannes

PR #250, opened 2017-11-21
Update on the README.md on binary distributions not for mac/windows

alexmyczko:
Using links with repology and debian packaging


ahojnnes:
Thanks @alexmyczko

PR #271, opened 2018-01-07
Fix CameraModel tuples in read_model.py according to camera_models.h

shinsumicco:
`model_name` and `num_params` of `CameraModel` namedtuples in `scripts/python/read_model.py` had not been identical with the ones in `src/base/camera_models.h`, so that sparse reconstruction outputs had not been parsed correctly.
I have corrected `read_model.py` according to `camera_models.h`.


ahojnnes:
LGTM, thanks for the fix.

PR #314, opened 2018-02-25
1-to-1 Matching for Spatial Verification

trueprice:
Along with the other changes we've recently pushed, this change provides a significant increase in mAP versus pure retrieval on VGG's Oxford5k with a 200k-word vocabulary learned from Paris6k, (end-to-end in COLMAP, using upright affine features). I reverted the change of using affine feature geometry because, in the end, it didn't seem to offer too much of an improvement over the similarity feature geometry already implemented.

I've also pushed out a Python script to compute mAP from COLMAP's output (assumes VGG's label format). There are other small changes throughout, including switching the IDF weighting to use log from log1p (minor change that could be reverted), new command-line arguments for vocab tree checks and enabling verification, and an option to save the index to speed up future retrievals.


scripts/python/compute_retrieval_map.py
```diff
@@ -0,0 +1,107 @@
+import os
```
ahojnnes: Let's move this script to the vote-and-verify repository?

src/exe/colmap.cc
```diff
@@ -1476,8 +1478,10 @@ int RunVocabTreeRetriever(int argc, char** argv) {
   options.AddDefaultOption("database_image_list_path",
                            &database_image_list_path);
   options.AddDefaultOption("query_image_list_path", &query_image_list_path);
-  options.AddDefaultOption("num_images", &num_images);
-  options.AddDefaultOption("num_verifications", &num_verifications);
```
ahojnnes: Why did you remove the ``num_verifications``? For large databases you probably only want to verify a shortlist of images?

src/feature/matching.h
```diff
@@ -55,8 +55,11 @@ struct SequentialMatchingOptions {
   // be significantly bigger than the sequential matching overlap.
   int loop_detection_num_images = 30;
 
-  // Number of images to spatially verify for re-ranking.
-  int loop_detection_num_verifications = 0;
+  // Number of nearest-neighbor checks to use in retrieval.
+  int loop_detection_num_checks = 256;
+
+  // Whether to run spatial verification afer matched image retrieval.
```
ahojnnes: after

src/feature/matching.h
```diff
@@ -72,8 +75,11 @@ struct VocabTreeMatchingOptions {
   // Number of images to retrieve for each query image.
   int num_images = 100;
 
-  // Number of images to spatially verify for re-ranking.
-  int num_verifications = 0;
+  // Number of nearest-neighbor checks to use in retrieval.
+  int num_checks = 256;
+
+  // Whether to run spatial verification afer matched image retrieval.
```
ahojnnes: after

src/retrieval/visual_index.h
```diff
@@ -187,7 +186,11 @@ template <typename kDescType, int kDescDim, int kEmbeddingDim>
 void VisualIndex<kDescType, kDescDim, kEmbeddingDim>::Add(
     const IndexOptions& options, const int image_id, const GeomType& geometries,
     const DescType& descriptors) {
-  CHECK(image_ids_.count(image_id) == 0);
+  //CHECK(image_ids_.count(image_id) == 0);
```
ahojnnes: Remove commented code.

src/retrieval/vote_and_verify.cc
```diff
@@ -17,6 +17,7 @@
 #include "retrieval/vote_and_verify.h"
 
 #include <array>
+#include <forward_list>
```
ahojnnes: Unused include?

src/retrieval/visual_index.h
```diff
@@ -232,69 +235,219 @@ void VisualIndex<kDescType, kDescDim, kEmbeddingDim>::QueryWithVerification(
...
       continue;
     }
 
-    // Collect matches for all features of current image.
+    // Enforce 1-to-1 matching: Build Fibonacci heaps for the query and database
+    // features, ordered by the minimum number of matches per feature. We'll
+    // select these matches one at a time. For convenience, we'll also pre-sort
+    // the matched feature lists by matching score.
+
+    typedef boost::heap::fibonacci_heap<std::pair<int, int>> FibonacciHeapType;
```
ahojnnes: Is this faster than ``std::priority_queue``?

src/exe/colmap.cc
```diff
@@ -1466,8 +1467,9 @@ int RunVocabTreeRetriever(int argc, char** argv) {
   std::string vocab_tree_path;
   std::string database_image_list_path;
   std::string query_image_list_path;
-  int num_images = -1;
-  int num_verifications = 0;
+  std::string output_index_path;
+  retrieval::VisualIndex<>::QueryOptions query_options;
+  bool spatial_verification = false;
```
ahojnnes: It might be cleaner to add this to the ``QueryOptions``, but I can do that at some point later as well.

src/retrieval/inverted_index.h
```diff
@@ -379,22 +407,18 @@ void InvertedIndex<kDescType, kDescDim,
     inverted_file.ComputeIDFWeight(image_ids.size());
   }
 
-  const int max_image_id =
-      *std::max_element(image_ids.begin(), image_ids.end());
-
-  std::vector<double> self_similarities(max_image_id + 1, 0.0);
+  std::unordered_map<int, double> self_similarities;
```
ahojnnes: Since the number of images is known, ``self_similarities`` could be preallocated.

src/retrieval/inverted_index.h
```diff
@@ -329,10 +355,13 @@ void InvertedIndex<kDescType, kDescDim, kEmbeddingDim>::Read(
   ifs->read(reinterpret_cast<char*>(&num_images), sizeof(int32_t));
   CHECK_GE(num_images, 0);
 
-  normalization_constants_.resize(num_images);
-  for (int32_t image_id = 0; image_id < num_images; ++image_id) {
-    ifs->read(reinterpret_cast<char*>(&normalization_constants_[image_id]),
-              sizeof(float));
+  normalization_constants_.clear();
```
ahojnnes: Since the number of images is known, ``normalization_constants_`` could be preallocated.

scripts/python/compute_retrieval_map.py
trueprice: Works for me. It will probably be better noticed if we put it there, anyhow.

src/exe/colmap.cc
trueprice: The existing behavior is to run a [fixed number of verifications]
 (https://github.com/colmap/colmap/blob/dev/src/retrieval/visual_index.h#L247), so it made sense to me to just specify num_images and add a boolean flag for whether to also run spatial verification. If we want to run spatial verification on only a subset of retrieved images and then return all retrieved images, then we could keep num_verifications.

src/retrieval/vote_and_verify.cc
trueprice: Aha, yes, left over from the other EIC implementation.

src/retrieval/visual_index.h
trueprice: Yes, Fibonacci heaps allow for constant-time priority increases, which the [algorithm exploits] (https://github.com/trueprice/colmap/blob/76fb50fbd8ea44082a237396631c2c5add4f7519/src/retrieval/visual_index.h#L430) to update the priorities of other features once a feature is matched.

src/retrieval/inverted_index.h
trueprice: Yeah -- "std::unordered_map<int, double> self_similarities(image_ids.size())" should do the trick?

ahojnnes: Or .reserve()

src/retrieval/visual_index.h
ahojnnes: Learned something new. Thanks.

src/exe/colmap.cc
ahojnnes: aha, somehow I thought the previous behavior was like that. Makes sense. Let’s keep your versio.

ahojnnes:
Looks great, thanks!

PR #317, opened 2018-02-28
FOV-based undistortion

DmitriyKorchemkin:
This pull request adds additional mode for undistorting, which might be used with fisheye images.

I've tried to solve the following problems:
* Focal length: previously focal length was copied from original camera; for some camera models (i.e. OPENCV_FISHEYE) focal length parameters are not directly connected with focal length of undistorted camera
* Valid region estimation: frequently fisheye images contain valid image data only in some circular region; moreover, model validity (i.e. monothonicity of distance-to-principal-point and angle-to-optical axis) is not guaranteed outside of this region (due to no data available during calibration)

With `estimate_fov` enabled, field-of-view of undistorted camera is estimated by iteratively testing monotonicity of angle-distance relationship (optionally, clamped by `max_fov`; and horizontal or vertical field-of-view if `clip_by_hfov` or `clip_by_vfov` are enabled).

Then, undistorted camera focal length is set to estimated diagonal field-of-view.

During inscribed/circumscribed valid region estimation, border points are clamped to maximal valid radius estimated during FOV estimation.

`estimate_fov`, `clip_by_hfov`, `clip_by_vfov` are disabled by default; `max_fov` is set to 150 degrees.

### DIFF
--- a/src/base/undistortion.cc
+++ b/src/base/undistortion.cc
@@ -117,6 +117,35 @@ void WriteCOLMAPCommands(const bool geometric,
         << JoinPaths(workspace_path, output_prefix + "meshed.ply") << std::endl;
 }
 
+Eigen::Vector2d SelectPointOnRay(const Camera& camera,
+                                 const Eigen::Vector2d& origin,
+                                 const Eigen::Vector2d& target,
+                                 double max_length, double max_angle,
+                                 double max_horizontal_angle,
+                                 double max_vertical_angle) {
+  const Eigen::Vector2d diff = target - origin;
+  const Eigen::Vector2d dir = diff.normalized();
+
+  double l = 0.0, r = std::min(max_length, diff.norm());
+  for (int i = 0; i < 32; ++i) {
+    double m = (r + l) / 2.0;
+    const Eigen::Vector2d world_point = camera.ImageToWorld(origin + dir * m);
+    const double vertical_angle = std::atan(world_point[1]),
+                 horizontal_angle = std::atan(world_point[0]),
+                 full_angle = std::atan(world_point.norm());
+    const bool m_valid = vertical_angle < max_vertical_angle &&
+                         horizontal_angle < max_horizontal_angle &&
+                         full_angle < max_angle;
+
+    if (m_valid) {
+      l = m;
+    } else {
+      r = m;
+    }
+  }
+  return origin + dir * l;
+}
+
 }  // namespace
 
 COLMAPUndistorter::COLMAPUndistorter(const UndistortCameraOptions& options,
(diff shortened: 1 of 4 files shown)


ahojnnes:
Looks great! Could you post a few examples of the different options enabled/disabled and compared to the existing implementation? Could you also implement a few unit tests to ensure correct behavior?

DmitriyKorchemkin:
I've updated pull request with tests.

Below are some examples of fisheye image undistortion [using synthetic images for simplicity of undistorted image quality guesstimation]

# Source image

Fisheye-distorted image

image_0_default

## Undistortion: default

rectified-default

As you can see, this image contains significantly less data then initially available


## Undistortion: `estimate_fov=1`

rectified-estimate_fov

FOV is estimated and clipped to ~150 degrees (since `max_fov=150` by default).

This image depicts large portion if initial data, though there are some interpolation artifacts near the boundaries.


## Undistortion: `estimate_fov=1` and `max_fov=120`

rectified-max_fov_120

FOV is estimated and clipped to ~120 degrees (since `max_fov=120`).

This image depicts smaller portion if initial data, but there are less interpolation artifacts near the boundaries.

## Undistortion: `estimate_fov=1` and `clip_by_vfov=1`

rectified-clip_by_vfov

FOV is estimated and clipped to vertical FOV

## Undistortion: `estimate_fov=1`, `max_fov=179.9` and `clip_by_hfov=1`

rectified-clip_by_hfov

FOV is estimated and clipped to horizontal FOV (note that we also changed `max_fov` to its maximal value). These settings are invalid for the image, since the whole 180 degrees of half-subspace can be mapped horizontally. Default `max_fov` value prevents you from getting such results.

ahojnnes:
This looks great! Thanks. One final idea: I suggest to replace the current set of ``max_fov``, ``clip_by_hfov``, ``clip_by_vfov`` with ``max_horizontal_fov``, ``max_vertical_fov``. By default, both of them could be set to 150. What do you think?

src/base/undistortion.cc
```diff
@@ -634,22 +634,75 @@ Camera UndistortCamera(const UndistortCameraOptions& options,
...
-    undistorted_camera.SetFocalLengthY(camera.FocalLengthY());
+  // Copy focal length parameters, or estimate them
+  double max_valid_radius = std::numeric_limits<double>::max();
+  Eigen::Vector2d principal_point = Eigen::Vector2d::Zero();
+  if (options.estimate_fov) {
+    // Estimate maximal valid diagonal FOV (based on radius~angle monothonicity
+    const std::vector<size_t>& principal_idxs = camera.PrincipalPointIdxs();
+    CHECK_EQ(principal_idxs.size(), 2)
+        << "Only two principal point parameters supported.";
+    const double max_fov = options.max_fov * M_PI / 180.0;
```
ahojnnes: There is a function called `DegToRad` in ``util/math.h``.

src/base/undistortion.cc
```diff
@@ -634,22 +634,75 @@ Camera UndistortCamera(const UndistortCameraOptions& options,
...
+    for (int i = 1; i < max_radius; ++i) {
+      const Eigen::Vector2d world_point =
+          camera.ImageToWorld(i * corner_dir + principal_point);
+      const double phi = std::atan(world_point.norm());
+      if (phi <= valid_half_fov || 2.0 * phi > max_fov) break;
+      valid_half_fov = phi;
+      max_valid_radius = i;
+    }
+
+    double focal = rb_corner.norm() / 2.0 / std::tan(valid_half_fov);
```
ahojnnes: Make const

src/base/undistortion.cc
DmitriyKorchemkin: Yes, my solution looks a bit confusing:

* `max_fov` limits diagonal field-of-view by a user-specified constant
* `clip_by_vfov` [additionally] limits diagonal field-of-view by current horizontal field-of-view
* `clip_by_hfov` [additionally] limits diagonal field-of-view by current vertical field-of-view

I'll think how to make it more clear and update pull request

DmitriyKorchemkin:
Yes, my solution looks a bit confusing:

* `max_fov` limits diagonal field-of-view by a user-specified constant
* `clip_by_vfov` [additionally] limits diagonal field-of-view by current horizontal field-of-view
* `clip_by_hfov` [additionally] limits diagonal field-of-view by current vertical field-of-view

I'll think how to make it more clear and update pull request

DmitriyKorchemkin:
So far, I've decided to organize the process in a following way:

* `estimate_focal_length_from_fov` (boolean) controls if focal lengths are copied or estimated from angular measurements
* `max_fov`, `max_horizontal_fov` `max_vertical_fov` are applied independently of `estimate_focal_length_from_fov`, by marking pixels 'blank':
    *  `max_fov` controls maximal angle w.r.t. optical axis; default value is set to 179.9 in order to eliminate almost-parallel rays
    *  `max_horizontal_fov`, `max_vertical_fov` control  for angle between projections of ray and optical axis on a corresponding plane; default values are 180 degrees

With `max_fov=179.9` by default, I am able to obtain reasonable rectification on my fisheye dataset even without enabling focal length estimation and constraining horizontal/vertical fields of view.

Tuning additional parameters allows to decrease amount of interpolation artifacts present in the rectified images. 

It also seems reasonable to me to add "forced fixed rectified camera model" argument to rectification, should I?

ahojnnes:
@DmitriyKorchemkin Sorry for not following up on this pull request. I just have not had time to look into this again. I will get in touch sometime over the next weeks again. Thanks for your patience.

DmitriyKorchemkin:
@ahojnnes should I update my pull request in order to resolve these conflicts, or this feature makes too little sense?

Thanks

ahojnnes:
I think, this would be a very useful feature. If you have time, it would be great if you could rebase the pull request. Thanks!

ahojnnes:
@DmitriyKorchemkin Did you have time to revisit this? I think, this would be a very useful feature!

ahojnnes:
Sounds great, thanks!

DmitriyKorchemkin:
@ahojnnes What is the correct way to combine these changes with ones from #362 (rectangular ROI selection): should we estimate field of view of the whole source model, or only of selected ROI)?

Should I also add new options to `undistortion_widget.cc`?

ahojnnes:
I guess, field of view of whole model and then cropping. What do you think?

ahojnnes:
Please also update the undistortion widget and the command line interface with the new options. Thanks!

ahojnnes:
Thanks, finally had time to merge this. I made a few changes to get this integrated with the other PR that enabled ROI in the undistortion. If you have time, it would be great if you could verify those changes. Thanks!

ahojnnes:
@DmitriyKorchemkin

ahojnnes:
@DmitriyKorchemkin There seems to be a bug in the unit tests in `TestUndistortFOVEstimation`. The tests are passing, even though the values are not matching (expected vs undistorted focal lengths). Could you please double check and fix the issues?

PR #348, opened 2018-04-16
Unbreak packaging on BSD and non-x86_64

jbeich:
- cmake/CheckSSEExtensions.cmake magic can be replaced by `export CFLAGS="-march=native" CXXFLAGS="-march=native"`
- PBA defaults to SSE2 even with `-mno-sse` or on non-x86, see also [#890106](https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=890106)
- VLfeat [wants](https://github.com/vlfeat/vlfeat/blob/aead1c1eaca7/make/dll.mak#L42-L43) only specific files built with `-mavx` or `-msse2`

Tested on FreeBSD downstream: [10.3 amd64](https://ptpb.pw/CeRY), [10.3 i386](https://ptpb.pw/7v-R), [10.4 amd64](https://ptpb.pw/_pfM), [10.4 i386](https://ptpb.pw/D_Ql), [11.1 aarch64](https://ptpb.pw/VJMb), [11.1 amd64](https://ptpb.pw/6ybx), [11.1 armv6](https://ptpb.pw/Eu-d), [11.1 i386](https://ptpb.pw/k-ab), [12.0 aarch64](https://ptpb.pw/rgWi), [12.0 amd64](https://ptpb.pw/hsAm), [12.0 armv6](https://ptpb.pw/LOwM), [12.0 armv7](https://ptpb.pw/a6c7)


ahojnnes:
Fantastic! Thanks very much.

PR #362, opened 2018-05-09
Add rectangular ROI for undistortion

XieDufang:
Hello Johannes,

I have some datasets captured by cameras mounted on a vehicle.
After SfM, only the bottom half image (mostly road area) is needed, since I want to generate dense clouds of road surface.

Therefore I added rectangular ROI for undistortion. Hope it's helpful in different applications.

Mit freundlichen Gruessen！
Xie Dufang


src/ui/undistortion_widget.cc
```diff
@@ -42,6 +42,11 @@ UndistortionWidget::UndistortionWidget(QWidget* parent,
   AddOptionDouble(&undistortion_options_.max_scale, "max_scale", 0);
   AddOptionInt(&undistortion_options_.max_image_size, "max_image_size", -1);
   AddOptionDouble(&undistortion_options_.blank_pixels, "blank_pixels", 0);
+  AddOptionBool(&undistortion_options_.roi, "roi");
```
ahojnnes: I guess, we could get rid of this extra parameter by checking that all of the bounding box coordinates are 0/1?

XieDufang:
@ahojnnes 

Extra "roi" parameter has been removed.

The roi is enabled if min/max bounding box coordinates are greater/less than 0/1.

ahojnnes:
Thanks, I had a look at the logic again. We should make sure that the image has at least 1px in size. This might not be the case of the min/max values are very close. We should also make sure that rounding issues don’t cause the ROI to be outside of the original image size. Thanks! I could either implement those changes myself or I will wait for you. Please let me know.

XieDufang:
I will implement it this weekend

src/base/undistortion.cc
```diff
@@ -672,18 +678,48 @@ Camera UndistortCamera(const UndistortCameraOptions& options,
...
-  // Scale the image such the the boundary of the undistorted image.
-  if (camera.ModelId() != SimplePinholeCameraModel::model_id &&
-      camera.ModelId() != PinholeCameraModel::model_id) {
+  // Modify undistorted camera parameters based on ROI if enabled
+  size_t roi_min_c = 0;
+  size_t roi_min_r = 0;
+  size_t roi_max_c = camera.Width();
+  size_t roi_max_r = camera.Height();
+  
+  bool roi_enabled = false;
```
ahojnnes: You could make this a single statement ``const bool roi_enabled = ...``.

XieDufang:
OK. It is more readable. Thanks for the suggestion

XieDufang:
ROI validity check added ^_^

ahojnnes:
Looks great to me, thanks for the contribution!

PR #367, opened 2018-05-18
Fix for Issue#365

veloman-yunkan:
Because of this bug the principal point of the first camera was always set
to (0, 0). Principal point of every subsequent camera was being set to the
center of the previous image, which could be a problem for input images of
different size.


fishcu:
Thanks a lot. Does this work for camera models that optimize the principal point (like OPENCV should, at least)? Or is it going to be fixed at resolution / 2?

veloman-yunkan:
I haven't tested it except on a single flow that I am currently running. However the logic of that code seems quite simple to me and in my understanding it shouldn't break any previous functionality.

ahojnnes:
LGTM, thanks for the fix!

PR #393, opened 2018-06-20
Fix bug in python db.add_camera() function

SBCV:
Using "db.add_camera(...)" writes currently the parameters as np.float32 to the database file. This causes the following error when opening the database file with colmap: " database.cc:207 Check failed:  num_params == camera.NumParams() (2 vs. 4)" (because database.cc expects double values)


ahojnnes:
Great, thanks for spitting this bug and for the fix!

PR #396, opened 2018-06-20
Fix SiftGPU bug with skipped levels.

mgprt:
If a image level is skipped due to a limit on the number of features, its feature counter can still contain data from the last image. This can lead to invalid features being returned and in the worst case a heap corruption since the total feature number does not match the sum of features per level.


ahojnnes:
Great, thanks for the fix!

PR #399, opened 2018-06-22
Fix #392 and small fix for visibility vector

### LINKED ISSUE #392: Image Index VS Image ID
Hi Johannes,

the recently added function "dense point cloud visibility export" must be very useful for the following mesh refinement and/or texture mapping. It's a wonderful new feature.

In the comments you wrote that the "image index" is used instead of the original "image id" in the fusion step.

Actually, I found that the zero-based contiguous index is used in the whole mvs module (e.g. mvs::Model, mvs::Workspace, mvs::PatchMatch) for efficient data access. But in many variable names or function names the concept "image_id" is still being used.

To avoid confusion, I would like to fix the related variable naming in my fork. Can I create a pull request then?

Mit freundlichen Gruessen!
Xie Dufang

XieDufang:
1. Correct variable/function names for image index in the whole mvs module.
Rename variables/functions if they actually refer to contiguous zero-based index, not unordered id.

2. Clear/shrink visibility vector in accordance with point vector.
Small fix for visibility vector. Clear/shrink also visibility vector while clearing/shrinking fused point vector.


XieDufang:
The older commits (ROI in undistortion) are also listed here. But they have already been merged before. I think it's just because they were not merged directly from my commits but from your own commit.

Anyway, the actual file changes are only from the recent commits.

ahojnnes:
This looks great. One minor change request: Throughout colmap I use the convention that the plural of idx is idxs, so could you please rename indices to idxs? Thanks very much!

XieDufang:
Ok, I'll change "indices" to "idxs"

XieDufang:
"Indices" has been renamed to "idxs"    ^_^

ahojnnes:
Looks great, thanks!

Issue #401, opened 2018-06-25
Questions about loransac


is03wlei:
https://github.com/colmap/colmap/blob/dev/src/optim/loransac.h#L180

          for (const auto& local_model : local_models) {
            local_estimator.Residuals(X, Y, local_model, &residuals);
            CHECK_EQ(residuals.size(), X.size());

            const auto local_support =
                support_measurer.Evaluate(residuals, max_residual);

            // Check if non-locally optimized model is better.
            if (support_measurer.Compare(local_support, support)) {
              best_support = local_support;
              best_model = local_model;
              best_model_is_local = true;
            }
          }

Why not compare local_support with best_support?

is03wlei:
In this case, we couldn't pick up best model from local_models.

I looked up several local estimators. Most of them return only one model. So it works.
But if there is one local estimator returns multiple models. Then this is a issue.

ahojnnes:
That is a valid point and it should be fixed. I will do so in a few days. Thanks!

PR #407, opened 2018-06-28
fix issue of finding best local model

is03wlei:
Refer to https://github.com/colmap/colmap/issues/401 for more discussions.


ahojnnes:
This fix looks good but you should start off from the dev branch rather than the master branch.

PR #408, opened 2018-06-29
fix issue of finding best local model for loransac

is03wlei:
Refer to #401 for more discussions.


ahojnnes:
Thanks for the fix!

PR #413, opened 2018-07-03
Do not pass boost dependencies to linking lib when on Windows.

mgprt:
This prevents boost symbol redefinition errors as the symbols are already included in the COLMAP libs.


mgprt:
You might also consider providing absolute paths to the internal libraries or even generate imported targets. The use of `link_directories()` should not be necessary.

https://cmake.org/cmake/help/v3.11/command/link_directories.html

ahojnnes:
@mgprt If you have time to add the "imported target" feature that would be great?

ahojnnes:
Thanks for the fix.

PR #439, opened 2018-07-25
Add CGAL dependency to config file

mgprt:


mgprt:
I missed that before. Now it also handles the case when CGAL is enabled but not found.

ahojnnes:
Great, thanks!

PR #457, opened 2018-08-15
Allow partial image prefix for nested subfolder rig BA

fishcu:
I made this small change so that the following folder structure can be supported by rig BA:
```
dataset:
- sub_dataset1:
  - left:
    - img0001.png
    - ...
  - right:
- sub_dataset2:
  - left:
  - right:
...
```

This was previously not possible, because the "sub_dataset" string would prevent image names from matching.


fishcu:
I added a unit test. Did you mean the string function, or the modified way that the rig is initialized? I think string function is enough -- the other would be more of an integration test

ahojnnes:
I meant exactly what you added. Great, thanks very much.

PR #472, opened 2018-08-29
Automatically downgrade gcc for cuda compilation

DmitriyKorchemkin:
CUDA compilation is possible only with a limited set of gcc versions.

This pull request adds automatic "downgrading" gcc version (via `CUDA_HOST_COMPILER`) if `CUDA_HOST_COMPILER` is set to its default value (`CMAKE_C_COMPILER`).

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -196,6 +196,35 @@ if(CUDA_FOUND)
         # Do not show warnings if the architectures are deprecated.
         set(CUDA_NVCC_FLAGS "${CUDA_NVCC_FLAGS} -Wno-deprecated-gpu-targets")
 
+        # Automatically downgrade compiler, if required
+        if ("${CMAKE_CXX_COMPILER_ID}" STREQUAL "GNU"
+            AND "${CUDA_HOST_COMPILER}" STREQUAL "${CMAKE_C_COMPILER}")
+            # CUDA 9.x requires gcc <= 6.x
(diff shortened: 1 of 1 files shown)


alexmyczko:
please add support for cuda-10 as well?

DmitriyKorchemkin:
@alexmyczko do you know a way to get paths to the compilers without much hassle?

To be honest, I've abandoned this pull-request  just because auto-finding correct paths seemed too painful way, and just issuing warning seemed kinda useless (since you will get error from `nvcc` simply by ignoring this stuff completely).

alexmyczko:
That's what I added to README.Debian

You might want to rebuild with this added to debian/rules
 
```
# with cuda-10-0
override_dh_auto_configure:
        dh_auto_configure -- -DCUDA_SDK_ROOT_DIR=/usr/local/cuda-10.0 -DCUDA_TOOLKIT_ROOT_DIR=/usr/local/cuda-10.0
 
# or if you have cuda-10-1
override_dh_auto_configure:
        dh_auto_configure -- -DCUDA_SDK_ROOT_DIR=/usr/local/cuda-10.1 -DCUDA_TOOLKIT_ROOT_DIR=/usr/local/cuda-10.1
```

basically the two defines when you use nvidia cuda repos
```
$ cat /etc/apt/sources.list.d/cuda.list 
deb http://developer.download.nvidia.com/compute/cuda/repos/ubuntu1804/x86_64 /
```

PR #492, opened 2018-10-08
not refine camera intrinsics in local BA

is03wlei:


ahojnnes:
Thanks for your pull request, but this will not work, unfortunately. We just don't want to refine the intrinsics if not all of the images of a camera are included in the local bundle. If all the images of a camera are included, the intrinsics should be optimized.

is03wlei:
Should we extend refined_cameras to unordered_map and make use of it.
Besides, we need to decrease refined_cameras in DeRegisterImageEvent.
Although it would affect configurations of RefineAbsolutePose, it shouldn't be worse.

ahojnnes:
This will take more effort, because bundle adjustment currently does not allow to specify constant intrinsics for individual images.

is03wlei:
You mean we need to handle such case:
local bundle contains all images of camera 1 but some of images of camera 2.

ahojnnes:
@is03wlei Yes, exactly.

ahojnnes:
The issue is now fixed in the latest commit in the dev branch. Thanks.

PR #495, opened 2018-10-10
not refine extra camera params in RefineAbsolutePose when they have a…

is03wlei:
…lready been refined


ahojnnes:
Great, thanks for the fix.

PR #501, opened 2018-10-24
enable Qt::AA_UseHighDpiPixmaps for better less pixeled button icons

tompollok:


tompollok:
Sorry for my late response. I did not test it with COLMAP but in my own software.

If your icons are high dpi like 128x128 resolution then they look better on your GUI than without using AA_UseHighDpiPixmaps
If your icons are of low quality then you should not notice a difference.

Here a comparision. Both use the same icon files, but without AA_UseHighDpiPixmaps it looks worse.

grafik

I contributed this to a couple of open source applications that i use. Like CppCheck to improve the UX for 4K displays.

PR #516, opened 2018-11-08
Update export_to_visualsfm.py

tsattler:
Running the current version of the script leads to a crash since calling ```.reshape(-1, 4)``` in Line 103 leads to more keypoint positions that descriptors, resulting in an out-of-bound access in Line 114. Changing ```.reshape(-1, 4)``` to ```.reshape(-1, 6)``` should solve this problem as it takes into account that keypoints are stored with 6 attributes.


tsattler:
Not sure why the continuous integration is failing. My changes are in a python file that I don't think is actually checked.

ahojnnes:
Thanks!

PR #517, opened 2018-11-09
Binary descriptors export in export_to_visualsfm.py

tsattler:
This PR enables exporting the descriptors to the binary file format used by Visual SfM.


ahojnnes:
Thanks!

PR #518, opened 2018-11-15
compability fix for ubuntu 18.10

andijcr:
Hi, i was having trouble linking under the combination Ubuntu 18.10, cuda-8 and g++-5, so i made these two small changes.

The first commit unifies in the CMakeLists.txt the OPENGL_gl_LIBRARY and OPENGL_glu_LIBRARY into OPENGL_LIBRARIES, and in Debug build mode this builds fine.

the second commit removes the dependency to boost::regex, it was causing troubles in Release build mode.
I simply switched the c++11 mode on, and moved from boost::regex to std::regex.


alexmyczko:
it works fine with gcc-7 + cuda-10-0 on ubuntu 18.04 here

ahojnnes:
This has been resolved in another PR. Thanks.

Issue #36, opened 2016-10-01
Failed to build on Arch Linux.


mrbemani:
```
[ 45%] Linking CXX executable warp_test
../util/libutil.a(bitmap.cc.o): In function `char const* boost::re_detail_106100::re_is_set_member<char const*, char, boost::regex_traits<char, boost::cpp_regex_traits<char> >, unsigned int>(char const*, char const*, boost::re_detail_106100::re_set_long<unsigned int> const*, boost::re_detail_106100::regex_data<char, boost::regex_traits<char, boost::cpp_regex_traits<char> > > const&, bool)':
bitmap.cc:(.text._ZN5boost16re_detail_10610016re_is_set_memberIPKccNS_12regex_traitsIcNS_16cpp_regex_traitsIcEEEEjEET_S8_S8_PKNS0_11re_set_longIT2_EERKNS0_10regex_dataIT0_T1_EEb[_ZN5boost16re_detail_10610016re_is_set_memberIPKccNS_12regex_traitsIcNS_16cpp_regex_traitsIcEEEEjEET_S8_S8_PKNS0_11re_set_longIT2_EERKNS0_10regex_dataIT0_T1_EEb]+0x114): undefined reference to `boost::re_detail_106100::cpp_regex_traits_implementation<char>::transform_primary(char const*, char const*) const'
bitmap.cc:(.text._ZN5boost16re_detail_10610016re_is_set_memberIPKccNS_12regex_traitsIcNS_16cpp_regex_traitsIcEEEEjEET_S8_S8_PKNS0_11re_set_longIT2_EERKNS0_10regex_dataIT0_T1_EEb[_ZN5boost16re_detail_10610016re_is_set_memberIPKccNS_12regex_traitsIcNS_16cpp_regex_traitsIcEEEEjEET_S8_S8_PKNS0_11re_set_longIT2_EERKNS0_10regex_dataIT0_T1_EEb]+0x2b5): undefined reference to `boost::re_detail_106100::cpp_regex_traits_implementation<char>::transform(char const*, char const*) const'
collect2: error: ld returned 1 exit status
make[2]: *** [src/base/CMakeFiles/warp_test.dir/build.make:154: src/base/warp_test] Error 1
make[1]: *** [CMakeFiles/Makefile2:266: src/base/CMakeFiles/warp_test.dir/all] Error 2
make: *** [Makefile:139: all] Error 2
```

ahojnnes:
[...]

PR #520, opened 2018-11-16
remove boost::regex dependency

BertaBescos:
I was able to compile by removing the boost/regex dependency (I just replaced every boost:: for std:: in the bitmap.cc file, similar to what is suggested in this issue (https://github.com/colmap/colmap/issues/36), and removing the regex dependency in the CMakeLists.txt. Actually, since this is the only place where boost/regex is used, and since it seems to be of an issue for other people, I will make a pull request with this change if that's ok.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -35,25 +35,25 @@ project(COLMAP)
  
 set(COLMAP_VERSION "3.6")
 set(COLMAP_VERSION_NUMBER "3600")
- 
- 
+
+
 ################################################################################
 # Include CMake dependencies
 ################################################################################
- 
+
 set(CMAKE_MODULE_PATH ${CMAKE_CURRENT_SOURCE_DIR}/cmake)
- 
+
 include(CheckCXXCompilerFlag)
- 
+
 # Include helper macros and commands, and allow the included file to override
 # the CMake policies in this file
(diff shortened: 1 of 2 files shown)


mgprt:
Note that std::regex support was experimental and basically broken for GCC < 4.9 (see https://stackoverflow.com/a/12665408), so this could lead to bugs on older systems (e.g. Ubuntu 14.04, which is still supported).

tompollok:
Ubuntu 14.04 LTS is supported only until April 2019, which is very soon. There is already 16.04 and 18.04. I dont see why anybody should care to much about supporting old compilers in favour of using std library functionality.

ahojnnes:
I am against breaking Ubuntu 14.04 for no reason and even if we break Ubuntu 14.04, then we should make sure there is a warning or something, because this is a silent bug that is not visible to the user. The regex just doesn’t work. In any case, there should be a way to make this work both under 14.04 and 18.10 - worst case with separate ifdef branches.

Eberty:
It is possible make an ifdef to solve this issue?

attilaolah:
> I am against breaking Ubuntu 14.04 for no reason […]

I think the reason was the compilation error. That said, 14.04 is no longer supported, I don't think it's worth the complexity introduced by adding an `#ifdef` guard.

At some point, this pull request should probably be merged, and people who have trouble compiling it on old systems can still reverse-apply the patch.

ahojnnes:
Resolved with the latest commit in the dev branch. Thanks!

Issue #208, opened 2017-08-23
Implement masking of images in feature extraction


ahojnnes:

insikk:
I think this is really necessary feature when we want to construct 3d model of an object from photos taken with variety of background. 

@ahojnnes  I am trying to get 3d model of can from below images. I thinking manually masking out background may give me the 3d model. However, is there better way to achieve it with COLMAP software without mask? At least, can you give me some related papers so I can study it by myself?
![image](https://user-images.githubusercontent.com/2110359/33604004-d24abcac-d9f7-11e7-8cd4-2990076fdb12.png)

spriglet66:
This feature would be so useful! Is there a temporary work around like rubbing out the areas that I don't want colmap to identify as features such as; moving cars or people? Maybe deleting the areas in photoshop and saving as a png with alpha transparency then importing that into colmap?

Thanks!

Ross.

ahojnnes:
Hi, since there is growing interest in this feature, I am planning to implement this soon. Before starting, I would like to get some feedback to the following questions:
1) Is your final goal a sparse 3D reconstruction?
2) Is your final goal dense depth/normal maps?
3) Is your final goal a dense 3D reconstruction (point cloud or mesh)?
Please answer with 1), 2), 3), or a combination of them.

insikk:
@ahojnnes Thank you. I am planning to reconstruct sparse 3D only. 
By the way, what I want to know is that if there is some method to do SfM when we do not have explicit mask. Your experience in SfM may suggest us some research paper on this topic. 

With explicit mask, I was able to generate sparse 3d model. In this case, I was very luck that the dataset I used already contains ground-truth mask for target object. However, it is really rare case. 
![image](https://user-images.githubusercontent.com/2110359/34197719-015a9d1e-e5ab-11e7-8bea-64a25751af8d.png)
![image](https://user-images.githubusercontent.com/2110359/34197722-06295218-e5ab-11e7-96c5-fc06bbebdb65.png)

@spriglet66 If you have ground truth mask, then just use opencv to mask out background. 
For me, just masking out background was not enough, so I had to crop so the object fit in whole image as you can see in the above image.

ahojnnes:
[...]

PR #522, opened 2018-11-18
Implement masking of images in feature extraction

puzzlepaint:
This implements masking of images for feature extraction (by removing extracted features at masked-out image locations). Corresponds to issue #208 (regarding sparse features; does not implement masking for dense reconstruction).

Two new image extraction parameters are introduced:

* ImageReader.mask_path to specify a root directory for per-image masks:
```
  // Optional root path to folder which contains image masks. For a given image,
  // the corresponding mask must have the same sub-path below this root as the
  // image has below image_path. The filename must be equal, aside from the
  // added extension .png. For example, for an image image_path/abc/012.jpg, the
  // mask would be mask_path/abc/012.jpg.png. No features will be extracted in
  // regions where the mask image is black (pixel intensity value 0 in
  // grayscale).
```

* ImageReader.camera_mask_path to specify the file path of a mask that globally applies to all images:
```
  // Optional path to an image file specifying a mask for all images. No
  // features will be extracted in regions where the mask is black (pixel
  // intensity value 0 in grayscale).
```

Open points:

* I did not run clang-format, since the version I have installed outputs the following error:
```
[...]/colmap/.clang-format: Invalid argument
Can't find usable .clang-format, using LLVM style
YAML:21:29: error: invalid boolean
BreakBeforeBinaryOperators: None
                            ^~~~
```
* The two new options are not self-explanatory. In the options list widget for feature extraction, there should ideally be some documentation on them.
* The options list for feature extraction becomes quite long vertically. At some point it will reach or has reached a point where it won't fit on small screens. Introducing a scroll bar would help.

### DIFF
--- a/src/base/image_reader.h
+++ b/src/base/image_reader.h
@@ -44,8 +44,17 @@ struct ImageReaderOptions {
   // Path to database in which to store the extracted data.
   std::string database_path = "";
 
-  // Root path to folder which contains the image.
+  // Root path to folder which contains the images.
   std::string image_path = "";
+
+  // Optional root path to folder which contains image masks. For a given image,
+  // the corresponding mask must have the same sub-path below this root as the
+  // image has below image_path. The filename must be equal, aside from the
+  // added extension .png. For example, for an image image_path/abc/012.jpg, the
+  // mask would be mask_path/abc/012.jpg.png. No features will be extracted in
+  // regions where the mask image is black (pixel intensity value 0 in
+  // grayscale).
+  std::string mask_path = "";
 
   // Optional list of images to read. The list must contain the relative path
   // of the images with respect to the image_path.
@@ -72,6 +81,11 @@ struct ImageReaderOptions {
   // have focal length EXIF information, the focal length is set to the
   // value `default_focal_length_factor * max(width, height)`.
   double default_focal_length_factor = 1.2;
+  
+  // Optional path to an image file specifying a mask for all images. No
+  // features will be extracted in regions where the mask is black (pixel
+  // intensity value 0 in grayscale).
+  std::string camera_mask_path = "";
 
   bool Check() const;
 };
@@ -93,7 +107,7 @@ class ImageReader {
 
   ImageReader(const ImageReaderOptions& options, Database* database);
 
-  Status Next(Camera* camera, Image* image, Bitmap* bitmap);
+  Status Next(Camera* camera, Image* image, Bitmap* bitmap, Bitmap* mask);
   size_t NextIndex() const;
   size_t NumImages() const;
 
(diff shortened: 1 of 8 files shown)


src/util/misc.cc
```diff
@@ -114,6 +114,45 @@ std::string GetParentDir(const std::string& path) {
   return boost::filesystem::path(path).parent_path().string();
 }
 
+std::string GetRelativePath(const std::string& from, const std::string& to) {
```
ahojnnes: Ideally, we have a short small unit test in src/util/misc_test.cc

src/feature/extraction.cc
```diff
@@ -53,6 +53,33 @@ void ScaleKeypoints(const Bitmap& bitmap, const Camera& camera,
...
 
+void MaskKeypoints(const Bitmap& mask, FeatureKeypoints* keypoints,
+                   FeatureDescriptors* descriptors) {
+  unsigned int out_index = 0;
+  BitmapColor<uint8_t> color;
+  for (unsigned int i = 0; i < keypoints->size(); ++ i) {
+    if (mask.GetPixel(static_cast<int>(keypoints->at(i).x),
+                      static_cast<int>(keypoints->at(i).y),
+                      &color) &&
+        color.r == 0) {
```
ahojnnes: I feel like, if mask.GetPixel() == false, we also want to exclude it here? What do you think?

src/base/image_reader.h
```diff
@@ -44,9 +44,18 @@ struct ImageReaderOptions {
   // Path to database in which to store the extracted data.
   std::string database_path = "";
 
-  // Root path to folder which contains the image.
+  // Root path to folder which contains the images.
   std::string image_path = "";
 
+  // Optional root path to folder which contains image masks. For a given image,
```
ahojnnes: It would be great, if you could copy over this to the documentation to docs/faq.rst as well. I am sure this will be a feature that quite a few people will find useful :-)

src/feature/extraction.cc
```diff
@@ -64,6 +91,21 @@ SiftFeatureExtractor::SiftFeatureExtractor(
       image_reader_(reader_options_, &database_) {
   CHECK(reader_options_.Check());
   CHECK(sift_options_.Check());
+  
+  std::shared_ptr<Bitmap> camera_mask;
+  if (!reader_options_.camera_mask_path.empty()) {
+    camera_mask = std::shared_ptr<Bitmap>(new Bitmap(), [](Bitmap* b){
+      b->Deallocate();
```
ahojnnes: This shouldn't be necessary, because the destructor of Bitmap deallocates the data already.

src/feature/extraction.cc
puzzlepaint: If the mask size differs from the image size, maybe a warning should be printed? The user might have mixed up the assignment of images to masks.

ahojnnes: That’s a good idea. The other issue that I had in mind is that SIFT can also detect features outside of the image sometimes. In any case, we should also check for valid masks as you suggested.

PR #545, opened 2018-12-27
Fix image viewer widget.

is03wlei:


ahojnnes:
Thanks for the pull request, but the intention of the viewer is to show the image with and without keypoints. The two different colors for keypoints indicate whether they are triangulated or not. To make them better distinguishable, we could change the color from red to green. What was the motivation for the change here?

is03wlei:
Sorry, I thought it was to switch between 2D and 3D points.
As you said, now it's very difficult to see 3D points among lots of 2D points.

ahojnnes:
No problem, thanks for bringing this up anyway. If you have an idea on how to improve the visualization, a new PR or enhancement proposal would be welcome.

PR #563, opened 2019-01-30
Fix pathname for delaunay_mesher example in CLI docs

drkoller:
This PR fixes an error in the `delaunay_mesher` command-line example in the documentation.


ahojnnes:
Thanks for the fix!

PR #577, opened 2019-02-17
fix compare_points

sniklaus:
This is a really minor fix but I might as well make you aware of it. Huge thanks Johannes!


ahojnnes:
Thanks for the fix!

PR #590, opened 2019-03-08
Update maps to unordered for faster access/emplace times

mdaiter:
Just caught my eye: unordered_maps will let you create faster `emplace` and `count` commands (due to hashing) and avoid the unnecessary ordering that `std::map` includes.


ahojnnes:
Doesn’t seem like this should be any bottleneck here. For small sized maps (like this one), std::map is often faster. Did you profile the performance here?

PR #591, opened 2019-03-13
 Return squared cosine distance as residuals in GP3PEstimator. 

mgprt:
This makes the GP3PEstimator consistent with all other estimators which return the squared distance (and RANSAC using the squared threshold for comparison).


ahojnnes:
Excellent, thank you!

PR #601, opened 2019-04-05
Fix PLY file bug inherited from old version of PoissonRecon

drkoller:


ahojnnes:
Thanks for the fix, great!

PR #643, opened 2019-07-05
[image_undistorter] Skip already undistorted images

fishcu:
See title. 
To be consistent with other behavior in COLMAP, such as feature extraction and patch match stereo, if the output file exists, the processing should be skipped.

### DIFF
--- a/src/base/undistortion.cc
+++ b/src/base/undistortion.cc
@@ -217,6 +217,11 @@ void COLMAPUndistorter::Undistort(const size_t reg_image_idx) const {
   const std::string output_image_path =
       JoinPaths(output_path_, "images", image.Name());
 
+  // Early out if output exists
+  if (ExistsFile(output_image_path)) {
+    return;
+  }
+
   Bitmap distorted_bitmap;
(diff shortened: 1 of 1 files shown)


ahojnnes:
I intentionally avoided this, because it has led to problems for people in the past, where they reconstruct the same scene with different parameters and use the same undistortion folder. In this case, the undistorted images do not align with the 3D model and the dense reconstruction will be terrible. Thanks for your contribution but I believe it will be better to just rewrite the undistorted images. Is this a big time saver for you? In that case, we could create a new boolean option “ignore_existing_images” and set it to false.

fishcu:
I see. Yes, I have found a (to be honest, niche) use case where this can save substantial amounts of time.

Could you advise me how to implement that boolean option correctly (for example, all the points where it needs to be registered or documented), or could you add that yourself?
It sounds like a good solution (-- keep it at "true" by default to avoid the issue you obvserved).

ahojnnes:
Closing this as wontfix. Thanks for your suggested contribution.

PR #665, opened 2019-08-29
check --input_path and --output_path when running bundle_adjuster command

elmirjagudin:
When running bundle_adjust cli command, check that specified
--input_path and --output_path directories exist.

This way we get a nice error message right at the start,
instead of getting a more cryptic error message after
potentially lengthy processing.


ahojnnes:
Looks great, thanks!

PR #695, opened 2019-09-26
signed/unsigned int color bug in matlab/read_model

schedldave:
Color values are converted to signed 8bit integer so the range is limited from 0 to 127 (-128 to 127, but negative colors are not used). 
The correct type should be UNSIGNED 8bit integer (uint8) so the range is 0 to 255. 
This code change fixes this behaviour.


ahojnnes:
Thank you for the fix!

PR #743, opened 2019-11-27
Fix bug in GpuMat::FillWithScalar() implementation

drkoller:
In the COLMAP MVS code, the current implementation of `GpuMat::FillWithScalar()` does not properly handle `GpuMat` objects containing multi-byte values. So, this usage of the method:

https://github.com/colmap/colmap/blob/f3d7aae3fd287546edac68d346592175a28f892e/src/mvs/patch_match_cuda.cu#L1633-L1635

does not work as intended, since the `GpuMat<float>` 4-byte matrix values cannot be set by simply calling `cudaMemset()`, which only sets single byte values. Additionally, the method does not properly take the pitch of the `GpuMat` into account.

This pull request fixes these bugs in `GpuMat::FillWithScalar()` by calling a kernel to set the matrix values rather than trying to use `cudaMemset()`.

### DIFF
--- a/src/mvs/gpu_mat.h
+++ b/src/mvs/gpu_mat.h
@@ -124,6 +124,17 @@ class GpuMat {
 #ifdef __CUDACC__
 
 namespace internal {
+ 
+template <typename T>
+__global__ void FillWithScalarKernel(GpuMat<T> output, const T value) {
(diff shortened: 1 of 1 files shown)


ahojnnes:
Indeed, this is a bug. Thanks for catching this!

PR #751, opened 2019-12-04
Clarify about image id being the same as in the database

oleg-alexandrov:
This section starts by suggesting the user should create a text file with the cameras, and then run feature detection and matching, then use this text file to get the user cameras. The reality is more complicated. The text file should have image ids that agree with the database created during feature detection and matching. This text is clarifying that.

### DIFF
--- a/doc/faq.rst
+++ b/doc/faq.rst
@@ -121,6 +121,11 @@ Example of images.txt::
 
     4 0.698777 0.714625 -0.023996 0.021129 -0.048184 0.004529 -0.313427 2 image0004.png
 
+Each image above must have the same image id (first column) as in the database (next step). 
+This database can be inspected either in the GUI (selecting from the menu ``Database management''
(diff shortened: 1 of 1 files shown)


doc/faq.rst
```diff
@@ -121,6 +121,11 @@ Example of images.txt::
 
     4 0.698777 0.714625 -0.023996 0.021129 -0.048184 0.004529 -0.313427 2 image0004.png
 
+Each image above must have the same image id (first column) as in the database (next step). 
+This database can be inspected either in the GUI (selecting from the menu ``Database management''
+from under ``Processing''), or, one can create a reconstruction with colmap and later export 
+it as as text in order to see the images.txt file it creates.
```
ahojnnes: as as

oleg-alexandrov:
Thanks. I fixed the typo.

ahojnnes:
Thank you!

PR #753, opened 2019-12-05
Add write_model functions for python scripts

ClementPinard:
This commit adds functions to write models using python, be it binary of textual
Tests have been added too
Both have been renamed for their new functionality

### DIFF
--- a/scripts/python/read_write_model.py
+++ b/scripts/python/read_write_model.py
@@ -34,6 +34,7 @@ import sys
 import collections
 import numpy as np
 import struct
+import argparse
 
 
 CameraModel = collections.namedtuple(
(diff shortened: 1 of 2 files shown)


scripts/python/read_write_model.py
```diff
@@ -296,17 +453,28 @@ def rotmat2qvec(R):
     return qvec
 
 
+parser = argparse.ArgumentParser(description='Read and write COLMAP binary and text models')
```
ahojnnes: No need to run this code when importing the model. Otherwise these arguments will be injected in other scripts importing this module. Can you move it to the main function?

ahojnnes:
Looks great, left one minor comment. Thanks!

ClementPinard:
Hi thanks for you comment !

I updated my code with several improvements in addition to your suggestion :
- Unifying notations
- not using the dictionnary key to get the id, but the object `.id` attribute, so that the function caller can have any key system they want

Clément

ahojnnes:
Looks great, thanks for the contribution!

PR #755, opened 2019-12-05
Fix bug in GpuMat::Rotate() implementation

drkoller:
In the COLMAP MVS code, the current implementation of `GpuMat::Rotate()` does not properly handle `GpuMat` objects for which the row pitch (in bytes) is not a multiple of the matrix data type size. So, this usage of the method:

https://github.com/colmap/colmap/blob/06d546bd9bd285263b285b60de19953bb96f9f13/src/mvs/patch_match_cuda.cu#L1672-L1678

does not work as intended, since the `GpuMatPRNG` matrix value data type (`curandState`) is 48 bytes long, which does not divide evenly into the pitch value returned by the `cudaMallocPitch()` memory allocation for many matrix sizes.

This pull request fixes this bug in `GpuMat::Rotate()`. Similar bugs may exist in other `GpuMat` methods (such as `Transpose()`, `FlipHorizontal()`), but they do not affect COLMAP behavior, and I haven't addressed them.

### DIFF
--- a/src/mvs/cuda_rotate.h
+++ b/src/mvs/cuda_rotate.h
@@ -67,8 +67,8 @@ __global__ void CudaRotateKernel(T* output_data, const T* input_data,
   int output_x = input_y;
   int output_y = width - 1 - input_x;
 
-  output_data[output_y * output_pitch + output_x] =
-      input_data[input_y * input_pitch + input_x];
+  *((T*)((char*)output_data + output_y * output_pitch) + output_x) =
+      *((T*)((char*)input_data + input_y * input_pitch) + input_x);
 }
 
 }  // namespace internal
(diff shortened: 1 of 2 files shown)


ahojnnes:
Thanks very much for this fix. Out of curiosity, does it make any difference in the computed depth maps?

ahojnnes:
@drkoller I applied the same fix for the transpose/flip functions. Thanks again.

drkoller:
> Thanks very much for this fix. Out of curiosity, does it make any difference in the computed depth maps?

Yes, the depth maps generated with the bug fix are slightly different, although they don't appear to necessarily be "better" to the naked eye. With the buggy version of `GpuMat::Rotate()`, the `rand_state_map` was getting getting slightly corrupted during rotation, due to copying of uninitialized memory and misaligned memory accesses. But since `rand_state_map` just contains state info for the PRNG, the effect of the bug is just that the stream of generated random numbers used in `patch_match_stereo` had somewhat less randomness than expected from the usual cuRAND algorithm.

Here's a comparison of photometric depth map images as computed before and after the bug fix, using the first image from your ETH3D "living_room" dataset as reference image, at 1000x667 resolution. Before bug fix:

depth-map-before

After bug fix:

depth-map-after

You can observe some minor variations due to the different random values, but no big deal!

ahojnnes:
Good to know, thank you again!

Issue #121, opened 2017-05-19
Difference between SIMPLE_PINHOLE and PINHOLE


xianggao1989:
Hi,
I am using the Windows binary version of COLMAP and benefit a lot from it. Could you tell me the difference between the camera model of SIMPLE_PINHOLE and PINHOLE?
Thank you very much!
Best regards,
Xiang Gao from NLPR, CASIA

ahojnnes:
One vs two focal length parameters, which is useful if you have non-square pixels/shear. Always stick to the most simple model if it is sufficient.

xianggao1989:
Get it and thanks for your reply!

PR #777, opened 2019-12-28
Clarify pinhole vs simple pinhole

oleg-alexandrov:
I think it is useful to clarify a bit how these two differ. I know camera_models.h explains it, but it looks easier to just spell it out. It also addresses the question in https://github.com/colmap/colmap/issues/121.


ahojnnes:
Thanks!

PR #783, opened 2020-01-15
Configure the parameters of read_dense.py

mihaimorariu:


scripts/python/read_dense.py
```diff
@@ -50,17 +55,40 @@ def read_array(path):
     return np.transpose(array, (1, 0, 2)).squeeze()
 
 
+def parse_args():
+    parser = argparse.ArgumentParser()
+    parser.add_argument("-d", "--depth_map",
+                        help="path to depth map", type=str, required=True)
+    parser.add_argument("-n", "--normal_map",
+                        help="path to normal map", type=str, required=True)
+    parser.add_argument("--min_depth", help="minimum visualization depth",
```
ahojnnes: I would rename this to min_depth_percentile / max_depth_percentile and also adjust the help description, because these are not absolute depth values.

scripts/python/read_dense.py
```diff
@@ -29,7 +31,10 @@
 #
 # Author: Johannes L. Schoenberger (jsch-at-demuc-dot-de)
 
+import argparse
 import numpy as np
+import os
+import pylab as plt
```
ahojnnes: This should stay in the main function, otherwise matplotlib becomes a hard dependency when importing this file from somewhere else.

scripts/python/read_dense.py
mihaimorariu: Agreed. What do you think about adding the option for selecting absolute values as well?

scripts/python/read_dense.py
mihaimorariu: Is this file going to be imported anywhere else? I thought it is meant for visualization purposes, not part of a library.

scripts/python/read_dense.py
```diff
@@ -50,17 +55,42 @@ def read_array(path):
...
-        "path/to/dense/stereo/normal_maps/image1.jpg.photometric.bin")
+    if not os.path.exists(args.depth_map):
+        raise FileNotFoundError("File not found: {}".format(args.depth_map))
+
+    if not os.path.exists(args.normal_map):
+        raise FileNotFoundError("File not found: {}".format(args.normal_map))
 
-    import pylab as plt
+    depth_map = read_array(args.depth_map)
+    normal_map = read_array(args.normal_map)
```
mihaimorariu: Currently, the normal map is read, but not being used. I would either display it together with the depth or remove this line. What do you think?

scripts/python/read_dense.py
ahojnnes: I use this file in my personal scripts for parsing dense outputs.

scripts/python/read_dense.py
ahojnnes: Agreed.

ahojnnes: I would add a visualization of the normal map, just to show that it can be loaded using the same function as well.

scripts/python/read_dense.py
ahojnnes: Absolute values are relatively meaningless here, because the SFM reconstruction has arbitrary scale. I would stick to percentiles.

mihaimorariu:
Thank you for your feedback!

ahojnnes:
Thanks, looks great!

PR #791, opened 2020-01-29
Make permission bits consistent

wzel:
*.sh, *.py, and *.bat get 755.
Anything else gets 644.


ahojnnes:
Thanks for the fix. This PR requires resolving a lot of merge conflicts, so I went ahead and fixed the issues in another PR.

PR #821, opened 2020-03-05
Handle existing camera

ClementPinard:
Was using python database script to make a db file with already calibrated cameras, and a bunch of pictures associated to them.

The feature extractor works great but it tries to create a new camera for images that are already in the database. The result is the database gets new cameras in the db file, with no image linked to them.

This PR makes sure that it does not happen.

Also, I could see that the actual default value of prior_tvec (which should probably be called prior_gps or something, since it gets converted after) is nan and not 0, so I changed the database python script accordingly to avoid errors in spatial matcher.

Lastly, genuine question, is there a plan to use qvec prior in the future ? It' possible to get a rough odometry estimation with several IMU enabled camera such as drone videos or late generations of GoPros. If you have some pointers for papers using prior position and orientation estimation for photogrammetry other than for spatial matching, i'd be happy to read them !

Thanks for your help !

Clément


ClementPinard:
Hi, is this PR considered ? I actually need this modif in my own project (otherwise, tons of unnecessary cameras are created). When I'll share my project, I'll point the README to my personal branch, but I'd rather point to the master in order to keep the other developments up to date :)

thanks !

ahojnnes:
Looks good to me, thanks for the improvement! Sorry for the slow response.

PR #868, opened 2020-05-08
force check and sort image_list

DaniilSNikulin:
In documentation not required what `image_list`, which contains in `image_list_path` must be sorted. However, in realization this is expected and need for `single_camera_per_folder` mode.


ahojnnes:
Thanks for the fix.

PR #869, opened 2020-05-09
Increased sparse solver threshold

zpillio:
Found that BA convergence sloves down when it whiches to ceres::ITERATIVE_SCHUR solver type. (You can recognise it from the ceres log by checking ls_iter column. It contains when 1 for non iterative method is used, but goes up to maximum 100 for iterative method.)
ceres::ITERATIVE_SCHUR is kept, because for large problems that's faster, but the value of kMaxNumImagesDirectSparseSolver is increased from 1000 to 10000.


ahojnnes:
This PR contains many more changes than just changing the Ceres thresholds. Also, these thresholds very much depend on the sparsity pattern of the problem. I did quite some evaluations on these trade-offs and would like to keep the current defaults unless you can show the opposite on a really large set of datasets.

zpillio:
Sorry that was a mistake. I created a wrong poll request.

PR #882, opened 2020-05-23
Geometry.h add missing <cstdio> include

atcw:
Does not compile otherwise because "FILE does not name a type" error on compilation in PoissonRec/Geometry.h


ahojnnes:
Thanks for the fix, already resolved through another PR.

PR #887, opened 2020-05-30
Custom matching speedup

S-o-T:
Default value for ImagePairsMatchingOptions.block_size should be the same as the number of image pairs to be matched based on default value of ExhaustiveMatchingOptions.block_size (block_size * (block_size - 1) / 2). Low value of this option decrease potential utilization of hardware resources. Also, utilization can be increased by filtering out non unique image pairs.

### DIFF
--- a/src/feature/matching.cc
+++ b/src/feature/matching.cc
@@ -1439,6 +1439,7 @@ void ImagePairsFeatureMatcher::Run() {
 
   std::string line;
   std::vector<std::pair<image_t, image_t>> image_pairs;
+  std::unordered_set<colmap::image_pair_t> image_pairs_set;
   while (std::getline(file, line)) {
     StringTrim(&line);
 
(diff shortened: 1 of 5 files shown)


src/feature/matching.cc
```diff
@@ -1467,8 +1468,16 @@ void ImagePairsFeatureMatcher::Run() {
       continue;
     }
 
-    image_pairs.emplace_back(image_name_to_image_id.at(image_name1),
-                             image_name_to_image_id.at(image_name2));
+    image_pairs_set.insert(
```
ahojnnes, line 1471: Instead of forward/backward converting and doing the loop twice, you could simply check whether image_pairs_set already contains the pair and then directly emplace_back to image_pairs

ahojnnes:
Thanks, looks good. I added a minor suggestion for improvement.

Issue #886, opened 2020-05-29
compilation errors in 3.6-dev.3 in pba_generated_ProgramCU


3ddiddE:
**Describe the bug**
[  2%] Building NVCC (Device) object lib/PBA/CMakeFiles/pba.dir/pba_generated_ProgramCU.cu.o
Scanning dependencies of target lsd
Scanning dependencies of target flann
Scanning dependencies of target graclus
[  2%] Building C object lib/LSD/CMakeFiles/lsd.dir/lsd.c.o
[  2%] Building CXX object lib/FLANN/CMakeFiles/flann.dir/flann.cpp.o
[  2%] Building C object lib/Graclus/CMakeFiles/graclus.dir/metisLib/balance.c.o
[  2%] Building C object lib/Graclus/CMakeFiles/graclus.dir/metisLib/bucketsort.c.o
[  2%] Building C object lib/Graclus/CMakeFiles/graclus.dir/metisLib/ccgraph.c.o
[  4%] Building C object lib/Graclus/CMakeFiles/graclus.dir/metisLib/coarsen.c.o
[  4%] Building C object lib/Graclus/CMakeFiles/graclus.dir/metisLib/compress.c.o
[  4%] Linking C static library liblsd.a
[  4%] Built target lsd
Scanning dependencies of target poisson_recon
[  4%] Building CXX object lib/PoissonRecon/CMakeFiles/poisson_recon.dir/CmdLineParser.cpp.o
[  4%] Building C object lib/Graclus/CMakeFiles/graclus.dir/metisLib/debug.c.o
[  4%] Building CXX object lib/PoissonRecon/CMakeFiles/poisson_recon.dir/Factor.cpp.o
/usr/bin/../lib64/gcc/x86_64-pc-linux-gnu/10.1.0/../../../../include/c++/10.1.0/type_traits(1396): error: type name is not allowed

/usr/bin/../lib64/gcc/x86_64-pc-linux-gnu/10.1.0/../../../../include/c++/10.1.0/type_traits(1396): error: type name is not allowed

/usr/bin/../lib64/gcc/x86_64-pc-linux-gnu/10.1.0/../../../../include/c++/10.1.0/type_traits(1396): error: identifier "__is_same" is undefined

[  4%] Building C object lib/Graclus/CMakeFiles/graclus.dir/metisLib/estmem.c.o
[  4%] Building C object lib/Graclus/CMakeFiles/graclus.dir/metisLib/fm.c.o
[...]

PR #888, opened 2020-06-01
add missing include <stdio.h> to lib/PoissonRecon/Geometry.h

Fytch:
Would not compile otherwise with GCC 10.1.0:
```
In file included from /tmp/tmp.ij35ejNwXQ/colmap/lib/PoissonRecon/Geometry.cpp:28:
/tmp/tmp.ij35ejNwXQ/colmap/lib/PoissonRecon/Geometry.h:344:2: error: ‘FILE’ does not name a type
  344 |  FILE* _fp;
      |  ^~~~
In file included from /tmp/tmp.ij35ejNwXQ/colmap/lib/PoissonRecon/Geometry.cpp:28:
/tmp/tmp.ij35ejNwXQ/colmap/lib/PoissonRecon/Geometry.h:36:1: note: ‘FILE’ is defined in header ‘<cstdio>’; did you forget to ‘#include <cstdio>’?
   35 | #include "Hash.h"
  +++ |+#include <cstdio>
   36 | 
```

Also mentioned in #886.

### DIFF
--- a/lib/PoissonRecon/Geometry.h
+++ b/lib/PoissonRecon/Geometry.h
@@ -32,6 +32,7 @@ DAMAGE.
 #include <math.h>
 #include <vector>
 #include <stdlib.h>
+#include <stdio.h>
 #include "Hash.h"
 
 template<class Real>


Fytch:
There's also pull request #882 but that one proposes to include `<cstdio>`, the C++ variant of the header. As `FILE` is used without the `std::` namespace qualifier (which would be necessary for the C++ variants of the C headers) and as the other C headers in the same source file are being included in their C variant, I propose to include the C variant here, too.

ahojnnes:
Thanks for the fix!

PR #889, opened 2020-06-01
Catch std::invalid_argument const& instead of std::exception in misc.cc.

Fytch:
Firstly, catching a polymorphic type like `std::exception` by value causes
warnings. Secondly, we don't want to accidentally catch `std::bad_alloc`,
we only want to catch the errors potentially generated by `std::stoi`, both
of which derive from `std::invalid_argument`:
https://en.cppreference.com/w/cpp/string/basic_string/stol#Exceptions

### DIFF
--- a/src/util/misc.cc
+++ b/src/util/misc.cc
@@ -245,7 +245,7 @@ std::vector<int> CSVToVector(const std::string& csv) {
     }
     try {
       values.push_back(std::stoi(elem));
-    } catch (std::exception) {
+    } catch (std::invalid_argument const&) {
       return std::vector<int>(0);
     }
   }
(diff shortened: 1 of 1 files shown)


src/util/misc.cc
```diff
@@ -245,7 +245,7 @@ std::vector<int> CSVToVector(const std::string& csv) {
     }
     try {
       values.push_back(std::stoi(elem));
-    } catch (std::exception) {
+    } catch (std::invalid_argument const&) {
       return std::vector<int>(0);
```
ahojnnes, line 249: Can you move the const before the type? Or just apply clang-format and it will do it for you.

ahojnnes:
Thanks for the fix, happy to merge after the cosmetic issue is addressed.

ahojnnes:
Applying the change myself. Thanks!

PR #897, opened 2020-06-12
Filter observations with negative depth

hiakru:


src/base/reconstruction.cc
```diff
@@ -1356,30 +1365,49 @@ size_t Reconstruction::FilterPoints3DWithLargeReprojectionError(
 
     for (const auto& track_el : point3D.Track().Elements()) {
       const class Image& image = Image(track_el.image_id);
-      const class Camera& camera = Camera(image.CameraId());
-      const Point2D& point2D = image.Point2D(track_el.point2D_idx);
-      const double squared_reproj_error = CalculateSquaredReprojectionError(
```
ahojnnes, line 1361: `CalculateSquaredReprojectionError` already checks for positive depth. It returns DBL_MAX, if the point is behind the camera.

hiakru:
Oooooh, my bad

Issue #898, opened 2020-06-12
【Bug】Retriangulate after AdjustGlobalBundle?


hiakru:
Describe:
Should we consider do Retriangulation after AdjustGlobalBundle?

Change From
https://github.com/colmap/colmap/blob/9949d4e702f783fbecbe8c31477a8451c2de436a/src/controllers/incremental_mapper.cc#L118
to
![image](https://user-images.githubusercontent.com/33151098/84481239-62d84a00-acc8-11ea-83e1-378eb9aef512.png)

Reason:
I think it's reasonable that images could generate new point3D, as the pose of images changed after global bundle.

PR #907, opened 2020-06-17
Retriangulate after AdjustGlobalBundle

hiakru:
https://github.com/colmap/colmap/issues/898
I think it's reasonable that images could generate new point3D, as the pose of images changed after global bundle.


ahojnnes:
Retriangulation has very relaxed triangulation thresholds. “Retriangulation” with tight thresholds is already covered by the CompleteAndMerge() function. Thanks for the idea, but I don’t think this is a good idea.

PR #943, opened 2020-08-04
Fix for EPnP estimator

vlarsson:
I was doing some benchmarking with absolute pose solvers and found that COLMAP's implementation of EPnP was returning garbage in roughly 25% of cases. After comparing with the OpenGV implementation I found the problem. I have not done any extensive testing after the fix, but I think it should work better now.


ahojnnes:
Thanks, looks good!

PR #949, opened 2020-08-11
BuildImageModel: use std::vector instead of numbered arguments

Pascal-So:
I needed to make this change anyway in my fork and it looks quite upstreamable so I might as well send you this PR. I hope that this is useful and that I'm not missing some obvious reason why the previous version was implemented like that.

### DIFF
--- a/src/ui/model_viewer_widget.cc
+++ b/src/ui/model_viewer_widget.cc
@@ -71,12 +71,8 @@ inline Eigen::Vector4f IndexToRGB(const size_t index) {
 void BuildImageModel(const Image& image, const Camera& camera,
                      const float image_size, const Eigen::Vector4f& plane_color,
                      const Eigen::Vector4f& frame_color,
-                     TrianglePainter::Data* triangle1,
-                     TrianglePainter::Data* triangle2, LinePainter::Data* line1,
-                     LinePainter::Data* line2, LinePainter::Data* line3,
-                     LinePainter::Data* line4, LinePainter::Data* line5,
-                     LinePainter::Data* line6, LinePainter::Data* line7,
-                     LinePainter::Data* line8) {
+                     std::vector<TrianglePainter::Data>& triangle_data,
+                     std::vector<LinePainter::Data>& line_data) {
   // Generate camera dimensions in OpenGL (world) coordinate space.
   const float kBaseCameraWidth = 1024.0f;
   const float image_width = image_size * camera.Width() / kBaseCameraWidth;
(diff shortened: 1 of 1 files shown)


src/ui/model_viewer_widget.cc
```diff
@@ -896,27 +894,22 @@ void ModelViewerWidget::UploadImageData(const bool selection_mode) {
...
     if (!selection_mode) {
-      line_data.push_back(line1);
-      line_data.push_back(line2);
-      line_data.push_back(line3);
-      line_data.push_back(line4);
-      line_data.push_back(line5);
-      line_data.push_back(line6);
-      line_data.push_back(line7);
-      line_data.push_back(line8);
+      for (const LinePainter::Data& line : image_line_data) {
```
ahojnnes, line 905: These for loops could just become:
```
line_data.insert(line_data.end(), image_line_data.begin(), image_line_data.end());
```

ahojnnes:
Thanks, looks like a nice simplification. I added a small suggestion.

Pascal-So:
Oh right of course, thank you for catching this!

Looking at this again now, I think that I'll probably change BuildImageModel to just accept two output iterators and pass a nop output iterator for the line data if selection_mode is active. That way we never have to copy anything around at all. This will have to be after my exam later this week though.

Pascal-So:
This version now seems like the simpler option to me to avoid a copy than doing anything with back_inserter.

My usecase is that I'm displaying more information per image using additional lines, but I hope the change makes sense for you even in the standard setting.

src/ui/model_viewer_widget.cc
```diff
@@ -71,12 +71,9 @@ inline Eigen::Vector4f IndexToRGB(const size_t index) {
 void BuildImageModel(const Image& image, const Camera& camera,
                      const float image_size, const Eigen::Vector4f& plane_color,
                      const Eigen::Vector4f& frame_color,
-                     TrianglePainter::Data* triangle1,
-                     TrianglePainter::Data* triangle2, LinePainter::Data* line1,
-                     LinePainter::Data* line2, LinePainter::Data* line3,
-                     LinePainter::Data* line4, LinePainter::Data* line5,
-                     LinePainter::Data* line6, LinePainter::Data* line7,
-                     LinePainter::Data* line8) {
+                     std::vector<TrianglePainter::Data>& triangle_data,
```
ahojnnes, line 74: output arguments should be pointers (Google C++ coding guideline).

src/ui/model_viewer_widget.cc
```diff
@@ -108,84 +105,76 @@ void BuildImageModel(const Image& image, const Camera& camera,
       Eigen::Vector4f(-image_width, -image_height, focal_length, 1);
 
   // Image plane as two triangles.
+  triangle_data.reserve(triangle_data.size() + 2);
```
ahojnnes, line 108: Remove this, as it will lead to extremely O(N^2) allocation cost.

src/ui/model_viewer_widget.cc
```diff
@@ -108,84 +105,76 @@ void BuildImageModel(const Image& image, const Camera& camera,
...
-      PointPainter::Data(bl(0), bl(1), bl(2), frame_color(0), frame_color(1),
-                         frame_color(2), frame_color(3));
-  line8->point2 =
-      PointPainter::Data(tl(0), tl(1), tl(2), frame_color(0), frame_color(1),
-                         frame_color(2), frame_color(3));
+                         plane_color(2), plane_color(3)));
+
+  if (show_lines) {
+    // Frame around image plane and connecting lines to projection center.
+    line_data.reserve(line_data.size() + 8);
```
ahojnnes, line 128: Remove the reserve, same issue as above

src/ui/model_viewer_widget.cc
Pascal-So: I might be wrong on this but I think it isn't O(N^2). We're reserving the total required size at the beginning of ModelViewerWidget::UploadImageData and ModelViewerWidget::UploadMovieGrabberData, which means that the reserve here is a noop. But I agree that this ends up being unnecessary in this case, if we assume that BuildImageModel will only be called from these places.

Pascal-So: And even if someone ends up using BuildImageModel without having reserved the total required size at the beginning, it probably still makes sense to keep it here because otherwise the allocation would happen during emplace back which isn't any better I guess.

src/ui/model_viewer_widget.cc
Pascal-So: I'm not too familiar with the Google C++ coding guidelines, but it seems like [this section here](https://google.github.io/styleguide/cppguide.html#Inputs_and_Outputs) recommends references. Am I checking in the wrong place?

Or do you mean that I should consider line_data to be an optional output parameter and then remove the show_lines parameter and instead pass nullptr as line_data when not required?

src/ui/model_viewer_widget.cc
ahojnnes: If we ever use this function somewhere else then this will end up as O(N^2). If the emplace_back() detects that the vector is out of capacity it will just increase the vector by a certain factor.

Pascal-So: Ah I see, you're right of course, I'll fix that tonight when I get home.

src/ui/model_viewer_widget.cc
ahojnnes: Apparently they updated their guidelines and I am pretty sure the old convention was to use pointers. In any case, it's the convention across the entire COLMAP code base, so we should follow it here as well. Thanks.

Pascal-So: Okay makes sense. In this case it's probably best to add null-checks and to remove the show_lines parameter. Same here, I'll fix this as soon as I get home.

ahojnnes:
Great, thanks very much and sorry for the delay in getting this merged.

PR #984, opened 2020-09-18
script for modifying fused results

SBCV:


scripts/python/read_write_fused_vis.py
```diff
@@ -0,0 +1,119 @@
...
+# ARISING IN ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE
+# POSSIBILITY OF SUCH DAMAGE.
+#
+# Author: Johannes L. Schoenberger (jsch-at-demuc-dot-de)
+
+import os
+import struct
+import collections
+import numpy as np
+import pandas as pd
```
ahojnnes, line 38: Unused

scripts/python/read_write_fused_vis.py
```diff
@@ -0,0 +1,119 @@
...
+# SUBSTITUTE GOODS OR SERVICES; LOSS OF USE, DATA, OR PROFITS; OR BUSINESS
+# INTERRUPTION) HOWEVER CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN
+# CONTRACT, STRICT LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE)
+# ARISING IN ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE
+# POSSIBILITY OF SUCH DAMAGE.
+#
+# Author: Johannes L. Schoenberger (jsch-at-demuc-dot-de)
+
+import os
+import struct
```
ahojnnes, line 35: Unused

scripts/python/read_write_fused_vis.py
```diff
@@ -0,0 +1,119 @@
...
+import collections
+import numpy as np
+import pandas as pd
+from pyntcloud import PyntCloud
+
+from read_write_model import read_next_bytes
+from read_write_model import write_next_bytes
```
ahojnnes, line 42: Merge with previous import

ahojnnes:
Looks great, thanks. Left some minor comments for you.

SBCV:
Changed the PR according to your comments and updated the PR accordingly. I guess the two "unused" comments referred both to "import struct"?

ahojnnes:
Pandas seems like a heavy dependency just for enabling structured numpy arrays. The same should be possible with numpy out of the box?

SBCV:
I leveraged the `pandas` package here, because it is internally used by the `pyntcloud` package (and will be automatically installed with `pip install pyntcloud`).

I've also considered other libraries than `pyntcloud`.

For example, in the past I've used [plyfile](https://github.com/dranjan/python-plyfile), which has no other dependencies. However, the interface of `plyfile` is really convoluted and a lot of custom code is required for writing points from and to plyfiles. See this this [file](https://github.com/SBCV/Blender-Addon-Photogrammetry-Importer/blob/master/photogrammetry_importer/file_handlers/ply_file_handler.py) for example. In addition, I've read that `pyntcloud` is three times [faster](https://stackoverflow.com/questions/36920562/python-plyfile-vs-pymesh).

I also considered `Open3D` for reading/writing ply files. However, Open3D [does not support](https://github.com/intel-isl/Open3D/issues/1325) to configure the types in the generated output file (32 bit float vs 64 bit float). Thus, with Open3D it is not possible to compare the results, i.e.
```
    assert filecmp.cmp(path_to_fused_ply_input, path_to_fused_ply_output)
    assert filecmp.cmp(path_to_fused_ply_vis_input, path_to_fused_ply_vis_output)
```
Thus, I think that the `pyntcloud` package is a valid choice - and therefore the usage of `pandas` not too bad. What do you think?

ahojnnes:
Thanks for the explanations, makes sense to me. Merging and thanks for the contribution.

PR #985, opened 2020-09-19
Fix bugs of sift feature matching

whuaegeanse:
1. Fix bug in sift feature matching, when features cannot be extracted from images lacking texture.

2. Fix bug in sift feature matching, when the number of feature points extracted from an image lacking texture is less than 2.

### DIFF
--- a/src/feature/matching.cc
+++ b/src/feature/matching.cc
@@ -240,6 +240,16 @@ void FeatureMatcherCache::Setup() {
       cache_size_, [this](const image_t image_id) {
         return database_->ReadDescriptors(image_id);
       }));
+
+  keypoints_exists_cache_.reset(new LRUCache<image_t, bool>(
+      images.size(), [this](const image_t image_id) {
+        return database_->ExistsKeypoints(image_id);
+      }));
+
+  descriptors_exists_cache_.reset(new LRUCache<image_t, bool>(
+      images.size(), [this](const image_t image_id) {
+        return database_->ExistsDescriptors(image_id);
+      }));
 }
 
 const Camera& FeatureMatcherCache::GetCamera(const camera_t camera_id) const {
@@ -279,12 +289,12 @@ std::vector<image_t> FeatureMatcherCache::GetImageIds() const {
 
 bool FeatureMatcherCache::ExistsKeypoints(const image_t image_id) {
   std::unique_lock<std::mutex> lock(database_mutex_);
-  return database_->ExistsKeypoints(image_id);
+  return keypoints_exists_cache_->Get(image_id);
 }
 
 bool FeatureMatcherCache::ExistsDescriptors(const image_t image_id) {
   std::unique_lock<std::mutex> lock(database_mutex_);
-  return database_->ExistsDescriptors(image_id);
+  return descriptors_exists_cache_->Get(image_id);
 }
 
 bool FeatureMatcherCache::ExistsMatches(const image_t image_id1,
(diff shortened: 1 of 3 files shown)


src/feature/sift.cc
```diff
@@ -206,22 +206,24 @@ void FindBestMatchesOneWayFLANN(
...
+
   const size_t kNumNearestNeighbors = 2;
   const size_t kNumTreesInForest = 4;
 
-  indices->resize(query.rows(), std::min(kNumNearestNeighbors,
-                                         static_cast<size_t>(database.rows())));
-  distances->resize(
-      query.rows(),
-      std::min(kNumNearestNeighbors, static_cast<size_t>(database.rows())));
+  if (kNumNearestNeighbors > database.rows()) {
```
ahojnnes, line 216: It's not really relevant for the COLMAP codebase, but technically we would still populate the indices/distances output and just skip the FLANN code block below.

src/feature/matching.cc
```diff
@@ -277,6 +277,16 @@ std::vector<image_t> FeatureMatcherCache::GetImageIds() const {
   return image_ids;
 }
 
+bool FeatureMatcherCache::ExiststKeypoints(const image_t image_id) {
+  std::unique_lock<std::mutex> lock(database_mutex_);
+  return database_->ExistsKeypoints(image_id);
+}
+
+bool FeatureMatcherCache::ExiststDescriptors(const image_t image_id) {
```
ahojnnes, line 285: Exists (not Existst)

src/feature/matching.cc
```diff
@@ -277,6 +277,16 @@ std::vector<image_t> FeatureMatcherCache::GetImageIds() const {
   return image_ids;
 }
 
+bool FeatureMatcherCache::ExiststKeypoints(const image_t image_id) {
```
ahojnnes, line 280: Exists (not Existst)

src/feature/matching.cc
```diff
@@ -450,6 +472,18 @@ void GuidedSiftCPUFeatureMatcher::Run() {
         continue;
       }
 
+      if (!cache_->ExiststKeypoints(data.image_id1) ||
+          !cache_->ExiststKeypoints(data.image_id2)) {
+        CHECK(output_queue_->Push(data));
+        continue;
+      }
+
+      if (!cache_->ExiststDescriptors(data.image_id1) ||
```
ahojnnes, line 481: Could be combined with the ExistsKeypoints() tests above.

src/feature/matching.cc
```diff
@@ -510,6 +544,18 @@ void GuidedSiftGPUFeatureMatcher::Run() {
...
```
ahojnnes, line 553: Could be combined with the ExistsKeypoints() tests above.

ahojnnes:
Thanks, looks good to me. I left some minor comments that could be addressed before merging.

whuaegeanse:
> Thanks, looks good to me. I left some minor comments that could be addressed before merging.

Following your suggestion, I made a new commit.

ahojnnes:
Thanks, one more comment: it would be good to use an LRU cache similar to keypoints_/descriptors_cache_ for the new Exists methods to avoid disk access to the database.

whuaegeanse:
> Thanks, one more comment: it would be good to use an LRU cache similar to keypoints_/descriptors_cache_ for the new Exists methods to avoid disk access to the database.

Good idea. I have learnt a lot from you. Thank you for providing such a great open source SfM engine.

ahojnnes:
Thanks, this looks great.

PR #997, opened 2020-10-03
fix camera model query

Pascal-So:
The return value of StringReplace was not being used and therefore the query was done with the name still containing spaces and dashes.


ahojnnes:
Thanks for the fix.

PR #1038, opened 2020-11-02
search src images for patch_match from all set, not only referenced subset

### LINKED ISSUE #1034: Strange behavior in `manual specification of source images during dense reconstruction`
If specify `__all__` in `patch_match.cfg`, then not all images will be as source, but only subset of reference images.

https://github.com/colmap/colmap/blob/c4714b2c6856d0e2afd3a993ff224f2f00634ebc/src/mvs/patch_match.cc#L314-L321

Is this ok? I can provide PR for code changes.

DaniilSNikulin:
fixes #1034


ahojnnes:
Thanks and sorry for delay in reviewing these changes.

PR #1039, opened 2020-11-03
Remove deprecated qt foreach

UncleGene:
QT foreach is on deprecation path (https://codereview.qt-project.org/c/qt/qtbase/+/147363) and already may not be available in qt depending on build configuration.
Change to C++11+ range for should be forward-compatible, and as far as I can see C++11 is the current minimum version


ahojnnes:
Thanks!

PR #1040, opened 2020-11-04
Ensure DecomposeHomographyMatrix() always returns rotations

daithimaco:
DecomposeHomographyMatrix() sometimes returns reflection matrices instead of rotations. 

This results in invalid quaternions in two view geometry estimation:
https://github.com/colmap/colmap/blob/91795d028736031901f2fdb1d2cd77956cddd0fc/src/estimators/two_view_geometry.cc#L209

Adding a fix here which just involves making sure the homography matrix has positive determinant before applying the formulas from the paper.

Also added a unit test, and confirmed it fails about half the time without this change.


ahojnnes:
Great, thank you!

PR #1058, opened 2020-11-19
Add image_list_path option to mapper gui

Dawars:


Dawars:
The list doesn't get refreshed when the reconstruction is paused and then resumed but neither do the other parameters so I think this is the expected behavior.

Maybe it would be interesting in the future to create a file browser window to select the images.

Issue #865, opened 2020-05-08
sparse reconstruction show is not normal when i use amd r5 3500u , but in nvida is normal 


pdycp3:
**Describe the bug**

**To Reproduce**

**Expected behavior**

**Screenshots**

**Desktop (please complete the following information):**

**Smartphone (please complete the following information):**

**Additional context**

tsattler:
The description that you provide is not detailed enough to determine what the problem is. Please be more precise. Ideally, show the two reconstructions.

pdycp3:
thanks for your comment, I use south building dataset, doing reconstruction in colmap, the sparse model look like a wrong model, i convert this model into .ply format, check it in other software,and it look like a right model ,i don't know why

tsattler:
This is still not sufficient information to determine what might be the problem. Can you describe how the model looks wrong? Or even better, post a screenshot here?

pdycp3:
![image](https://user-images.githubusercontent.com/45677507/81407241-a1309580-916d-11ea-8d1e-b2b13d54d114.png)

pdycp3:
![image](https://user-images.githubusercontent.com/45677507/81407502-17cd9300-916e-11ea-8bc2-b0c7fd08a1fe.png)
ply in other software

tsattler:
If you are loading a model in the text format, please have a look at this issue: #766

pdycp3:
![image](https://user-images.githubusercontent.com/45677507/81407836-a7734180-916e-11ea-9bc6-c659e35e6b3d.png)
thanks for your help,this is my gpu information ,I don't know wether  it can help you locate the problem

schizz99:
I have the same problem and it appears to be the same as the unresolved issue #218. Only seems to happen with AMD cards, and does occur with updated drivers. The corrupted rendering occurs during any reconstruction, including original reconstruction and when opening existing reconstructions.

Issue #218, opened 2017-09-09
Reconstruction on AMD cards yields wrong results


MrMinimal:
According to this [conversation on Google Groups](https://groups.google.com/forum/#!topic/colmap/pvSB8SxZ3b4) there is a problem with reconstruction in the current build.

The results with the provided south-building dataset look like this:

![broken](https://user-images.githubusercontent.com/8739690/30244150-a261f3da-95b8-11e7-867f-7eaf95834788.png)

The one thing all commenters seem to have in common is an AMD card (multiple architechtures).

I have a new Vega card as well in case I can be any help in finding the problem.

ahojnnes:
Seems like the underlying reconstruction process succeeded but something is wrong with the visualization of the model in the GUI. Did you install the latest GPU drivers on your system?

MrMinimal:
Yes, the drivers are up to date.

Looking at the output in Meshlab, I get the following
![image](https://user-images.githubusercontent.com/8739690/30387201-1b762994-98ac-11e7-8ae6-eb2eada098ab.png)

Whereas it should look more like this (Done via VisualSFM):
![image](https://user-images.githubusercontent.com/8739690/30387279-4cf9b562-98ac-11e7-9a65-04b28e16f992.png)

Granted, refining etc. is most likely not going to be enough. Any further help I can provide on this one?

ahojnnes:
Unfortunately, I do not know what is going on in your case. Something is clearly wrong even with the reconstruction. How did you produce the dense reconstruction on your system with Colmap?

MrMinimal:
The steps for the sparse reconstruction were

- Feature extraction
- Feature matching
- Start reconstruction

and tried automatic reconstruction as well, which yielded the same results.

For the dense reconstruction, I used openMVS.

**Good news:** Meshlab shows the correct sparse point cloud if exported from Colmap.
So you were right with the assumption that it is just a display error in AMD cards.

The next problem I have is dense reconstruction does not work in Colmap. When clicking "stereo" a "No CUDA available" dialog pops up. I am using the dev-windows-no-cuda build.

Any ideas or should I rather make a new issue with this one?

ahojnnes:
COLMAP requires a CUDA-enabled GPU for dense reconstruction. Alternatives are to use PMVS/CMPMVS. The documentation describes how to do this. Let me know if you have questions.

MrMinimal:
Will do, thanks for the help!

MrMinimal:
[...]

PR #1079, opened 2020-12-18
Fix AMD/Windows GUI visualization bug

drkoller:
Throughout the lifetime of the COLMAP software, users on the Windows platform with AMD display drivers have reported problems visualizing 3D geometry in the main display window of the COLMAP GUI. See for example:
- #218 Reconstruction on AMD cards yields wrong results
- #865 Sparse reconstruction show is not normal when I use amd r5 3500u
- [Discussion thread on COLMAP Google Groups](https://groups.google.com/g/colmap/c/pvSB8SxZ3b4)

The source of these problems is an OpenGL programming bug resulting in mismatched vertex attribute indices (locations) between the shader code and the calling code. For example, this current COLMAP code:

https://github.com/colmap/colmap/blob/0aea04cbd9bc5042a70beb5618cd1e52977e13ac/src/ui/line_painter.cc#L84-L92

sets the vertex position attribute to correspond to location 0, and the vertex color attribute to location 1. However, these attribute locations are not specified in the GLSL shaders, and so different OpenGL implementations are free to assign different mappings for these vertex attributes. While Nvidia drivers seem to use choose locations 0 and 1 for the vertex position and color attributes, respectively, AMD drivers on Windows seem to be using the opposite configuration (location 0 for color, location 1 for position) when the COLMAP shaders are compiled. So, the vertex data is misinterpreted by the shaders on AMD cards.

There are several different ways in OpenGL to harmonize the vertex attribute locations with the shaders (`glGetAttribLocation()` queries, `glBindAttribLocation()` binding, `GL_ARB_explicit_attrib_location` in the shader code, etc.). In this PR, I've fixed the bug by using similar `QOpenGLShaderProgram` methods to those already in use by COLMAP, but calling them with the corresponding attribute names from the shader programs, rather than the hard-coded indices of 0 and 1. With this change, the vertex attributes are properly matched, and 3D geometry is rendered as expected on both NVidia and AMD systems.

### DIFF
--- a/src/ui/line_painter.cc
+++ b/src/ui/line_painter.cc
@@ -82,14 +82,14 @@ void LinePainter::Upload(const std::vector<LinePainter::Data>& data) {
                 static_cast<int>(data.size() * sizeof(LinePainter::Data)));
 
   // in_position
-  shader_program_.enableAttributeArray(0);
-  shader_program_.setAttributeBuffer(0, GL_FLOAT, 0, 3,
+  shader_program_.enableAttributeArray("a_pos");
+  shader_program_.setAttributeBuffer("a_pos", GL_FLOAT, 0, 3,
                                      sizeof(PointPainter::Data));
 
   // in_color
-  shader_program_.enableAttributeArray(1);
-  shader_program_.setAttributeBuffer(1, GL_FLOAT, 3 * sizeof(GLfloat), 4,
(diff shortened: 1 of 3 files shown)


ahojnnes:
Fantastic, thanks very much for tracking down this issue!

PR #1084, opened 2020-12-30
include colmap_cuda in COLMAP_LIBRARIES when compiled with cuda

### LINKED ISSUE #1065: use colmap as a library problem
Hello, thank you very much for your great work.I've used colmap directly on GUI and command line, and all ther have achieved good results.
But when I followed [(https://colmap.github.io/install.html#library)],I met the problem ,I have been looking for a long time and haven't found a suitable method. Where is the problem?(my CMakeLists.txt and hello_world.cc just the same as the [(https://colmap.github.io/install.html#library)])
Here is the error log.

```
(base) magus@magus:~/zxp_workspace/test2/build$ make
Scanning dependencies of target hello_world
[ 50%] Building CXX object CMakeFiles/hello_world.dir/hello_world.cc.o
[100%] Linking CXX executable hello_world
/usr/local/share/colmap/../../lib/colmap/libcolmap.a(matching.cc.o)：in function‘colmap::SiftFeatureMatcher::SiftFeatureMatcher(colmap::SiftMatchingOptions const&, colmap::Database*, colmap::FeatureMatcherCache*)’：
matching.cc:(.text+0x4d11)：‘colmap::GetNumCudaDevices()’Undefined reference
collect2: error: ld returned 1 exit status
CMakeFiles/hello_world.dir/build.make:136: recipe for target 'hello_world' failed
make[2]: *** [hello_world] Error 1
CMakeFiles/Makefile2:67: recipe for target 'CMakeFiles/hello_world.dir/all' failed
make[1]: *** [CMakeFiles/hello_world.dir/all] Error 2
Makefile:83: recipe for target 'all' failed
make: *** [all] Error 2
```

ClementPinard:
fixes #1065 
As discussed there, the example given in https://colmap.github.io/install.html#library does not work if colmap was compiled with cuda support because it doesnt link against the cuda_colmap library, even if the cuda functions are not used.

This PR tries tries to fix the problem by adding colmap_cuda to the COLMAP_LIBRARIES variable so that the end user does not have to worry about it.


ahojnnes:
Thanks!

PR #1103, opened 2021-02-02
Store relative poses in two_view_geometry table

Ahmed-Salama:
Small change to persist two view geometry relative pose (`qvec` and `tvec`) in the corresponding `two_view_geometries` table.

COLMAP's database has a `two_view_geometries` table which serializes instances of the `TwoViewGeometry` class, but it doesn't store the relative pose, which adds the burden of having to re-compute after reading from the database.

Also extended `base/database_test` to test the read and write.


ahojnnes:
Looks perfect. Nice feature addition. Thanks!

PR #1114, opened 2021-02-11
Integration of PatchmatchNet into colmap by implementing the evaluation in LibTorch 

anmatako:
Example of PatchmatchNet integration with COLMAP using LibTorch.

-  Added PatchMatchNet implementation as alternative to standard patch-match
  - New functionality controlled by the new `--method learned` option in `patch_match_stereo`
  - Additional options were added in `PatchMatchOptions` to load a checkpoint and a parameter dictionary
- Created inheritance structure for `PatchMatch` (base class) and `PatchMatchCuda` and `PatchMatchNet` (derived) to facilitate the choice of processing method
- All PatchmatchNet torch-related files are under to `colmap_cuda` library due to their CUDA dependency


CMakeLists.txt
```diff
@@ -130,6 +131,11 @@ if(CGAL_FOUND)
     message(STATUS "  Libraries : ${CGAL_LIBRARY}")
 endif()
 
+if (TORCH_ENABLED)
+    set(CMAKE_TORCHLIB_PATH ${CMAKE_CURRENT_SOURCE_DIR}/lib/libtorch)
```
anmatako, line 135: The CMake config is based on existence of LibTorch as part of COLMAP lib. Did not include in the PR as this should likely change and require it as an external dependency

scripts/torch/param_dictionary.txt
```diff
@@ -0,0 +1,212 @@
+feature.stage1.0.conv.weight, module.feature.conv0.conv.weight
```
anmatako, line 1: This whole file is a hack to allow loading the existing checkpoint from Pytorch. The names of the parameters do not match up exactly. Ideally there should be a checkpoint created with the correct names and then this "translation" can go away.

src/exe/CMakeLists.txt
```diff
@@ -37,3 +37,12 @@ endif()
...
 set_target_properties(colmap_exe PROPERTIES OUTPUT_NAME colmap)
+
+if(IS_MSVC)
+  file(GLOB TORCH_DLLS "${CMAKE_TORCHLIB_PATH}/lib/*.dll")
+  add_custom_command(TARGET colmap_exe
+                     POST_BUILD
+                     COMMAND ${CMAKE_COMMAND} -E copy_if_different
+                     ${TORCH_DLLS}
+                     $<TARGET_FILE_DIR:colmap_exe>)
+endif()
```
anmatako, line 48: This was recommended by the LibTorch setup to copy the DLLs alongside the executable. I'm not sure if COLMAP does that already, so it may be redundant.

src/mvs/patch_match.cc
```diff
@@ -153,31 +158,10 @@ void PatchMatch::Check() const {
   }
 }
 
-void PatchMatch::Run() {
-  PrintHeading2("PatchMatch::Run");
-
-  Check();
-
-  patch_match_cuda_.reset(new PatchMatchCuda(options_, problem_));
```
anmatako, line 161: Removed this in favor of an inheritance structure where `PatchMatchCuda` and the new `PatchMatchNet` both inherit from `PatchMatch`. Not sure what was the intent of using `PatchMatchCuda` inside a `unique_ptr` so if that breaks something important I could change it back.

src/mvs/patch_match_cuda.h
```diff
@@ -50,17 +50,17 @@
...
                  const PatchMatch::Problem& problem);
-  ~PatchMatchCuda();
+  virtual ~PatchMatchCuda();
 
-  void Run();
+  virtual void Run() override;
 
-  DepthMap GetDepthMap() const;
-  NormalMap GetNormalMap() const;
-  Mat<float> GetSelProbMap() const;
```
anmatako, line 63: This seemed to be unused

anmatako:
Abandon this in favor of newer PR that should be able to get merged in public repo

PR #1115, opened 2021-02-11
Fix runtime crash when sparsesuite is missing from ceres

anmatako:
The Ceres built-in variable `sparse_linear_algebra_library_type ` is set correctly during build, so checking the existence of sparse support before using the `SPARSE_SCHUR` method avoids a runtime crash in bundle adjustment


ahojnnes:
LGTM, thank you!

PR #1132, opened 2021-02-23
Update GPSTransform calculations to improve accuracy

anmatako:


src/base/gps.cc
```diff
@@ -38,23 +38,20 @@ namespace colmap {
...
-      f_ = 0.003352810681182;
-      break;
     case WGS84:
-      a_ = 6378137;
-      b_ = 6.356752314245179e+06;
-      f_ = 0.003352810664747;
+      a_ = 6378137.0;
+      // Using flattening factor and semi-minor axis calculation from Bing maps.
+      // New calculations are marginally more accurate than the original values.
+      f_ = 1.0 / 298.257222101;
```
Freeverc, line 45: Accurate f of WGS84 is 1.0 / 298.257223563.  You are using a wrong value here,which comes from GRS80(1.0 / 
 298.2572221008827112431628366). Please refer to https://en.wikipedia.org/wiki/World_Geodetic_System

src/base/gps_test.cc
```diff
@@ -66,10 +66,10 @@ BOOST_AUTO_TEST_CASE(TestEllToXYZWGS84) {
...
                    11 + 34. / 60 + 11.77179 / 3600, 561.1509);
   std::vector<Eigen::Vector3d> ref_xyz;
-  ref_xyz.emplace_back(4.177239709042750e6, 0.855153779923415e6,
-                       4.728267404769168e6);
-  ref_xyz.emplace_back(4.177218660452103e6, 0.855175931344048e6,
-                       4.728281850382507e6);
+  ref_xyz.emplace_back(4.1772397090808507e6, 0.85515377993121441e6,
+                       4.7282674046563692e6);
+  ref_xyz.emplace_back(4.1772186604902023e6, 0.8551759313518483e6,
+                       4.7282818502697079e6);
```
Freeverc, line 73: The test case is for GRS80 not WGS84!

Freeverc (commented):
The gps test is in wrong logic now.

src/base/gps.cc
ahojnnes: @Freeverc Thanks for detecting this issue. Would you be able to submit a fix to revert back ?

Freeverc: That's ok. Fix submited.

PR #1134, opened 2021-02-23
Add CopyFile utility that can copy or hard/soft-link files

anmatako:


anmatako:
This will be leveraged in subsequent PRs to skip undistortion of already undistorted images.

PR #1135, opened 2021-02-23
Update camera intrinsics with new camera_updater command

anmatako:
This is a limited command to update the intrinsics for SIMPLE_RADIAL cameras (focal length, principal point x, and principal point y). The input is a CSV file with 4 entries per line: `<camera_id>, <focal length>, <PP x>, <PP y>`. It also sets the `prior_focal_length` field to `true`.


anmatako:
Updated the functionality to now handle all camera models

PR #1136, opened 2021-02-23
Allow cleanup of SQLite tables using new database_cleaner command

anmatako:
Added functionality to clear SQLite tables and also call `VACUUM` on `Database::Close` to reduce the DB file size after cleanup.

### DIFF
--- a/src/base/database.h
+++ b/src/base/database.h
@@ -201,6 +201,21 @@ class Database {
   void DeleteInlierMatches(const image_t image_id1,
                            const image_t image_id2) const;
 
+  // Clear all database tables
+  void ClearAllTables() const;
+
+  // Clear the entire cameras table
+  void ClearCameras() const;
+
+  // Clear the entire images, keypoints, and descriptors tables
+  void ClearImages() const;
+
+  // Clear the entire descriptors table
+  void ClearDescriptors() const;
+
+  // Clear the entire keypoints table
+  void ClearKeypoints() const;
+
   // Clear the entire matches table.
   void ClearMatches() const;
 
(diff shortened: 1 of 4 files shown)


src/exe/colmap.cc
```diff
@@ -226,6 +226,49 @@ int RunColorExtractor(int argc, char** argv) {
   return EXIT_SUCCESS;
 }
 
+int RunDatabaseCleaner(int argc, char** argv) {
+  int cleanup_type = -1;
+
+  OptionManager options;
+  options.AddRequiredOption("cleanup_type", &cleanup_type,
+                            "{0: All, 1: Images, 2: Features, 3: Matches}");
```
ahojnnes, line 234: It might be more natural to expose these options as strings, such that I could write `colmap database_cleaner --database_path path/to/db --type all` instead of `--type 0`.

anmatako: Not sure why I used numeric values in the first place. I'll update them to strings.

src/base/database.h
```diff
@@ -201,6 +201,21 @@ class Database {
...
+  void ClearCameras() const;
+
+  // Clear the entire images, keypoints, and descriptors tables
+  void ClearImages() const;
+
+  // Clear the entire descriptors table
+  void ClearDescriptors() const;
+
+  // Clear the entire keypoints table
+  void ClearKeypoints() const;
```
ahojnnes, line 217: Would be good to add basic unit tests for these methods in the database_test.cc file.

anmatako: Added testing for clearing tables as part of the existing unit tests. If that's not desirable I can write individual tests for this functionality.

PR #1140, opened 2021-02-23
Extend SceneClustering to support non-hierarchical (flat) clusters

anmatako:


ahojnnes:
Would be great to add some unit tests to scene_clustering_test.cc to make sure this new functionality works as expected.

anmatako:
> Would be great to add some unit tests to scene_clustering_test.cc to make sure this new functionality works as expected.

Unit tests added for flat clustering with and without overlaps.

anmatako:
Tests will have to be revised. It appears that the graph cut is not fully deterministic in the order it assigns labels to clusters; the cluster sizes and the IDs in each cluster are as expected but the order may differ, e.g. cluster 0 on Windows may be cluster 1 in Ubuntu. Not sure why this is not being an issue with the unit tests for hierarchical clustering. New tests pass on Windows and Mac but fail in Ubuntu.

anmatako:
Made the new flat clustering deterministic sorting by descending cluster size and using min image ID as tie breaker. Now the test should be consistent on all environments.

PR #1143, opened 2021-02-23
Add auto_matcher command for feature matching

anmatako:
`auto_matcher` is a convenience feature matching command that chooses the best feature matching method based on the given option. The choices in order are:
1. ImagePairsFeatureMatcher: if the `--ImagePairsMatching.match_list_path` is set to a valid image pairs file that includes all images in the DB
2. SpatialFeatureMatcher: if the image pairs file is not given or is malformed in any way and the DB contains GPS priors for the images
3. ExhaustiveFeatureMatcher: This is a fallback that should always work if neither of the two earlier choices are usable. It is also used when the max matching distance of the spatial matcher is larger than the max distance of images in the DB; this happens because in such a case the spatial matcher devolves to exhausting matching anyway, and leveraging the batches of the exhaustive matcher speeds things up a bit

### DIFF
--- a/src/exe/colmap.cc
+++ b/src/exe/colmap.cc
@@ -37,6 +37,7 @@
 #include <boost/property_tree/ptree.hpp>
 
 #include "base/similarity_transform.h"
+#include "base/gps.h"
 #include "controllers/automatic_reconstruction.h"
 #include "controllers/bundle_adjustment.h"
 #include "controllers/hierarchical_mapper.h"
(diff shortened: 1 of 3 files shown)


src/exe/colmap.cc
```diff
@@ -176,6 +177,132 @@ int RunAutomaticReconstructor(int argc, char** argv) {
   return EXIT_SUCCESS;
 }
 
+static bool UseSpatialMatcher(const std::string& database_path,
```
ahojnnes, line 180: Let's replace static with an anonymous namespace.

src/exe/colmap.cc
```diff
@@ -176,6 +177,132 @@ int RunAutomaticReconstructor(int argc, char** argv) {
...
+    return true;
+  } else {
+    std::cout << "Max matching distance exceeds max image distance; using "
+                 "exhaustive matching"
+              << std::endl;
+    return false;
+  }
+}
+
+static bool UseImagePairsMatcher(const std::string& database_path,
```
ahojnnes, line 223: Same here, remove static and move it to an anonymous namespace.

PR #1144, opened 2021-02-24
Add fixed extrinsics in rig config

anmatako:
The rig config can now support fields for relative extrinsics and use these values instead of estimating relative poses from the reconstruction. To use the new behavior the rig config must contain the new "location" and "orientation" properties and the flag estimate_rig_relative_poses must be false (`--estimate_rig_relative_poses 0`) when calling `rig_bundle_adjuster`

### DIFF
--- a/src/exe/colmap.cc
+++ b/src/exe/colmap.cc
@@ -1633,22 +1633,36 @@ int RunPointTriangulator(int argc, char** argv) {
 //       {
 //           "camera_id": 1,
 //           "image_prefix": "left1_image"
+//           "location": [0, 0, 0],
+//           "orientation": [1, 0, 0, 0]
 //       },
 //       {
 //           "camera_id": 2,
 //           "image_prefix": "left2_image"
+//           "location": [0, 0, 0],
+//           "orientation": [0, 1, 0, 0]
 //       },
 //       {
 //           "camera_id": 3,
 //           "image_prefix": "right1_image"
+//           "location": [0, 0, 0],
+//           "orientation": [0, 0, 1, 0]
 //       },
 //       {
 //           "camera_id": 4,
 //           "image_prefix": "right2_image"
+//           "location": [0, 0, 0],
+//           "orientation": [0, 0, 0, 1]
 //       }
 //     ]
 //   }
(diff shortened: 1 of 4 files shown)


src/exe/colmap.cc
```diff
@@ -1633,23 +1633,37 @@ int RunPointTriangulator(int argc, char** argv) {
...
 //           "camera_id": 4,
 //           "image_prefix": "right2_image"
+//           "location": [0, 0, 0],
+//           "orientation": [0, 0, 0, 1]
 //       }
 //     ]
 //   }
 // ]
 //
+// The "camera_id" and "image_prefix" fields are required, whereas the
```
ahojnnes, line 1661: * It would be good to call location->rel_tvec and orientation->rel_qvec for consistency with the rest of the code base.
* It would be good to describe the coordinate convention of these parameters, i.e., rel_qvec/rel_tvec transform coordinates from rig to camera coordinate space.

anmatako: Renamed the fields and added a bit more to the comment below. Let me know if more explanation is needed.

src/exe/colmap.cc
```diff
@@ -1699,8 +1712,28 @@ std::vector<CameraRig> ReadCameraRigConfig(
...
       image_prefixes.push_back(camera.second.get<std::string>("image_prefix"));
-      camera_rig.AddCamera(camera_id, ComposeIdentityQuaternion(),
-                           Eigen::Vector3d(0, 0, 0));
+      Eigen::Vector3d rel_tvec;
+      Eigen::Vector4d rel_qvec;
+      int index = 0;
+      auto rel_tvec_node = camera.second.get_child_optional("rel_tvec");
+      if (rel_tvec_node) {
+        for (const auto& node : rel_tvec_node.get()) {
+          rel_tvec[index] = node.second.get_value<double>();
```
jcanizales, line 1721: You missed the `++` after `index` 🙂  Same for loading `rel_qvec` below.

anmatako: Good catch on that one; I'll post a fix.

PR #1146, opened 2021-02-24
ImageReder new option and bug fix in GPS priors

anmatako:
Default behavior of `ImageReader` is now changed and it will attempt to reuse the same camera model if the model exif tag and image size remains the same. The old default of one camera per image can now be used through the explicit option `--ImageReader.single_camera_per_image 1`. Convenience input `--camera_mode` was added in `feature_extractor` to set the correct flags for `ImageReader`.

Fixed bug when reading GPS coordinates from EXIF tag where the Lat/Lon reference tag (N or S, and E or W) was being ignored resulting in incorrect GPS coordinates for locations outside the NE quadrant; coordinates were always positive instead of negative Lat for south hemisphere and negative Lon for West hemisphere.


src/util/bitmap.h
```diff
@@ -164,7 +165,7 @@ class Bitmap {
 
   // Rescale image to the new dimensions.
   void Rescale(const int new_width, const int new_height,
-               const FREE_IMAGE_FILTER filter = FILTER_BILINEAR);
+               const FREE_IMAGE_FILTER filter = FILTER_CATMULLROM);
```
ahojnnes, line 168: Why this change? Seems unrelated to the other changes in this PR. I found bilinear to work quite well for rescaling. Note that FreeImage applies an appropriate low-pass filter depending on the scaling factor, so it's not just a pure bilinear interpolation of the pixels.

src/util/bitmap.h
```diff
@@ -145,6 +145,7 @@ class Bitmap {
 
   // Extract EXIF information from bitmap. Returns false if no EXIF information
   // is embedded in the bitmap.
+  bool ExifCameraModel(std::string& camera_model) const;
```
ahojnnes, line 148: The convention in COLMAP is to use pointers for output arguments.

ahojnnes: In other words, this should be std::string*

src/exe/colmap.cc
```diff
@@ -499,12 +499,16 @@ bool VerifyCameraParams(const std::string& camera_model,
...
+enum class CameraMode { AUTO = 0, SINGLE = 1, PER_FOLDER = 2, PER_IMAGE = 3 };
+
 int RunFeatureExtractor(int argc, char** argv) {
   std::string image_list_path;
+  int camera_mode = (int)CameraMode::AUTO;
 
   OptionManager options;
   options.AddDatabaseOptions();
   options.AddImageOptions();
+  options.AddDefaultOption("camera_mode", &camera_mode);
```
ahojnnes, line 511: Needs documentation.

How does this interact with the following definitions in option_manager.cc?
```
AddAndRegisterDefaultOption("ImageReader.single_camera",
                              &image_reader->single_camera);
  AddAndRegisterDefaultOption("ImageReader.single_camera_per_folder",
                              &image_reader->single_camera_per_folder);
```

src/util/bitmap.h
anmatako: Yes that was pulled accidentally from the diff. I will revert it to its prior setting.

src/util/bitmap.h
anmatako: Fixed

src/base/image_reader.cc
```diff
@@ -181,15 +181,30 @@ ImageReader::Status ImageReader::Next(Camera* camera, Image* image,
       return Status::CAMERA_SINGLE_DIM_ERROR;
     }
 
+    //////////////////////////////////////////////////////////////////////////////
+    // Read camera model and check for consistency if it exists
+    //////////////////////////////////////////////////////////////////////////////
+    std::string camera_model = "Undefined";
+    bitmap->ExifCameraModel(&camera_model);
```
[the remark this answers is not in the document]
anmatako: That is a good suggestion. So far the model/make tags along with width and height have provided uniqueness for the data available to us. Including the focal length does make sense and if there are other tags that could be used I can include those as well.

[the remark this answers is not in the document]
ahojnnes: It would be great, if you could add focal length to the "unique hashing". A short documentation of the trade-offs of using "auto" could also be great. Thanks! Then I am happy to merge this.

ahojnnes, line 188: If I read this code here correctly, then this would always select a single camera, if the `camera_model == "Undefined"`? This doesn't seem right.

src/util/bitmap.cc
```diff
@@ -279,6 +279,27 @@ bool Bitmap::InterpolateBilinear(const double x, const double y,
   return false;
 }
 
+bool Bitmap::ExifCameraModel(std::string* camera_model) const {
+  // Read camera make and model
+  std::string make_str;
+  std::string model_str;
+  std::string focal_length;
+  *camera_model = "";
+  if (ReadExifTag(FIMD_EXIF_MAIN, "Make", &make_str)) {
```
ahojnnes, line 288: I think each if block here should have an else { return false } and in the negative case also set camera_model = ""

ahojnnes, line 288: I think each if block here should have an else { return false } and in the negative case also set camera_model = ""

src/base/image_reader.cc
ahojnnes: We should also check whether we successfully retrieved the camera_model by checking the return value of ExifCamerModel

anmatako: Due to the "hashing" by make, model, focal length, and width+height the call to `ExifCameraModel` will always return something meaningful even if it's just the "WidthxHeight" descriptor. Thus, even for missing EXIF tags it will still only group together images with the same dimensions (which I think is fine in this case) and won't force a single camera for the whole model.

Returning false from `ExifCameraModel` when an EXIF tag is missing will break the setup since the `cameras_` is a map indexed by `camera_model` so we do want this to always succeed for the `AUTO` mode to work correctly. Returning an empty string would essentially be forcing the use of a single camera model in the `AUTO` mode when EXIF tags are missing. But, maybe that's a more acceptable behavior; I'm open to handle this in whichever way makes most sense to you.

ahojnnes: Okay, so correct me if I am wrong, but the way I understand the current behavior: if only width_height are available, then it will use the same camera for all images that only have width_height EXIF fields and where they are equal. The behavior I want is that: if any of the make/model/focal/width/height fields are missing, instantiate a new/independent camera for that image. I really only want to group them to a single camera, if all the fields are present and their values match exactly.

anmatako: OK I see your take now; I can certainly do it this way. It will take a bit of restructuring but it will get there.

anmatako: Now the logic should work correctly; tested locally and results agreed with expectation of camera models. The map used to keep track of existing camera models now gets populated only if the camera model is valid, thus cameras where the model cannot be found will not get grouped together.

PR #1147, opened 2021-02-24
Support more formats in model_converter

anmatako:
- model_converter changes:
  - Added support for 2 new formats (CAM and Recon3D)
  - New option `remove_image_prefix` allows de-registering of images before conversion
  - New option `skip_distortion` can be used to export the model without the distortion parameters (set to 0)

### DIFF
--- a/src/base/reconstruction.h
+++ b/src/base/reconstruction.h
@@ -272,13 +272,71 @@ class Reconstruction {
   void ImportPLY(const std::vector<PlyPoint>& ply_points);
 
   // Export to other data formats.
+
+  // Exports in NVM format http://ccwu.me/vsfm/doc.html#nvm. Only supports
+  // SIMPLE_RADIAL camera model when exporting distortion parameters. When
+  // skip_distortion == true it supports all camera models with the caveat that
+  // it's using the mean focal length which will be inaccurate for camera models
+  // with two focal length parameters.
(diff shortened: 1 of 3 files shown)


src/base/reconstruction.h
```diff
@@ -269,9 +272,12 @@ class Reconstruction {
   void ImportPLY(const std::vector<PlyPoint>& ply_points);
 
   // Export to other data formats.
-  bool ExportNVM(const std::string& path) const;
-  bool ExportBundler(const std::string& path,
-                     const std::string& list_path) const;
+  bool ExportNVM(const std::string& path, bool skip_distortion = false) const;
+  bool ExportCam(const std::string& path, bool skip_distortion = false) const;
```
ahojnnes, line 276: What is the "cam" format? What is the "Recon3D" format?

Maybe we can add a one-liner to each method here, e.g.:
```
// VisualSFM data format.
bool ExportNVM(const std::string& path, bool skip_distortion = false) const;
// Bundler data format.
bool ExportBundler(const std::string& path,	  bool ExportCam(const std::string& path, bool skip_distortion = false) const;
etc.
```

anmatako: Added detailed descriptions for new and existing formats. Let me know if that looks OK now.

ahojnnes: I made some minor modifications. skip_distortion also ignores distortion effects and is not just a problem with the focal length.

PR #1129, opened 2021-02-22
Add PatchMatchNet module for MVS and calculation of normals from depth

anmatako:
This work mainly integrated PatchMatchNet functionality in colamp using a TorchScript pre-trained module. Additionally it introduces functionality to calculate normal maps from depth maps since PatchMatchNet evaluation does not create normal maps as part of its process. More details about the changes:
- Colmap can compile with Torch support to enable PatchMatchNet. For this the pre-compiled LitTorch library needs to be downloaded from here https://pytorch.org/ on the desired configuration for GPU or CPU-only and the archive extracted under `<colmap-root>/lib/` thus creating a `libtorch` subfolder. Then CMake should be able to find the dependency and set the correct compilation flags.
- PatchMatchNet can now be enabled from `patch_match_stereo` by setting the `mvs_module_path` option to a valid TorchScript module. One such module is included as part of this PR in `<colmap-root>\mvs-modules\patchmatchnet-module.pt`
  - The TorchScript interface is fairly generic using the following input structure: `(images: List[Tensor], intrisics: Tensor, extrinsics: Tensor, depth_params: Tensor)` with the output being a tuple of `(depth: Tensor, confidence: Tensor)`. Thus any module that subscribes to that input/output format for forward evaluation can be used instead.
- Functionality of standard patch-match remains unchanged. There is now an inheritance structure used to select between standard and PMNet processing
- Normal maps are now not required for stereo fusion. If missing they will be calculated from the depth maps themselves. This is needed to accommodate PMNet processing that does not produce normal maps as part of the estimation work.
  - Note that use of calculated normal maps can be forced even for standard patch-match processing through the use of a new stereo fusion option `--StereoFusion.calculate_normals`.
- Confidence maps can now be used for stereo fusion and they are optional. If missing a confidence of 1 is assumed everywhere. This is also added to make use of the confidence maps that are created as part of PMNet estimation.
- New method for finding related images for fusion based on triangulation scoring is introduced and can be enabled with the option `--StereoFusion.use_triangulation_scoring`. This is included for parity with PatchMatchNet that has this method for finding related images instead of the colmap default. (useful for comparing results between colmap and Python)

PR #1148, opened 2021-02-24
Parallelize stereo fusion; needs pre-loading of entire workspace

anmatako:
Stereo fusion can now be done in parallel as long as the entire workspace gets pre-loaded. A new `NoCacheWorkspace` class can do that when initializing fusion. Parallel execution is the new default so to use the existing functionality with cache and sequential fusion `stereo_fusion` must be used with `--StereoFusion.use_cache 1`.

Also added options to:
- Use image masks that excludes pixels from the depth maps. Work similarly to masks in `ImageReaderOptions`
- Restrict the fused point cloud within a specified bounding box described from its two 3D corner points. The bounding box is in plan text format with space separated numeric values for the corner points, e.g. `<x1> <y1> <z1> <x2> <y2> <z2>.

### DIFF
--- a/src/exe/colmap.cc
+++ b/src/exe/colmap.cc
@@ -44,6 +44,7 @@
 #include "feature/extraction.h"
 #include "feature/matching.h"
 #include "feature/utils.h"
+#include "mvs/fusion.h"
 #include "mvs/meshing.h"
 #include "mvs/patch_match.h"
 #include "retrieval/visual_index.h"
@@ -267,6 +268,7 @@ int RunStereoFuser(int argc, char** argv) {
   std::string pmvs_option_name = "option-all";
   std::string output_type = "PLY";
   std::string output_path;
+  std::string bbox_path;
 
   OptionManager options;
   options.AddRequiredOption("workspace_path", &workspace_path);
@@ -278,6 +280,7 @@ int RunStereoFuser(int argc, char** argv) {
   options.AddDefaultOption("output_type", &output_type,
                             "{BIN, TXT, PLY}");
   options.AddRequiredOption("output_path", &output_path);
+  options.AddDefaultOption("bbox_path", &bbox_path);
   options.AddStereoFusionOptions();
   options.Parse(argc, argv);
 
@@ -295,6 +298,20 @@ int RunStereoFuser(int argc, char** argv) {
                  "'photometric' and 'geometric'."
               << std::endl;
     return EXIT_FAILURE;
+  }
+
+  if (!bbox_path.empty()) {
+    std::ifstream file(bbox_path);
+    if (file.is_open()) {
+      auto& min_bound = options.stereo_fusion->bounds.first;
+      auto& max_bound = options.stereo_fusion->bounds.second;
+      file >> min_bound(0) >> min_bound(1) >> min_bound(2);
+      file >> max_bound(0) >> max_bound(1) >> max_bound(2);
+      file.close();
+    } else {
+      std::cout << "WARN: Invalid bounds path: \"" << bbox_path
+                << "\" - continuing without bounds check" << std::endl;
+    }
   }
 
   mvs::StereoFusion fuser(*options.stereo_fusion, workspace_path,
--- a/src/mvs/fusion.cc
+++ b/src/mvs/fusion.cc
@@ -83,6 +83,7 @@ int FindNextImage(const std::vector<std::vector<int>>& overlapping_images,
 void StereoFusionOptions::Print() const {
 #define PrintOption(option) std::cout << #option ": " << option << std::endl
   PrintHeading2("StereoFusion::Options");
+  PrintOption(mask_path);
   PrintOption(max_image_size);
   PrintOption(min_num_pixels);
   PrintOption(max_num_pixels);
@@ -91,7 +92,12 @@ void StereoFusionOptions::Print() const {
   PrintOption(max_depth_error);
   PrintOption(max_normal_error);
   PrintOption(check_num_images);
+  PrintOption(use_cache);
   PrintOption(cache_size);
+  const auto& min_bound = bounds.first.transpose().eval();
+  const auto& max_bound = bounds.second.transpose().eval();
+  PrintOption(min_bound);
+  PrintOption(max_bound);
 #undef PrintOption
 }
 
@@ -151,12 +157,27 @@ void StereoFusion::Run() {
   }
 
   workspace_options.max_image_size = options_.max_image_size;
+  // Saving calculated maps only when the option is set (useful for debugging)
+  // or when using cache to avoid potential multiple calculations of the same
+  // map.
+  workspace_options.save_calculated_maps =
+      options_.save_calculated_maps || options_.use_cache;
   workspace_options.image_as_rgb = true;
   workspace_options.cache_size = options_.cache_size;
   workspace_options.workspace_path = workspace_path_;
   workspace_options.workspace_format = workspace_format_;
   workspace_options.input_type = input_type_;
 
+  const auto image_names = ReadTextFileLines(JoinPaths(
+      workspace_path_, workspace_options.stereo_folder, "fusion.cfg"));
+  int num_threads = 1;
+  if (options_.use_cache) {
+    workspace_.reset(new Workspace(workspace_options));
+  } else {
+    workspace_.reset(new NoCacheWorkspace(workspace_options));
+    workspace_->Load(image_names);
+    num_threads = omp_get_max_threads();
+  }
   workspace_.reset(new Workspace(workspace_options));
 
   if (IsStopped()) {
@@ -176,6 +197,18 @@ void StereoFusion::Run() {
     overlapping_images_ = model.GetMaxOverlappingImagesFromPMVS();
   }
 
+  fusion_queue_.resize(num_threads);
+  fused_point_x_.resize(num_threads);
+  fused_point_y_.resize(num_threads);
+  fused_point_z_.resize(num_threads);
+  fused_point_nx_.resize(num_threads);
+  fused_point_ny_.resize(num_threads);
+  fused_point_nz_.resize(num_threads);
+  fused_point_r_.resize(num_threads);
+  fused_point_g_.resize(num_threads);
+  fused_point_b_.resize(num_threads);
+  fused_point_visibility_.resize(num_threads);
+
   used_images_.resize(model.images.size(), false);
   fused_images_.resize(model.images.size(), false);
   fused_pixel_masks_.resize(model.images.size());
@@ -185,8 +218,6 @@ void StereoFusion::Run() {
   inv_P_.resize(model.images.size());
   inv_R_.resize(model.images.size());
 
-  const auto image_names = ReadTextFileLines(JoinPaths(
-      workspace_path_, workspace_options.stereo_folder, "fusion.cfg"));
   for (const auto& image_name : image_names) {
     const int image_idx = model.GetImageIdx(image_name);
 
@@ -206,9 +237,7 @@ void StereoFusion::Run() {
 
     used_images_.at(image_idx) = true;
 
-    fused_pixel_masks_.at(image_idx) =
-        Mat<bool>(depth_map.GetWidth(), depth_map.GetHeight(), 1);
-    fused_pixel_masks_.at(image_idx).Fill(false);
+    InitFusedPixelMask(image_idx, depth_map.GetWidth(), depth_map.GetHeight());
 
     depth_map_sizes_.at(image_idx) =
         std::make_pair(depth_map.GetWidth(), depth_map.GetHeight());
@@ -235,6 +264,8 @@ void StereoFusion::Run() {
             .transpose();
   }
 
+  std::cout << "Starting fusion with " << num_threads << " threads"
+            << std::endl;
   size_t num_fused_images = 0;
   for (int image_idx = 0; image_idx >= 0;
        image_idx = internal::FindNextImage(overlapping_images_, used_images_,
@@ -246,27 +277,30 @@ void StereoFusion::Run() {
     Timer timer;
     timer.Start();
 
-    std::cout << StringPrintf("Fusing image [%d/%d]", num_fused_images + 1,
-                              model.images.size())
+    std::cout << StringPrintf("Fusing image [%d/%d] with index %d",
+                              num_fused_images + 1, model.images.size(),
+                              image_idx)
               << std::flush;
 
     const int width = depth_map_sizes_.at(image_idx).first;
     const int height = depth_map_sizes_.at(image_idx).second;
     const auto& fused_pixel_mask = fused_pixel_masks_.at(image_idx);
 
-    FusionData data;
-    data.image_idx = image_idx;
-    data.traversal_depth = 0;
-
-    for (data.row = 0; data.row < height; ++data.row) {
-      for (data.col = 0; data.col < width; ++data.col) {
-        if (fused_pixel_mask.Get(data.row, data.col)) {
+#pragma omp parallel for schedule(dynamic) num_threads(num_threads)
+    for (int row = 0; row < height; ++row) {
+      for (int col = 0; col < width; ++col) {
+        if (fused_pixel_mask.Get(row, col)) {
           continue;
         }
+        FusionData data;
+        data.image_idx = image_idx;
+        data.traversal_depth = 0;
+        data.row = row;
+        data.col = col;
 
-        fusion_queue_.push_back(data);
-
-        Fuse();
+        const int tid = omp_get_thread_num();
+        fusion_queue_[tid].push_back(std::move(data));
+        Fuse(tid);
       }
     }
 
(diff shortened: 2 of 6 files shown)


anmatako:
This will not build for now since it has dependencies on #1129

src/mvs/fusion.cc
```diff
@@ -292,31 +326,54 @@ void StereoFusion::Run() {
...
+        mask.GetPixel(col, row, &color);
+        fused_pixel_mask.Set(row, col, color.r == 0);
+      }
+    }
+  } else {
+    fused_pixel_mask.Fill(false);
+  }
+}
+
+void StereoFusion::Fuse(int tid) {
```
ahojnnes, line 352: tid->thread_id

src/mvs/fusion.cc
```diff
@@ -383,25 +440,35 @@ void StereoFusion::Fuse() {
     // Set the current pixel as visited.
     fused_pixel_mask.Set(row, col, true);
```
ahojnnes, line 441: You will end up with race conditions here, because you simultaneously read/write to the masks from multiple threads.

anmatako: Yes this is true, and ideally this should have been an atomic update, but unfortunately the OpenMP standard (at least on Windows) does not allow atomic operations with function calls, only simple updates like `x = val` or `x += val`, etc. Wrapping this in a critical section kills the performance since all threads get blocked instead of just the ones that may conflict so this was also a no-go from a performance standpoint.

I've let that race condition exist since all threads are only attempting to set the same value (i.e. make the mask value `true`) so they're not adversely impacting the result of fusion. Worst case, by my understanding, is that we reconsider a pixel that should have already been masked out as visited. This causes some potential duplication of work, but from a performance standpoint its practically negligible, and definitely much preferred to using a critical section.

If there is a better way to go about it I can update accordingly. If it so happens that atomic works better in Linux/Mac I can add the `#pragma omp atomic` with a specific guard to skip it on windows. Otherwise I would have to change the data type of the mask from `Mat<bool>` to `std::vector` or something similar so I can directly access the data without a need for a function call.

anmatako: I think I managed to make this work with `#pragma omp atomic` by accessing the underlying pointer of `Mat<bool>`. Of course I'll have to performance test it on my end to ensure there is no adverse performance impact.

src/mvs/fusion.cc
anmatako: Fixed

src/mvs/fusion.cc
anmatako: What I tried won't work either. Apparently `std::vector<bool>` does not have a `.data()` member, unlike other types of vectors. I managed to make it work by converting the mask to `Mat<int>` instead of `Mat<bool>` and doing this:
```
const int pixel_index = depth_map_sizes_.at(image_idx).first * row + col;
#pragma omp atomic
fused_pixel_mask.GetPtr()[pixel_index]++;
```

Note that I'm using increment and not assignment here since Windows only support OpenMP 2.0 and atomic only supports `+=`, `-=`, `++`, and `--`.

If you think that's a valid way of handling this then I'll go ahead and update the PR. Thankfully making this atomic does not impact performance for parallel execution, at least in the few datasets I tested.

src/exe/colmap.cc
```diff
@@ -267,6 +268,7 @@ int RunStereoFuser(int argc, char** argv) {
   std::string pmvs_option_name = "option-all";
   std::string output_type = "PLY";
   std::string output_path;
+  std::string bbox_path;
```
ahojnnes, line 271: The bounding box code here seems unrelated to the rest of the PR?

src/mvs/fusion.cc
```diff
@@ -246,27 +277,30 @@ void StereoFusion::Run() {
...
-        fusion_queue_.push_back(data);
-
-        Fuse();
+        FusionData data;
+        data.image_idx = image_idx;
+        data.traversal_depth = 0;
+        data.row = row;
+        data.col = col;
+
+        const int tid = omp_get_thread_num();
```
ahojnnes, line 301: thread_id

src/mvs/fusion.cc
```diff
@@ -381,27 +438,40 @@ void StereoFusion::Fuse() {
...
 
     // Set the current pixel as visited.
-    fused_pixel_mask.Set(row, col, true);
+    bool* mask_ptr = fused_pixel_mask.GetPtr();
+    const int width = depth_map_sizes_.at(image_idx).first;
+#pragma omp atomic
+    mask_ptr[row * width + col] = true;
+
+    // Pixels out of bounds are filtered
+    if (xyz(0) < options_.bounds.first(0) ||
```
ahojnnes, line 447: bounds -> bbox?

src/exe/colmap.cc
ahojnnes: Ignore my comment. I just saw that "bounds" is used further below.

src/mvs/fusion.cc
ahojnnes: I think your suggested fix doesn't resolve the actual problem causing potentially many duplicated fused points. I think this effect is exacerbated by the fact that you start spawning threads for each pixel row, i.e. nearby pixels will be processed by different threads, which are very likely to be fused into the same point.

I think we could minimize this effect by using some striding between different threads. For example:
```
const int kRowStride = max(1, height / num_threads);
#pragma omp parallel for schedule(dynamic) num_threads(num_threads)
for (int row_start = 0; row_start < height; row_start += kRowStride) {
    const int row_end = min(height, row_start + kRowStride);
    for (int row = row_start; row < row_end; ++row) {
        for (int col = 0; col < width; ++col) {
            ...
        }
    }
}
```
In this case, elements in the fusion queue of different threads are further apart and are less likely to race with each other.

I was not so much concerned about the atomic update, which will obviously be a problem for `bool`, which is typically implemented using bitshifting. We could switch to `Mat<char>` and just use `mask.Set(row, col, 1)`.

src/mvs/fusion.cc
```diff
@@ -246,27 +277,30 @@ void StereoFusion::Run() {
...
     const auto& fused_pixel_mask = fused_pixel_masks_.at(image_idx);
 
-    FusionData data;
-    data.image_idx = image_idx;
-    data.traversal_depth = 0;
-
-    for (data.row = 0; data.row < height; ++data.row) {
-      for (data.col = 0; data.col < width; ++data.col) {
-        if (fused_pixel_mask.Get(data.row, data.col)) {
+#pragma omp parallel for schedule(dynamic) num_threads(num_threads)
```
ahojnnes, line 289: Any OpenMP code needs to be wrapped in `#ifdef OPENMP_ENABLED`, because it's an optional dependency in Python. Alternatively and preferably, we could use the ThreadPool in `util/threading.h`. We could implement a

src/mvs/fusion.cc
```diff
@@ -151,12 +157,27 @@ void StereoFusion::Run() {
...
 
+  const auto image_names = ReadTextFileLines(JoinPaths(
+      workspace_path_, workspace_options.stereo_folder, "fusion.cfg"));
+  int num_threads = 1;
+  if (options_.use_cache) {
+    workspace_.reset(new Workspace(workspace_options));
+  } else {
+    workspace_.reset(new NoCacheWorkspace(workspace_options));
+    workspace_->Load(image_names);
+    num_threads = omp_get_max_threads();
```
ahojnnes, line 179: It would be nice to make the number of threads configurable in the options. It could be initialized to -1 and you use `GetEffectiveNumThreads` to determine the number of threads.

src/mvs/fusion.cc
anmatako: Yes, you are right, the current setup with dynamic scheduling does suffer from potential duplications. I think an easier approach could be to configure it with a batch size of 10 or more so that different threads start further apart; this avoids the need for manual management of offsets. I can experiment with a few different batch sizes and depth map sizes and see if there's some value that works well enough.

And yes doing `Mat<char>` for mask should work even without a specific guard for the atomic update, which will keep things simpler.

These two I can do and quickly update, but changing the format of fusion to work with the existing thread pool will take a while as it's not clear to me what kind of setup it will need to process the problems, and I don't want to make the approach too verbose. But, I'll definitely attempt this at some point.

src/mvs/fusion.cc
anmatako: Got it; I'll wrap it as such, and use your suggestion for an attempt to use the existing ThreadPool.

src/mvs/fusion.cc
anmatako: Yes, that's a good suggestion and an easy change to make :)

src/mvs/fusion.cc
anmatako: Will change

src/mvs/fusion.cc
anmatako: So I just tested the current setup (chunk size of 1) against the single threaded run and multi-threaded runs with chunk sizes of 10, 20, and 50 (that should be the same as setting a row stride).

Interestingly, the processing time goes slightly up when increasing the chunk size, and the reconstructed points are slightly less than the single threaded point clouds, with the exception of chunk size of 1. Below is a summary of the findings.

- Reference single threaded run: Runtime 11.5 min; points: 7,570,936

The comparison results show the relative number of points and the % change:
1. Chunk size 1 (current schedule): Runtime: 1.45 min; points: +20,568 (+0.27%)
2. Chunk size 10: Runtime: 1.50 min; points: -13353 (-0.18%)
3. Chunk size 20: Runtime: 1.66 min; points: -6130 (-0.08%)
4. Chunk size 50: Runtime: 2.09 min; points: -2552 (-0.03%)

Those differences are really small, thankfully, and in light of that I think it would make sense to configure the scheduling with a chunk size of 10.

src/mvs/fusion.cc
anmatako: Quick question on wrapping with `#ifdef OPENMP_ENABLED`. I thought that the `#pragma`s are ignored in cases where they cannot be resolved (i.e. OpenMP missing and finding a `#pragma omp` directive). Is this not the case with the colmap setup?

src/mvs/fusion.cc
ahojnnes: Chunk size of 10 seems like a good trade off. Thanks!

src/mvs/fusion.cc
ahojnnes: It will certainly trigger an unknown pragma warning. If you then also treat warnings as errors, compilation will fail.

src/mvs/fusion.cc
ahojnnes: No need to do the thread pool - was just a suggestion.

anmatako: Seeing how colmap does not natively use OpenMP in favor of the ThreadPool, I'm afraid that a solution with OpenMP would pollute the code with quite a few guards checking if OpenMP is available.

I think I have a clear idea now on how to go about this with the ThreadPool, so I'll give it a try and update the PR. Should be cleaner overall with that approach.

anmatako:
Replaced OpenMP with ThreadPool for both fusion and workspace. Also removed all the members from Fusion that were only really used internally in the `Fuse` function. Now there's much less of the ugliness with the `thread_id` indexing. Performance took a minor hit, but overall it's not significant.

src/exe/colmap.cc
```diff
@@ -297,6 +300,20 @@ int RunStereoFuser(int argc, char** argv) {
...
   }
 
+  if (!bbox_path.empty()) {
+    std::ifstream file(bbox_path);
+    if (file.is_open()) {
+      auto& min_bound = options.stereo_fusion->bounding_box.first;
+      auto& max_bound = options.stereo_fusion->bounding_box.second;
+      file >> min_bound(0) >> min_bound(1) >> min_bound(2);
+      file >> max_bound(0) >> max_bound(1) >> max_bound(2);
+      file.close();
```
ahojnnes, line 310: .close() is not necessary as the destructor will do the job.

src/mvs/workspace.h
```diff
@@ -42,12 +42,17 @@
 namespace colmap {
 namespace mvs {
 
+class ConfidenceMap;
+
 class Workspace {
```
ahojnnes, line 47: Maybe it's better to called to `NoCache` version `Workspace` and the cached one as `CachedWorkspace`.

src/mvs/workspace.cc
```diff
@@ -234,5 +235,88 @@ void ImportPMVSWorkspace(const Workspace& workspace,
...
 }
 
+void NoCacheWorkspace::Load(const std::vector<std::string>& image_names) {
+  const size_t num_images = model_.images.size();
+  bitmaps_.resize(num_images);
+  depth_maps_.resize(num_images);
+  confidence_maps_.resize(num_images);
+  normal_maps_.resize(num_images);
+
+  auto LoadWorkspaceData = [&, this](int image_idx) {
```
ahojnnes, line 245: const int image_idx (in COLMAP we use const even for basic types)

src/mvs/workspace.h
anmatako: OK sounds good, I'll flip the inheritance structure around.

src/mvs/workspace.cc
```diff
@@ -234,5 +235,88 @@ void ImportPMVSWorkspace(const Workspace& workspace,
...
+      // Estimate normal map from depth when the map is not given
+      normal_maps_[image_idx]->EstimateFromDepth(
+          *depth_maps_[image_idx], model_.images.at(image_idx).GetInvK());
+      if (options_.save_calculated_maps) {
+        normal_maps_[image_idx]->Write(normal_map_path);
+      }
+    }
+  };
+
+  int num_threads = GetEffectiveNumThreads(options_.num_threads);
```
ahojnnes, line 297: const

src/mvs/workspace.cc
```diff
@@ -234,5 +235,88 @@ void ImportPMVSWorkspace(const Workspace& workspace,
...
+  ThreadPool thread_pool(num_threads);
+  Timer timer;
+  timer.Start();
+
+  std::cout << StringPrintf("Loading workspace data with %d threads...",
+                            num_threads)
+            << std::endl;
+  for (int i = 0; i < image_names.size(); ++i) {
```
ahojnnes, line 305: size_t i

src/mvs/fusion.h
```diff
@@ -120,41 +137,36 @@ class StereoFusion : public Thread {
...
 
   struct FusionData {
-    int image_idx = kInvalidImageId;
-    int row = 0;
-    int col = 0;
-    int traversal_depth = -1;
+    int image_idx;
+    int row;
+    int col;
+    int traversal_depth;
```
ahojnnes, line 153: Can you restore the previous default values here? It's generally good practice to keep these to avoid uninitialized memory, I think.

src/mvs/fusion.cc
```diff
@@ -418,57 +500,52 @@ void StereoFusion::Fuse() {
         continue;
       }
 
-      next_data.image_idx = next_image_idx;
-
       const Eigen::Vector3f next_proj =
           P_.at(next_image_idx) * xyz.homogeneous();
-      next_data.col = static_cast<int>(std::round(next_proj(0) / next_proj(2)));
-      next_data.row = static_cast<int>(std::round(next_proj(1) / next_proj(2)));
+      int next_col = static_cast<int>(std::round(next_proj(0) / next_proj(2)));
```
ahojnnes, line 505: const and also next_row

src/mvs/fusion.cc
```diff
@@ -292,35 +341,61 @@ void StereoFusion::Run() {
...
+        mask.GetPixel(col, row, &color);
+        fused_pixel_mask.Set(row, col, color.r == 0 ? 1 : 0);
+      }
+    }
+  } else {
+    fused_pixel_mask.Fill(0);
+  }
+}
+
+void StereoFusion::Fuse(int thread_id, int image_idx, int row, int col) {
```
ahojnnes, line 367: please make the parameters const

src/mvs/fusion.cc
```diff
@@ -235,7 +256,28 @@ void StereoFusion::Run() {
             .transpose();
   }
 
+  std::cout << StringPrintf("Starting fusion with %d threads", num_threads)
+            << std::endl;
+  ThreadPool thread_pool(num_threads);
+
+  const int kRowStride = 10;
```
ahojnnes, line 263: It would be great to add a short explanation, why we use a stride here.

src/mvs/fusion.cc
```diff
@@ -246,40 +288,47 @@ void StereoFusion::Run() {
...
-                              fused_points_.size())
+                              total_fused_points)
               << std::endl;
   }
 
-  fused_points_.shrink_to_fit();
-  fused_points_visibility_.shrink_to_fit();
+  fused_points_.reserve(total_fused_points);
+  fused_points_visibility_.reserve(total_fused_points);
+  for (size_t task_id = 0; task_id < task_fused_points_.size(); ++task_id) {
```
ahojnnes, line 320: task_id -> thread_id for consistency

src/mvs/fusion.cc
ahojnnes: Looks great, thanks.

ahojnnes:
@anmatako You will have to merge with master. note that I modularized exe/colmap.cc into individual source files, since it grew way beyond what we should put into a single source file.

src/mvs/fusion.h
anmatako: Sounds good; I only removed them because now we never create a default `FusionData` struct and we only use it with the constructor through `emplace_back`

anmatako:
> @anmatako You will have to merge with master. note that I modularized exe/colmap.cc into individual source files, since it grew way beyond what we should put into a single source file.

yes now that we are getting closer with these changes I'll do a merge from master and also resolve some other issues with confidence maps and calculated normals that were introduced from the PatchMatchNet PR.

I'll do a good cleanup and make sure this builds standalone.

src/mvs/workspace.h
```diff
@@ -63,34 +67,78 @@ class Workspace {
...
+  inline const Options& GetOptions() const { return options_; }
 
-  const Options& GetOptions() const;
+  inline const Model& GetModel() const { return model_; }
 
-  const Model& GetModel() const;
-  const Bitmap& GetBitmap(const int image_idx);
-  const DepthMap& GetDepthMap(const int image_idx);
-  const NormalMap& GetNormalMap(const int image_idx);
+  virtual inline const Bitmap& GetBitmap(const int image_idx) {
```
ahojnnes, line 77: I suggest moving these method definitions to the .cc file. Since it's a virtual method, it won't be able to inline these anyway.

ahojnnes:
@anmatako Thanks, the PR looks great now. I trust you checked correctness that these changes lead to equivalent (+- small changes) fusion results as before? I left two minor comments, then I am happy to merge. Thanks!

anmatako:
> @anmatako Thanks, the PR looks great now. I trust you checked correctness that these changes lead to equivalent (+- small changes) fusion results as before? I left two minor comments, then I am happy to merge. Thanks!

Thanks! I've tested quite a lot in our pipelines and the differences are really minor compared to single threaded fusion; less than 0.1% in the number fused points and visually nothing stood out as an obvious outlier.

ahojnnes:
Thanks very much. This is a great improvement!

PR #1166, opened 2021-03-09
Issue 1033 prior intrinsic for multiple cameras

### LINKED ISSUE #1033: Pass intrinsic calibration as prior information
Is there a way to pass colmap an intrinsic calibration for specific images and even maybe prevent optimization during BA?

DaniilSNikulin:
fixes #1033 

Extended functionality by allowing users to specify a prior intrinsics for multiple cameras.
Also extended documentation for new functionality.

API was changed, but saved backward compatibility.

Code was hand tested,  but unit tests are not exists. I don't now how create tests for CLI module.

### DIFF
--- a/doc/cli.rst
+++ b/doc/cli.rst
@@ -158,8 +158,10 @@ the available options, e.g.::
           --database_path arg
           --image_path arg
           --image_list_path arg
+          --camera_params_per_folder_path arg
           --ImageReader.camera_model arg (=SIMPLE_RADIAL)
           --ImageReader.single_camera arg (=0)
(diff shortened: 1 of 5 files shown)


MartinEthier:
@DaniilSNikulin What's the current status of this? It would be a useful change seeing as the current way to do this involves directly editing the database.

DaniilSNikulin:
Hi @MartinEthier 
I wasn't working with Colmap for several years.
I guess current PR is not actual anymore. (at least because of merge conflicts)

And also I don't think, that I can actualize it, because I lost my competitions with Colmap and afraid introduce error.

PR #1167, opened 2021-03-09
Add new auto_mapper command

anmatako:
The new `auto_mapper` command is a convenience command that combines mapper, triangulator, and bundle adjuster as needed based on the provided options. It can do the following:
1. Standard mapper with optional point triangulator and rig BA if a rig config path is provided
2. Hierarchical mapper followed by point triangulator and rig or standard BA
3. Read existing reconstruction followed by point triangulator and rig or standard BA

Also, added new option `skip_color_extraction` in both `auto_mapper` and `point_triangulator` that avoids the call to `Reconstruction::ExtractColorsForAllImages` after triangulation since that's a pretty slow part of the process (single-threaded only) and is not necessary when we don't care about colors of the sparse point clouds


src/exe/sfm.cc
```diff
@@ -47,6 +47,293 @@
...
+//            frame001.png
+//            frame002.png
+//            ...
+//        right2_image/...
+//            frame000.png
+//            frame001.png
+//            frame002.png
+//            ...
+//
+std::vector<CameraRig> ReadCameraRigConfig(const std::string& rig_config_path,
```
anmatako, line 131: No code changes here; just moved it from below to group all helper functions together.

ahojnnes:
I like the skip_color_extraction but I am not 100% sure about the logic in the auto_mapper command to be honest. I don't want to encourage people to rely too much on the hierarchical_mapper, which will not work too well when the input consists of internet images, etc.

anmatako:
> I like the skip_color_extraction but I am not 100% sure about the logic in the auto_mapper command to be honest. I don't want to encourage people to rely too much on the hierarchical_mapper, which will not work too well when the input consists of internet images, etc.

I can see the concern about relying too much on auto_mapper. Has worked well enough for me so far since our datasets are more cohesive, but that may not generalize well enough to be broadly used.

I can leave the refactoring and the new `skip_color_extraction` for point triangulator and remove the `auto_mapper` command from this PR. Another alternative would be to make the default hierarchical mapper threshold really high (for the auto_mapper command only) to prevent accidental usage, such that `auto_mapper` would default to `mapper` unless options are explicitly set to enable the `hierarchical_mapper`.

Either approach is fine by me.

PR #1169, opened 2021-03-09
Refactoring and new functionality in Reconstruction class

anmatako:
New functionality:
- `ComputeBoundsAndCentroid`: Computes the bounding box (two corner points) of the reconstruction and the 3D point centroid based on the min and max percentiles. Similarly to how `Normalize` works.
- `ComputeCentroid`: Computes the centroid based on the min and max percentiles.
- `ComputeBoundingBox`: Computes the bounding box as two corner points based on the min and max percentiles.
- `Crop`: Crops the reconstruction based on the given bounding box. It does not modify the current reconstruction; the cropped result is an output by reference instead. The output reconstruction only includes the 3D points within the bounding box. It includes all images and cameras of the original reconstruction, but only the images that see 3D points inside the bounding box are registered.

Modified functionality:
- `Normalize`: Works same as before but now leverages the new `ComputeBoundsAndCentroid` function.
- `Align` and `AlignRobust`: These are now templated on a bool flag that determines whether or not the scale is estimated when estimating the similarity transform. Also they can export the transform used in alignment since it's now provided as input by reference.


anmatako:
This new functionality is leveraged in additional model manipulation commands that will be part of subsequent PR.

src/base/reconstruction.h
```diff
@@ -192,15 +207,19 @@ class Reconstruction {
...
+             const int min_common_images, SimilarityTransform3& tform);
 
   // Robust alignment using RANSAC.
+  template <bool kEstimateScale = true>
   bool AlignRobust(const std::vector<std::string>& image_names,
                    const std::vector<Eigen::Vector3d>& locations,
                    const int min_common_images,
-                   const RANSACOptions& ransac_options);
+                   const RANSACOptions& ransac_options,
+                   SimilarityTransform3& tform);
```
anmatako, line 221: I can convert the `tform` input to pointer with `nullptr` default if that's more compatible with the existing colmap style.

src/base/reconstruction.h
```diff
@@ -178,9 +181,21 @@ class Reconstruction {
...
+  std::pair<Eigen::Vector3d, Eigen::Vector3d> ComputeBoundingBox(double p0 = 0.0, double p1 = 1.0) const;
+
   // Apply the 3D similarity transformation to all images and points.
   void Transform(const SimilarityTransform3& tform);
 
+  // Creates a cropped reconstruction using the input bounds as corner points
+  // of the bounding box containing the included 3D points of the new reconstruction.
+  // Only the cameras and images of the included points are registered.
+  bool Crop(const std::pair<Eigen::Vector3d, Eigen::Vector3d>& boundary,
+            Reconstruction& reconstruction) const;
```
anmatako, line 197: Not sure if getting the output by reference is better than simply returning the cropped reconstruction. Was trying to avoid unnecessary copies of large objects, but maybe it's better to change this after all.

Open to suggestions on how to change that if needed.

src/base/reconstruction.cc
```diff
@@ -427,6 +431,34 @@ void Reconstruction::Transform(const SimilarityTransform3& tform) {
   }
 }
 
+bool Reconstruction::Crop(
+    const std::pair<Eigen::Vector3d, Eigen::Vector3d>& boundary,
```
ahojnnes, line 435: bbox

src/base/reconstruction.cc
```diff
@@ -335,26 +334,57 @@ void Reconstruction::Normalize(const double extent, const double p0,
...
+    double p0, double p1) const {
+  auto bound = ComputeBoundsAndCentroid(p0, p1, false);
+  return std::make_pair(std::get<0>(bound), std::get<1>(bound));
+}
+
+std::tuple<Eigen::Vector3d, Eigen::Vector3d, Eigen::Vector3d>
+Reconstruction::ComputeBoundsAndCentroid(double p0, double p1,
+                                         bool use_images) const {
+
+  size_t num_elements = use_images ? reg_image_ids_.size() : points3D_.size();
```
ahojnnes, line 368: const

src/base/reconstruction.cc
```diff
@@ -335,26 +334,57 @@ void Reconstruction::Normalize(const double extent, const double p0,
...
+  if (num_elements == 0)
+  {
```
ahojnnes, line 370: clang format

src/base/reconstruction.cc
```diff
@@ -335,26 +334,57 @@ void Reconstruction::Normalize(const double extent, const double p0,
...
+}
+
+std::pair<Eigen::Vector3d, Eigen::Vector3d> Reconstruction::ComputeBoundingBox(
+    double p0, double p1) const {
+  auto bound = ComputeBoundsAndCentroid(p0, p1, false);
+  return std::make_pair(std::get<0>(bound), std::get<1>(bound));
+}
+
+std::tuple<Eigen::Vector3d, Eigen::Vector3d, Eigen::Vector3d>
+Reconstruction::ComputeBoundsAndCentroid(double p0, double p1,
```
ahojnnes, line 365: const params

src/base/reconstruction.cc
```diff
@@ -335,26 +334,57 @@ void Reconstruction::Normalize(const double extent, const double p0,
...
+                             -scale * std::get<2>(bound));
+  Transform(tform);
+}
+
+Eigen::Vector3d Reconstruction::ComputeCentroid(double p0, double p1) const {
+  return std::get<2>(ComputeBoundsAndCentroid(p0, p1, false));
+}
+
+std::pair<Eigen::Vector3d, Eigen::Vector3d> Reconstruction::ComputeBoundingBox(
+    double p0, double p1) const {
```
ahojnnes, line 359: const params

src/base/reconstruction.cc
```diff
@@ -335,26 +334,57 @@ void Reconstruction::Normalize(const double extent, const double p0,
...
+  } else {
+    scale = extent / old_extent;
+  }
+
+  SimilarityTransform3 tform(scale, ComposeIdentityQuaternion(),
+                             -scale * std::get<2>(bound));
+  Transform(tform);
+}
+
+Eigen::Vector3d Reconstruction::ComputeCentroid(double p0, double p1) const {
```
ahojnnes, line 354: const params

src/base/reconstruction.h
ahojnnes: You can directly return a Reconstruction object. Copy elision will ensure that there is no overhead when returning the obejct.

anmatako:
I'll add some unit tests for the new functionality in the next update and it should be good then.

src/base/reconstruction_test.cc
```diff
@@ -366,12 +362,6 @@ BOOST_AUTO_TEST_CASE(TestCrop) {
   BOOST_CHECK(recon2.IsImageRegistered(1));
   BOOST_CHECK(recon2.IsImageRegistered(2));
   BOOST_CHECK(!recon2.IsImageRegistered(3));
-  BOOST_CHECK_LE(
```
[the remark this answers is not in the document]
anmatako: It was not a stable check since the points are in an unordered map and there is no guarantee that they will be included in the same order as they appeared in the original reconstruction. The test was passing on Windows and failing on all other platforms due to this non-deterministic behavior.

I decided to remove this test instead of writing more tedious code along the lines of ensuring that at least one of the included points matches the target and also it's not double counted.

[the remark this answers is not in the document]
ahojnnes: Okay, thanks.

Issue #1161, opened 2021-03-08
Memory leak because of using thread_local std::mt19937* PRNG


whuaegeanse:
Hi @ahojnnes 

[PRNG](https://github.com/colmap/colmap/blob/dev/src/util/random.h#L44) is a thread_local variable, and there is no memory leak problem when SetPRNGSeed is not called. Otherwise, if the SetPRNGSeed function is called in the thread, the PRNG will cause a memory leak because no function to release the PRNG's memory is called.

Here I refer to the thread storage duration mentioned in cppreference.
`thread storage duration. The storage for the object is allocated when the thread begins and deallocated when the thread ends. Each thread has its own instance of the object. Only objects declared thread_local have this storage duration. thread_local can appear together with static or extern to adjust linkage. See Non-local variables and Static local variables for details on initialization of objects with this storage duration.`[cppreference](https://en.cppreference.com/w/cpp/language/storage_duration)

Maybe using std::unique_ptr can help us to delay the creation of PRNG and avoid memory leaks. Because at the end of the thread, the destructor of std::unique_ptr will be called, and the memory corresponding to the PRNG will be released.

If you think this suggestion is reasonable, I will be happy to fix this.

PR #1170, opened 2021-03-10
Fix the memory leak caused by not releasing the memory of the PRNG at the end of the thread

whuaegeanse:
Fix the memory leak caused by not releasing the memory of the PRNG at the end of the thread as mentioned by #1161.


ahojnnes:
Thank you for this fix. I think there is a way to avoid the compiler errors. Please have a look at: https://github.com/whuaegeanse/colmap/pull/1. Let's see what the CI build does but we can safely drop support for older versions of GCC and other compilers at this point. Several years have passed since I originally wrote this code.

whuaegeanse:
@ahojnnes  There seems to be no problem. But I have another question, does the extern variable PRNG need to be initialized to nullptr? There seems to be an implicit conversion and copy construction. Is there a certain overhead?

ahojnnes:
No, indeed, there is no need to initialize to the nullptr. We can drop it. Thanks.

PR #1177, opened 2021-03-14
Update model_aligner functionality

anmatako:
`model_aligner` has been extended with the following additional functionality:
1. Can use either an image list (existing method) or the location priors from the database (new functionality) for alignment. Using the new `database_path` option will use the database instead of an image list
2. Can optionally skip the scale estimation during alignment. This is controlled by the new `estimate_scale` flag.
3. Can optionally export the alignment transform in a text file as a 4x4 row-major matrix. This is enabled by providing a valid path for the output transform file through the new `transform_path` option.
4. Can perform 5 types of alignment: "plane", "ecef", "enu", "enu-unscaled", and "custom".  

The different alignment types work as:
1. custom: This is the same as the existing method where the image locations are used as-is for alignment
2. plane: This does not use any reference images. It simply aligns the model to the principal plane of the 3D points (estimated with SVD), such that the x-y plane aligns with the principal plane
3. ecef: This mode assumes that the image locations (from file or database) are provided as GPS Lat/Lon/Alt coordinates (WGS84) and then converts them to the ECEF coordinate systems and aligns the model. This results in a model with global orientation expressed in physical units (meters).
4. enu: This mode also assumes that the image locations are provided as GPS coordinates. It aligns the model with the true ENU ground plane at the model's 3D points centroid, such that x points East, y points North, and z points Up. The model scale is in physical units (meters).
5. enu-unscaled: Does the same alignment as the "enu" mode with the only difference that the model's original scale is preserved. This is useful for model display or to avoid numerical instabilities for large extent models.


src/exe/model.cc
```diff
@@ -40,6 +41,99 @@
 namespace colmap {
 namespace {
 
+// Aligns the reconstruction to the plane defined by running PCA on the 3D
+// points. The model centroid is at the origin of the new coordinate system
+// and the X axis is the first principal component with the Y axis being the
+// second principal component
+void AlignToPrincipalPlane(Reconstruction& recon, SimilarityTransform3& tform) {
```
ahojnnes, line 48: pointers for modified output parameters

src/exe/model.cc
```diff
@@ -40,6 +41,99 @@
...
+  // Perform SVD on the 3D points to estimate the ground plane basis
+  auto centroid = recon.ComputeCentroid();
```
ahojnnes, line 50: const Eigen::Vector3d

src/exe/model.cc
```diff
@@ -40,6 +41,99 @@
...
+  Eigen::MatrixXd points(3, recon.NumPoints3D());
+  int pidx = 0;
+  for (const auto point : recon.Points3D()) {
```
ahojnnes, line 53: auto&

src/exe/model.cc
```diff
@@ -40,6 +41,99 @@
...
+    points.col(pidx++) = point.second.XYZ() - centroid;
+  }
+  Eigen::Matrix3d basis =
```
ahojnnes, line 56: const

src/exe/model.cc
```diff
@@ -40,6 +41,99 @@
...
+  Eigen::Matrix3d rot_mat;
+  rot_mat << basis.col(0), basis.col(1), basis.col(0).cross(basis.col(1));
+  rot_mat.transposeInPlace();
+
+  tform = SimilarityTransform3(1.0, RotationMatrixToQuaternion(rot_mat),
+                               -rot_mat * centroid);
+
+  // if camera plane ends up below ground then flip basis vectors and create new
+  // transform
+  class Image* test_img = new class Image(recon.Images().begin()->second);
```
ahojnnes, line 67: memory leak 

I also don't see a reason for not creating the object on the stack?

src/exe/model.cc
```diff
@@ -40,6 +41,99 @@
...
+  recon.Transform(tform);
+}
+
+// Aligns the reconstruction to the local ENU plane orientation. Rotates the
+// reconstruction such that the x-y plane aligns with the ENU tangent plane at
+// the point cloud centroid and translates the origin to the centroid.
+// If unscaled == true, then the original scale of the model remains unchanged.
+void AlignToENUPlane(Reconstruction& recon, SimilarityTransform3& tform,
+                     bool unscaled) {
+  Eigen::Vector3d centroid = recon.ComputeCentroid();
```
ahojnnes, line 85: const

src/exe/model.cc
```diff
@@ -40,6 +41,99 @@
...
+  double scale = unscaled ? 1.0 / tform.Scale() : 1.0;
+  tform = SimilarityTransform3(scale, RotationMatrixToQuaternion(rot_mat),
+                               -(scale * rot_mat) * centroid);
+  recon.Transform(tform);
+}
+
+void ReadFileCameraLocations(const std::string& ref_images_path,
+                             std::vector<std::string>& ref_image_names,
+                             std::vector<Eigen::Vector3d>& ref_locations) {
+  auto lines = ReadTextFileLines(ref_images_path);
```
ahojnnes, line 109: const

src/exe/model.cc
```diff
@@ -40,6 +41,99 @@
...
+  for (const auto& line : lines) {
+    std::stringstream line_parser(line);
+    std::string image_name = "";
```
ahojnnes, line 112: Remove `= ""`

src/exe/model.cc
ahojnnes, line 48: Could be nice to move this function to estimators/coordinate_frame.h and also add a basic unit test.

src/exe/model.cc
```diff
@@ -40,6 +41,99 @@
...
+  }
+
+  recon.Transform(tform);
+}
+
+// Aligns the reconstruction to the local ENU plane orientation. Rotates the
+// reconstruction such that the x-y plane aligns with the ENU tangent plane at
+// the point cloud centroid and translates the origin to the centroid.
+// If unscaled == true, then the original scale of the model remains unchanged.
+void AlignToENUPlane(Reconstruction& recon, SimilarityTransform3& tform,
```
ahojnnes, line 83: Same here, could be nice to move this function to estimators/coordinate_frame.h and also add a basic unit test.

ahojnnes (changes requested):

src/exe/model.cc
anmatako: I think I was having a weird compilation issue here. I changed it to `Image test_img = recon.Images().begin()->second` and now it works fine. Thanks for pointing this out.

ahojnnes:
Thanks Antonios for the fixes, LGTM.

PR #1178, opened 2021-03-14
Add new model_transformer command

anmatako:
The new command reads a sparse or dense model and a similarity transform from file and then writes the transformed model in the same format.


src/base/similarity_transform.cc
```diff
@@ -194,6 +196,19 @@ SimilarityTransform3::SimilarityTransform3(const double scale,
   transform_.matrix() = matrix;
 }
 
+SimilarityTransform3::SimilarityTransform3(const std::string& path) {
```
ahojnnes, line 199: Seems like this is redundant to the file reading functionality from the model_cropper/splitter PR?

anmatako: Yes it is. I did not want to make dependent PRs so I included the change here as well. I will change to the static `FromFile` as agreed in the other PR #1179 and resolve the conflict (if it appears) in whichever gets merged last.

ahojnnes:
This will need a merge with master. Thanks.

PR #1179, opened 2021-03-14
Add new model_cropper and model_splitter commands

anmatako:
`model_cropping`: Reads model from `input_path`, crops using the values provided in `boundary` and writes cropped model in `output_path`. It can support GPS coordinates (WGS84 lat/lon/alt) in the boundary descriptor if the model transform to ECEF is provided in the `gps_transform_path` option. The `boundary` option should be in one of two formats:
1. x1,y1,z1,x2,y2,z2: 6 boundary coordinates defining the corner points of the bounding box. Either in model coordinates or GPS coordinates if the `gps_transform_path` is provided.
2. p0,p1: 2 percentiles (min, max) of the 3D points extent, similarly to how percentiles are used in `Reconstruction::Normalize`

`model_splitter`: Reads model from `input_path`, splits the model in parts according to the `split_type` and `split_params` options, and writes the model parts as sub-folders in `output_path`. The valid options for splitting are:
1. `split_type="parts"`, `split_params="n1,n2,n3"`: Splits the model in `n1*n2*n3` equal parts, each with extent `ex/n1, ey/n2, yz/n3`.
2. `split_type="extent"`, `split_params="ex,ey,ez"`: Splits the model in equal parts such that the extent of each part does not exceed (ex, ey, ez) in the (x, y, z) axes respectively, e.g. for a model with extent (11, 12, 4) using split_params="5,5,5" will lead in a split of 9 parts, each with extent of (3.3333, 4, 4). The extent can be expressed in physical units (meters) if the `gps_transform_path` is provided.
3. `split_type="tiles"`, `split_params="<path to tiles_text_file>"`: Splits the model according to the entries in `tiles_text_file`; this file contains entries (one per line) of the form: `TILE_KEY X1 Y1 Z1 X2 Y2 Z2`, where TILE_KEY is used as a name for the subfolder where the output is stored and the 6 coordinates that follow specify the bounding box for cropping. The coordinates can be in model units or GPS coordinates if `gps_transform_path` is provided.

Additional options can be used to:
- Reject models from the output based on number of registered images (`min_reg_images`), the number of 3D points (`min_num_points`) and the proportion of output area covered by the model (`min_area_ratio`).
- Create overlap between the output parts (`overlap_ratio`) expressed as a percentage of the extent.


src/base/similarity_transform.cc
```diff
@@ -185,6 +187,19 @@ SimilarityTransform3::SimilarityTransform3(
     const Eigen::Transform<double, 3, Eigen::Affine>& transform)
     : transform_(transform) {}
 
+SimilarityTransform3::SimilarityTransform3(const std::string& path) {
```
ahojnnes, line 190: Instead create a static method "FromFile" and add a unit test for it.

src/exe/colmap.cc
```diff
@@ -124,9 +124,11 @@ int main(int argc, char** argv) {
   commands.emplace_back("model_analyzer", &RunModelAnalyzer);
   commands.emplace_back("model_comparer", &RunModelComparer);
   commands.emplace_back("model_converter", &RunModelConverter);
+  commands.emplace_back("model_cropping", &RunModelCropping);
```
ahojnnes, line 127: model_cropper

src/exe/model.cc
```diff
@@ -459,4 +606,188 @@ int RunModelOrientationAligner(int argc, char** argv) {
...
+
+  PrintHeading2("Applying split and writing reconstructions");
+  const size_t num_parts = bounds.size();
+  std::cout << StringPrintf(" => Splitting to %d parts", num_parts)
+            << std::endl;
+
+
+  const bool use_tile_keys = split_type == "tiles";
+
+  auto SplitRecon = [&](int idx) {
```
ahojnnes, line 745: const int

ahojnnes (changes requested):
It will be good to document new commands here as all: https://colmap.github.io/cli.html#commands.

anmatako:
> It will be good to document new commands here as all: https://colmap.github.io/cli.html#commands.

Yes I'll make sure to add them there for better reference. Also where is a good place to add example usage for the new commands and the ones I'm extending?

PR #1234, opened 2021-06-12
Modify RefineAbsolutePose to return covariance

oldshuren:
Covariances are important for determining the confidence of localization of the query images from colmap models. So it is very useful to return the covariance values.


ahojnnes:
This feature was added through another PR. Thanks.

Issue #1239, opened 2021-06-20
COLMAP model aligner not working properly


andreas-at-talon:
Hello. COLMAP has been working perfectly on old VM but we are trying to install on new VM. Every part from feature extraction to dense reconstruction works perfectly on new VM apart from model aligner. On the old VM we have generated a .ply Point Cloud that is perfect and perfectly alligned. However when I run same script with colmap Linux commands everything works apart from model aligner. I can see this because of comparing final .plys between OLD and NEW VM, seeing the differences is only the orientation and position of the two point clouds. Something weird is happening with colmap model aligner command. I can also see this by comparing the images.bin of the OLD and NEW georefenced directory by looking at the tvec of each image.

Another thing to note is the output of 'colmap aligner' command. On OLD VM when it worked perfectly, output of model aligner looked like this:

Aligning reconstruction
-----------------------
 => Using 86 reference images
 => Alignment succeeded
 => Alignment error: 0.353243 (mean), 0.318907 (median)

However when I run same command on NEW VM we get this:

Aligning reconstruction to principal plane
------------------------------------------
 => Alignment succeeded

This is either a problem of not being able to find georefenced_images.txt, or having an issue reading the .txt file. Even if I take .txt file from OLD VM to NEW VM and run colmap aligner it returns same problem. I know for a fact .txt file is in correct format with no difference in image names, so I am not sure what the problem is, and output of colmap model aligner does not help with any info. Just to make sure, I have even same COLMAP version and commit as old VM and same gcc g++ compilers as well. Any insight into this would be extremely appreciated.

Another note:
Just to test, I made georefenced_images.txt file completely empty, and ran same command, and it produecs same output as NEW VM above which is weird. I have even set .txt file permissions to 0777 so there is no permission error either. 

Thank you.

tsattler:
It seems that the latest version of Colmap introduced different types of alignment (see https://github.com/colmap/colmap/blob/e25479aaced4aef517c109c61e5343ae5126eddf/src/exe/model.cc#L214) and that the default type is aligning the model with the dominant plane. I believe setting the `--alignment_type` flag to `custom` will go to the previous behavior.

tsattler:
The problem should be fixed and I think we can close this issue.

PR #1240, opened 2021-06-21
Backwards compatibility of model_aligner

tsattler:
In previous versions of Colmap, the `model_aligner` would align a 3D model to a set of camera poses. In this version, not setting the `alignment_type` option defaults to aligning the model with the dominant plane, which is not the behavior users would expect given previous versions (see #1239 ). I suggest to keep backwards compatibility by keeping the original behavior as the default.


ahojnnes:
Thanks!

PR #1257, opened 2021-07-12
Compute reprojection error in generalized absolute solver

Skydes:
Currently the generalized absolute pose solver computes the squared cosine similarity:
https://github.com/colmap/colmap/blob/cf4a39cee879a1c1000bac7f92cdc024c4c4c523/src/estimators/generalized_absolute_pose.cc#L319-L321

This is not consistent with the single-camera solver, which uses the squared reprojection error, and makes it tricky to tune the RANSAC inlier threshold. This PR allows the GP3P solver to optionally instead compute the reprojection error.

Catch: in-class initialization of non-static data members requires C++>=11, but afaik COLMAP already assumed C++11.
@mihaidusmanu @ahojnnes @mgprt

### DIFF
--- a/src/estimators/generalized_absolute_pose.cc
+++ b/src/estimators/generalized_absolute_pose.cc
@@ -314,11 +314,19 @@ void GP3PEstimator::Residuals(const std::vector<X_t>& points2D,
 
       const double x_0 = points2D[i].xy(0);
       const double x_1 = points2D[i].xy(1);
-      const double inv_x_norm = 1 / std::sqrt(x_0 * x_0 + x_1 * x_1 + 1);
 
-      const double cosine_dist =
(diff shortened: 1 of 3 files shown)


mgprt:
Non-static data member initialization was apparently added in 4.7 (https://gcc.gnu.org/projects/cxx-status.html#cxx11) and COLMAP requires at least 4.9 on Linux. 
Not sure about MSVC.

ahojnnes:
Thanks Paul. Have not had time to look into this until now. One problem with this approach arises when the intrinsics of the different cameras in the rig are different. Then the single reprojection threshold in normalized coordinates will correspond to different pixel values in the different cameras. It might be worth noting that in a comment somewhere.

Skydes:
Thanks @ahojnnes for the comment. This is indeed a problem and I don't see an easy way to get around this. I thus added a warning to the docstring. I usually compute the normalized threshold using the average of the focal lengths over all cameras in the rig.

PR #1275, opened 2021-08-17
Modifying scripts/python/flickr_downloader.py to create files with correct extensions

snavely:
Prior to this change, all downloaded files will be saved with a .jpg extension, regardless of the actual file type or extension of the file. This change retains the extension of the file.


ahojnnes:
Thanks Noah!

PR #1317, opened 2021-10-25
Fix compiler warnings reported by GCC

ahojnnes:


src/feature/sift.cc
```diff
@@ -36,7 +36,6 @@
 #include <memory>
 
 #include "FLANN/flann.hpp"
-#include "GL/glew.h"
```
jhacsonmeza, line 39: Hi @ahojnnes I was trying to compile COLMAP with the option `GUI_ENABLED` disabled on a machine without CUDA, and I obtained an error: `GL_LUMINANCE` not defined. I was checking this previous discussion https://github.com/colmap/colmap/pull/1165 and I think that this line it's important for this case. I added `#include "GL/glew.h"` and compilation worked. I don't know if you are aware of this problem.

ahojnnes: @jhacsonmeza Thanks. I reverted the change. See: https://github.com/colmap/colmap/pull/1329

PR #1329, opened 2021-11-09
Add missing include in case CUDA/GUI is not available

ahojnnes:


ducha-aiki:
But if one purposedly turns off GUI by changing option 
https://github.com/colmap/colmap/blob/dev/CMakeLists.txt#L61
the error of not finding OpenGL thing comes back

PR #1332, opened 2021-11-15
Added descriptor normalization argument to feature_extractor.

mihaidusmanu:
Allows users to select between L1_ROOT and L2 at runtime in the feature_extractor (for benchmarking purposes). Default stays RootSIFT as it usually performs better.


ahojnnes:
Thanks Mihai!

PR #1361, opened 2021-12-29
Prevent a divide by zero corner case.

rmbrualla:
This was observed to lead to a SIGILL error. It could be due to num_inliers being zero, or that inlier_ration ^ kMinNumSamples being so small, that it flushes to zero.


ahojnnes:
Thanks again.

PR #1363, opened 2021-12-31
Keep precision in the process of storing in text.

whuaegeanse:
When writing variables to std::ostringstream, we will loose precision of double variables if output manipulators were not applied.


ahojnnes:
Thanks for the fix. However, I don’t think we need the std::fixed and can save us some bytes on disk. Also, I suggest to also use the full 17 digits precision for the 2D points to ensure identical results to the binary format. Thanks!

ahojnnes:
Thanks!

PR #1365, opened 2022-01-02
Require Qt in COLMAPConfig only if GUI is enabled

Skydes:
PR https://github.com/colmap/colmap/pull/1165 introduced the CMake option `GUI_ENABLED` to disable to GUI but did not disable the Qt requirement in the exported `COLMAPConfig.cmake` file.


ahojnnes:
Great, thanks Paul!

PR #1366, opened 2022-01-03
Expose exe internals

Skydes:
In pycolmap we now bind some of the exe functions like RunPointTriangulator. See [pycolmap/pipeline.cc](https://github.com/colmap/pycolmap/blob/master/pipeline.cc). We therefore need COLMAP to expose some exe code that is currently hidden in anonymous namespace or argc-argv functions. Most of this code is likely useful for others and worthy to be publicly exposed, but src/exe is probably not the right place for it.

Any advice on where these bits should be moved?

@ahojnnes @mihaidusmanu @Phil26AT

### DIFF
--- a/src/exe/feature.cc
+++ b/src/exe/feature.cc
@@ -73,38 +73,7 @@ bool VerifySiftGPUParams(const bool use_gpu) {
   return true;
 }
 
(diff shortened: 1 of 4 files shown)


ahojnnes:
I am fine with the proposed changes. Thanks!

PR #1369, opened 2022-01-04
Fix inliers matches extraction in EstimateUncalibrated function.

ferreram:
Currently, the inliers found with the fundamental matrix are always the one used in EstimateUncalibrated(), even if the homography is selected as a better fit.  This commit fixes this behavior.


ahojnnes:
Thanks for the fix !

PR #1371, opened 2022-01-07
Update ModelAligner to handle GPS and custom coords. and more

ferreram:
- Adds GPS conversions to and from ENU coords. (GPS -> ENU, ECEF -> ENU, ENU -> GPS, ENU -> ECEF).

Update of ModelAligner function:  
- Now handles either GPS or custom coordinates as reference coordinates for alignment. 
-  Alignment to pure ENU frame offered as an option. 
-  Aligning reconstruction to ref coords. origin offered as an option.  
- Fix the alignment to plane case if no ref or database provided.  
- Documentation added for the function.


src/exe/model.cc
```diff
@@ -309,13 +393,35 @@ int RunModelAligner(int argc, char** argv) {
...
-                      alignment_type == "enu-unscaled");
+                      alignment_type == "enu-plane-unscaled");
     }
   }
 
+  if (merge_origins) {
+    const Image* first_image =
+        reconstruction.FindImageWithName(ref_image_names[0]);
+
+    const Eigen::Vector3d first_img_position = ref_locations[0];
```
ahojnnes, line 407: const ref

src/exe/model.cc
```diff
@@ -309,13 +393,35 @@ int RunModelAligner(int argc, char** argv) {
...
+
+    const Eigen::Vector3d trans_align =
+        first_img_position - first_image->ProjectionCenter();
+
+    SimilarityTransform3 origin_align(1.0, ComposeIdentityQuaternion(),
```
ahojnnes, line 412: const

src/exe/model.cc
```diff
@@ -124,22 +126,53 @@ void ReadFileCameraLocations(const std::string& ref_images_path,
...
-      gps_locations.push_back(image.TvecPrior());
+      ref_locations.push_back(image.TvecPrior());
     }
   }
-  ref_locations = gps_transform.EllToXYZ(gps_locations);
+
+  if (ref_is_gps) {
+    GPSTransform gps_transform(GPSTransform::WGS84);
+    if (alignment_type != "enu") {
+      std::cout << "\n Converting Alignment Coordinates from GPS (lat/lon/alt) "
```
ahojnnes, line 164: nit: Remove the beginning/trailing spaces in the logged text here and elsewhere. Thanks.

src/base/gps_test.cc
```diff
@@ -165,3 +165,133 @@ BOOST_AUTO_TEST_CASE(TestXYZToEllToXYZ_WGS84) {
...
+  const double alt0 = ell[0](2);
+
+  // Get ENU from Ell
+  const auto enu = gps_tform.EllToENU(ell);
+
+  // Get XYZ from ENU
+  const auto xyz = gps_tform.ENUToXYZ(enu, lat0, lon0, alt0);
+
+  for (size_t i = 0; i < ell.size(); ++i) {
+    BOOST_CHECK(std::abs(xyz[i](0) - ref_xyz[i](0)) < 1e-8);
```
ahojnnes, line 293: Could be BOOST_CHECK(xyz[i].isApprox(ref_xyz[i], 1e-8))

src/base/gps.cc
```diff
@@ -120,4 +120,79 @@ std::vector<Eigen::Vector3d> GPSTransform::XYZToEll(
...
+  std::vector<Eigen::Vector3d> xyz = EllToXYZ(ell);
+
+  // GPS reference origin
+  const double lat0 = ell[0](0);
+  const double lon0 = ell[0](1);
+
+  return XYZToENU(xyz, lat0, lon0);
+}
+
+std::vector<Eigen::Vector3d> GPSTransform::XYZToENU(
```
ahojnnes, line 135: I don't understand how this is meant to be used. The first point is used as the reference but then we require the input of the same point also as lat0/lon0?

src/base/gps.h
```diff
@@ -55,6 +55,21 @@ class GPSTransform {
   std::vector<Eigen::Vector3d> XYZToEll(
       const std::vector<Eigen::Vector3d>& xyz) const;
 
+  std::vector<Eigen::Vector3d> EllToENU(
+      const std::vector<Eigen::Vector3d>& ell) const;
+
+  std::vector<Eigen::Vector3d> XYZToENU(const std::vector<Eigen::Vector3d>& xyz,
+                                        const double lat0,
+                                        const double lon0) const;
```
ahojnnes, line 63: Seems like this should be a private method?

src/base/gps.h
```diff
@@ -55,6 +55,21 @@ class GPSTransform {
   std::vector<Eigen::Vector3d> XYZToEll(
       const std::vector<Eigen::Vector3d>& xyz) const;
 
+  std::vector<Eigen::Vector3d> EllToENU(
```
ahojnnes, line 58: It seems like that the convention is that the first point in the input list is considered as the origin/ref? It would seem more natural to require this point explicitly, equivalent to the ENUToXYZ method?

ahojnnes (changes requested):
Thanks for your contribution, I left a few comments for you.

src/exe/model.cc
ferreram: Thanks, it's done

src/exe/model.cc
ferreram: Thanks, it's done

src/exe/model.cc
ferreram: Thanks, it's done

src/base/gps_test.cc
ferreram: Thanks, I ended up doing this for all test functions in gps_test.cc

src/base/gps.cc
ferreram: I guess this is solved by the change in EllToENU function now explicitly requiring the origin lat/lon coords?

src/base/gps.h
ferreram: I don't really know here. It's not used yet anywhere else yet but might become handy for later use / applications? Also this will give me some troubles for the test functions ^^

src/base/gps.h
ferreram: Right, I have applied the modification.

src/base/gps.cc
ahojnnes: LGTM, thanks.

src/exe/model.cc
```diff
@@ -111,11 +111,33 @@ void WriteBoundingBox(const std::string& reconstruction_path,
   }
 }
 
+void ConvertCameraLocations(const bool ref_is_gps,
+                            const std::string& alignment_type,
+                            std::vector<Eigen::Vector3d>& ref_locations) {
```
ahojnnes, line 116: Output parameters should be pointers.

src/exe/model.cc
```diff
@@ -111,11 +111,33 @@ void WriteBoundingBox(const std::string& reconstruction_path,
...
+  if (ref_is_gps) {
+    GPSTransform gps_transform(GPSTransform::WGS84);
```
ahojnnes, line 118: Can be const?

ahojnnes (changes requested):
Small comments. Maybe you can also run another clang-format before final commit, as it seems some formatting got messed up in the last changes. Thanks for your PR!

ahojnnes (approved):

PR #1408, opened 2022-01-27
Adding user-specified option for reconstructing purely planar scene. …

ferreram:
… If option is set, estimation of homography-based two-view geometry for matching will be forced.

In my tests, this option significantly improves the quality of the matching step when reconstructing planar scenes.

Feel free to merge it or not as it might be a bit specific.

### DIFF
--- a/src/ui/feature_matching_widget.cc
+++ b/src/ui/feature_matching_widget.cc
@@ -132,7 +132,8 @@ void FeatureMatchingTab::CreateGeneralOptions() {
                                  "multiple_models");
   options_widget_->AddOptionBool(&options_->sift_matching->guided_matching,
                                  "guided_matching");
-
+  options_widget_->AddOptionBool(&options_->sift_matching->planar_scene,
+                                 "pure_planar_scene");
   options_widget_->AddSpacer();
 
   QScrollArea* options_scroll_area = new QScrollArea(this);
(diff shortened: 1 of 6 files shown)


src/util/option_manager.cc
```diff
@@ -320,6 +320,8 @@ void OptionManager::AddMatchingOptions() {
                               &sift_matching->multiple_models);
   AddAndRegisterDefaultOption("SiftMatching.guided_matching",
                               &sift_matching->guided_matching);
+  AddAndRegisterDefaultOption("SiftMatching.pure_planar_scene",
```
ahojnnes, line 323: nit: pure_planar_scene -> planar_scene

src/ui/feature_matching_widget.cc
```diff
@@ -132,7 +132,8 @@ void FeatureMatchingTab::CreateGeneralOptions() {
                                  "multiple_models");
   options_widget_->AddOptionBool(&options_->sift_matching->guided_matching,
                                  "guided_matching");
-
+  options_widget_->AddOptionBool(&options_->sift_matching->planar_scene,
+                                 "pure_planar_scene");
```
ahojnnes, line 136: planar_scene

ahojnnes (changes requested):

ahojnnes:
Thanks, looks good to me.

PR #1409, opened 2022-01-27
Sfm-GPS BA

ferreram:
Hey @ahojnnes ,

I have been working on directly adding prior motions (e.g. from GPS) into the Bundle Adjustment steps of Colmap.

It allows to use the priors on cameras' position directly in the incremental reconstruction as well as in pure BA.

This not ready to be merged yet but I wanted to know if you would feel like adding this option to Colmap and if so, I'd be glad to get your opinion on it.

### DIFF
--- a/src/base/cost_functions.h
+++ b/src/base/cost_functions.h
@@ -281,9 +281,11 @@ class PositionAlignCostFunction {
         inv_std_y_(1. / std_xyz(1)),
         inv_std_z_(1. / std_xyz(2)) {}
 
-  static ceres::CostFunction* Create(const Eigen::Vector3d& pos_xyz) {
+  static ceres::CostFunction* Create(
+      const Eigen::Vector3d& pos_xyz,
+      const Eigen::Vector3d& std_xyz = Eigen::Vector3d::Ones()) {
     return (new ceres::AutoDiffCostFunction<PositionAlignCostFunction, 3, 4, 3>(
-        new PositionAlignCostFunction(pos_xyz)));
+        new PositionAlignCostFunction(pos_xyz, std_xyz)));
   }
 
   template <typename T>
--- a/src/base/cost_functions.h
+++ b/src/base/cost_functions.h
@@ -292,18 +292,71 @@ class PositionAlignCostFunction {
   bool operator()(const T* const qvec, const T* const tvec,
                   T* residuals) const {
     // Get Pose in world frame (cam -> world transform)
-    Eigen::Matrix<T, 3, 3, Eigen::RowMajor> R;
-    ceres::QuaternionToRotation(qvec, R.data());
+    Eigen::Matrix<T, 3, 3, Eigen::RowMajor> Rciw;
+    ceres::QuaternionToRotation(qvec, Rciw.data());
 
-    const Eigen::Matrix<T, 3, 1> t(tvec[0], tvec[1], tvec[2]);
+    const Eigen::Matrix<T, 3, 1> tciw(tvec[0], tvec[1], tvec[2]);
 
-    const Eigen::Matrix<T, 3, 3, Eigen::RowMajor> Rwc = R.transpose();
-    const Eigen::Matrix<T, 3, 1> twc = T(-1.) * Rwc * t;
+    const Eigen::Matrix<T, 3, 1> twci = T(-1.) * (Rciw.transpose() * tciw);
 
     // Weighted Position error.
-    residuals[0] = T(inv_std_x_) * (twc[0] - T(pos_x_));
-    residuals[1] = T(inv_std_y_) * (twc[1] - T(pos_y_));
-    residuals[2] = T(inv_std_z_) * (twc[2] - T(pos_z_));
+    residuals[0] = T(inv_std_x_) * (twci[0] - T(pos_x_));
+    residuals[1] = T(inv_std_y_) * (twci[1] - T(pos_y_));
+    residuals[2] = T(inv_std_z_) * (twci[2] - T(pos_z_));
+
+    return true;
+  }
+
+ private:
+  const double pos_x_;
+  const double pos_y_;
+  const double pos_z_;
+
+  const double inv_std_x_;
+  const double inv_std_y_;
+  const double inv_std_z_;
+};
+
+// Position alignment cost function with Global Rotation Aligner
+// (position meas. expected to be in world frame).
+class PositionAlignGloblaRotCostFunction {
+ public:
+  explicit PositionAlignGloblaRotCostFunction(
+      const Eigen::Vector3d& pos_xyz,
+      const Eigen::Vector3d& std_xyz = Eigen::Vector3d::Ones())
+      : pos_x_(pos_xyz(0)),
+        pos_y_(pos_xyz(1)),
+        pos_z_(pos_xyz(2)),
+        inv_std_x_(1. / std_xyz(0)),
+        inv_std_y_(1. / std_xyz(1)),
+        inv_std_z_(1. / std_xyz(2)) {}
+
+  static ceres::CostFunction* Create(
+      const Eigen::Vector3d& pos_xyz,
+      const Eigen::Vector3d& std_xyz = Eigen::Vector3d::Ones()) {
+    return (new ceres::AutoDiffCostFunction<PositionAlignGloblaRotCostFunction,
+                                            3, 4, 4, 3>(
+        new PositionAlignGloblaRotCostFunction(pos_xyz, std_xyz)));
+  }
+
+  template <typename T>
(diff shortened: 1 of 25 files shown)


ahojnnes:
I will have a more detailed look, but this would be a quite useful feature. Do you have some test data to share that could be used to validate that this works?

src/base/cost_functions.h
```diff
@@ -267,6 +267,110 @@ class RelativePoseCostFunction {
...
+  }
+
+  template <typename T>
+  bool operator()(const T* const qvec, const T* const tvec,
+                  T* residuals) const {
+    // Get Pose in world frame (cam -> world transform)
+    Eigen::Matrix<T, 3, 3, Eigen::RowMajor> Rciw;
+    ceres::QuaternionToRotation(qvec, Rciw.data());
+
+    const Eigen::Matrix<T, 3, 1> tciw(tvec[0], tvec[1], tvec[2]);
```
ahojnnes, line 298: Seems like we could instead invert the quaternion and directly rotate the point instead of going through a rotation matrix?

src/base/cost_functions.h
```diff
@@ -267,6 +267,110 @@ class RelativePoseCostFunction {
   const double y2_;
 };
 
+// Position alignment cost function (position meas. expected to be in world
+// frame).
+class PositionAlignCostFunction {
+ public:
+  explicit PositionAlignCostFunction(
+      const Eigen::Vector3d& pos_xyz,
+      const Eigen::Vector3d& std_xyz = Eigen::Vector3d::Ones())
```
ahojnnes, line 276: We should CHECK_GT(std, 0)

EwertzJN:
Hello all,
I am very interested in this extension for colmap and would like to test it already on my own dataset. Could you maybe give me instructions on how to build a pipeline that takes into account the GPS prior positions of my images? Thanks a lot in advance!

ferreram:
@EwertzJN you can use the branch https://github.com/ferreram/colmap/tree/sfm_gps_ba_aligner_4_clem

What you need to provide then is either images with their GPS positions set in the EXIF or update the database once you have applied the features extraction step.  You can update the database with python, check the database.py script in colmap repository.  Here is the SQL command that you can apply from python:
`
    # Update database.
    for img_name, nav_val in zip(img_names, nav_vals):
        db.execute("UPDATE images SET prior_tx=?, prior_ty=?, prior_tz=? WHERE name=?;", \
                (nav_val[0], nav_val[1], nav_val[2], img_name))
`

EwertzJN:
First, thank you very much for the quick feedback.
My images are indeed equipped with geodata via EXIF, so the first row of the database after feature extraction looked like this, for example:
`1,000028_pitch1_yaw1.jpg,1,,,,,40.439502777777776,-79.99679722222223,295`

After that, I did an exhaustive feature matching (default settings) and started the reconstruction process, selecting here in addition to the default settings under `Reconstruction options` → `Bundle` → `Prior Motion Parameters` `use_prior_motion` and `is_gps(…)`. Unfortunately, the reconstruction process did not lead to any result.
Maybe you could help me again, I would be very grateful. Enclosed, you will also find the [log](https://github.com/colmap/colmap/files/8161680/log.txt) of the colmap gui.

EwertzJN:
I have now been able to calculate a result, but on a different data set. Could it be that the images I want to use are possibly of too poor quality (the images are stitched together from shots taken by an omnidirectional camera and do not have a high resolution)? Or maybe there is something wrong with the EXIF data? Attached is a selection of the images.
What surprises me then, however, is that the reconstruction without the use of the GPS prior produces a result, albeit a partially poor result. The idea was in fact to improve these bad results by using the GPS-Prior.

000028_pitch1_yaw5
000028_pitch1_yaw9
000044_pitch2_yaw1
000044_pitch2_yaw6
000061_pitch2_yaw11
000078_pitch2_yaw12
000079_pitch1_yaw3
000095_pitch1_yaw2
000096_pitch1_yaw12
000154_pitch1_yaw1

EwertzJN:
I tried a little bit more and activated `use_enu_coords(...)` which initially leads to a more reasonable output. At least this is what this [log](https://github.com/colmap/colmap/files/8224558/log.txt) suggests.
Unfortunately, however, the program always terminates with this error at this point:

```
F0310 14:49:53.567759 14702 parameter_block.h:186] Check failed: UpdatePlusJacobian() Manifold::PlusJacobian computation failed for x: -nan -nan -nan -nan
*** Check failure stack trace: ***
    @     0x7f6b31ec81c3  google::LogMessage::Fail()
    @     0x7f6b31ecd25b  google::LogMessage::SendToLog()
    @     0x7f6b31ec7ebf  google::LogMessage::Flush()
    @     0x7f6b31ec86ef  google::LogMessageFatal::~LogMessageFatal()
    @     0x55d1be352fa3  ceres::internal::ParameterBlock::SetManifold()
    @     0x55d1be34d56c  ceres::internal::ProblemImpl::InternalSetParameterization()
    @     0x55d1be34d72d  ceres::internal::ProblemImpl::SetParameterization()
    @     0x55d1bd91cafe  colmap::BundleAdjuster::AddImageToProblem()
    @     0x55d1bd91c0c0  colmap::BundleAdjuster::SetUp()
    @     0x55d1bd91b73b  colmap::BundleAdjuster::Solve()
    @     0x55d1bda33760  colmap::IncrementalMapper::AdjustGlobalBundle()
    @     0x55d1bd758bb7  colmap::(anonymous namespace)::AdjustGlobalBundle()
    @     0x55d1bd7590c8  colmap::(anonymous namespace)::IterativeGlobalRefinement()
    @     0x55d1bd75c412  colmap::IncrementalMapperController::Reconstruct()
    @     0x55d1bd75ac35  colmap::IncrementalMapperController::Run()
    @     0x55d1bdab31c8  colmap::Thread::RunFunc()
    @     0x55d1bdabaebe  std::__invoke_impl<>()
    @     0x55d1bdabad81  std::__invoke<>()
    @     0x55d1bdabac4f  _ZNSt6thread8_InvokerISt5tupleIJMN6colmap6ThreadEFvvEPS3_EEE9_M_invokeIJLm0ELm1EEEEvSt12_Index_tupleIJXspT_EEE
    @     0x55d1bdababc1  std::thread::_Invoker<>::operator()()
    @     0x55d1bdabab50  std::thread::_State_impl<>::_M_run()
    @     0x7f6b1646ade4  (unknown)
    @     0x7f6b16590609  start_thread
    @     0x7f6b16115163  clone
Signal: SIGABRT (Aborted)
```

Therefore, I took a closer look at the program flow with a debugger. I came to the conclusion that the problem is that `qvec_data` in line 441 of `src/optim/bundle_adjustment.cc` is set to a vector containing `-nan`.
This is because on line 670 of `src/base/reconstruction.h`, `transform` is assigned a non-invertible matrix, which causes the reconstruction model transformation in the next line to assign `[-nan -nan -nan -nan]` to all `qvec` (see `SimilarityTransform3::TransformPose(Eigen::Vector4d* qvec, Eigen::Vector3d* tvec)` on line 215 of `src/base/similarity_transform.cc`). How it can happen that `transform` in line 670 of `src/base/reconstruction.h` is assigned with a non-invertible matrix at all, I unfortunately could not find out yet.
For reference, this is the non-invertible matrix I was talking about:
```
[          0,          0,          0,      24.04]
[          0,          0,          0,     -25.29]
[          0,          0,          0, -9.552e-05]
[          0,          0,          0,          1]
```

src/optim/bundle_adjustment.cc
```diff
@@ -261,9 +261,70 @@ bool BundleAdjuster::Solve(Reconstruction* reconstruction) {
...
+              << Median(vini_err) << " / " << Mean(vini_err) << " / "
+              << StdDev(vini_err) << "\n";
+
+    if (src.size() > 3) {
+      SimilarityTransform3 tform;
+      // bool success = reconstruction->Align(src, dst, &tform);
+      RANSACOptions ransac;
+      ransac.max_error = (options_.motion_prior_xyz_std * 3.).norm();
+      const int min_num_inliers = 3;
+      bool success = reconstruction->AlignRobust(src, dst, min_num_inliers,
```
EwertzJN, line 290: I believe that this call is problematic if all vectors in `dst` are equal, because then an attempt may be made within the function to invert a non-invertible matrix.

EwertzJN, line 290: I believe that this call is problematic if all vectors in `dst` are equal, because then an attempt may be made within the function to invert a non-invertible matrix.

ferreram:
@EwertzJN thanks for your comments.  I have been quite busy lately but I'll try to find some time to get back into this soon.

shubham-monarch:
Why was this branch deleted?

Issue #1413, opened 2022-01-29
GUI no response for a while after opening and closing database manager


json87:
GUI no response for a while after opening and closing database manager. I compare the current code with v3.6, and found that there is one more line in database.cc 

SQLITE3_EXEC(database_, "VACUUM", nullptr);

I have no more knowledge about sqlite db and cannot determine whether the issue is caused by this line.

**To Reproduce**
1. Go to 'GUI'
2. Click on 'database manager'
3. Close 'database manager'
4. See GUI is busy for a while.

**Desktop (please complete the following information):**
 - OS: [win10]
 - COLMAP: [v3.7]

PR #1414, opened 2022-01-31
Only apply sqlite vacuum command when elements are deleted from the database.

ferreram:
Fix for #1413 . Vacuum command was applied whenever the close function is called leading to important delay whenever accessing the database.  As the vacuum command is meant to decrease the size of db files by deleting allocated chunks of memory not used anymore, the PR only applies it when elements are deleted from the database.

### DIFF
--- a/src/base/database.cc
+++ b/src/base/database.cc
@@ -286,7 +286,10 @@ void Database::Open(const std::string& path) {
 void Database::Close() {
   if (database_ != nullptr) {
     FinalizeSQLStatements();
-    SQLITE3_EXEC(database_, "VACUUM", nullptr);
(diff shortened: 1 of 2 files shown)


ahojnnes:
Setting mutable on fields is feels dangerous, as we are giving the false impression of not mutating the state of the object, which might be problematic when concurrently accessing it. I guess, we could ignore this problem in this specific case, as concurrent database access is anyway prohibited. Thanks for the fix.

wwh301:
Hi!
This fix a great problem!
But we guess why it needs 5~6 seconds to response, when opening "Database management"->"Overlapping images".
Do you met the same situation?
Thank you so much!

Issue #1447, opened 2022-03-04
-nan Changed observations in Global bundle adjustment


zhywanna:
I got some trouble dusing colmap mapper

==============================================================================
Global bundle adjustment
==============================================================================

  => Completed observations: 0
  => Merged observations: 0
  => Filtered observations: 0
  => Changed observations: -nan

==============================================================================
Global bundle adjustment
==============================================================================

  => Completed observations: 0
  => Merged observations: 0
  => Filtered observations: 0
  => Changed observations: -nan

==============================================================================
Global bundle adjustment
==============================================================================

  => Completed observations: 0
  => Merged observations: 0
  => Filtered observations: 0
  => Changed observations: -nan
  => Filtered images: 0

as you can see Changed observations: -nan

tsattler:
You get the nan because nothing changed and this special case is not caught in the computation of the output (https://github.com/colmap/colmap/blob/dev/src/controllers/incremental_mapper.cc#L91). Note that this is just a bug in the computation of the statistics. This does not mean there are nans in the bundle adjustment.

PR #1449, opened 2022-03-05
Fix bug in the computation of the statistics Global/Local BA

whuaegeanse:
Fix bug in the computation of the statistics for Global/Local BA.
The bug is reported in #1447


ahojnnes:
Thanks for the fix!

PR #1488, opened 2022-04-12
Use masks for stereo fusion on automatic reconstruction

ibrarmalik:
Hey @ahojnnes

The automatic reconstruction does not make use of the depth map masking (#1148), which I think makes sense if we have the masks available. Just a minor PR!


ahojnnes:
Thanks, makes sense.

PR #1491, opened 2022-04-13
Fix memory leaks in Ply.h

alanb-sony:
`PlyWritePolygons` leaks some memory.
There is also an incorrect use of `delete[]` with memory allocated with `malloc` in `PlyReadPolygons`


ahojnnes:
Thanks! Is this a fix that you copied over from the official poisson recon library or is it your custom fix?

ahojnnes:
This has been fixed.

PR #1494, opened 2022-04-14
Replace deprecated Eigen nonZeros() call for most recent Eigen versions.

nackjaylor:
Hi team, 

Small change proposed which allows COLMAP to be built from source on the most recent versions of Eigen.

Removal of nonZeros() call which was deprecated in https://gitlab.com/libeigen/eigen/-/commit/08da52eb8537107e2853452bb13c369856d1f84a corresponding to a redundant function call to size().

Issue: COLMAP does not build on Eigen 3.4.90 (and presumably sub-releases since above commit).
Fix: Adjust nonZeros() call to be size().

This is a somewhat interesting issue, because if COLMAP intends to check the number of non-zero elements, the original implementation seemingly did not do that. If the intention was indeed to check non-zero elements a custom check of the matrices is needed pending an appropriate fix in Eigen.

More details on the Eigen front here:
https://gitlab.com/libeigen/eigen/-/issues/2382

Happy to discuss any other workarounds you can think of.

Cheers,
Jack

### DIFF
--- a/src/exe/model.cc
+++ b/src/exe/model.cc
@@ -765,10 +765,10 @@ int RunModelOrientationAligner(int argc, char** argv) {
     const Eigen::Matrix3d frame = EstimateManhattanWorldFrame(
         frame_estimation_options, reconstruction, *options.image_path);
 
-    if (frame.col(0).size() == 0) {
+    if (frame.col(0).lpNorm<1>() == 0) {
       std::cout << "Only aligning vertical axis" << std::endl;
       tform = RotationFromUnitVectors(frame.col(1), Eigen::Vector3d(0, 1, 0));
-    } else if (frame.col(1).size() == 0) {
+    } else if (frame.col(1).lpNorm<1>() == 0) {
       tform = RotationFromUnitVectors(frame.col(0), Eigen::Vector3d(1, 0, 0));
       std::cout << "Only aligning horizontal axis" << std::endl;
     } else {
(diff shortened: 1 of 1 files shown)


ahojnnes:
Thanks for the fix, but the new logic is incorrect. The size of the column is always 3. If `.nonZeros()` is broken in latest Eigen versions, then we probably want something like `.abs().sum()` instead ?

nackjaylor:
Thanks @ahojnnes !

Apologies, I should have been clearer: my initial fix was merely removing the redundant function call since as far as I can tell nonZeros() has always returned size() (i.e. nonZeros has always been misleading). I've taken this comment from the Eigen repo issue and included below which implies this:

> Antonio Sánchez [@cantonios](https://gitlab.com/cantonios) · [4 months ago](https://gitlab.com/libeigen/eigen/-/issues/2382#note_744458095)
> Owner
> 
> Probably. Going back through blames, Gael introduced it over 12 years ago when creating Array.h, which then got factored out into DenseBase.h. It has always returned .size() for dense arrays. It's not particularly useful for dense objects, and I'd be surprised if it's ever actually used. My vote would be to remove it and see if anything breaks.

Checking through the file history on Gitlab to some random old commits, this looks to be the case. As you say, the size of these in COLMAP is always 3, which I believe means that the condition was never satisfied and those lines of code might not have ever run.

In any case: your proposed fix works, I've just used the inbuilt l1-norm to wrap it up together.

ahojnnes:
Thanks!

PR #1540, opened 2022-06-11
Avoid calling copy constructors of  FeatureKeypoints and FeatureDescriptors

whuaegeanse:
Use std::shared_ptr in LRUCache for FeatureKeypoints and FeatureDescriptors.
Avoid (or reduce) calling copy constructors of FeatureKeypoints and FeatureDescriptors when accessing FeatureKeypoints and FeatureDescriptors.


ahojnnes:
Thanks, looks great.

PR #1541, opened 2022-06-13
Added a check if the image_id is part of the correspondence graph.

humenbergerm:
If it is not, and the image_registrator tries to set an observation from that image to triangulated, it will crash with an error in CorrespondenceGraph::FindCorrespondences because it tries to access an image in the graph that does not exist.

"terminating with uncaught exception of type std::out_of_range: unordered_map::at: key not found"
This can happen if you use, for example, a reconstruction that was created by only using the point_triangulator and not the entire SFM pipeline.

### DIFF
--- a/src/base/reconstruction.cc
+++ b/src/base/reconstruction.cc
@@ -108,14 +108,19 @@ void Reconstruction::SetUp(const CorrespondenceGraph* correspondence_graph) {
   // If an existing model was loaded from disk and there were already images
   // registered previously, we need to set observations as triangulated.
   for (const auto image_id : reg_image_ids_) {
-    const class Image& image = Image(image_id);
-    for (point2D_t point2D_idx = 0; point2D_idx < image.NumPoints2D();
-         ++point2D_idx) {
-      if (image.Point2D(point2D_idx).HasPoint3D()) {
-        const bool kIsContinuedPoint3D = false;
-        SetObservationAsTriangulated(image_id, point2D_idx,
-                                     kIsContinuedPoint3D);
+    if (correspondence_graph->ExistsImage(image_id)) {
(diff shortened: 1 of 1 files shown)


ahojnnes:
Hi, thanks for the fix! Could you explain the scenario under which this would happen? It seems like this can only happen when the correspondence graph is inconsistent with the reconstruction. The code was written with the intent that the correspondence graph always contains a superset of the reconstruction (admittedly, that not being documented in any way).

humenbergerm:
Hi! I think this is a special case that occurs for us when we want to register new images in an existing reconstruction that was created outside of COLMAP and when the corresponding db does not contain two-view geometries for the images involved (reference images). That is the case, e.g., if we use RGBD images (incl. 6DOF poses) to create the 3D points instead of triangulating keypoint matches. In order to register new images in this reconstruction using the image_registrator, we need to create feature matches between the reference images and the new images (and import them into the db in order to create the two-view geometries). Since the two-view geometries between the reference images are missing, the image_registrator will create the correspondence graph only with the ones between the reference images and the new images. It can then happen that some reference images are not connected and those ones will be ignored by my fix. At least that is how I understand it. 
An alternative solution would be to create the missing reference two-view geometries and to import them as well. This would probably be a bit of an overkill since those two-view geometries would not be used for anything else.
What do you think?

ahojnnes:
Sorry for the slow response and thanks for the context. I need to read this code carefully again, as I wrote this many years ago. Let me try to find some time to understand but, given that you log out a clearly visible warning, I don't see a big concern with this change. Anyway, let me try to find some time and understand again :-)

ahojnnes:
I believe this is not relevant anymore and is now fixed. Please reopen if still relevant. Thank you.

PR #1565, opened 2022-07-16
Enable writing relative poses into database

AIBluefisher:
In this PR, we are able to write relative poses into the database when performing feature matching.

### DIFF
--- a/src/controllers/incremental_mapper.cc
+++ b/src/controllers/incremental_mapper.cc
@@ -350,7 +350,7 @@ bool IncrementalMapperController::LoadDatabase() {
   PrintHeading1("Loading database");
 
   // Make sure images of the given reconstruction are also included when
-  // manually specifying images for the reconstrunstruction procedure.
+  // manually specifying images for the reconstruction procedure.
   std::unordered_set<std::string> image_names = options_->image_names;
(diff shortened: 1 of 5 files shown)


ahojnnes:
I appreciate the contribution. In practice, recomputing the relative pose from the already stored essential matrix is very fast. Can you elaborate a little more why you need to store the relative poses? Currently, I am skeptical that we want to merge this into the main codebase. Thanks.

AIBluefisher:
Sure, the relative pose computation is efficient. But we still have solid reasons to save the relative poses: (1) Sometimes I want to make a comparison with global SfM methods. If the scene graph is pretty large, we still need to wait for minutes to recompute all the relative poses; (2) I noticed that in the initialization procedure of incremental SfM, the relative poses are re-estimated by RANSAC, which sometimes makes the results difficult to reproduce due to the randomness. Obviously, the RANSAC is everywhere in the incremental SfM algorithm, explicitly storing the relative poses can at least make the initialization step reproducible, and thus we can make a more fair comparison.

ahojnnes:
Apologies for the lack of response in this PR. This has now been addressed in another PR: https://github.com/colmap/colmap/pull/1774

PR #1568, opened 2022-07-18
Update vcpkg version to fix CI pipeline

ahojnnes:


XuChengHUST:
Hi Johannes, colmap windows CI pipeline is broken due to the upstream dependency `metis` link is not valid any more.
microsoft/vcpkg#25609
microsoft/vcpkg#25764

ahojnnes:
@XuChengHUST Thanks, I noticed that as well and was hoping to fix it with this PR. It appears that upstream was also still broken but is fixed now.

Issue #1573, opened 2022-07-20
Fix MacOS build from source docs


catapulta:
**Describe the bug**
Code snippet has wrong path to Qt5 in Mac M1.

**To Reproduce**
1. Execute:
```
brew install \
    git \
    cmake \
    boost \
    eigen \
    freeimage \
    glog \
    gflags \
    metis \
    suite-sparse \
    ceres-solver \
    qt5 \
    glew \
    cgal
```

2. Execute:
```
git clone https://github.com/colmap/colmap.git
cd colmap
git checkout dev
mkdir build
cd build
cmake .. -DQt5_DIR=/usr/local/opt/qt/lib/cmake/Qt5
make
sudo make install
```
3. Boost will warn of no Qt5:
```
-- Boost include dirs: /opt/homebrew/include
-- Boost libraries:    
CMake Warning at CMakeLists.txt:123 (find_package):
  By not providing "FindQt5.cmake" in CMAKE_MODULE_PATH this project has
  asked CMake to find a package configuration file provided by "Qt5", but
  CMake did not find one.

  Could not find a package configuration file provided by "Qt5" (requested
  version 5.4) with any of the following names:

    Qt5Config.cmake
    qt5-config.cmake

  Add the installation prefix of "Qt5" to CMAKE_PREFIX_PATH or set "Qt5_DIR"
  to a directory containing one of the above files.  If "Qt5" provides a
  separate development package or SDK, be sure it has been installed.
```

4. When executing `colmap gui` we get `ERROR: Cannot start colmap GUI; colmap was built without GUI support or QT dependency is missing.`

**Desktop:**
 - OS: macOS Monterey 12.14 on Apple M1

ahojnnes:
Thanks for the fix. I am currently blocked to merge it because of vcpkg issues but will merge as soon as that is unblocked. Closing this issue as resolved.

PR #1574, opened 2022-07-20
use default qt5 brew install directory #1573

catapulta:


ahojnnes:
I guess the behavior of homebrew changed. I have not worked actively on a Mac recently, but I remember that it used to be installed under /usr/local/ in the past. Can you confirm? Thanks!

catapulta:
Hi @ahojnnes I used homebrew as directed in the installation instructions, and Qt5 path is `/opt/homebrew/opt/qt@5/lib/cmake/Qt5`. 
Also, notice the `qt@5` -- apparently this happens when there's an existing Qt5 installation in the system, so homebrew adds the @.
Please let me know if I can help further.

Caveat: I used an M1 Mac (ARM CPU), not sure if that changes things in homebrew. However, it seems like M1 will be the standard going forward.

ahojnnes:
Thanks for confirming and your PR.

PR #1576, opened 2022-07-22
checkout the expected version of ceres solver

scott-vsi:
There is no `--tag` option for `git clone` only `--branch` [1]. What is happening is git seems to be ignoring the `--tag` flag and naming the clone 2.1.0. But it is *not* checking out that version. It checks out master. 

[1] https://git-scm.com/docs/git-clone


ahojnnes:
Thanks.

PR #1632, opened 2022-09-13
Add OpenMP to COLMAP_EXTERNAL_LIBRARIES if enabled

logchan:
OpenMP libraries are missing from `COLMAP_EXTERNAL_LIBRARIES` (and therefore `COLMAP_LIBRARIES`) when OpenMP is enabled. As a result [pycolmap](https://github.com/colmap/pycolmap) won't link to OpenMP and produces this runtime error: https://github.com/colmap/pycolmap/issues/62#issuecomment-1185492546

Don't know why it only happens to arm macs though.


ahojnnes:
LGTM, thanks.

PR #1639, opened 2022-09-21
Fix broken loading of image masks on macOS

buesma:
OS: macOS 12.5.1
COLMAP version: 3.8

I experienced the problem that my image masks where not used during feature extraction. Debugging of the mask loading process showed that function `GetRelativePath(options_.image_path, image_path)` had a faulty behaviour:

- Inputs (example):
   - options_.image_path: /users/buesma/colmap-ws/images/
   - image_path: /users/buesma/colmap-ws/images/cam1/image0001
- Retval (expected): cam1/image0001
- Retval (actual): ../cam1/image0001

This could be related to issues of the BOOST filesystem library with paths with trailing slashs under macOS (probably related boostorg/filesystem#76). The wrong path lead to `ExistsFile(mask_path)`returning `False`, and therefore to not loading the image mask.

As `options_.image_path` is fully contained in `image_path`, the (for macOS) faulty function for computation of relative paths is not needed at all. In fact, the relative path is already computed with simple string operations  for `image->Name()` (see line 105).

Independent of this problem, `mask_path` was not processed the same way as `image_path`, before being assigned to `options`. This could have led to issues in the future.

### DIFF
--- a/src/base/image_reader.cc
+++ b/src/base/image_reader.cc
@@ -54,6 +54,8 @@ ImageReader::ImageReader(const ImageReaderOptions& options, Database* database)
   // Ensure trailing slash, so that we can build the correct image name.
   options_.image_path =
       EnsureTrailingSlash(StringReplace(options_.image_path, "\\", "/"));
+  options_.mask_path =
+      EnsureTrailingSlash(StringReplace(options_.mask_path, "\\", "/"));
 
   // Get a list of all files in the image path, sorted by image name.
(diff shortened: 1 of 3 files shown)


buesma:
@ahojnnes Is there anything preventing this PR from being merged (other than probably low priority). Please feel free to let me know if something has to be further improved or if my changes need additional justification.

ahojnnes:
Thanks for the fix and apologies for the slow feedback.

PR #1642, opened 2022-09-28
Fix inconsistent image resizing between CPU/GPU implementations of SIFT

Yzhbuaa:
**This is an important fix which could improves reconstruction quality in many cases.**

Fix the issue of inconsistent resize behavior between CPU version of SIFT and siftGPU.

The 'max_image_size' is passed to '-maxd' flag of siftGPU. The '-maxd' flag means _maximum working dimension of SIFT algorithm_. The problem is that the _maximum working dimension of SIFT algorithm_ is twice the maximum of the image width/height(The SIFT algorithm first doubles the width and height of its input image).

The issue will cause that many reconstructions which used siftGPU features suffer from potentially lower feature quality and less feature numbers due to unexpected image down-sampling.

Check siftGPU manual for more details.

image

### DIFF
--- a/src/feature/sift.cc
+++ b/src/feature/sift.cc
@@ -793,14 +793,13 @@ bool CreateSiftGPUExtractor(const SiftExtractionOptions& options,
   sift_gpu_args.push_back("-v");
   sift_gpu_args.push_back("0");
 
-  // Fixed maximum image dimension.
-  // Note the max dimension of siftGPU is the maximum dimension of the
-  // first scale in the pyramid(which is the 'first_octave')
-  const double compensation_ratio = options.first_octave < 0 ?
-                              std::pow(2, -options.first_octave) : 1;
+  // Set maximum image dimension.
+  // Note the max dimension of SiftGPU is the maximum dimension of the
+  // first octave in the pyramid (which is the 'first_octave').
+  const double compensation_ratio = std::pow(2, -options.first_octave);
   sift_gpu_args.push_back("-maxd");
-  sift_gpu_args.push_back(
-      std::to_string(options.max_image_size * compensation_ratio));
+  sift_gpu_args.push_back(std::to_string(static_cast<int>(
+      std::ceil(options.max_image_size * compensation_ratio))));
 
   // Keep the highest level features.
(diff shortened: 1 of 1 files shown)


ahojnnes:
While I am aware of discrepancies between CPU and GPU implementations, I am not so sure that this fix is correct. I would imagine that one has to consider the upsampling factor for the first scale in the pyramid to do this calculation correctly?

Yzhbuaa:
I think you are right, we need to take the 'first_octave' option into account to unify the downsampling ratio setting between CPU and GPU implementation. By default('first_octave = -1'), if we set the 'max_image_size' to 3200, and we provide images whose dimension are 1920 * 1080. The CPU implementation would not perform downsamping to the original images(which is the work of ImageResizerThread), while the GPU implementation would downsample the original images on its side(cause 3840 exceeds the threshold 3200 we provided).

src/feature/sift.cc
```diff
@@ -793,14 +793,13 @@ bool CreateSiftGPUExtractor(const SiftExtractionOptions& options,
...
 
-  // Fixed maximum image dimension.
-  // Note the max dimension of siftGPU is the maximum dimension of the
-  // first scale in the pyramid(which is the 'first_octave')
-  const double compensation_ratio = options.first_octave < 0 ?
-                              std::pow(2, -options.first_octave) : 1;
+  // Set maximum image dimension.
+  // Note the max dimension of SiftGPU is the maximum dimension of the
+  // first octave in the pyramid (which is the 'first_octave').
+  const double compensation_ratio = std::pow(2, -options.first_octave);
```
Yzhbuaa, line 799: Hello Johannes, I am a little confused by this line's modification. No image up-sampling happens in the GPU version SIFT when 'first_ocatve' > 1.  In that case, image resizing happens only in ImageResizerThread. So I think the compensation_ratio should be 1  when 'first_ocatve' > 1. Otherwise, image will be downsampled more than we've 
 expected.

ahojnnes: Yes, I overlooked the fact that SIFT doesn't handle the upsampling. Should be fixed here: https://github.com/colmap/colmap/pull/1642

Issue #1431, opened 2022-02-18
Colmap on linux without display and with GPU


pratibhashinde:
Hello everyone,

I am new to colmap. I installed it using "conda install -c conda-forge colmap". When the installation is done, I tried to run colmap automatic reconstructor. But it is giving an error : qt:qpa:plugin: could not connect to display. I cannot connect to the display  as I am on the server and already having 2 GPU to run. 

Can anyone tell me how to use colmap on linux server with gpu and without display?

Thanks.

tsattler:
You can call the individual stages manually as they can run without an attached display. See https://colmap.github.io/cli.html for examples.

marc-micatka:
Is there a special flag to force this behavior? Either during installation or running?
I'm experiencing the same issue with a Linux server that I'm trying to run COLMAP on remotely.

I run the following (changing DATASET_PATH correctly):
```
$  conda install -c conda-forge colmap

$ DATASET_PATH=/path/to/project

$ colmap automatic_reconstructor \
    --workspace_path $DATASET_PATH \
    --image_path $DATASET_PATH/images
```

But I get the following error:
```qt.qpa.xcb: could not connect to display 
qt.qpa.plugin: Could not load the Qt platform plugin "xcb" in "" even though it was found.
This application failed to start because no Qt platform plugin could be initialized. Reinstalling the application may fix this problem.
```

So I try and run the CLI step by step:
```
$ colmap feature_extractor \
   --database_path $DATASET_PATH/database.db \
   --image_path $DATASET_PATH/images

qt.qpa.xcb: could not connect to display 
qt.qpa.plugin: Could not load the Qt platform plugin "xcb" in "" even though it was found.
```

Am I doing something wrong? Should I install via a different method?

nicktasios:
I'm having the same issue and I also installed colmap using conda.

micat001:
I solved the issue, as it were, by using The docker environment provided and running the gui locally to see the results.

tb2-sy:
> I solved the issue, as it were, by using The docker environment provided and running the gui locally to see the results.

Could you please explain in detail how to solve it?

marc-micatka:
I couldn't get COLMAP to install with GUI support on a server. Instead, I'm using the docker environment provided in this repo to run COLMAP and output the dense reconstruction. So there's not many steps involved in the solution. The [link here](https://github.com/colmap/colmap/tree/dev/docker) explains how to do it.

marc-micatka:
[...]

PR #1654, opened 2022-10-11
Add CMake flag FORCE_OPENGL_ENABLED

clintlombard:
This allows building Colmap with GUI_ENABLED=OFF, but still allowing OpenGL support.

Related to #1431


src/feature/sift.cc
```diff
@@ -36,9 +36,7 @@
 #include <memory>
 
 #include "FLANN/flann.hpp"
-#if !defined(GUI_ENABLED) && !defined(CUDA_ENABLED)
```
clintlombard, line 39: I may be mistaken, but having these here doesn't makes sense because `GL_LUMINANCE` and `GL_UNSIGNED_BYTE` in `ExtractSiftFeaturesGPU` won't be defined. This causes compilation errors.

ahojnnes:
Thanks and apologies for the slow response. I think, the change looks good, but I would suggest to revert the option and call it DISABLE_OPENGL, which would be off by default?

RemiLacroix-IDRIS:
Note that OpenGL handling is generally a bit weird. The option says `Whether to enable OpenGL, if available` but the package is always flagged as required in `CMakeLists.txt`: `find_package(OpenGL REQUIRED)`.

kuuhakuJ:
thx, it works!

Issue #1451, opened 2022-03-07
Colmap cmake .. error


HandsomeFYM:
Hello,
I want to install the colmap and I follow the tutorials at https://colmap.github.io/install.html.
It seems to successfully install the ceres-solver. 
And then, in the colmap install part, when I run the "cmake .. ", the following error occurred.

fan@fan-System-Product-Name:~/workspace/colmap/build$ cmake ..
-- The C compiler identification is GNU 9.3.0
-- The CXX compiler identification is GNU 9.3.0
-- Check for working C compiler: /usr/bin/cc
-- Check for working C compiler: /usr/bin/cc -- works
-- Detecting C compiler ABI info
-- Detecting C compiler ABI info - done
-- Detecting C compile features
-- Detecting C compile features - done
-- Check for working CXX compiler: /usr/bin/c++
-- Check for working CXX compiler: /usr/bin/c++ -- works
-- Detecting CXX compiler ABI info
-- Detecting CXX compiler ABI info - done
-- Detecting CXX compile features
-- Detecting CXX compile features - done
-- Looking for pthread.h
-- Looking for pthread.h - found
-- Performing Test CMAKE_HAVE_LIBC_PTHREAD
-- Performing Test CMAKE_HAVE_LIBC_PTHREAD - Failed
-- Looking for pthread_create in pthreads
-- Looking for pthread_create in pthreads - not found
-- Looking for pthread_create in pthread
-- Looking for pthread_create in pthread - found
-- Found Threads: TRUE  
-- Found CXSparse: /usr/include/suitesparse (found suitable version "3.2.0", minimum required is "3.2.0") 
CMake Warning (dev) at /usr/local/lib/cmake/Ceres/FindSuiteSparse.cmake:115 (if):
  Policy CMP0057 is not set: Support new IN_LIST if() operator.  Run "cmake
  --help-policy CMP0057" for policy details.  Use the cmake_policy command to
  set the policy and suppress this warning.

  IN_LIST will be interpreted as an operator when the policy is set to NEW.
  Since the policy is not set the OLD behavior will be used.
Call Stack (most recent call first):
  /usr/share/cmake-3.16/Modules/CMakeFindDependencyMacro.cmake:47 (find_package)
  /usr/local/lib/cmake/Ceres/CeresConfig.cmake:183 (find_dependency)
  CMakeLists.txt:90 (find_package)
This warning is for project developers.  Use -Wno-dev to suppress it.

[...]

PR #1672, opened 2022-10-25
add `libmetis-dev` to solve `METIS_INCLUDE_DIRS`.

FavorMylikes:
```log
Please set them or make sure they are set and tested correctly in the CMake files:
METIS_INCLUDE_DIRS
```

https://github.com/colmap/colmap/issues/1451#issuecomment-1062720119


ahojnnes:
Thanks

PR #1693, opened 2022-11-16
Use FindCUDAToolkit when available.

hanseuljun:
Using find_package resulted in missing CUDA::cudart, CUDA::cublas, etc, from my machine.
WSL Ubuntu 20, CMake 3.24, CUDA 12.

Using find_package(CUDAToolkit) solved this issue. Also, FindCUDA is departed since CMake 3.10!

Have not tested this in other machines and new to using CUDA, so maybe this might be a wrong approach to this problem...

references:
https://cmake.org/cmake/help/latest/module/FindCUDA.html
https://cmake.org/cmake/help/latest/module/FindCUDAToolkit.html

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -116,7 +116,11 @@ endif()
 
 set(CUDA_MIN_VERSION "7.0")
 if(CUDA_ENABLED)
-    find_package(CUDA ${CUDA_MIN_VERSION} QUIET)
+    if(${CMAKE_VERSION} VERSION_LESS "3.17.0")
+	find_package(CUDA ${CUDA_MIN_VERSION} QUIET)
+    else()
+	find_package(CUDAToolkit ${CUDA_MIN_VERSION} QUIET)
+    endif()
 endif()
 
 if(GUI_ENABLED)


ahojnnes:
Thank you.

takeru:
Failed to build on my Ubuntu 22.04.

ahojnnes:
@takeru Could you post an error Message or would you have time to submit a fix? Thank you.

fadenb:
> Failed to build on my Ubuntu 22.04.

While the build does not fail for me the method fails to build something with CUDA support:

When I force `find_package(CUDA ${CUDA_MIN_VERSION} QUIET)` it is properly detected:
```
-- Autodetected CUDA architecture(s):  8.6 8.6 8.6 8.6
-- Enabling CUDA support (version: 11.8, archs: sm_86)
```

Using `find_package(CUDAToolkit ${CUDA_MIN_VERSION} QUIET)` results in:
```
-- Disabling CUDA support
```

I am on Ubuntu 22.04.

takeru:
> While the build does not fail for me the method fails to build something with CUDA support:

That's what I meant.

lnexenl:
Also failed on Arch Linux, with CUDA 11.8 and cmake 3.25.0

PR #1698, opened 2022-11-22
Fixed a crash due to inconsistent undistortion

SomeAlphabetGuy:
When distortion coefficients are really small but not zero, IsUndistorted() may return true, but undistorted image size may still be different from the original image size. In this situation the mismatch between the image size in the model and the actual image size causes the patch match stereo step to crash. This change skips camera undistortion when the coefficients are very small, which makes it consistent with image undistortion and fixes the crash.


ahojnnes:
Thanks!

Issue #1773, opened 2023-01-27
Exporting tvec and qvec from two_view_geometries in database


yanxke:
I see that F, E, and H are all populated but qvec and tvec seem to be all zeroes.  Is there a way to populate these and extract them out?  I see that https://github.com/colmap/colmap/pull/1103 is supposed to store them, but they're not there in my data.

yanxke:
Tracing through the code, it looks like qvec and tvec aren't updated until the incremental mapper is called.  I was hoping they'd be written after feature matching.

yanxke:
Closing as the code change is merged.

PR #1774, opened 2023-01-28
Option to store relative pose between two cameras in database

yanxke:
See https://github.com/colmap/colmap/issues/1773.  So that the values can be written into the DB after feature matching.


ahojnnes:
Thanks for your contribution. I have the same concerns as with the referenced PR #1565. Could we maybe add this feature behind an flag "compute_relative_pose" (which defaults to false) in the feature matching options?

yanxke:
Yes, please see the latest commit.  It seems like the SIFT matching options is overloaded to contain other options and I couldn't find a better place to put this.

ahojnnes:
Yes, the options are not clearly delineated. This looks good to me. Thank you!

src/estimators/two_view_geometry.cc
```diff
@@ -361,6 +361,10 @@ void TwoViewGeometry::EstimateCalibrated(
                         num_inliers, *best_inlier_mask, options)) {
       config = ConfigurationType::WATERMARK;
     }
+
+    if (options.compute_relative_pose) {
+      EstimateRelativePose(camera1, points1, camera2, points2);
```
ahojnnes, line 366: it seems like we could also add this to EstimateUncalibrated/EstimateHomography?

ahojnnes (changes requested):

src/estimators/two_view_geometry.cc
yanxke: Sure, added.

ahojnnes (approved):

PR #1814, opened 2023-02-26
fix take the same image for construct pair

tornadomeet:
if ```i==0```, then ```image_idx2 is equal image_idx1```, the image pair is redundancy


tornadomeet:
i found code in SiftFeatureMatcher, so close the pr
```c++
if (image_pair.first == image_pair.second) {
      continue;
    }
```

PR #1866, opened 2023-03-29
Manual extrinsics calculation

gainikussainova:
The change is:

- calculation of camera extrinsics matrices using features matches generated by colmap and camera intrinsics matrices.
to run:
python database.py --database_path path_to_colmap_database.db_file_from_colmap_with_matched_features --outdir file_with_output_extrinsics
The database.db file I used with colmap matched features is in input_data/database.db

Result shows that rotation matrix was reconstructed right: L2 error: 2.2580556793279634. 
But translation vector has scaling factor, so the L2 error is 7660005.091985998. 
input_data/estimated_extrinsics.txt has the calculated extrinsics for every camera.

- shell script to run colmap reconstruction using known camera intrinsics matrices using cli 
to run: 
shell run_with_known_intrinsics.sh
Intrinsics.txt should have the camera intrinsics in the colmap format. The input_data/intrinsics.txt has the example of the file.


ahojnnes:
Thanks for your contribution. After looking at the code, I am trying to understand why we would merge this change into the main branch. Can you elaborate a little more why you think this will be valuable for others? Thank you.

PR #1896, opened 2023-04-28
check Z_index correctly in ReadPly

countywest:
When reading a ply file, it seems properties x,y,z should exist in the header. Correct checking "z" property is missing or not.


ahojnnes:
Thanks for the fix!

PR #1923, opened 2023-05-17
Log the name of the file that causes Mat::Read() to checkfail

SomeAlphabetGuy:
Some of the CHECK statements in Mat::Read() do not log the name of the file that triggers them. This can cause crashes that are difficult to debug.


SomeAlphabetGuy:
The Windows build has failed for reasons not related to this commit.

ahojnnes:
Thanks for the PR, will merge once I find time to resolve the Windows CI issue.

PR #1947, opened 2023-06-04
Remove unnecessary custom Eigen aligned allocator macros

ahojnnes:
Not required anymore with newer C++ standards.


nnop:
Eigen [says](https://eigen.tuxfamily.org/dox/group__TopicStructHavingEigenMembers.html) that if using the C++17 standard, alignment issues can be ignored. But colmap is using C++14. Is that OK?

PR #1954, opened 2023-06-08
Fix Ceres version test

drkoller:
In #1477, checks of the Ceres version were added to choose between the newer ceres::Manifold or the deprecated ceres::LocalParameterization interface, e.g.:

https://github.com/colmap/colmap/blob/bce8f13727960d9aa4a21bc87524ed11f4706f4f/src/colmap/base/cost_functions.h#L298

However, this test will incorrectly return False for Ceres versions 3.0, 4.0, etc. This PR fixes this bug so that True is returned for all Ceres versions 2.1 and beyond.


ahojnnes:
Great, thanks for the fix.

Issue #1953, opened 2023-06-05
Question about IncrementalMapper::findLocalBundle


coffeeancode:
Hi!

I got a bit confused when looking at the code in `IncrementalMapper::findLocalBundle`, line 1040, that loops over `image.Points2D()` to fill `shared_points3D`? Shouldn't it instead loop through the 2D-points in `overlapping_image.Points2D()`, to find the 3D-points shared with the reference image? As it is now (I might have missed something though), the vector `shared_points3D` will be filled with the same values every time it's computed.

wesleyliwei:
fixed in latest code ,please check, and close the issue, if solved~

coffeeancode:
Looks good to me, hopefully this doesn't cause any weird regressions, in which case I'd be even more confused.

PR #1963, opened 2023-06-16
[Bug] fix finding shared points3D in FindLocalBundle

wesleyliwei:
the latest logic will always use the new coming image observed 3D points, this is not shared.
Related issue: https://github.com/colmap/colmap/issues/1953
@ahojnnes could you help double check? thanks~


ahojnnes:
Thank you, this fixes a genuine bug.

Issue #1946, opened 2023-06-01
I cannot run rig_bundle_adjuster


mrKornWij:
here is the result in Terminal after running 'colmap rig_bundle_adjuster' :

==============================================================================
Camera rig configuration
==============================================================================

Estimating relative rig poses
-----------------------------
Need at least one snapshot with an image of camera 4 and the reference camera 2 to compute its relative pose in the camera rig
WARN: Failed to estimate rig poses from reconstruction; cannot use rig BA

==============================================================================
Rig bundle adjustment
==============================================================================

F0601 14:09:22.119477  7115 sfm.cc:702] Check failed: bundle_adjuster.Solve(&reconstruction, &camera_rigs) 
*** Check failure stack trace: ***
    @     0x7fbce84f90cd  google::LogMessage::Fail()
    @     0x7fbce84faf33  google::LogMessage::SendToLog()
    @     0x7fbce84f8c28  google::LogMessage::Flush()
    @     0x7fbce84fb999  google::LogMessageFatal::~LogMessageFatal()
    @     0x558072e82f86  (unknown)
    @     0x558072e3c863  (unknown)
    @     0x7fbce43b4c87  __libc_start_main
    @     0x558072e4073a  (unknown)
Aborted (core dumped)

mrKornWij:
Here is my json file:

[
	{
     		"ref_camera_id": 2,
     		"cameras":
     		[
			{
           			"camera_id": 1,
           			"image_prefix": "back_cam",
           			"rel_tvec": [0, 0, 0],
           			"rel_qvec": [0, 0, 1, 0]
       			},
       			{
           			"camera_id": 2,
           			"image_prefix": "front_cam",
           			"rel_tvec": [0, 0, 0],
           			"rel_qvec": [1, 0, 0, 0]
       			},
			{
           			"camera_id": 3,
           			"image_prefix": "left_cam",
           			"rel_tvec": [0, 0, 0],
           			"rel_qvec": [0.7071, 0, -0.7071, 0]
			},
       			{
           			"camera_id": 4,
           			"image_prefix": "right_cam",
           			"rel_tvec": [0, 0, 0],
           			"rel_qvec": [0.7071, 0, 0.7071, 0]
       			}
     		]
	}
]

mrKornWij:
My inputs are:
1. txtfile model (cameras.txt include 4 cameras ,images.txt: manually write qvec tvec cam_id, points3d.txt)
2. camera rig config file (json)
[...]

PR #1978, opened 2023-06-25
[feature] print verbose information for model analyzer

wesleyliwei:
When solving the issue, I found the model analyzer didn't print model detail, I add an option to print the detail camera and image information. I think it's useful tooling enhancement.

https://github.com/colmap/colmap/issues/1946
would continue follow up the issue.


wesleyliwei:
An example result:
image

wesleyliwei:
@ahojnnes Could you help check? thanks~

PR #1984, opened 2023-06-28
Add a missing include to compile with gcc13

EstebanDugueperoux2:
- Add a missing include to compile with gcc13 on Fedora 38


EstebanDugueperoux2:
@Hi @ahojnnes,

I don't understand why some two builds failed, in annotations the hyperlink "[Check failure on line 168 in Build log](https://github.com/colmap/colmap/pull/1984/files#annotation_12305271792)" seems incorrect, it don't redirect to build log.

PR #1988, opened 2023-06-29
Speed up snapshot construct in RigBundleAdjuster

wesleyliwei:
1. Too many loops for snapshot setup in ReadCameraRigConfig, could break if ref camera found.
2. the same for another palce in ComputeRelativePoses function , could break if ref camera found, although there is only one ref image.


ahojnnes:
Thx.

PR #2000, opened 2023-07-04
Added a virtual destructor to Sampler

SomeAlphabetGuy:
Any superclass that has virtual methods should also have a virtual destructor, otherwise there's a risk of undefined behavior (if an object of a subclass is deleted through a pointer to the superclass).


ahojnnes:
Thanks, good catch. Need to figure out why clang-tidy doesn't catch this one.

ahojnnes:
Shouldn't happen again: https://github.com/colmap/colmap/pull/2001

Issue #1813, opened 2023-02-24
Hello,I have the error with nvcc fatal : Unknown option 'fPIC'


bobododosjl:
sjl@sjl-System-Product-Name:~/GNC/colmap/build$ make -j8
[ 18%] Built target vlfeat
[ 20%] Built target lsd
[ 25%] Built target poisson_recon
[ 32%] Built target sift_gpu
[ 34%] Built target pba
[ 34%] Building CUDA object src/CMakeFiles/colmap_cuda.dir/mvs/gpu_mat_prng.cu.o
[ 34%] Building CUDA object src/CMakeFiles/colmap_cuda.dir/mvs/gpu_mat_ref_image.cu.o
nvcc fatal   : Unknown option 'fPIC'
nvcc fatal   : Unknown option 'fPIC'
[ 34%] Building CUDA object src/CMakeFiles/colmap_cuda.dir/mvs/patch_match_cuda.cu.o
src/CMakeFiles/colmap_cuda.dir/build.make:75: recipe for target 'src/CMakeFiles/colmap_cuda.dir/mvs/gpu_mat_prng.cu.o' failed
make[2]: *** [src/CMakeFiles/colmap_cuda.dir/mvs/gpu_mat_prng.cu.o] Error 1
make[2]: *** 正在等待未完成的任务....
src/CMakeFiles/colmap_cuda.dir/build.make:90: recipe for target 'src/CMakeFiles/colmap_cuda.dir/mvs/gpu_mat_ref_image.cu.o' failed
make[2]: *** [src/CMakeFiles/colmap_cuda.dir/mvs/gpu_mat_ref_image.cu.o] Error 1
nvcc fatal   : Unknown option 'fPIC'
src/CMakeFiles/colmap_cuda.dir/build.make:119: recipe for target 'src/CMakeFiles/colmap_cuda.dir/mvs/patch_match_cuda.cu.o' failed
make[2]: *** [src/CMakeFiles/colmap_cuda.dir/mvs/patch_match_cuda.cu.o] Error 1
CMakeFiles/Makefile2:633: recipe for target 'src/CMakeFiles/colmap_cuda.dir/all' failed
make[1]: *** [src/CMakeFiles/colmap_cuda.dir/all] Error 2
Makefile:135: recipe for target 'all' failed
make: *** [all] Error 2
sjl@sjl-System-Product-Name:

mmbannert:
[...]

Issue #1753, opened 2023-01-10
nvcc fatal   : Unknown option 'fPIC'


yyspyysp:
When I use  make command I get this error

`[ 90%] Building CXX object src/CMakeFiles/colmap.dir/colmap_autogen/UYX5XTB5RZ/qrc_resources.cpp.o
[ 90%] Linking CXX static library libcolmap.a
[ 90%] Built target colmap
[ 90%] Building CUDA object src/CMakeFiles/colmap_cuda.dir/mvs/gpu_mat_prng.cu.o
nvcc fatal   : Unknown option 'fPIC'
make[2]: *** [src/CMakeFiles/colmap_cuda.dir/mvs/gpu_mat_prng.cu.o] Error 1
make[1]: *** [src/CMakeFiles/colmap_cuda.dir/all] Error 2
make: *** [all] Error 2
`
How can I fix this error?Thanks

yyspyysp:
No. I have tried some online methods but still no solution, what is the problem you are experiencing?




尹一帆
***@***.***



&nbsp;




------------------&nbsp;原始邮件&nbsp;------------------
发件人:                                                                                                                        "colmap/colmap"                                                                                    ***@***.***&gt;;
发送时间:&nbsp;2023年1月11日(星期三) 下午4:39
***@***.***&gt;;
***@***.******@***.***&gt;;
主题:&nbsp;Re: [colmap/colmap] nvcc fatal   : Unknown option 'fPIC' (Issue #1753)





 
@yyspyysp
 Hi, I got the same problem with you. Do you solve this problem?

spicy-dog:
@ahojnnes hello, could you give some suggestions to fix this error? Thanks!

rmadan2:
Hi, I am facing the same issue as well. Can the -fPIC flag be removed from CMakeLists.txt?

spicy-dog:
您好，您的邮件我已经收到了，谢谢

ahojnnes:
Which version of CUDA/GCC are you using to compile colmap?

rmadan2:
CUDA: Cuda compilation tools, release 10.1, V10.1.243
GCC: gcc (Ubuntu 9.4.0-1ubuntu1~20.04.1) 9.4.0

rmadan2:
P.S: I tried removing the -fPIC flag, and it compiled all the way to 96%, (sift_gpu compiles as well), but then it crashes with errors arising here: 
```
[...]

PR #2022, opened 2023-07-10
Remove check on Qt version

Skydes:
I'm on Ubuntu 18.04 and seeing the extra `-fPIC` flag added by Qt 5.9.5, as reported in https://github.com/colmap/colmap/issues/1753 https://github.com/colmap/colmap/issues/1813. Upgrading to cmake 3.26 does not help. I think that it's safer to remove the flag regardless of the Qt version.


ahojnnes:
Are you stuck with 18.04? I was planning to not support 18.04 going forward, as its outdated default compiler/packages hold development back due to lack of C++14, etc.

Skydes:
Nope I indeed need to update very soon...

ahojnnes:
I would rather add the 5.9.5 version explicitly, as it does appear to be fixed for the Qt version under 20.04 and 22.04. It sounds a little dangerous to mess with the default flags across all versions.

ahojnnes:
Looked at the cmake bug again and also the special casing here. I think we can go ahead as proposed by you. Thanks.

Issue #2061, opened 2023-07-25
fix the poses of a part of images(treat them as ground truth poses), while optimizing the poses of the remaining images 


xuxumiao777:
**_Servus!_**
Thank you for your excellent work. Now I have a set of images which are required to get poses. But in my experiment setting, a part of images' poses are already got, they are treated as ground truth, However the other images' poses are unknown. So, I am wondering if there is a way to optimize the unknown poses in this colmap framework while keep the ground truth poses unchanged. Thanks:)

tsattler:
You can use the `point_triangulator` to build a 3D model for the images where you have known poses. You can then use the resulting 3D model as input to the `mapper` and use the command line option to fix the camera poses of the images already in the reconstruction.

xuxumiao777:
@tsattler I tried to use option --Mapper.fix_existing_images 1 to fix the ground truth poses when using mapper. However, when I have done this, I found the ground truth poses have been optimized and modified. So how it happens?
Here is my whole procedure:
![image](https://github.com/colmap/colmap/assets/107257126/50485fad-6d0a-4750-9482-33e422b82137)
I have two dataset folders one called flowers_split1 and the other called flowers_split2. The poses of flowers_split1 are known and then fixed, the poses are also calculated by colmap, so it has database.db and sparse model. The poses of flowers_split2 are unknown and to be calculated by colmap.
## I run in the following procedure.
### feature extraction
`colmap feature_extractor --database_path ./dataset/flowers_split1/database.db --image_path ./dataset/flowers_split2/images --image_list_path ./dataset/flowers_split2/image-list.txt --ImageReader.single_camera 1`
###feature matching
`colmap vocab_tree_matcher --database_path ./dataset/flowers_split1/database.db --VocabTreeMatching.vocab_tree_path ./dataset/vocab_tree_flickr100K_words256K.bin --VocabTreeMatching.match_list_path ./dataset/flowers_split2/image-list.txt`
###mapping
`colmap mapper --database_path ./dataset/flowers_split1/database.db --image_path ./dataset/flowers_split2/images --input_path ./dataset/flowers_split1/sparse/0 --output_path ./dataset/new2 --Mapper.fix_existing_images 1`


After that, I found the poses of flowers_split1 have been modified and not fixed as expected.

xuxumiao777:
## Update
I have checked why translation has been changed but rotation remain unchanged. 
### First Try
[...]

Issue #1260, opened 2021-07-13
Using known camera poses for reconstruction


y-zheng18:
If part of  (not all) the images is provided with known camera poses, how can I use those pose information for reconstruction? for example, which command line should I use to import those camera poses and run reconstruction? It seems --point_triangulator
 needs all the camera poses.

tsattler:
You can use the `point_triangulator` to build a model for those images for which you know the intrinsics and extrinsics and then use the resulting model as input to the `mapper` to continue the reconstruction process.

y-zheng18:
Thanks! But using mapper to continue the reconstruction can change the known camera parameters since global bundle adjustment can influence those camera poses. How can I fix the models and camera poses when using mapper to continue the reconstruction?

tsattler:
There is a parameter for the mapper where you can fix the existing camera poses.

y-zheng18:
Thanks a lot. But when I use command **_colmap mapper --database_path xxx --image_path xxx --output_path xxx --input_path xxx --Mapper.fix_existing_images 1_**, the reconstruction camera rotation remains fixed but translation changed. Did I miss something?

Jentuuh:
I have the same issue, my rotations remain fixed but my translations keep being updated. Does this happen due to the bundle adjustment? Is there a way to keep both the translation and rotation fixed for the existing camera poses?

lkskstlr:
@y-zheng18, @Jentuuh this could be due to normalization, see here: https://github.com/colmap/colmap/issues/664#issuecomment-526578328

xuxumiao777:
> @y-zheng18, @Jentuuh this could be due to normalization, see here: [#664 (comment)](https://github.com/colmap/colmap/issues/664#issuecomment-526578328)
@lkskstlr @tsattler @Jentuuh @y-zheng18 
Have you found a solution to keep both the translation and rotation fixed?

PR #2080, opened 2023-07-31
No reconstruction normalization when fix_exisiting_images option is set to true

ferreram:
Prevent applying a normalization step to the reconstruction after a Global BA step when the user set existing images as fixed in RunMapper(). 

Should fix issues #2061 #1260


ahojnnes:
I am a little hesitant with this proposed change, as it might be problematic when starting from a fixed set of images that is extended by a lot of other images. At some point, the reconstruction may grow to a point where numerical stability may become a problem. An alternative could be to sim3 align the final reconstruction to the fixed set of images before serializing the reconstruction back to disk.

ferreram:
I understand your concern.  I will work on this and open a new pull request when done.

PR #2095, opened 2023-08-02
Transform back to existing images positions after mapper processing if set fixed

ferreram:


ahojnnes:
Thanks, I made some minor adjustments.

Issue #2101, opened 2023-08-07
gtest/gtest.h needed even though TESTS_ENABLED=NO


NeroBurner:
**Describe the bug**
Following the installation instructions on linux (Ubuntu 22.04) https://colmap.github.io/install.html fails at compiling `testing.cc`

```sh
[2/111] Building CXX object src/colmap/util/CMakeFiles/colmap_util.dir/testing.cc.o
FAILED: src/colmap/util/CMakeFiles/colmap_util.dir/testing.cc.o 
/usr/bin/c++ -DBOOST_ALL_NO_LIB -DBOOST_FILESYSTEM_DYN_LINK -DCOLMAP_GUI_ENABLED -DCOLMAP_OPENGL_ENABLED -DCOLMAP_OPENMP_ENABLED -DGFLAGS_IS_A_DLL=0 -DGOOGLE_GLOG_DLL_DECL="" -DGOOGLE_GLOG_DLL_DECL_FOR_UNITTESTS="" -DQT_CORE_LIB -DQT_GUI_LIB -DQT_NO_DEBUG -DQT_OPENGL_LIB -DQT_WIDGETS_LIB -I/home/nero/repos/external/colmap/src -isystem /usr/include/eigen3 -isystem /usr/include/x86_64-linux-gnu/qt5 -isystem /usr/include/x86_64-linux-gnu/qt5/QtCore -isystem /usr/lib/x86_64-linux-gnu/qt5/mkspecs/linux-g++ -isystem /usr/include/x86_64-linux-gnu/qt5/QtOpenGL -isystem /usr/include/x86_64-linux-gnu/qt5/QtWidgets -isystem /usr/include/x86_64-linux-gnu/qt5/QtGui -Wno-maybe-uninitialized -Wall -std=c++14 -O3 -DNDEBUG -std=gnu++14 -fPIC -MD -MT src/colmap/util/CMakeFiles/colmap_util.dir/testing.cc.o -MF src/colmap/util/CMakeFiles/colmap_util.dir/testing.cc.o.d -o src/colmap/util/CMakeFiles/colmap_util.dir/testing.cc.o -c /home/nero/repos/external/colmap/src/colmap/util/testing.cc
/home/nero/repos/external/colmap/src/colmap/util/testing.cc:39:10: fatal error: gtest/gtest.h: No such file or directory
   39 | #include <gtest/gtest.h>
      |          ^~~~~~~~~~~~~~~
compilation terminated.
```

[...]

PR #2102, opened 2023-08-07
cmake: add testing.cc to colmap_util only if TESTS_ENABLED=ON

NeroBurner:
Conditionally add the files `testing.h` and `testing.cc` to the library target `colmap_util` only if `TESTS_ENABLED=ON`.

Otherwise one needs the `gtest/gtest.h` header available, for example by installing the Debian package `libgtest-dev`, even though no tests are actually built.

Using the `target_sources()` function (Introduced with CMake 3.1`) to conditionally add the two testing files to the util target.

https://cmake.org/cmake/help/latest/command/target_sources.html

Fixes: https://github.com/colmap/colmap/issues/2101

### DIFF
--- a/src/colmap/util/CMakeLists.txt
+++ b/src/colmap/util/CMakeLists.txt
@@ -41,7 +41,6 @@ COLMAP_ADD_LIBRARY(
         ply.h ply.cc
         sqlite3_utils.h
         string.h string.cc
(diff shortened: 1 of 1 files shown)


ahojnnes:
Thanks, LGTM.

PR #2126, opened 2023-08-22
Update Linux install documentation with new branch policy

joshuaoreilly:
Not sure if this is actually correct, but the `dev` branch doesn't exist anymore, and [this](https://github.com/colmap/colmap/commit/58ba46f81db47e2982545f763698f312d793e9e3) commit seems to suggest that it's formally no longer being used.


ahojnnes:
Thank you!

PR #2139, opened 2023-09-04
Update CMakeLists.txt

sharon846:
When installing colmap on deep learning cluster, I found that the colmap needs to know some GPU architecture parameter. When using native, it auto find and configures to the current GPU data.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -84,6 +84,7 @@ endif()
 ################################################################################
 
 set(COLMAP_FIND_QUIETLY FALSE)
+set(CMAKE_CUDA_ARCHITECTURES "native")  # Set the desired CUDA architecture(s)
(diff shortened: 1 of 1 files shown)


sharon846:
It’s not relevant here, it’s just a small required fix to support multiple GPUs..

sharon846:
The failing reasons are weird and it does not seem my changes had something to do with that. I changed one line which auto configures the GPU being used.

Skydes:
The is already indicated in the doc:

https://github.com/colmap/colmap/blob/5c1d58a085920c41cf2d9c892da8041b1fc0c86d/doc/install.rst?plain=1#L112-L115

So `cmake .. -GNinja -DCMAKE_CUDA_ARCHITECTURES=native`. I agree that making it a default value makes sense.

PR #2147, opened 2023-09-05
Add Sim3d::FromMatrix

sarlinpe:


src/colmap/geometry/sim3.h
```diff
@@ -60,6 +60,12 @@ struct Sim3d {
     return matrix;
   }
 
+  inline void FromMatrix(const Eigen::Matrix3x4d& matrix) {
```
ahojnnes, line 63: In terms of interface, it will be nicer to have
```
static Sim3d FromMatrix(const Eigen::Matrix3x4d& matrix) {...}
```

ahojnnes: It will be good to have a round-trip test for this new method using ToMatrix/FromMatrix and expecting near equality.

Skydes: Done, ready for review.

PR #2161, opened 2023-09-11
Add cuda accelerate in bundle adjustment with ceres >= 2.1

Richard-coder:
6 times faster on gpu. cpu: i9-13900K(F), gpu: 4090

### DIFF
--- a/src/colmap/controllers/incremental_mapper.cc
+++ b/src/colmap/controllers/incremental_mapper.cc
@@ -224,6 +224,7 @@ BundleAdjustmentOptions IncrementalMapperOptions::LocalBundleAdjustment()
 #if CERES_VERSION_MAJOR < 2
   options.solver_options.num_linear_solver_threads = num_threads;
 #endif  // CERES_VERSION_MAJOR
+  options.use_cuda = ba_use_cuda;
   options.print_summary = true;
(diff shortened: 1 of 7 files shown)


Skydes:
From the Ceres doc:
> Currently this support is limited to using the dense linear solvers that ship with CUDA. As a result GPU acceleration can be used to speed up DENSE_QR, DENSE_NORMAL_CHOLESKY and DENSE_SCHUR.

So using CUDA or not is independent from `linear_solver_type` and yields improvements only for small problems (`num_images <= kMaxNumImagesDirectDenseSolver`) - at least until sparse solvers are GPU-accelerated as well (https://github.com/ceres-solver/ceres-solver/issues/759).

Richard-coder:
Thanks, I create this request because in my experiments, for scenes around 2000 images(sparse problem with DENSE_SCHUR), `mapper` and `bundle_adjuster` is faster with gpu. Thanks for your reply, it seems better to close this reuqest and wait ceres 2.2 support sparse solvers

pwais:
cc @sarlinpe It looks like CERES just recently got at least cuDSS sparse cholesky https://github.com/ceres-solver/ceres-solver/commit/6fb3dae4eeef855568e47ebbb29a8ba4f3c9153f

PR #2165, opened 2023-09-11
Expose exe as libs

sarlinpe:
- Rename the executable target `colmap_exe` -> `colmap_main`.
- Add lib target `colmap_exe` to expose some internals of exe/{feature,sfm} for pycolmap - done in https://github.com/colmap/colmap/pull/1368 but the target was deleted in a recent commit.
- Expose the internals of RunModelComparer

### DIFF
--- a/src/colmap/exe/model.h
+++ b/src/colmap/exe/model.h
@@ -29,7 +29,22 @@
 //
 // Author: Johannes L. Schoenberger (jsch-at-demuc-dot-de)
 
+#pragma once
+
+#include "colmap/estimators/alignment.h"
+#include "colmap/geometry/sim3.h"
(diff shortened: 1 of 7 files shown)


src/colmap/exe/model.h
```diff
@@ -29,8 +29,23 @@
...
 
+#pragma once
+
+#include "colmap/estimators/alignment.h"
+#include "colmap/geometry/sim3.h"
+#include "colmap/scene/reconstruction.h"
+
 namespace colmap {
 
+int CompareModels(const Reconstruction& reconstruction1,
```
ahojnnes, line 40: The return value should be bool?

sarlinpe: Good catch, fixed.

PR #2172, opened 2023-09-14
Migrate to glog logging

ahojnnes:


src/colmap/feature/sift.cc
```diff
@@ -31,10 +31,6 @@
 
 #include "colmap/feature/sift.h"
 
-#if defined(COLMAP_GPU_ENABLED)
-#include <GL/glew.h>
```
ahojnnes, line 35: Submit separately.

PR #2174, opened 2023-09-15
Add EstimateGeneralizedAbsolutePose

sarlinpe:
- Created `estimators/generalized_pose` as `estimators/pose` is getting too crowded
- The logic for the unique inlier support was written by @mihaidusmanu


src/colmap/estimators/generalized_pose.cc
```diff
@@ -0,0 +1,372 @@
...
+#include "colmap/sensor/models.h"
+#include "colmap/util/misc.h"
+
+#include <Eigen/Core>
+#include <ceres/ceres.h>
+
+namespace colmap {
+namespace {
+
+const bool lowerVector3d(const Eigen::Vector3d& v1, const Eigen::Vector3d& v2) {
```
ahojnnes, line 51:
```suggestion
bool LowerVector3d(const Eigen::Vector3d& v1, const Eigen::Vector3d& v2) {
```

src/colmap/estimators/generalized_pose.cc
```diff
@@ -0,0 +1,372 @@
...
+    } else {
+      return false;
+    }
+  } else {
+    return false;
+  }
+}
+
+const bool equalVector3d(const Eigen::Vector3d& v1, const Eigen::Vector3d& v2) {
+  return ((v1 - v2).norm() < 1e-5);
```
ahojnnes, line 68: v1.isApprox(v2) should be better?

src/colmap/estimators/generalized_pose.cc
```diff
@@ -0,0 +1,372 @@
...
+}
+
+// Measure the support of a model by counting the number of unique inliers
+// (e.g., visible 3D points), number of inliers, and summing all inlier
+// residuals. Each sample should have an associated id. Samples with the same id
+// are only counted once in num_unique_inliers. The support is better if it has
+// more unique inliers, more inliers, and a smaller residual sum.
+struct UniqueInlierSupportMeasurer {
```
ahojnnes, line 76: This should be something that is implemented under optim/ with unit tests.

src/colmap/estimators/generalized_pose.cc
```diff
@@ -0,0 +1,372 @@
...
+    }
+  }
+
+ private:
+  std::vector<size_t> sample_ids_;
+};
+
+}  // namespace
+
+bool EstimateGeneralizedAbsolutePose(
```
ahojnnes, line 139: I know these tests were missing before, but it could be great to add a unit test for this. I want to at least not add to further gaps in code coverage with new functionality being added.

src/colmap/estimators/generalized_pose.cc
sarlinpe: Done.

src/colmap/estimators/generalized_pose.cc
sarlinpe: Done.

src/colmap/estimators/generalized_pose.cc
sarlinpe: Done.

src/colmap/estimators/generalized_pose.cc
sarlinpe: Done for generalized_pose.

src/colmap/estimators/generalized_pose.h
```diff
@@ -0,0 +1,97 @@
...
+bool EstimateGeneralizedAbsolutePose(
+    const RANSACOptions& options,
+    const std::vector<Eigen::Vector2d>& points2D,
+    const std::vector<Eigen::Vector3d>& points3D,
+    const std::vector<size_t>& camera_idxs,
+    const std::vector<Rigid3d>& cams_from_rig,
+    const std::vector<Camera>& cameras,
+    Rigid3d* rig_from_world,
+    size_t* num_inliers,
+    std::vector<char>* inlier_mask);
```
sarlinpe, line 67: @ahojnnes  Why are inlier masks defined as `vector<char>` instead of `vector<bool>`?

sarlinpe:
PTAL.

src/colmap/estimators/generalized_pose.h
ahojnnes: vector<bool> uses a bit mask under the hood, so each read or write requires some additional shuffling of bits. I vaguely remember benchmarking this in the past and then using char instead. Not sure if this applies here though. I would keep it as is, unless we have opposite evidence from some profiling.

src/colmap/estimators/generalized_pose.cc
```diff
@@ -0,0 +1,306 @@
...
+#include "colmap/sensor/models.h"
+#include "colmap/util/misc.h"
+
+#include <Eigen/Core>
+#include <ceres/ceres.h>
+
+namespace colmap {
+namespace {
+
+bool LowerVector3d(const Eigen::Vector3d& v1, const Eigen::Vector3d& v2) {
```
ahojnnes, line 52: `(v1 < v2).any()` would be the same?

src/colmap/estimators/generalized_pose.cc
```diff
@@ -0,0 +1,306 @@
...
+    const std::vector<size_t>& camera_idxs,
+    const std::vector<Rigid3d>& cams_from_rig,
+    const std::vector<Camera>& cameras,
+    Rigid3d* rig_from_world,
+    size_t* num_inliers,
+    std::vector<char>* inlier_mask) {
+  CHECK_EQ(points2D.size(), points3D.size());
+  CHECK_EQ(points2D.size(), camera_idxs.size());
+  CHECK_EQ(cams_from_rig.size(), cameras.size());
+  CHECK_GE(*std::min_element(camera_idxs.begin(), camera_idxs.end()), 0);
```
ahojnnes, line 83: This should probably also work:
```suggestion
  CHECK_GE(std::min(camera_idxs), 0);
```

src/colmap/estimators/generalized_pose.cc
```diff
@@ -0,0 +1,306 @@
...
+  unique_points3D.erase(
+      std::unique(unique_points3D.begin(),
+                  unique_points3D.end(),
+                  [](const Eigen::Vector3d& v1, const Eigen::Vector3d& v2) {
+                    return v1.isApprox(v2, 1e-5);
+                  }),
+      unique_points3D.end());
+  std::vector<size_t> point3D_ids;
+  point3D_ids.reserve(points3D.size());
+  for (const auto& p3D : points3D) {
```
ahojnnes, line 122: nit: for consistency in the repo
```suggestion
  for (const auto& point3D : points3D) {
```

src/colmap/estimators/generalized_pose.cc
```diff
@@ -0,0 +1,306 @@
...
+    point3D_ids.push_back(std::lower_bound(unique_points3D.begin(),
```
ahojnnes, line 123: Since we use isApprox above to prune redundant points, but here we use the exact LowerVector3d, I suspect that you may end up with unique_points3D.end() as a return value, if the last point was pruned. In this case, you will end up with a point3D_id that is out of bounds?

src/colmap/estimators/generalized_pose.cc
```diff
@@ -0,0 +1,306 @@
...
+  std::sort(unique_points3D.begin(), unique_points3D.end(), LowerVector3d);
+  unique_points3D.erase(
+      std::unique(unique_points3D.begin(),
+                  unique_points3D.end(),
+                  [](const Eigen::Vector3d& v1, const Eigen::Vector3d& v2) {
+                    return v1.isApprox(v2, 1e-5);
+                  }),
+      unique_points3D.end());
+  std::vector<size_t> point3D_ids;
+  point3D_ids.reserve(points3D.size());
```
ahojnnes, line 121:
```suggestion
  unique_point3D_ids.reserve(points3D.size());
```

src/colmap/estimators/generalized_pose_test.cc
```diff
@@ -0,0 +1,174 @@
...
+
+namespace colmap {
+namespace {
+
+void BuildGeneralizedCameraProblem(Rigid3d& gt_rig_from_world,
+                                   std::vector<Eigen::Vector2d>& points2D,
+                                   std::vector<Eigen::Vector3d>& points3D,
+                                   std::vector<size_t>& camera_idxs,
+                                   std::vector<Rigid3d>& cams_from_rig,
+                                   std::vector<Camera>& cameras) {
```
ahojnnes, line 54: nit

```suggestion

struct GeneralizedCameraProblem {
    Rigid3d gt_rig_from_world;
  std::vector<Eigen::Vector2d> points2D;
  std::vector<Eigen::Vector3d> points3D;
  std::vector<size_t> camera_idxs;
  std::vector<Rigid3d> cams_from_rig;
  std::vector<Camera> cameras;
};

GeneralizedCameraProblem BuildGeneralizedCameraProblem() {
```

src/colmap/optim/support_measurement.h
```diff
@@ -56,6 +56,38 @@ struct InlierSupportMeasurer {
...
+
+    // The number of inliers.
+    // This is still needed for determining the dynamic number of iterations.
+    size_t num_inliers = 0;
+
+    // The sum of all inlier residuals.
+    double residual_sum = std::numeric_limits<double>::max();
+  };
+
+  void SetSampleIds(const std::vector<size_t>& sample_ids) {
```
ahojnnes, line 77:
```suggestion
  void SetUniqueSampleIds(const std::vector<size_t>& sample_ids) {
```

src/colmap/optim/support_measurement.h
```diff
@@ -56,6 +56,38 @@ struct InlierSupportMeasurer {
...
+  }
+
+  // Compute the support of the residuals.
+  Support Evaluate(const std::vector<double>& residuals, double max_residual);
+
+  // Compare the two supports and return the better support.
+  bool Compare(const Support& support1, const Support& support2);
+
+ private:
+  std::vector<size_t> sample_ids_;
```
ahojnnes, line 88:
```suggestion
  std::vector<size_t> unique_sample_ids_;
```

src/colmap/estimators/generalized_pose.cc
```diff
@@ -0,0 +1,306 @@
...
+  options.Check();
+  if (points2D.size() == 0) {
+    return false;
+  }
+
+  const double max_error_px = options.max_error;
+  CHECK_GT(max_error_px, 0.0);
+
+  std::vector<GP3PEstimator::X_t> rig_points2D(points2D.size());
+  double error_threshold_camera = 0.;
```
ahojnnes, line 95: nit: could be extracted into separate helper function `ComputeMaxErrorInCamera` in anonymous namespace

src/colmap/estimators/generalized_pose.cc
sarlinpe: No, here we explicitly order along the x axis first, then y, then z.

src/colmap/estimators/generalized_pose.cc
sarlinpe: `std::min` only works for initializer lists. `std::ranges:min` works but requires C++ 20.

src/colmap/estimators/generalized_pose.cc
sarlinpe: Done.

src/colmap/estimators/generalized_pose.cc
sarlinpe: Done.

src/colmap/estimators/generalized_pose_test.cc
sarlinpe: Done.

src/colmap/optim/support_measurement.h
sarlinpe: Done.

src/colmap/optim/support_measurement.h
sarlinpe: Done.

src/colmap/estimators/generalized_pose.cc
sarlinpe: Yes this is quite error-prone. I've replaced it with the argmin of the distance to all distinct points.

src/colmap/estimators/generalized_pose.cc
sarlinpe: Done.

sarlinpe:
PTAL.

src/colmap/estimators/generalized_pose_test.cc
```diff
@@ -0,0 +1,165 @@
...
+  EXPECT_TRUE(EstimateGeneralizedAbsolutePose(ransac_options,
+                                              problem.points2D,
+                                              problem.points3D,
+                                              problem.camera_idxs,
+                                              problem.cams_from_rig,
+                                              problem.cameras,
+                                              &rig_from_world,
+                                              &num_inliers,
+                                              &inlier_mask));
+  EXPECT_EQ(num_inliers, num_points - gt_num_outliers);
```
sarlinpe, line 123: TODO: compute the number of unique inliers from the number of unique 3D points.

sarlinpe:
All tests now pass - PTAL.

PR #2178, opened 2023-09-16
Restore GLEW include

sarlinpe:
Since https://github.com/colmap/colmap/pull/2176, building with `CUDA_ENABLED=ON GUI_ENABLED=OFF` fails due to undfined `GL_*` symbols:
https://github.com/colmap/colmap/blob/e7014366d750ba5a7e05956bc16617af84588cbd/src/colmap/feature/sift.cc#L652-L656


ahojnnes:
Oh l man, thanks. It could be good to have one of the two cuda builds disable the GUI to catch this in CI?

sarlinpe:
Yes!

PR #2193, opened 2023-09-22
Removing clustering_options_ and mapper_options_ in Hierarchical Mapper Controller

Serenitysmk:
Hi, 

since `clustering_options_` and `mapper_options_` are moved into HierachicalMapperController::Options. They are no longer used, therefore they can be deleted.


ahojnnes:
Thanks!

PR #2238, opened 2023-10-25
Update reconstruction.cc

RayShark0605:
The 'num_tri_corrs' is already of type 'size_t', so the original CHECK here to ensure it's non-negative is redundant. However, considering that a decrement operation is to be performed on 'num_tri_corrs', it is necessary to ensure it is greater than 0 before decrementing.


ahojnnes:
Thanks.

PR #2252, opened 2023-11-02
Setting clear_points to true per default in point_triangulator

tsattler:
Making sure that image_ids between a model and the database are consistent is a pain. The `clear_points` flag takes care of this by assigning the correct image_ids to images (consistent with the database) based on filenames. This makes things much easier for casual users, who don't need to read out the image_ids from the database. I would argue that enabling this by default is the better choice (especially since in many cases, one would re-triangulate from scratch anyways).

The PR sets the default value to true. It also adds a bit of documentation about the behavior described above to the `--help` argument.

If you prefer to keep `clear_points` set to false by default, I would suggest to add another flag that makes recomputing image_ids default behavior.


ahojnnes:
Thanks, that makes sense!

PR #2288, opened 2023-12-03
Use minimal solvers from poselib

ahojnnes:


CMakeLists.txt
```diff
@@ -272,8 +272,8 @@ endif()
 configure_file("${CMAKE_CURRENT_SOURCE_DIR}/cmake/CMakeUninstall.cmake.in"
                "${CMAKE_CURRENT_BINARY_DIR}/CMakeUninstall.cmake"
                IMMEDIATE @ONLY)
-add_custom_target(uninstall COMMAND ${CMAKE_COMMAND} -P ${CMAKE_CURRENT_BINARY_DIR}/CMakeUninstall.cmake)
-set_target_properties(uninstall PROPERTIES FOLDER ${CMAKE_TARGETS_ROOT_FOLDER})
+add_custom_target(uninstall_colmap COMMAND ${CMAKE_COMMAND} -P ${CMAKE_CURRENT_BINARY_DIR}/CMakeUninstall.cmake)
```
ahojnnes, line 275: Need to resolve conflicting target in poselib.

ahojnnes: @vlarsson Maybe we can call the poselib one "uninstall_poselib" for consistency?

vlarsson: Sure. I setup a PR here https://github.com/PoseLib/PoseLib/pull/86 that should be merged to master soon.
@pablospe setup the CMake for poselib and I don't really know exactly how it works, so it's probably best if he has a look as well :)

pablospe: Interesting problem. I will see there is a different solution. 

I was thinking we could check in poselib if the `uninstall` is already defined:
```
# Check if the uninstall target already exists
if(NOT TARGET uninstall)
    # Define the uninstall target
    add_custom_target(uninstall
        COMMAND ...
    )
endif()
```
and it would be responsibly of the user (in this case, colmap, to uninstall poselib properly).

----

One question. How do you include `poselib` into `colmap`? FYI, Poselib supports being used as submodule (see [here](https://github.com/PoseLib/PoseLib#use-library-as-dependency-in-an-external-project)). But I think there is another alternative it is not included in poselib `README` yet but relevant for this case, which it is:
```
include(FetchContent)
FetchContent_Declare(
    poselib
    GIT_REPOSITORY <URL-of-PoseLib>
    GIT_TAG <desired-commit-or-tag>
    # Other options as necessary
)
FetchContent_MakeAvailable(poselib)
```

This is the default for google test (what [they recommend](https://google.github.io/googletest/quickstart-cmake.html#set-up-a-project) for external project after cmake 3.11). This will download the code during compilation time, avoiding the need to install it, which makes the uninstall irrelevant.

src/thirdparty/CMakeLists.txt
```diff
@@ -37,10 +37,21 @@ endif()
...
 add_subdirectory(VLFeat)
+
+include(FetchContent)
+FetchContent_Declare(PoseLib
+  GIT_REPOSITORY    https://github.com/PoseLib/PoseLib.git
+  GIT_TAG           c23af0c51924e3337f7bb193bb72899ec6dd4d3d
+)
+message(STATUS "Configuring PoseLib...")
+FetchContent_MakeAvailable(PoseLib)
+message(STATUS "Configuring PoseLib... done")
```
sarlinpe, line 55: @pablospe Here. IMHO the alternative (PoseLib as submodule) is less convenient because it requires users to install an additional dependency or to add `--recursive` when cloning colmap (which is easy to forget).

pablospe: yes, you are right and I agree. I was actually suggesting `FetchContent`.

pablospe: Well... actually thinking it could be an option and see the opinions.

pablospe: Viktor's PR has been merged into master, see https://github.com/PoseLib/PoseLib/pull/86

S-o-T: Please, consider a scenario where someone is working on modifications into PoseLib which are used/assessed in context of colmap and such work is performed on local machine.

If i understand correctly, in case of `FetchContent` (if we ignore relatively new `Dependency Providers` functionality) it would be necessary to modify a call to `FetchContent_Declare` in order to point to a local folder with modified PoseLib sources, also, for those who use IDE, working on such sources would require one more IDE context.

In case of PoseLib dependency satisfied via git submodule no "temporary" modifications to colmap's cmake scripts would be required and such work could be performed in single IDE context.

pablospe: I am curious about what you meant exactly by ignoring `Dependency Providers`. Maybe you can expand more on this, I was reading the doc and it seems that it is not ignored (but I am not sure what you refer):
https://cmake.org/cmake/help/latest/command/cmake_language.html#dependency-providers

My concern with `FetchContent` is how it plays with code autocomplete (like in vscode or another IDE), since you need to configure it first with cmake before the tools can understand the PoseLib code.

S-o-T: Sorry for a bit misleading wording, i have meant that one could use a `Dependency Providers` mechanism to override default behavior and to point cmake into standalone PoseLib instance which state could be preserved/shared between multiple configurations. The part about "ignoring" was to highlight the most probable action for a user facing the scenario i proposed to consider.

Regarding your concern about IDE features that becomes available only after cmake project is configured - i agree that this is another angle of the same issue.

ahojnnes:
On exact same two-view geometries, P3P from poselib vs. P3P from colmap doesn't seem to make any difference in practice:
```
$ ./src/colmap/exe/colmap model_analyzer --path ~/data/south-building/poselib/
I1216 13:33:07.592862 48175 model.cc:429] Cameras: 1
I1216 13:33:07.592973 48175 model.cc:430] Images: 128
I1216 13:33:07.592978 48175 model.cc:431] Registered images: 128
I1216 13:33:07.592980 48175 model.cc:433] Points: 61160
I1216 13:33:07.592981 48175 model.cc:434] Observations: 326980
I1216 13:33:07.592985 48175 model.cc:436] Mean track length: 5.346305
I1216 13:33:07.592995 48175 model.cc:438] Mean observations per image: 2554.531250
I1216 13:33:07.592999 48175 model.cc:441] Mean reprojection error: 0.512072px

$ ./src/colmap/exe/colmap model_analyzer --path ~/data/south-building/main/
I1216 13:33:13.860183 48237 model.cc:429] Cameras: 1
I1216 13:33:13.860323 48237 model.cc:430] Images: 128
I1216 13:33:13.860327 48237 model.cc:431] Registered images: 128
I1216 13:33:13.860330 48237 model.cc:433] Points: 61160
I1216 13:33:13.860332 48237 model.cc:434] Observations: 326980
I1216 13:33:13.860337 48237 model.cc:436] Mean track length: 5.346305
I1216 13:33:13.860347 48237 model.cc:438] Mean observations per image: 2554.531250
I1216 13:33:13.860350 48237 model.cc:441] Mean reprojection error: 0.512072px
```

ahojnnes:
Essentially identical results on E2E metrics when including poselib solvers in two-view geometry estimation:
```
joschonb@CPC-josch-OWPVG:~/dev/colmap/build$ ./src/colmap/exe/colmap model_analyzer --path ~/data/south-building/poselib-with-matching/
I1216 18:21:10.982481 109487 model.cc:429] Cameras: 1
I1216 18:21:10.982582 109487 model.cc:430] Images: 128
I1216 18:21:10.982586 109487 model.cc:431] Registered images: 128
I1216 18:21:10.982589 109487 model.cc:433] Points: 82357
I1216 18:21:10.982590 109487 model.cc:434] Observations: 482856
I1216 18:21:10.982595 109487 model.cc:436] Mean track length: 5.862962
I1216 18:21:10.982606 109487 model.cc:438] Mean observations per image: 3772.312500
I1216 18:21:10.982610 109487 model.cc:441] Mean reprojection error: 0.615324px
$ ./src/colmap/exe/colmap model_analyzer --path ~/data/south-building/main-with-matching/
I1216 18:21:16.408385 109543 model.cc:429] Cameras: 1
I1216 18:21:16.408489 109543 model.cc:430] Images: 128
I1216 18:21:16.408493 109543 model.cc:431] Registered images: 128
I1216 18:21:16.408496 109543 model.cc:433] Points: 82330
I1216 18:21:16.408499 109543 model.cc:434] Observations: 482931
I1216 18:21:16.408504 109543 model.cc:436] Mean track length: 5.865796
I1216 18:21:16.408514 109543 model.cc:438] Mean observations per image: 3772.898438
I1216 18:21:16.408517 109543 model.cc:441] Mean reprojection error: 0.615088px
```

ahojnnes:
Assuming unknown focal length, the results are also virtually identical. Overall reconstruction runtime is ~5% lower.
```
$ ./src/colmap/exe/colmap model_analyzer --path ~/data/south-building/main-with-focal-estimation/
I1217 15:56:08.626852 171591 model.cc:429] Cameras: 1
I1217 15:56:08.626966 171591 model.cc:430] Images: 128
I1217 15:56:08.626971 171591 model.cc:431] Registered images: 128
I1217 15:56:08.626973 171591 model.cc:433] Points: 82334
I1217 15:56:08.626976 171591 model.cc:434] Observations: 482940
I1217 15:56:08.626981 171591 model.cc:436] Mean track length: 5.865621
I1217 15:56:08.626991 171591 model.cc:438] Mean observations per image: 3772.968750
I1217 15:56:08.626996 171591 model.cc:441] Mean reprojection error: 0.615133px
$ ./src/colmap/exe/colmap model_analyzer --path ~/data/south-building/poselib-with-focal-estimation
I1217 15:56:00.155539 171526 model.cc:429] Cameras: 1
I1217 15:56:00.155656 171526 model.cc:430] Images: 128
I1217 15:56:00.155660 171526 model.cc:431] Registered images: 128
I1217 15:56:00.155663 171526 model.cc:433] Points: 82330
I1217 15:56:00.155665 171526 model.cc:434] Observations: 482936
I1217 15:56:00.155670 171526 model.cc:436] Mean track length: 5.865857
I1217 15:56:00.155681 171526 model.cc:438] Mean observations per image: 3772.937500
I1217 15:56:00.155685 171526 model.cc:441] Mean reprojection error: 0.615118px
```

vlarsson:
> Assuming unknown focal length, the results are also virtually identical. Overall reconstruction runtime is ~5% lower.
> 
> ```
> $ ./src/colmap/exe/colmap model_analyzer --path ~/data/south-building/main-with-focal-estimation/
> I1217 15:56:08.626852 171591 model.cc:429] Cameras: 1
> I1217 15:56:08.626966 171591 model.cc:430] Images: 128
> I1217 15:56:08.626971 171591 model.cc:431] Registered images: 128
> I1217 15:56:08.626973 171591 model.cc:433] Points: 82334
> I1217 15:56:08.626976 171591 model.cc:434] Observations: 482940
> I1217 15:56:08.626981 171591 model.cc:436] Mean track length: 5.865621
> I1217 15:56:08.626991 171591 model.cc:438] Mean observations per image: 3772.968750
> I1217 15:56:08.626996 171591 model.cc:441] Mean reprojection error: 0.615133px
> $ ./src/colmap/exe/colmap model_analyzer --path ~/data/south-building/poselib-with-focal-estimation
> I1217 15:56:00.155539 171526 model.cc:429] Cameras: 1
> I1217 15:56:00.155656 171526 model.cc:430] Images: 128
> I1217 15:56:00.155660 171526 model.cc:431] Registered images: 128
> I1217 15:56:00.155663 171526 model.cc:433] Points: 82330
> I1217 15:56:00.155665 171526 model.cc:434] Observations: 482936
> I1217 15:56:00.155670 171526 model.cc:436] Mean track length: 5.865857
> I1217 15:56:00.155681 171526 model.cc:438] Mean observations per image: 3772.937500
> I1217 15:56:00.155685 171526 model.cc:441] Mean reprojection error: 0.615118px
> ```

If you have a single camera, is the unknown focal length solver used? I assumed it would be running P3P once the camera is in the reconstruction? Also comparing to pycolmap at least, H matrix estimation was the problem where I saw the biggest improvement in terms of runtime. (iirc this is run for the geometric verification as well?) Might be worthwhile to replace this solver too.

ahojnnes:
@vlarsson I hardcoded to always use p4pf in the experiment above. I will take a look at replacing the homography solver. Thanks for the hint.

ahojnnes:
@vlarsson I double confirmed and my experimental results above are correct. In both cases, I forced focal length estimation for each camera from scratch. The runtime is still faster, because we don't have to run 20 P3Ps in parallel but only a single RANSAC. p4pf likely also works better than the sampling based approach when the initial guess of the focal length is very far off from the true focal length.

ahojnnes:
@vlarsson I noticed that your homograph solver doesn't normalize the image coordinates, which may not behave well for images with large image resolution due to squaring of pixel coordinates in the design matrix (e.g., for images with size of ~10'000 pixels, the matrix will have values of ~100M). Did you not observe any issues in this case?

vlarsson:
> @vlarsson I noticed that your homography solver doesn't normalize the image coordinates, which may not behave well for images with large image resolution due to squaring of pixel coordinates in the design matrix (e.g., for images with size of ~10'000 pixels, the matrix will have values of ~100M). Did you not observe any issues in this case?
> 
> **EDIT**: I did some quick and dirty experiments and it seems like the normalization of points is not necessary. I assume it accounts for a significant fraction of the overhead of the colmap homography solver...

I have not seen any issues with this either, but I have also not done extensive experiments on it :) My gut feeling is that normalisation does not matter so much for the minimal 4p problem, but is mostly important if you want to run DLT with more points.

src/thirdparty/CMakeLists.txt
```diff
@@ -37,10 +37,21 @@ endif()
...
 add_subdirectory(PoissonRecon)
 if(GPU_ENABLED)
     add_subdirectory(SiftGPU)
 endif()
 add_subdirectory(VLFeat)
+
+include(FetchContent)
+FetchContent_Declare(PoseLib
+  GIT_REPOSITORY    https://github.com/ahojnnes/PoseLib.git
+  GIT_TAG           33f7e865dae05124c0c34f1255fb33b0007f8176
```
ahojnnes, line 51: Change to latest commit tag in upstream repo.

ahojnnes:
@tsattler Do you have any datasets that could be suitable to evaluate the changes in this branch? I ran the changes against some of the standard datasets from UNC and it all looks good. Do you have anything more suitable like your Aachen dataset to check in particular the new p4pf solver against? Cheers.

tsattler:
@ahojnnes  Sure, I can do this. Can look into this in a week or so once I finished CVPR reviews and other things. If I don't reply in some time, can you please ping me?

ahojnnes:
Great, thank you. No urgency. I will try my best to remember :-)

tsattler:
Actually, playing with focal length solvers has been on my TODO list for a while. Is it only the P4Pf solver or also something that estimates the radial distortion?

ahojnnes:
Only focal length. I just swapped out the existing sampling+p3p with poselib's p4pf solver. As a next step, we could also try to estimate radial distortion but I wanted to keep this for a future pull request.

vlarsson:
@tsattler I am also interested in this. Also for the two-view case. Let's talk more :)

tsattler:
@vlarsson Sounds good. We can try to setup a benchmark for these methods together.

ahojnnes:
@tsattler Gentle reminder for the above 🙂😉

tsattler:
@ahojnnes Thanks for the reminder. It is on my TODO list and climbing higher in terms of priorities :) 
What are the most important parts to cover in terms of experiments? Which solvers? Which types of experiments?

ahojnnes:
I am quite confident about all solvers except for the new p4pf estimator. The Aachen dataset could be useful?

ahojnnes:
@vlarsson @tsattler I now tested the performance of this branch, in particular the p4pf solver in replacement of the sampling based approach, and it appears that p4pf performs worse than sampling on a number of datasets with unknown focal length from the internet. Given my current doubts about the quality of p4pf, I am probably going to go ahead with this PR but undo the p4pf changes.

ahojnnes:
For example, on the cornell arts quad dataset, I get the following with p4pf:
image
versus the following for colmap's sampling based approach:
image

It's not that either of them are perfect but the p4pf one is visibly worse.

S-o-T:
@ahojnnes i would like to one more time kindly ask to consider not using cmake's `FetchContent` as a mean to satisfy PoseLib dependency and rather integrate PoseLib either as a git submodule or as an externally provided library to be discovered via `find_package`.

ahojnnes:
@vlarsson I recall that you were interested in investigating the issue about p4pf. Did you have time to do that?

ahojnnes:
@S-o-T Consuming poselib as a submodule is not easily possible without a significant refactor of poselib itself due to the build rule and include folder structure. Consuming poselib as a package through find_package appears to have no benefit over the Fetch_Content approach in terms of being able to easily modify poselib sources while compiling colmap? Could you please clarify your proposal?

pablospe:
> Consuming poselib as a submodule is not easily possible without a significant refactor of poselib itself due to the build rule and include folder structure

I am wondering what exactly it is not working with poselib as submodule, I am interested to fix it. I remember making sure it was working some time ago:
https://github.com/PoseLib/PoseLib/pull/17

Is this perhaps because the pybind is now a submodule of poselib?

S-o-T:
@ahojnnes thanks for considering this issue.

> Consuming poselib as a package through find_package appears to have no benefit over the Fetch_Content approach in terms of being able to easily modify poselib sources while compiling colmap?

My approach to work with colmap is to use IDE that for each set of cmake options creates/uses individual build directories. Given that there are usually more then one set of cmake options that are actively worked with (e.g. Release\Debug, options controlling whether to enable/compile specific parts of colmap etc), each of this directories would contain its private copy of poselib downloaded by cmake via Fetch_Content during configure phase. Thus, actively switching between such configurations of colmap build, while working on modifications to poselib sources, becomes somewhat impossible (assuming no mechanism to automagically sync such changes). But if poselib would've been used as an internal third-party (via submodule)/external dependency, then i could work on changes into poselib independently of particular colmap cmake options set.

Anyway, i believe that this could be solved by a relatively small patch that i am capable to maintain locally, so i think that this should be addressed only if there are other users of colmap that might be facing the same issue.

vlarsson:
> @vlarsson I recall that you were interested in investigating the issue about p4pf. Did you have time to do that?

I should finally have some time to look into this second week of July. I also have a backlog of todos in PoseLib that I want to get into during the summer :)

vlarsson:
> For example, on the cornell arts quad dataset, I get the following with p4pf: image versus the following for colmap's sampling based approach: image
> 
> It's not that either of them are perfect but the p4pf one is visibly worse.

@ahojnnes Do still have the Arts Quad database? If so could you share it. It would save me some compute.

vlarsson:
I've done some initial testing. I ran the reconstruction of ArtsQuad (with the main branch version) and dumped all of the correspondences from the registration where estimate_focal_length was set. I put together a quick (LO-)RANSAC in poselib for unknown focal and it seems to work okay (at least metrics wise). 

image

So first two rows are standard sampling based estimator, before and after the RefineAbsolutePose call. Focal length errors are relative (%) to the "ground truth" focal length provided in the dataset (maybe they are from bundler?). 

I've tried some different things.
- Comparing different solvers 
- Vanilla RANSAC vs LO RANSAC
- Minimize reproj error on the minimal sample (MinRefine). Since it's over determined it will not satisfy the 4 points exactly.
- Filter minimal samples based on reprojection error on the 4 points (MinFilter).

So far the impact has been quite minor. I am not sure yet why your reconstruction was failing. The P4P_RSC above should be fairly close to what you had. I will try to integrate this and re-run the reconstruction with this.

ahojnnes:
Thanks very much @vlarsson for the experiments. This suggests your solvers should be superior to the sampling based approach. I am not sure where the input focal lengths are coming from. These might be extracted from EXIF information as well. I am looking forward to see how this works on the reconstruction. I am wondering whether the P4Pf/P5Pf approach sometimes fails catastrophically and thus leads to a few bad cameras that lead to the bad overall results. Such a behavior would not necessarily be captured by the stats you shared above? The sampling based approach is fundamentally limited to a fixed range of focal lengths and cannot blow up completely?

vlarsson:
The reconstruction with the poselib estimator looked okay, however the vanilla colmap reconstruction I got had larger artifacts compared to the result you posted above. I wonder if this scene is a bit unstable and maybe we should not draw too much conclusions from it. I will try to setup more controlled experiments to figure out what is actually working well.

Another thing I noted was that running the vanilla colmap reconstruction with SIMPLE_PINHOLE instead of SIMPLE_RADIAL completely broke the reconstruction. Once I get a more proper benchmark up and running I will look into adding distortion estimation as well.

> Thanks very much @vlarsson for the experiments. This suggests your solvers should be superior to the sampling based approach. I am not sure where the input focal lengths are coming from. These might be extracted from EXIF information as well. I am looking forward to see how this works on the reconstruction. I am wondering whether the P4Pf/P5Pf approach sometimes fails catastrophically and thus leads to a few bad cameras that lead to the bad overall results. Such a behavior would not necessarily be captured by the stats you shared above? The sampling based approach is fundamentally limited to a fixed range of focal lengths and cannot blow up completely?

This seems reasonable to me. The sampling is essentially a form of regularization on the focal lengths. I think doing some basic filtering on focal lengths in the minimal solver could have a similar effect.  I will look into this as well.

vlarsson:
Some updates on this. To setup a more structured benchmark I downloaded 432 scenes from MegaScenes that had at least 100 images registered. I dropped 25% of the images, retriangulated the map and re-registered the dropped images with image_registrator, dumping the correspondences and the output of COLMAPs EstimateAbsolutePose and RefineAbsolutePose, forcing focal length estimation for all images. This gave me ~32k query images. The ground truth is MegaScenes is far from perfect though and there are many broken reconstructions, but hopefully it contains some signal.

After a lot of experimentation and trying a bunch of different things, my take aways so far has been

- P3.5Pf is slower than P4Pf but seems to have slightly less failures
- P5Pf is fast and gives good solutions when it works, but images with few inlier matches (low number of matches and low inlier ratio) it can struggle
- For some reason I don't understand inlier counting seems more robust than MSAC scoring for this problem. I ended up doing a hybrid scoring with simply summing MSAC and outlier count (see here https://github.com/PoseLib/PoseLib/blob/absolute_pose_unknown_focal_clean/PoseLib/robust/estimators/absolute_pose.cc#L124)
- Filtering minimal samples with poor reprojection error was sometimes better and sometimes worse
- Same with refinement on the minimal sample (especially for P5Pf it seems that this breaks sometimes)

I put a cleaned up version of the robust estimator in poselib here https://github.com/PoseLib/PoseLib/tree/absolute_pose_unknown_focal_clean

For this version we get the following metrics on the MegaScenes queries
image

Metrics: Focal length errors are relative (%) and we look at AUC@{5,10,20}%. Rotation error is degrees. We don't look at translations as we done have any scale. For inlier ratios we also show quantiles for 5%, 1% and 0.1%, to look at the hardest instances. Since the GT is a bit fishy, I think inlier ratio and MSAC score are the most meaninful.

So here `poselib-pnpf` should be compared with `COLMAP_EAP` which is the sampling based estimator in COLMAP. The cameras in the reconstruction are SIMPLE_RADIAL, so I ran COLMAP's RefineAbsolutePose (`RAP`) afterwards, as is also done during registration (as solvers only give focal length).  `poselib-pnp` is running calibrated PnP with the "GT" camera intrinsics.

I've integrated the latest version into COLMAP here 
https://github.com/vlarsson/colmap/tree/users/vlarsson/poselib-solvers
and I am now re-running the ArtsQuad reconstruction.

Based on the large difference between the initial and the +RAP versions above I think it might be necessary to integrate distortion estimation into the RANSAC as well.

src/thirdparty/CMakeLists.txt
ahojnnes: I now converged to introducing a "FETCH_POSELIB" option, which allows to consume poselib using find_package.

B1ueber2y (approved):
Thanks. LGTM from my side.

src/colmap/estimators/pose.cc
```diff
@@ -84,83 +51,46 @@ bool EstimateAbsolutePose(const AbsolutePoseEstimationOptions& options,
...
+  custom_ransac_options.max_error =
+      camera->CamFromImgThreshold(options.ransac_options.max_error);
 
-  // Scale output camera with best estimated focal length.
-  if (options.estimate_focal_length && *num_inliers > 0) {
-    for (const size_t idx : camera->FocalLengthIdxs()) {
-      camera->params[idx] *= focal_length_factor;
+  if (options.estimate_focal_length) {
+    // TODO(joschonb): Implement non-minimal solver for LORANSAC refinement.
+    RANSAC<P4PFEstimator> ransac(custom_ransac_options);
```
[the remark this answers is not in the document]
ahojnnes: This should always be called by the RefineAbsolutePose, which refines on all inliers. As I understood your earlier experiments, this yielded essentially identical results to using LORANSAC?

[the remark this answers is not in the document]
vlarsson: In the initial experiments on ArtsQuad instances this seemed to be the case, but in your initial tests you had failures there and I am not sure how this would be different. I did not run this exact comparison on the larger dataset (from MegaScenes). I can quickly see if I can get these numbers.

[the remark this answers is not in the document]
ahojnnes: Thank you for the experiment. Above, I thought that poselib-pnpf+RAP vs colmap-EAP+RAP would be essentially a comparison of this branch vs main. Probably, I misunderstood. Let's see your numbers to be sure.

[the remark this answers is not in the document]
vlarsson: Ah yes, so poselib-pnpf is with LO-RANSAC.

Here are the results without LO in the RANSAC. So poselib-pnpf and poselib-pnpf-NO_LO are exactly the same exact in each call for LO it just returns without doing anything.

image

I think the ransac parameters might've been different in this run from the last (as they are all a bit slower), but I guess it should not matter for the comparison.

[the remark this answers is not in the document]
ahojnnes: Thank you very much. With these experiments, I conclude that we can move ahead with the current code and leave further improvements to the future. It will become easier once you merge your changes in the poselib dev branch to main. Then, we can consider directly moving to the higher-level poselib::estimate_absolute_pose functions, even though we need to be careful because of possible discrepancies of the camera models.

PR #2300, opened 2023-12-18
Replace result_of with invoke_result_t

sarlinpe:
`std::result_of` is being deprecated in C++17 and removed in C++20 https://en.cppreference.com/w/cpp/types/result_of

### DIFF
--- a/src/colmap/util/threading.h
+++ b/src/colmap/util/threading.h
@@ -43,13 +43,6 @@
 
 namespace colmap {
 
-template <class func_t, class... args_t>
-#ifdef __cpp_lib_is_invocable
-using result_of_t = std::invoke_result_t<func_t, args_t...>;
-#else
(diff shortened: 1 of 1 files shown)


src/colmap/util/threading.h
```diff
@@ -38,10 +38,18 @@
...
+#include <type_traits>
 #include <unordered_map>
 
 namespace colmap {
 
+template <class func_t, class... args_t>
+#ifdef __cpp_lib_is_invocable
+using result_of_t = std::invoke_result_t<func_t, args_t...>;
+#else
+using result_of_t = typename std::result_of<func_t(args_t...)>::type;
```
ahojnnes, line 50: using in the header will make this name available in any file that includes this header. I suggest to move this under the ThreadPool class below. thx

PR #2301, opened 2023-12-18
Allow getters FocalLength{X,Y} for isotropic models

sarlinpe:
Simplify the higher level logic, especially for Python users (https://github.com/colmap/pycolmap/issues/199).

By the way: what is the rationale for using debug checks `DCHECK_*` instead of `CHECK_*` in `scene/camera.h` getters and setters?

### DIFF
--- a/src/colmap/scene/camera.h
+++ b/src/colmap/scene/camera.h
@@ -160,11 +160,7 @@ double Camera::FocalLengthX() const {
 
 double Camera::FocalLengthY() const {
   const span<const size_t> idxs = FocalLengthIdxs();
-  size_t idx = 1;
-  if (idxs.size() == 1) {
-    idx = 0;
-  }
-  return params[idxs[idx]];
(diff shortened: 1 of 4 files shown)


src/colmap/scene/camera.h
```diff
@@ -155,14 +155,16 @@ double Camera::FocalLength() const {
...
 
 double Camera::FocalLengthY() const {
   const span<const size_t> idxs = FocalLengthIdxs();
-  DCHECK_EQ(idxs.size(), 2);
-  return params[idxs[1]];
+  size_t idx = 1;
+  if (idxs.size() == 1) {
+    idx = 0;
+  }
+  return params[idxs[idx]];
```
ahojnnes, line 167:
```suggestion
  return params[idxs[(idxs.size() == 1) ? 0 : 1]];
```

ahojnnes:
DCHECK has no runtime overhead when compiling in release (non-debug) mode.

PR #2321, opened 2024-01-06
Replace std::string with constant chars

whuaegeanse:
This PR is probably trivial and won't affect performance.
Since COLMAP_VERSION, COLMAP_COMMIT_ID, COLMAP_COMMIT_DATE, etc. are all string constants, there is no need to use std::string.


ahojnnes:
Thanks!

PR #2330, opened 2024-01-07
Fix warnings

whuaegeanse:
This PR attempts to fix warnings in bitmap, timer, ply, misc, sift and other files.


ahojnnes:
I assume this is an attempt to resolve the visual studio warnings about type truncation? I honestly question the value of these warnings. Just adding a static_cast isn't much better than silencing the warning altogether (imho)? The same cast is already happening implicitly anyway?

whuaegeanse:
> I assume this is an attempt to resolve the visual studio warnings about type truncation?

Yes.

> Just adding a static_cast isn't much better than silencing the warning altogether (imho)?
> The same cast is already happening implicitly anyway?

Here is an [answer](https://stackoverflow.com/questions/4497896/should-you-correct-compiler-warnings-about-type-conversions-using-explicit-typec) to a similar question, maybe we can refer to it.

```
Several reasons:

*) It may be the cause of an error and need an actual fix rather than just a cast. You won't know until you look.

*) Actual coding errors that manifest as warnings can get lost in the noise generated by hundreds of warnings

*) It makes it clear to other coders that you really did mean to use that variable of the wrong type/sign there. That it is deliberate.

*) It makes it clear and explicit that the type and/or signedness is being changed. If your variable names do not contain an indication of the type and signedness it may not be obvious that this is occurring.
```

src/colmap/util/endian.h
```diff
@@ -131,6 +131,26 @@ T NativeToBigEndian(const T x) {
   }
 }
 
+template <typename T1, typename T2>
+T2 LittleEndianToNativeCast(const T1 x) {
```
ahojnnes, line 135: I suggest to not introduce these new helpers but instead apply the static_cast explicitly on the call site.

ahojnnes:
Thanks for the response. I agree that we should take warnings seriously and fix them. In particular for the discussed narrowing type cast warnings, I am personally a little skeptical about their value. For example, GCC/Clang decided to not include these by default in `-Wall` and `-Wpedantic` but MSVC does. The changes in this PR are a net positive, so we should merge them, yet I am posing the question whether fixing the same issues across the codebase is a good investment of time.

PR #2344, opened 2024-01-11
Support decoding multi-byte characters in Python script

jot-jt:
This fixes an issue where read_images_binary() throws an exception when decoding image names with multi-byte characters.

UTF-8 encodes characters using one to four bytes, so decoding byte-by-byte sometimes won't work if a character is encoded with more than one byte. For instance, if a character is encoded using three bytes, all three bytes need to be passed into `.decode("utf-8")` for it to return the proper character. This pull request decodes the entire image name at once to avoid this issue.


ahojnnes:
Thanks!

PR #2347, opened 2024-01-12
Split Dockerfile in two stages: builder and runtime.

pablospe:
Splitting the Dockerfile in two stages reduces the size of the final image considerably, since all the build dependencies are not needed.

You can still create the tag for the stages separately:
```bash
docker build --target builder -t colmap:builder .
docker build --target runtime -t colmap:runtime .
```
But the default command (without `--target`) is the last stage, so:
```bash
docker build -t colmap:latest .
```
will refer to `--target runtime` because it is the last stage in the Dockerfile.

### DIFF
--- a/docker/Dockerfile
+++ b/docker/Dockerfile
@@ -1,43 +1,80 @@
-FROM nvidia/cuda:12.2.2-devel-ubuntu22.04
+ARG UBUNTU_VERSION=22.04
+ARG NVIDIA_CUDA_VERSION=12.3.1
+
+#
+# Docker builder stage.
+#
+FROM nvidia/cuda:${NVIDIA_CUDA_VERSION}-devel-ubuntu${UBUNTU_VERSION} as builder
 
 ARG COLMAP_GIT_COMMIT=main
-ARG CUDA_ARCHITECTURES=native
+ARG CUDA_ARCHITECTURES=50
 ENV QT_XCB_GL_INTEGRATION=xcb_egl
 
-# Prevent stop building ubuntu at time zone selection.  
+# Prevent stop building ubuntu at time zone selection.
 ENV DEBIAN_FRONTEND=noninteractive
 
 # Prepare and empty machine for building.
-RUN apt-get update && apt-get install -y \
-    git \
-    cmake \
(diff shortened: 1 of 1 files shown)


docker/Dockerfile
```diff
@@ -1,43 +1,80 @@
...
+ARG NVIDIA_CUDA_VERSION=12.3.1
+
+#
+# Docker builder stage.
+#
+FROM nvidia/cuda:${NVIDIA_CUDA_VERSION}-devel-ubuntu${UBUNTU_VERSION} as builder
 
 ARG COLMAP_GIT_COMMIT=main
-ARG CUDA_ARCHITECTURES=native
+ARG CUDA_ARCHITECTURES=50
```
pablospe, line 10: Should I leave `ARG CUDA_ARCHITECTURES=native`? It didn't work when I run it myself without the calls from the yaml file.

docker/Dockerfile
```diff
@@ -1,43 +1,80 @@
...
+        libfreeimage3 \
+        libgcc-s1 \
+        libgl1 \
+        libglew2.2 \
+        libgoogle-glog0v5 \
+        libqt5core5a \
+        libqt5gui5 \
+        libqt5widgets5
+
+COPY --from=builder /colmap/bin/colmap /colmap/bin/colmap
```
pablospe, line 79: @ahojnnes A follow-up question I had was whether I should have copied the entire installation directory. I only copied the binary from the `builder` stage. To clarify:
```
COPY --from=builder /colmap/ /colmap/
```

pablospe:
FYI, this PR reduces the docker image ~3GB :)

image

(image from https://hub.docker.com/r/colmap/colmap)

docker/Dockerfile
pablospe: I have addressed this comment in PR https://github.com/colmap/colmap/pull/2356

docker/Dockerfile
pablospe: I have addressed this comment in PR https://github.com/colmap/colmap/pull/2356

PR #2364, opened 2024-01-25
Simplify the creation of reprojection error cost functions

sarlinpe:


ahojnnes:
Great simplification, thank you.

PR #2365, opened 2024-01-26
Update VCPKG commit in Windows CI

sarlinpe:
The build of `jasper` is broken, a patch has been recently applied in vcpkg.


ahojnnes:
Thanks!

PR #2367, opened 2024-01-26
Migrate pycolmap

sarlinpe:
- I would prefer to keep the CI running on GitHub actions for now and move to Azure Pipelines only later. To minimize the binary sizes we only build a subset of COLMAP features (no GUI, no CGAL) so we anyway need separate build job. With caching it takes <20min to run.
- The code mostly follows conventions of COLMAP and passes all clang-tidy checks except `google-build-using-namespace` - we need to use `using namespace colmap;` because pybind doesn't play well with namespaces.
  - I will add a CI job for Clang-Tidy in a separate PR with `-checks=-google-build-using-namespace*` 

Will be addressed in later PRs:
- [ ] Move exception checks to COLMAP
- [ ] Bind `SetPRNGSeed` and remove its calls in `pycolmap/estimators`
- [ ] Replace calls to FreeImage in `pycolmap/feature/sift.h`
- [ ] CI with Clang-Tidy for pycolmap
- [ ] More tests

cc @mihaidusmanu @Phil26AT @tsattler


pycolmap/CMakeLists.txt
```diff
@@ -0,0 +1,27 @@
+cmake_minimum_required(VERSION 3.10)
+project(${SKBUILD_PROJECT_NAME} VERSION ${SKBUILD_PROJECT_VERSION})
+
+set(CMAKE_CUDA_ARCHITECTURES "native")
```
ahojnnes, line 4: Do we really want to hardcode this to native? This prevents us from build redistributable binaries (or wheels) that run on other architectures than the current machine?

pycolmap/CMakeLists.txt
```diff
@@ -0,0 +1,27 @@
...
+
+find_package(colmap REQUIRED)
+
+find_package(Python REQUIRED COMPONENTS Interpreter Development.Module)
+
+find_package(pybind11 REQUIRED)
+
+pybind11_add_module(pycolmap pycolmap/main.cc)
+target_include_directories(pycolmap PRIVATE ${PROJECT_SOURCE_DIR})
+target_link_libraries(pycolmap PRIVATE colmap::colmap freeimage::FreeImage glog::glog Ceres::ceres)
```
ahojnnes, line 25: Do we really need to explicitly link against freeimage glog ceres? I would have expected that these are propagated from colmap?

pycolmap/README.md
```diff
@@ -0,0 +1,296 @@
...
+# - options: dict or pycolmap.TwoViewGeometryOptions
+answer = pycolmap.estimate_calibrated_two_view_geometry(camera1, points1, camera2, points2)
+# Returns: pycolmap.TwoViewGeometry
+```
+
+The `TwoViewGeometryOptions` control how each model is selected. The output structure contains the geometric model, inlier matches, the relative pose (if `options.compute_relative_pose=True`), and the type of camera configuration, which is an instance of the enum `pycolmap.TwoViewGeometryConfiguration`.
+
+### Camera argument
+
+Some estimators expect a COLMAP camera object, which can be created as follow:
```
ahojnnes, line 238: nit: follows

pycolmap/ci/install-centos.sh
```diff
@@ -0,0 +1,64 @@
+#!/bin/bash
```
ahojnnes, line 1: nit: these scripts only install colmap but not pycolmap. Maybe we call the scripts install-colmap-centos.sh etc.?

pycolmap/pycolmap/estimators/alignment.h
```diff
@@ -0,0 +1,144 @@
...
+      "min_inlier_ratio"_a = 0.9);
+
+  m.def(
+      "align_reconstrution_to_locations",
+      [](const Reconstruction& src,
+         const std::vector<std::string>& image_names,
+         const std::vector<Eigen::Vector3d>& locations,
+         const int min_common_images,
+         const RANSACOptions& ransac_options) {
+        THROW_CHECK_GE(min_common_images, 3);
```
ahojnnes, line 98: I guess we still want to unify CHECKs to be exceptions also in COLMAP?

ahojnnes: Not for this PR, just aligning on remaining steps.

pycolmap/pycolmap/estimators/generalized_absolute_pose.h
```diff
@@ -0,0 +1,95 @@
...
+py::object PyEstimateAndRefineGeneralizedAbsolutePose(
+    const std::vector<Eigen::Vector2d>& points2D,
+    const std::vector<Eigen::Vector3d>& points3D,
+    const std::vector<size_t>& camera_idxs,
+    const std::vector<Rigid3d>& cams_from_rig,
+    std::vector<Camera>& cameras,
+    const RANSACOptions& ransac_options,
+    const AbsolutePoseRefinementOptions& refinement_options,
+    const bool return_covariance) {
+  SetPRNGSeed(0);
```
ahojnnes, line 28: This may be something that we want to set outside and not hardcode to a fixed seed?

pycolmap/pycolmap/feature/sift.h
```diff
@@ -0,0 +1,118 @@
...
+
+  sift_output_t Extract(const Eigen::Ref<const pyimage_t<uint8_t>>& image) {
+    THROW_CHECK_LE(image.rows(), options_.max_image_size);
+    THROW_CHECK_LE(image.cols(), options_.max_image_size);
+
+    const unsigned int bpp = 8;  // Grey.
+    const unsigned int width = image.cols();
+    const unsigned int scan_width = (bpp / 8) * width;
+    pyimage_t<uint8_t> image_copy = image;
+    FIBITMAP* bitmap_raw = FreeImage_ConvertFromRawBitsEx(
```
ahojnnes, line 49: It may be nicer to add a `Bitmap(uint8_t* data, int width, int height, bool as_rgb);` constructor, then you don't have to deal with freeimage in the pycolmap layer? I already changed the bitmap.h interface to fully encapsulate freeimage.

pycolmap/pycolmap/feature/sift.h
```diff
@@ -0,0 +1,118 @@
...
+      keypoints(i, 0) = keypoints_[i].x;
+      keypoints(i, 1) = keypoints_[i].y;
+      keypoints(i, 2) = keypoints_[i].ComputeScale();
+      keypoints(i, 3) = keypoints_[i].ComputeOrientation();
+    }
+
+    descriptors_t descriptors = descriptors_.cast<float>();
+    descriptors /= 512.0f;
+
+    return std::make_tuple(keypoints, descriptors);
```
ahojnnes, line 79: return `std::make_tuple(std::move(keypoints), std::move(descriptors))` will avoid copying the data.

pycolmap/pycolmap/geometry/homography_matrix.h
```diff
@@ -0,0 +1,41 @@
+#include "colmap/geometry/homography_matrix.h"
```
ahojnnes, line 1: Do we need a pragma once here and elsewhere?

pycolmap/pycolmap/log_exceptions.h
```diff
@@ -0,0 +1,147 @@
+#pragma once
```
ahojnnes, line 1: As a follow-up, we should probably move this file to colmap and replace any glog CHECK* macros with THROW_CHECK*? I can try to look into this as a follow-up.

pycolmap/pycolmap/scene/database.h
```diff
@@ -0,0 +1,47 @@
...
+      .def("pair_id_to_image_pair", &Database::PairIdToImagePair)
+      .def("read_camera", &Database::ReadCamera)
+      .def("read_all_cameras", &Database::ReadAllCameras)
+      .def("read_image", &Database::ReadImage)
+      .def("read_image_with_name", &Database::ReadImageWithName)
+      .def("read_all_images", &Database::ReadAllImages)
+      // ReadKeypoints
+      // ReadDescriptors
+      // ReadMatches
+      // ReadAllMatches
```
ahojnnes, line 38: These are TODO? If so, maybe add a TODO() comment?

scripts/format/clang_format.sh
```diff
@@ -36,7 +36,7 @@ fi
 root_folder=$(git rev-parse --show-toplevel)
 all_files=$( \
     git ls-tree --full-tree -r --name-only HEAD . \
-    | grep "^src/colmap.*\(\.cc\|\.h\|\.hpp\|\.cpp\|\.cu\)$" \
+    | grep "^\(src/colmap\|pycolmap/src\).*\(\.cc\|\.h\|\.hpp\|\.cpp\|\.cu\)$" \
```
ahojnnes, line 39: In terms of folder hierarchy, it can be nicer to do src/pycolmap instead of pycolmap/src?

ahojnnes (commented):
Thanks very much for this monumental effort. Instead of trying to migrate from Github CI to ADO pipelines, we could also try to convert the existing ones to github. Both run on the same machines (afaik) in the end but the Github ones integrate more nicely with direct log output, etc.

ahojnnes:
We should probably also setup vcpkg binary caching?

pycolmap/CMakeLists.txt
sarlinpe: Could we instead make it the default in `FindDependencies.cmake`? Most users that compile from source do so for their own machine only so they don't need to known about this flag.
https://github.com/colmap/colmap/blob/fb09740a390ea01806dc3945c18d031fa7814145/cmake/FindDependencies.cmake#L124-L128
```diff
    if(NOT DEFINED CMAKE_CUDA_ARCHITECTURES)
-        message(
-            FATAL_ERROR "You must set CMAKE_CUDA_ARCHITECTURES to e.g. 'native', 'all-major', '70', etc. "
-            "More information at https://cmake.org/cmake/help/latest/prop_tgt/CUDA_ARCHITECTURES.html")
+        set(CMAKE_CUDA_ARCHITECTURES "native")
    endif()
```

pycolmap/CMakeLists.txt
sarlinpe: They do propagate but we state them explicitly because they are directly used in pycolmap as well (not only through colmap internals).

pycolmap/pycolmap/scene/database.h
sarlinpe: Removed.

pycolmap/pycolmap/feature/sift.h
sarlinpe: Good catch, done.

pycolmap/pycolmap/estimators/generalized_absolute_pose.h
sarlinpe: I suggest to bind address this in a later PR: bind `SetPRNGSeed` and remove its calls in `pycolmap/estimators`

pycolmap/pycolmap/estimators/alignment.h
sarlinpe: Yes, will address in a later PR

pycolmap/pycolmap/geometry/homography_matrix.h
sarlinpe: Most files are meant to be included only once so I've added pragma once only on helpers that are included multiple times - is that ok?

pycolmap/pycolmap/log_exceptions.h
sarlinpe: Yes definitely, will address this in a later PR.

scripts/format/clang_format.sh
sarlinpe: Good catch, this change is incorrect, the code actually lives in `pycolmap/pycolmap`. It could indeed be nice to have it in `src/pycolmap/` to be closer to colmap, but I don't know how to configure CMake such that includes look like `"pycolmap/....h"` instead of `"src/pycolmap/....h"` while having the `CMakeLists.txt` in `pycolmap/` (it shouldn't conflict with COLMAP's main `CMakeLists.txt`)

pycolmap/pycolmap/feature/sift.h
sarlinpe: Yes this definitely needs a cleanup. We can address this in a later PR.

pycolmap/CMakeLists.txt
ahojnnes: Sure, we can.

pycolmap/pycolmap/estimators/generalized_absolute_pose.h
ahojnnes: Ok, wasn't aware this is not exposed through pycol map. Let's do it as a follow-up.

pycolmap/pycolmap/geometry/homography_matrix.h
ahojnnes: I would still add the pragma once. It doesn't hurt?

scripts/format/clang_format.sh
ahojnnes: I'll try to have a look later when I have access to a computer.

pycolmap/CMakeLists.txt
ahojnnes: got it.

scripts/format/clang_format.sh
ahojnnes: OK, maybe stupid idea, but how about src/pycolmap/pycolmap ?

ahojnnes: `target_include_directories(pycolmap PRIVATE ${PROJECT_SOURCE_DIR}/..)` doesn't work?

ahojnnes: I also feel like we could eventually just merge the two CMakeLists.txt and just have a "PYTHON_ENABLED" cmake flag that then automatically includes the pycolmap target?

.github/workflows/build-pycolmap.yml
```diff
@@ -0,0 +1,153 @@
...
+          # Fix: cibuildhweel cannot interpolate env variables.
+          CONFIG_SETTINGS="cmake.define.CMAKE_TOOLCHAIN_FILE=${CMAKE_TOOLCHAIN_FILE}"
+          CONFIG_SETTINGS="${CONFIG_SETTINGS} cmake.define.VCPKG_TARGET_TRIPLET=${VCPKG_TARGET_TRIPLET}"
+          CONFIG_SETTINGS="${CONFIG_SETTINGS} cmake.define.CMAKE_OSX_ARCHITECTURES=${CMAKE_OSX_ARCHITECTURES}"
+          echo "CIBW_CONFIG_SETTINGS_MACOS=${CONFIG_SETTINGS}" >> "$GITHUB_ENV"
+
+          # vcpkg binary caching
+          VCPKG_CACHE_DIR="${COMPILER_CACHE_DIR}/vcpkg"
+          VCPKG_BINARY_SOURCES="clear;files,${VCPKG_CACHE_DIR},readwrite"
+          echo "VCPKG_BINARY_SOURCES=${VCPKG_BINARY_SOURCES}" >> "$GITHUB_ENV"
```
sarlinpe, line 66: @ahojnnes We do have binary caching in VCPKG, managed explicitly through the cache. There is no cleanup mechanism so it size might grow over time if we update dependencies. We could instead use NuGet (but I don't know how to set this up) or the [GitHub action cache](https://learn.microsoft.com/en-us/vcpkg/users/binarycaching#gha) (which is cleaner but it's not well documented).

https://learn.microsoft.com/en-us/vcpkg/github-integration

sarlinpe:
PTAL. I would also prefer to have everything under GitHub CI instead of ADO, it looks easier to manage and better integrated. We do already have binary caching for vcpkg, see comment above.

pycolmap/CMakeLists.txt
```diff
@@ -0,0 +1,27 @@
...
+    set(CMAKE_C_FLAGS "${CMAKE_C_FLAGS} /MP")
+    set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} /MP")
+endif()
+
+
+find_package(colmap REQUIRED)
+
+find_package(Python REQUIRED COMPONENTS Interpreter Development.Module)
+
+find_package(pybind11 REQUIRED)
```
ahojnnes, line 21: It seems like we need a minimum pybind11 version. I just tried with default packages from apt under ubuntu2204 and getting various C++ compiler errors. It should be possible to directly enforce a minimum version here in the find_package call?

pycolmap/pycolmap/geometry/homography_matrix.h
sarlinpe: Done.

pycolmap/CMakeLists.txt
sarlinpe: The version of pybind is pinned in the pyproject.toml and gets installed with pip when running `python -m pip install .` - did you install pycolmap in a different way?

ahojnnes: I was just manually configuring CMake and installing pybind through apt.

sarlinpe: Pinned the pybind version in CMakeLists as well to avoid this issue.

pycolmap/CMakeLists.txt
sarlinpe: Done, updated the doc as well.

scripts/format/clang_format.sh
sarlinpe: Moved the pycolmap code to `src/`, changing `target_include_directories` indeed worked.

sarlinpe: As for merging the two CMakeLists: the build process of pycolmap is driven by scikit-build-core, which takes care of passing the right Python dependencies, calling cmake, installing to site-packages, etc. So to build colmap and pycolmap together at once we'd need to have it done by scikit-build-core. At the moment I don't see much benefits to this, so we could see in a later PR?

sarlinpe:
Feel free to merge if you're happy with it.

.github/workflows/build-pycolmap.yml
ahojnnes: OK, something to look into as a follow-up.

scripts/format/clang_format.sh
ahojnnes: Yes, I wasn't aware that the pycolmap build was not fully driven by cmake, so this makes sense.

PR #2371, opened 2024-01-28
Switch to exception checks

sarlinpe:
Use custom checks (written by @Phil26AT) that are Python friendly by throwing an exception instead of killing the program.

- [x] Replace `CHECK*` for the entire COLMAP
- [x] Replace `LOG(FATAL)` with exceptions
- [ ] Remove the unnecessary checks in pycolmap (in a future PR)


src/colmap/controllers/automatic_reconstruction.cc
```diff
@@ -49,9 +50,9 @@ AutomaticReconstructionController::AutomaticReconstructionController(
     : options_(options),
       reconstruction_manager_(std::move(reconstruction_manager)),
       active_thread_(nullptr) {
-  CHECK(ExistsDir(options_.workspace_path));
-  CHECK(ExistsDir(options_.image_path));
-  CHECK_NOTNULL(reconstruction_manager_);
+  THROW_CHECK(ExistsDir(options_.workspace_path));
+  THROW_CHECK(ExistsDir(options_.image_path));
+  THROW_CHECK(reconstruction_manager_);
```
ahojnnes, line 55: THROW_CHECK_NOTNULL

src/colmap/controllers/bundle_adjustment.cc
```diff
@@ -66,7 +67,7 @@ BundleAdjustmentController::BundleAdjustmentController(
     : options_(options), reconstruction_(std::move(reconstruction)) {}
 
 void BundleAdjustmentController::Run() {
-  CHECK_NOTNULL(reconstruction_);
+  THROW_CHECK(reconstruction_);
```
ahojnnes, line 70: NOTNULL

src/colmap/util/logging.h
```diff
@@ -101,4 +139,86 @@ bool __CheckOptionOpImpl(const char* file,
...
+void __ThrowCheckOpImpl(const char* file,
+                        const int line,
+                        const bool result,
+                        const T1& val1,
+                        const T2& val2,
+                        const char* val1_str,
+                        const char* val2_str,
+                        const char* op_str) {
+  if (!result) {
+    // TODO: cleanup
```
ahojnnes, line 214: There are various cleanup TODOs in the new code. What is the specific cleanup that is still left?

src/colmap/controllers/automatic_reconstruction.cc
sarlinpe: The current code inconsistently uses `CHECK` and `CHECK_NOTNULL` for pointers. Should we switch all to `CHECK_NOTNULL` for consistency?

ahojnnes: OK, I was just looking at the first file and saw the discrepancy.... :-) Unless the THROW_CHECK_NOTNULL would be much more informative, it doesn't really matter imo.

src/colmap/util/logging.h
sarlinpe: I've replaced the string streams with string concatenation, which should be faster.

src/colmap/image/warp.cc
```diff
@@ -57,9 +57,11 @@ void WarpImageBetweenCameras(const Camera& source_camera,
                              const Camera& target_camera,
                              const Bitmap& source_image,
                              Bitmap* target_image) {
-  CHECK_EQ(source_camera.width, source_image.Width());
-  CHECK_EQ(source_camera.height, source_image.Height());
-  CHECK_NOTNULL(target_image);
+  THROW_CHECK_EQ(source_camera.width,
+                 static_cast<size_t>(source_image.Width()));
```
sarlinpe, line 61: @ahojnnes  I had to fix quite a lot of these annoying warnings. Any reason why `Bitmap::{Width,Height}` return `int` instead of `size_t` ?

src/colmap/util/logging.h
```diff
@@ -61,6 +63,66 @@
...
+#define THROW_CHECK_NE_MSG(val1, val2, msg) \
+  THROW_CHECK_OP_MSG(!=, val1, val2, msg)
+#define THROW_CHECK_LE_MSG(val1, val2, msg) \
+  THROW_CHECK_OP_MSG(<=, val1, val2, msg)
+#define THROW_CHECK_LT_MSG(val1, val2, msg) \
+  THROW_CHECK_OP_MSG(<, val1, val2, msg)
+#define THROW_CHECK_GE_MSG(val1, val2, msg) \
+  THROW_CHECK_OP_MSG(>=, val1, val2, msg)
+#define THROW_CHECK_GT_MSG(val1, val2, msg) \
+  THROW_CHECK_OP_MSG(>, val1, val2, msg)
```
sarlinpe, line 124: I'm no too happy with these duplicated macros but I couldn't find a clean way to provide macro overload. Piping messages like glog with the `<<` syntax would be nicer but requires throwing in the destructor - I'm not convinced that we can easily make it safe.
https://github.com/google/glog/blob/v0.7.0-rc1/src/glog/logging.h#L1217-L1220
cc @Phil26AT

src/colmap/image/warp.cc
ahojnnes: No, not really. Can be changed.

PR #2372, opened 2024-01-28
Encapsulate freeimage usage from pycolmap in colmap bitmap

ahojnnes:


src/colmap/sensor/bitmap.cc
```diff
@@ -221,6 +205,41 @@ std::vector<uint8_t> Bitmap::ConvertToColMajorArray() const {
   return array;
 }
 
+std::vector<uint8_t> Bitmap::ConvertToRawBits() const {
```
sarlinpe, line 208: How does this differ to `ConvertToRowMajorArray/ConvertToColMajorArray`?

ahojnnes: The data may have a pitch that does not equal the width of the image to align the memory.

ahojnnes: This is how freeimage will internally allocate the memory for faster operations on the image.

PR #2376, opened 2024-01-29
Switch to exception checks - v2

sarlinpe:
- This is a rework of PR https://github.com/colmap/colmap/pull/2371 that is much less intrusive - we simply overwrite the glog macros to throw instead of aborting
- I checked with valgrind that this doesn't incur any memory leak
- Folks that depend on COLMAP not throwing exception can easily add an ifdef to disable this behavior
- I will remove the now-unnecessary pycolmap checks in a later PR

cc @Phil26AT


ahojnnes:
One drawback here. Imagine the following setup:
```
#include "colmap/util/logging.h"

#include <ExternalLibraryThatUsesGlogInHeader.h>
```
This will blow up when that external library uses glog's check macros in a way that our overrides don't support (e.g., `CHECK(condition) << "message";`.

This can also happen when colmap is installed as a library and other code uses glog. The v1 version of your PR added more code but does not have that problem?

sarlinpe:
As opposed to the previous PR, this one actually retains the syntax of glog by only overriding the internals. So `CHECK(condition) << "message";` remains valid and any library using it in headers will simply throw an exception instead of aborting. If you think that this is confusing, we can:
1. Make this behavior optional with a CMake option, or
2. Name the macros differently to retain the behavior of glog's original checks

ahojnnes:
If option (2) is possible, then that sounds great to me and nobody sees unexpected behavior when using glog macros?

sarlinpe:
Done, PTAL.

ahojnnes:
Great, thanks!

PR #2378, opened 2024-01-30
Add costs for pose graph optimization

sarlinpe:
I will add pycolmap bindings in a future PR

### DIFF
--- a/src/colmap/estimators/cost_functions.h
+++ b/src/colmap/estimators/cost_functions.h
@@ -43,6 +43,7 @@ template <typename T>
 using EigenVector3Map = Eigen::Map<const Eigen::Matrix<T, 3, 1>>;
 template <typename T>
 using EigenQuaternionMap = Eigen::Map<const Eigen::Quaternion<T>>;
+using EigenMatrix6d = Eigen::Matrix<double, 6, 6>;
 
 // Standard bundle adjustment cost function for variable
 // camera pose, calibration, and point parameters.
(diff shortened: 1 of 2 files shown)


whuaegeanse:
> I will add pycolmap bindings in a future PR

How does PGO integrate with incremental SfM? Are there any relevant papers?

sarlinpe:
We used it for our LaMAR benchmark https://lamar.ethz.ch/ https://github.com/microsoft/lamar-benchmark It is not yet used within COLMAP.

src/colmap/estimators/cost_functions.h
```diff
@@ -301,6 +302,111 @@ class SampsonErrorCostFunction {
...
+};
+
+// 6-DoF error between two absolute poses based on a measurement that is their
+// relative pose, with identical scale for the translation. The covariance is
+// defined in the reference frame of the camera j.
+// Its first and last three components correspond to the rotation and
+// translation errors, respectively.
+struct MetricRelativePoseErrorCostFunction {
+ public:
+  explicit MetricRelativePoseErrorCostFunction(const Rigid3d& j_from_i,
```
ahojnnes, line 363: explicit only useful for single arg constructor

src/colmap/estimators/cost_functions.h
```diff
@@ -301,6 +302,111 @@ class SampsonErrorCostFunction {
...
+                                            6,
+                                            4,
+                                            3,
+                                            4,
+                                            3>(
+        new MetricRelativePoseErrorCostFunction(j_from_i, covariance)));
+  }
+
+  template <typename T>
+  bool operator()(const T* const i_from_world_q,
```
ahojnnes, line 380: The following is probably more efficient, as inverting a quaternion is much cheaper as compared to applying it to a point:
```
template <typename T>
  bool operator()(const T* const i_from_world_q,
                  const T* const i_from_world_t,
                  const T* const j_from_world_q,
                  const T* const j_from_world_t,
                  T* residuals_ptr) const {
    const Eigen::Quaternion<T> j_from_i_q =
        EigenQuaternionMap<T>(j_from_world_q) *
        EigenQuaternionMap<T>(i_from_world_q).inverse();
    Eigen::Matrix<T, 3, 1> j_from_i_t =
        EigenVector3Map<T>(j_from_world_t) -
        j_from_i_q * EigenVector3Map<T>(i_from_world_t);

    const Eigen::Quaternion<T> mes_from_est_q =
        j_from_i_.rotation.cast<T>().inverse() * j_from_i_q;
    EigenQuaternionToAngleAxis(mes_from_est_q.coeffs().data(), residuals_ptr);

    Eigen::Map<Eigen::Matrix<T, 3, 1>> mes_from_est_t(residuals_ptr + 3);
    mes_from_est_t = j_from_i_.translation.cast<T>() - j_from_i_t;

    Eigen::Map<Eigen::Matrix<T, 6, 1>> residuals(residuals_ptr);
    residuals.applyOnTheLeft(sqrt_information_.template cast<T>());
    return true;
  }
```

src/colmap/estimators/cost_functions.h
```diff
@@ -301,6 +302,111 @@ class SampsonErrorCostFunction {
...
+    const Eigen::Quaternion<T> mes_from_est_q =
+        j_from_i_.rotation.cast<T>() * i_from_j_q;
+    EigenQuaternionToAngleAxis(mes_from_est_q.coeffs().data(), residuals_ptr);
+
+    Eigen::Map<Eigen::Matrix<T, 3, 1>> mes_from_est_t(residuals_ptr + 3);
+    mes_from_est_t = j_from_i_.translation.cast<T>() +
+                     j_from_i_.rotation.cast<T>() * i_from_j_t;
+
+    Eigen::Map<Eigen::Matrix<T, 6, 1>> residuals(residuals_ptr);
+    residuals.applyOnTheLeft(sqrt_information_.template cast<T>());
```
ahojnnes, line 401: We need to be very careful here about the definition of covariance here. Is it defined in frame i or frame j. It will be good to explicitly express this in the variable name of the covariance and sqrt_information variables (i.e., covariance_in_j)

src/colmap/estimators/cost_functions.h
ahojnnes: Same comment applies to the absolute pose error cost function above. If you compute j_from_i_q instead of i_from_j_q, you can save some compute.

src/colmap/estimators/cost_functions.h
```diff
@@ -301,6 +302,111 @@ class SampsonErrorCostFunction {
   const double y2_;
 };
 
+template <typename T>
+inline void EigenQuaternionToAngleAxis(const T* eigen_quaternion,
+                                       T* angle_axis) {
+  const T quaternion[4] = {eigen_quaternion[3],
```
ahojnnes, line 308: nit: Since you already map it to an Eigen::Quaternion, this could also be directly done using the Eigen object. Not sure if the implemented math is suitable for computing stable Jacobians with Jet though.

src/colmap/estimators/cost_functions.h
sarlinpe: I think that this isn't mathematically correct. If we write ${}_j\hat{T}_i$ the measured transform and ${}_iT_w$ and ${}_jT_w$ the parameters, the error in the reference frame $j$ of the measurement is
$$\Delta T = {}_j\hat{T}_i \cdot {}_iT_w \cdot {}_jT_w^{-1}\enspace.$$
For the translation, this amounts to
$$\Delta t = {}_j\hat{t}_i + {}_j\hat{R}_i \cdot {}_it_j\enspace.$$
This is basically the center of $j$ (parameter) in the reference frame of the measure $j$. What you wrote amounts to:
$$\Delta t = {}_j\hat{t}_i + {}_jR_i \cdot {}_it_j\enspace.$$
This makes the simplification ${}_j\hat{R}_i \approx {}_jR_i\ $, which does not necessarily hold.

src/colmap/estimators/cost_functions.h
sarlinpe: Done.

src/colmap/estimators/cost_functions.h
sarlinpe: The docstring mentions that the covariance is defined in the reference frame of the camera $j$. I updated the name of the covariance accordingly.

src/colmap/estimators/cost_functions.h
ahojnnes: Sorry, on my phone right now, and cannot right down the math. It looks like that in your notation my code does the following:
```
$$\Delta T = {}_jT_w \cdot {}_iT_w^{-1}  \cdot {}_j\hat{T}_i$$
```

sarlinpe: Reconsidering what I stated above: since Ceres use the left-hand convention for the quaternion retractation, the residual should indeed be as you describe. I'll update the code accordingly.

sarlinpe: Updated. This doesn't match your code (because of the approximation explained above) but is still efficient by expecting the constraint as ${}_i\hat{T}_j$ instead of ${}_j\hat{T}_i$. Also saved an inversion in `AbsolutePoseErrorCostFunction` by pre-computing the inverse.

ahojnnes: LGTM, thx

Issue #2349, opened 2024-01-15
Scaling issue in CameraRig::ComputeRigFromWorld()


nnop:
I found that in the `CameraRig::ComputeRigFromWorld()` function, the rig pose is directly composed using the pose of the reconstruction and the provided camera rig pose:
https://github.com/colmap/colmap/blob/fb3fa6a6962f8fee299d03b895cac2e1dd052104/src/colmap/scene/camera_rig.cc#L226-L227

However, the reconstruction is **up to an unknown scale**, while the provided camera rig pose is calibrated with an absolute scale (e.g., meters). This makes direct composition unreasonable. 
Shouldn't it be first calculating the scale ratio between the rig and the reconstruction, compensate for it, and then compose the two.
Actually, there is a function for calculating this scale, but I couldn't find any place where it is called.
Could this be a bug?

coffeeancode:
I haven't used the RigBA for a while, so something might have changed. But I used to have to manually correct the scale of the reconstruction, before running the RigBundleAdjuster. You can do this using the model_aligner for instance. Not sure why the ComputeRigFromWorldScale is never used though.

Also not that by default, the RigBundleAdjuster attempts to estimate the relative poses (using the scale from the reconstruction), unless you set --estimate_rig_relative_poses to False.

nnop:
Could you have a look at this problem? @tsattler

shubham-monarch:
Why did you mark this as completed? How did you solve this issue?

PR #2382, opened 2024-01-31
rescale reconstruction to the same scale with rigs

nnop:
Fix https://github.com/colmap/colmap/issues/2349

With rescaling, BA converges better and faster.


shubham-monarch:
Why did you close this? Does this fix work for you?

PR #2391, opened 2024-02-01
Fix warning C4722

whuaegeanse:
Vistual Studio 2022 outputs compiler warning C4722.

`warning C4722: 'colmap::LogMessageFatalThrow<std::invalid_argument>::~LogMessageFatalThrow<std::invalid_argument>': destructor never returns, potential memory leak`

This PR attempts to suppress the C4722 warning. But I'm not sure if this is a good idea.

[Compiler Warning (level 1) C4722](https://learn.microsoft.com/en-us/cpp/error-messages/compiler-warnings/compiler-warning-level-1-c4722?view=msvc-170)
```
'function' : destructor never returns, potential memory leak

The flow of control terminates in a destructor. The thread or the entire program will terminate and allocated resources may not be released. Furthermore, if a destructor will be called for stack unwinding during exception processing, the behavior of executable is undefined.

To resolve, remove the function call that causes the destructor to not return.
```

### DIFF
--- a/src/colmap/util/CMakeLists.txt
+++ b/src/colmap/util/CMakeLists.txt
@@ -88,6 +88,11 @@ COLMAP_ADD_TEST(
     LINK_LIBS colmap_util
 )
 COLMAP_ADD_TEST(
+    NAME logging_test
+    SRCS logging_test.cc
+    LINK_LIBS colmap_util
+)
+COLMAP_ADD_TEST(
     NAME misc_test
     SRCS misc_test.cc
     LINK_LIBS colmap_util
--- a/src/colmap/util/logging.h
+++ b/src/colmap/util/logging.h
@@ -138,6 +138,11 @@ inline std::string __MakeExceptionPrefix(const char* file, int line) {
          std::to_string(line) + "] ";
 }
 
+#ifdef _MSC_VER
+#pragma warning(push)
+#pragma warning(disable : 4722)
+#endif
+
 template <typename T>
 class LogMessageFatalThrow : public google::LogMessage {
  public:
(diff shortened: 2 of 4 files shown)


ahojnnes:
Throwing from a destructor is indeed not a good idea and will leak memory. @sarlinpe FYI

sarlinpe:
As mentioned in https://github.com/colmap/colmap/pull/2376, I checked with valgrind that this doesn't incur any memory leak. I agree that in general throwing from a constructor is bad design, but I think that for this case it's fine. The only scenario in which this could bite is the following, which causes an `std::terminate`:
```c++
std::string PrintingFn(...) {
    if (...) {
        LOG(FATAL_THROW) << "Error in PrintingFn";
    }
    return ...;
}

LOG(FATAL_THROW) << "Error: " << PrintingFn(...);
```
We don't have any such example in the codebase and it's easy to avoid. Besides, the throwing destructor is the only design (AFAIK) that allows the nice `<<` syntax.

ahojnnes:
I assume that address sanitizer (which is already executed as part of CI) should also catch these. Maybe we should have a few unit tests that exercise the new exception macros explicitly, so they get executed in our CI pipeline.

I am not sure I fully understand why the above example would cause a memory leak or std::terminate. I would have to dig into glog's code I guess?

sarlinpe:
Having tests covering the macro is a good idea, I can give it a try unless @whuaegeanse you have time.

Regarding the above example: `PrintingFn` throws an exception, the stack is unwound, `LogMessageFatalThrow` of the calling scope gets destructed, which throws a second exception. As the warning suggests, this has undefined behavior - on Unix this causes an `std::terminate`.

ahojnnes:
OK, got it, thanks for the explanation. I guess we could detect such cases (where our log exceptions are thrown recursively) and then not throw a second time, crash the process, or log out both. This could be done with an atomic boolean flag or so in the LogFatalThrowLogMessage class.

ahojnnes:
Not sure if there is another way to prevent the problem in the first place... Probably there is but requires time to dig.

sarlinpe:
We can detect such cases by checking `std::uncaught_exception`. I'll give it a try.

sarlinpe:
@ahojnnes @whuaegeanse Done, PTAL.

PR #2392, opened 2024-02-01
Disentangle the controller from threading and integrate the new logic into IncrementalMapperController

B1ueber2y:
Since we want to pybind substeps of the ``IncrementalMapperController`` in python in the end, we need to disentangle the threading-related function from the controller class, such that we can further abstract the controller and manipulate the same thread in python. 

This **DRAFT** feature contains several updates as follows **(need discussions on the naming and infrastructure. This PR is potentially dangerous)**:
- Remove the usage of timer from the Thread class, to prepare for separating the Run function into several steps in Python. Now the timer is initialized in the upper level at the start of the "Run" function. 
- Make a new ``BaseController`` class that inherits from a ``CoreController`` class, which supports all the previous features from the Thread class (https://github.com/B1ueber2y/colmap/blob/features/remove_thread_from_exe/src/colmap/util/threading_ctrl.h). While all the threading-related operations are set as callbacks with empty as default. When the threading is needed, one can use ``ControllerThread`` class which uses a controller class as its template class and implements the callbacks in the constructor. 
- Change the virtual method ``Run`` from private into public (such that we can bind substeps later). Change the calling at top level from ``Start()`` + ``Wait()`` into a single ``Run()`` (e.g., https://github.com/colmap/colmap/pull/2392/files#diff-a075407d92533f0fe38542d18c5b0d9a8110138218e77a67deff2db3bb2811beR113). Only ui needs threading-related functions.
- Change the parent of the IncrementalMapper into ``BaseController`` and update the ui by using ``ControllerThread``.

If approved we can later change all the classes (now this draft PR only updates ``IncrementalMapperController`` and the UI) into this new design by changing the parents of most controller from ``Thread`` to ``BaseController``, and only use ``ControllerThread`` class when threading-related functions are needed. 

@ahojnnes @sarlinpe


ahojnnes:
Thanks, I am not sure how this helps with the next steps that you are envisioning and thus cannot formulate an opinion on these changes. For me to understand a little better: Would the existing "Thread" class still exist after these changes or simplify?

src/colmap/util/threading_ctrl.h
```diff
@@ -0,0 +1,270 @@
...
+  };
+
+  BaseController();
+  virtual ~BaseController() = default;
+
+  ///////////////////////////////////////////////////
+  // Thread-related functions
+  ///////////////////////////////////////////////////
+  ThreadStatus status_;
+  ThreadStatus* GetThreadStatus() { return &status_; }
```
ahojnnes, line 137: return const ref to avoid manipulation outside?

src/colmap/util/threading_ctrl.h
```diff
@@ -0,0 +1,270 @@
...
+
+#ifdef __clang__
+#pragma clang diagnostic pop  // -Wkeyword-macro
+#endif
+
+// Reimplementation of threading with thread-related functions outside
+// controller Following util/threading.h
+
+// Core methods of a controller, wrapped by BaseController
+class CoreController {
```
ahojnnes, line 60: Not sure I understand why we need separate Core and BaseController classes instead of being the same?

B1ueber2y: The initial motivation is that one can instantiate a controller that does not support threading in the future. But I agree it feels like overcomplication. So I have just merged them together.

src/colmap/util/threading_ctrl.h
B1ueber2y: This function is removed with the new update, where all the thread status goes into ControllerThread, where they are made private.

B1ueber2y:
With the new update just now the controller and threading is further disentangled by registering the callbacks of the controller in the thread class. Now the controller only supports the threading-functions as follows:
- SignalValidSetup()
- SignalInValidSetup()
- CheckValidSetup()
- IsStopped()
- BlockIfPaused()

Note that now functions such as ``Start()``, ``IsStarted()``, ``Wait()``, etc is not supported in the controller, but can only be called in the ``ControllerThread`` class. 

Also, @sarlinpe suggested to combine ``BlockIfPaused()`` and ``IsStopped()`` together in a ``CheckIfStopped()`` method, such that we can modify this handle in python in the future outside the C++ threading design. I have made this possible and integrated into the ``IncrementalMapperController``, such that now in the ``IncrementalMapperController`` we only used this ``CheckIfStopped`` method.

@ahojnnes About the next steps, Now the new design is only integrated in the ``IncrementalMapperController`` as a draft. The initial plan is indeed replace the ``Thread`` class with ``ControllerThread`` in the end. But I also have no problem to keep the ``Thread`` class implementation for convenience (while it may cause confusion in the future). In the end ``BaseController`` will go into the controller folder, and ``ControllerThread`` can stay in the ui or be put into the threading header. If approved we can update all the controller with this new design and then we can split the Run() method further into substeps, such that we can support better and detailed bindings in Python.

B1ueber2y:
The polymorphism with ``active_thread_`` in the mapper controller can be further supported with the ``SetCheckIfStoppedFunc`` method from the controller (https://github.com/colmap/colmap/pull/2392/commits/f82b0022f6189988b43f8a7de469316b9b3765a9#diff-faf29b81ac33239a3b5179ec793d7f032660ae99f8cf1865341f7b4eb09476edR230), which is a callback that returns boolean value. This is also used in the IncrementalMapperController. 

This update has been done and stably integrated for the IncrementalMapperController with the latest update.

src/colmap/util/threading_ctrl.h
```diff
@@ -0,0 +1,259 @@
...
+namespace colmap {
+
+#ifdef __clang__
+#pragma clang diagnostic push
+#pragma clang diagnostic ignored "-Wkeyword-macro"
+#endif
+
+#ifdef __clang__
+#pragma clang diagnostic pop  // -Wkeyword-macro
+#endif
```
sarlinpe, line 54: Is this still needed?

src/colmap/util/threading_ctrl.cc
```diff
@@ -0,0 +1,110 @@
...
+  else
+    return false;
+}
+
+void BaseController::BlockIfPaused() { Callback(BLOCK_IF_PAUSED_CALLBACK); }
+
+bool BaseController::CheckIfStopped() {
+  BlockIfPaused();
+  return IsStopped();
+}
```
sarlinpe, line 103: Didn't we say we would combine pausing and blocking in a single callback?

src/colmap/util/threading_ctrl.h
```diff
@@ -0,0 +1,259 @@
...
+    STARTED_CALLBACK = INT_MIN,
+    FINISHED_CALLBACK,
+
+    // thread-related callbacks from here, only useful in the Thread class,
+    // empty as default
+    LOCK_MUTEX_CALLBACK,
+    BLOCK_IF_PAUSED_CALLBACK,
+    SIGNAL_SETUP_CALLBACK,
+    CHECK_VALID_SETUP_CALLBACK,
+  };
```
sarlinpe, line 74: Can't this go into the `ControllerThread` instead?

src/colmap/util/threading_ctrl.h
```diff
@@ -0,0 +1,259 @@
...
+  // Call back to the function with the specified name, if it exists.
+  void Callback(int id) const;
+
+  // Signal that the thread is setup. Only call this function once.
+  void SignalValidSetup();
+  void SignalInvalidSetup();
+
+ private:
+  bool setup_ = false;
+  bool setup_valid_ = false;
```
sarlinpe, line 123: Can't this go into the `ControllerThread` instead? By adding an additional callback `signal_valid_setup_fn` similar to `check_if_stopped_fn`.

src/colmap/util/threading_ctrl.h
```diff
@@ -0,0 +1,259 @@
...
+      std::unique_lock<std::mutex> lock(mutex_);
+    });
+    controller_->AddCallback(BaseController::BLOCK_IF_PAUSED_CALLBACK, [&]() {
+      std::unique_lock<std::mutex> lock(mutex_);
+      if (status_.paused) {
+        status_.pausing = true;
+        pause_condition_.wait(lock);
+        status_.pausing = false;
+      }
+    });
```
sarlinpe, line 191: This is redundant with `CheckIfStoppedFunc`, no?

src/colmap/util/threading_ctrl.cc
```diff
@@ -0,0 +1,110 @@
...
+  for (const auto& callback : callbacks_.at(id)) {
+    callback();
+  }
+}
+
+void BaseController::RunFunc() {
+  Callback(STARTED_CALLBACK);
+  Run();
+  Callback(LOCK_MUTEX_CALLBACK);
+  Callback(FINISHED_CALLBACK);
```
sarlinpe, line 68: Call these in `ControllerThread::RunFunc`? The callback logic can then be moved to `ControllerThread`

sarlinpe:
@B1ueber2y This design is much cleaner than the previous one. I think that we can move more threading logic out of the `BaseController` - see my comments.

src/colmap/util/threading_ctrl.h
B1ueber2y: These two flags are not necessarily related to threading. For the ``SignalValidSeup`` and ``SignalInvalidSetup`` we need a boolean variable in the controller for checking.

src/colmap/util/threading_ctrl.h
B1ueber2y: Same to the previous one. If we combine them together this is not needed.

src/colmap/util/threading_ctrl.cc
sarlinpe: `BlockIfPaused()` is called at 3 locations only and always right before `IsStopped()`. `IsStopped()` is called more often, so this change would make the pausing more responsive.

src/colmap/util/threading_ctrl.h
sarlinpe: IIUC `CheckValidSetup` exists only because we can't easily propagate errors in threading. It's used only in nested threads in feature extraction/matching. When there is no threading involved, it is useless - one should rather return a flag or throw an exception. So I think that the booleans should be part of the threading class.

Regarding the signaling callbacks: if we make them explicit callbacks (like `signal_valid_setup_fn`) then we can merge them into a single one that accepts a boolean (setup is valid or not).

B1ueber2y:
I have made a new update on the ControllerThread by directly inheriting from Thread. The only added feature is
- ``AddControllerCallback`` which adds controller-specific callbacks (e.g., https://github.com/colmap/colmap/pull/2392/files#diff-517534814bb88b451c2f5f4856136755190c58987f0870fa50939d2115124704R626)
- combine ``BlockIfPaused()`` and ``IsStopped`` into a ``CheckIfStopped`` function in the controller

In this design we wont deprecate the ``Thread`` class, but can still migrate most controller class into the new design to disentangle the threading. This feels like a far more elegant way to me and also the update becomes more stable without significant changes.

src/colmap/util/threading_ctrl.h
B1ueber2y: Removed.

src/colmap/util/threading_ctrl.cc
B1ueber2y: OK I have merged them together.

src/colmap/util/threading_ctrl.h
B1ueber2y: With the new design this method is not needed in the controller anymore

src/colmap/util/threading_ctrl.cc
B1ueber2y: With the new design this is back to the thread.

src/colmap/util/threading_ctrl.h
B1ueber2y: Yes this is not needed after merging.

src/colmap/util/threading_ctrl.h
B1ueber2y: Not needed after the new design

src/colmap/util/threading_ctrl.h
```diff
@@ -0,0 +1,123 @@
+// Copyright (c) 2023, ETH Zurich and UNC Chapel Hill.
```
sarlinpe, line 1: Maybe rename this file (including `ControllerThread`) to `controllers/base_controller.h`?

src/colmap/util/threading_ctrl.h
```diff
@@ -0,0 +1,123 @@
...
+  // do BlockIfPaused() every time before checking IsStopped()
+  bool IsStopped() {
+    BlockIfPaused();
+    return Thread::IsStopped();
+  }
+
+  // add a controller-specific callback
+  void AddControllerCallback(int id, const std::function<void()>& func) {
+    controller_->AddCallback(id, func);
+  }
```
sarlinpe, line 116: I would prefer exposing `controller_` as const with a getter such that we're not tied to the interface of `AddCallback` (which might change in the future).

src/colmap/util/threading_ctrl.h
```diff
@@ -0,0 +1,123 @@
...
+//
+template <class Controller>
+class ControllerThread : public Thread {
+  // check if the Controller class is inherited from BaseController
+  static_assert(std::is_base_of<BaseController, Controller>::value,
+                "The controller needs to be inherited from BaseController");
+
+ public:
+  explicit ControllerThread(std::shared_ptr<Controller> controller) {
+    controller_ = controller;
```
sarlinpe, line 102: Use initializer list and `std::move`

src/colmap/util/threading_ctrl.h
B1ueber2y: Nice idea. This is done.

src/colmap/util/threading_ctrl.h
B1ueber2y: Done.

src/colmap/util/threading_ctrl.h
B1ueber2y: This is done. But do we want to put ControllerThread elsewhere?

sarlinpe: I think that it can stay there for now

src/colmap/controllers/base_controller.h
```diff
@@ -0,0 +1,121 @@
...
+// INTERRUPTION) HOWEVER CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN
+// CONTRACT, STRICT LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE)
+// ARISING IN ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE
+// POSSIBILITY OF SUCH DAMAGE.
+
+#pragma once
+
+#include "colmap/util/logging.h"
+#include "colmap/util/threading.h"
+#include "colmap/util/timer.h"
```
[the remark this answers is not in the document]
B1ueber2y: Done. Thanks

src/colmap/controllers/base_controller.h
```diff
@@ -0,0 +1,121 @@
...
+
+#include <atomic>
+#include <climits>
+#include <functional>
+#include <future>
+#include <list>
+#include <queue>
+#include <thread>
+#include <type_traits>
+#include <unordered_map>
```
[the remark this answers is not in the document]
B1ueber2y: This is cleaned. Thanks

src/colmap/controllers/base_controller.h
```diff
@@ -0,0 +1,112 @@
...
+template <class Controller>
+class ControllerThread : public Thread {
+  // check if the Controller class is inherited from BaseController
+  static_assert(std::is_base_of<BaseController, Controller>::value,
+                "The controller needs to be inherited from BaseController");
+
+ public:
+  explicit ControllerThread(std::shared_ptr<Controller> controller)
+      : controller_(std::move(controller)) {
+    controller_->SetCheckIfStoppedFunc([&]() { return IsStopped(); });
```
sarlinpe, line 94: AFAIK here you can also write `SetCheckIfStoppedFunc(std::bind(&IsStopped, this))`

B1ueber2y: Yeah but I guess lambda is fine. Do we want to change this? I can do it

sarlinpe: I don't care much

src/colmap/controllers/base_controller.h
```diff
@@ -0,0 +1,112 @@
...
+// INTERRUPTION) HOWEVER CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN
+// CONTRACT, STRICT LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE)
+// ARISING IN ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE
+// POSSIBILITY OF SUCH DAMAGE.
+
+#pragma once
+
+#include "colmap/util/threading.h"
+
+#include <functional>
```
sarlinpe, line 34: I believe that you do need `<list>` here because of `callbacks_`

B1ueber2y: Indeed. Fixed.

sarlinpe:
LGTM, great work, thanks a lot!

sarlinpe (approved):

ahojnnes:
Thanks, looks great. I assume you will follow up with equivalent changes for other controllers?

PR #2399, opened 2024-02-05
Move reconstruction IO utils to a new file

sarlinpe:
We could consider deprecating the NVM, CAM, Recon3D, VRML, and Bundler formats at later point.


ahojnnes:
Beautiful, thank you. Not for this PR, but we should at some point maybe follow up with some basic roundtrip unit tests that generate a synthetic reconstruction and then write/read those to local test folder using CreateTestDir().

sarlinpe:
Yes definitely!

PR #2406, opened 2024-02-09
Adapt all the controllers to inherit from BaseController rather than Thread (except for feature extraction and matching)

B1ueber2y:

- Adapt all the controllers except for feature extraction and matching to use the new logic. Each controller inherits from BaseController, and in the ui a ControllerThread is instantiated.
- Move ControllerThread in the util, such that there is no "Thread"-related practice in the other folder. Maybe we should separate class ``Thread`` from ``ThreadPool`` and ``JobQueue`` at ``util/threading.h`` and move ``Thread`` class and ``ControllerThread`` class completely into the ui folder.

Now feature extraction and feature matching still inherits from the Thread class since in Python we will use different logics. However, because of this the ``AutomaticReconstructionController`` needs to inherit from Thread as well to support the Stopping function for the feature extractor and matcher. While this can potentially be resolved by adding ``SignalInvalidSetup``, ``SignalValidSetup``, and ``CheckValidSetup`` into the controller with a boolean valued callback, this might be a little overcomplication in the class design since this is not relevant to pybind for now. It depends on whether we want to totally disentangle the Thread outside the ui @ahojnnes @sarlinpe 

If needed I can make the update on the feature extraction and matcher later as well in a separate PR.


B1ueber2y:
OK unfortunately keeping the ``BaseController`` class in the ``controllers`` folder and inheriting it in the undistorters will result in cyclic dependency between the ``controllers`` folder and ``image`` folder, so I have to move the ``BaseController`` implementation to ``util`` folder as well.

sarlinpe:
`undistortion.cc` is huge (1.1k LOC), it would make sense to split it into a controller (that would go into `controllers/`) and the actual undistortion utils.

B1ueber2y:
> `undistortion.cc` is huge (1.1k LOC), it would make sense to split it into a controller (that would go into `controllers/`) and the actual undistortion utils. This would also make it easier to write bindings for it.

Sure sounds good. But then how about the mvs if we put ``BaseController`` back to the ``controllers`` folder? Patchmatch stereo and fusion also uses the controller, making the ``controller`` folder a dependency for the ``mvs``. This also does not feel like an ideal dependency graph conceptually.

src/colmap/util/controller_thread.h
```diff
@@ -0,0 +1,73 @@
...
+// Helper class to create single threads with simple controls
+// Similar usage as ``Thread`` class in util/threading.h except for
+// initialization. e.g.,
+//
+// std::shared_ptr<Controller> controller = std::make_shared<Controller>(args);
+// std::unique_ptr<ControllerThread<Controller>> thread =
+// std::make_unique<ControllerThread<Controller>>(controller);
+//
+//
+template <class Controller>
```
ahojnnes, line 46: Wondering why this needs to be templated? We are only accessing virtual methods of the abstract interface? We could just pass a shared/unique_ptr of a BaseController here instead and avoid the template?

src/colmap/util/controller_thread.h
```diff
@@ -0,0 +1,73 @@
...
+
+ public:
+  explicit ControllerThread(std::shared_ptr<Controller> controller)
+      : controller_(std::move(controller)) {
+    controller_->SetCheckIfStoppedFunc([&]() { return IsStopped(); });
+  }
+  ~ControllerThread() = default;
+
+  // get the handle to the controller in ControllerThread
+  const std::shared_ptr<Controller> GetController() { return controller_; }
```
ahojnnes, line 60: The const here is not very useful?

src/colmap/util/controller_thread.h
```diff
@@ -0,0 +1,73 @@
...
+  // check if the Controller class is inherited from BaseController
+  static_assert(std::is_base_of<BaseController, Controller>::value,
+                "The controller needs to be inherited from BaseController");
+
+ public:
+  explicit ControllerThread(std::shared_ptr<Controller> controller)
+      : controller_(std::move(controller)) {
+    controller_->SetCheckIfStoppedFunc([&]() { return IsStopped(); });
+  }
+  ~ControllerThread() = default;
```
ahojnnes, line 57: Not needed?

src/colmap/util/controller_thread.h
B1ueber2y: This needs to be templated because we sometimes need to access the controller-specific functions from the ControllerThread, e.g. https://github.com/B1ueber2y/colmap/blob/features/controllers/src/colmap/ui/dense_reconstruction_widget.cc#L418-L420. The ``GetController`` method needs the template.

ahojnnes: OK, got it.

src/colmap/util/controller_thread.h
B1ueber2y: Thanks. This is removed.

src/colmap/util/controller_thread.h
B1ueber2y: Why is this not useful? I personally think returning a const shared_ptr makes sense in this case.

ahojnnes: Did you mean to write shared_ptr<const Controller> ? Returning a const shared_ptr has limited utility IMHO, as it only prevents the user from calling reset or other mutating shared ptr methods. Since the class still holds a copy of the shared ptr that doesn't do much harm though? Maybe I misunderstand the intent of the current usage though.

B1ueber2y: Ah OK thanks for the thoughts. No i indeed meant const shared_ptr<Controller>. Just that generally I would return a const pointer for a getter method, but I also agree that the utility is very limited. This is updated now.

ahojnnes (approved):

PR #2411, opened 2024-02-11
Migrate Ubuntu CI pipelines from ADO to Github

ahojnnes:
Other platforms to follow as separate PRs. The final PR will then remove the ADO pipeline definitions.


.github/workflows/build-ubuntu.yml
```diff
@@ -0,0 +1,272 @@
...
+      - name: Check code format
+        run: |
+          if [ "${{ matrix.config.checkCodeFormat }}" != "true" ]; then
+            exit 0
+          fi
+          set +x -euo pipefail
+          sudo apt-get update && sudo apt-get install -y clang-format-14 black
+          ./scripts/format/clang_format.sh
+          ./scripts/format/black.sh
+          git diff --name-only | xargs -r printf "##vso[task.LogIssue type=error;sourcepath=%s;linenumber=1;columnnumber=1;]Invalid code format, run scripts/format/*.sh\n"
```
sarlinpe, line 121: This logging syntax seems to be ADO-specific. In GH actions the equivalent is described here: https://docs.github.com/en/actions/using-workflows/workflow-commands-for-github-actions#setting-an-error-message

.github/workflows/build-ubuntu.yml
```diff
@@ -0,0 +1,272 @@
...
+          cd ../doc/sample-project
+          mkdir build
+          cd build
+          colmap_DIR=${{ github.workspace }}/build/install/share/colmap cmake .. \
+            -GNinja \
+            -DCMAKE_CUDA_ARCHITECTURES=50
+          ninja
+          ./hello_world --message "world"
+
+      - name: Run tests
```
sarlinpe, line 216:
```suggestion
      - name: Run unit tests
```

sarlinpe (approved):
Thanks a lot! Added some minor comments. I assume that the macOS and windows pipelines will be migrated too?

.github/workflows/build-ubuntu.yml
ahojnnes: Thx, good catch.

.github/workflows/build-ubuntu.yml
ahojnnes: I'll keep it as is, since these are not just unit tests but also integration tests on the controller level with synthetic data, etc.

sarlinpe (approved):

PR #2413, opened 2024-02-13
Improve bindings of Database

sarlinpe:
Related to this PR: `FeatureKeypoint`/`FeatureKeypoints` and `FeatureMatch`/`FeatureMatches` are not Python-friendly, would it be OK to replace all their occurrences with their blob counterparts? NxD matrices are directly translated to numpy arrays. This would simplify the COLMAP code as well and also make it cleaner to support non-affine features.

### DIFF
--- a/src/colmap/scene/database.h
+++ b/src/colmap/scene/database.h
@@ -194,7 +194,7 @@ class Database {
                     const FeatureMatches& matches) const;
   void WriteMatches(image_t image_id1,
                     image_t image_id2,
-                    FeatureMatchesBlob blob) const;
+                    const FeatureMatchesBlob& blob) const;
   void WriteTwoViewGeometry(image_t image_id1,
                             image_t image_id2,
                             const TwoViewGeometry& two_view_geometry) const;
--- a/src/colmap/scene/database.h
+++ b/src/colmap/scene/database.h
@@ -134,9 +134,8 @@ class Database {
   inline static image_pair_t ImagePairToPairId(image_t image_id1,
                                                image_t image_id2);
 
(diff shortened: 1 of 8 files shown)


src/pycolmap/scene/database.h
```diff
@@ -8,36 +8,137 @@ using namespace colmap;
 using namespace pybind11::literals;
 namespace py = pybind11;
 
+class DatabaseTransactionWrapper {
+ public:
+  explicit DatabaseTransactionWrapper(Database* database)
+      : database_(database){};
```
ahojnnes, line 14:
```suggestion
      : database_(database) {}
```
nit

src/colmap/scene/database.h
```diff
@@ -177,11 +186,15 @@ class Database {
...
                       const FeatureKeypoints& keypoints) const;
+  void WriteKeypoints(image_t image_id, const FeatureKeypointsBlob& blob) const;
   void WriteDescriptors(image_t image_id,
                         const FeatureDescriptors& descriptors) const;
   void WriteMatches(image_t image_id1,
                     image_t image_id2,
                     const FeatureMatches& matches) const;
+  void WriteMatches(image_t image_id1,
+                    image_t image_id2,
+                    FeatureMatchesBlob blob) const;
```
ahojnnes, line 197:
```suggestion
                    const FeatureMatchesBlob& blob) const;
```

sarlinpe: `SwapFeatureMatchesBlob` a few lines later modifies it in-place, so better copy in the argument.

ahojnnes (dismissed):

ahojnnes (approved):

PR #2421, opened 2024-02-14
Simplify IncrementalMapperController

sarlinpe:
Major simplification of `IncrementalMapperController:Reconstruct`. This should make it easier to rewrite it in Python (next step). The logical behavior of the function is unchanged.
cc @B1ueber2y

### DIFF
--- a/src/colmap/exe/sfm.cc
+++ b/src/colmap/exe/sfm.cc
@@ -419,7 +419,7 @@ int RunPointTriangulatorImpl(
     const std::string& database_path,
     const std::string& image_path,
     const std::string& output_path,
-    const IncrementalMapperOptions& mapper_options,
+    const IncrementalMapperOptions& options,
     const bool clear_points,
     const bool refine_intrinsics) {
   PrintHeading1("Loading database");
@@ -430,12 +430,11 @@ int RunPointTriangulatorImpl(
     Timer timer;
     timer.Start();
     const Database database(database_path);
-    const size_t min_num_matches =
-        static_cast<size_t>(mapper_options.min_num_matches);
+    const size_t min_num_matches = static_cast<size_t>(options.min_num_matches);
     database_cache = DatabaseCache::Create(database,
                                            min_num_matches,
-                                           mapper_options.ignore_watermarks,
-                                           mapper_options.image_names);
+                                           options.ignore_watermarks,
+                                           options.image_names);
 
     if (clear_points) {
       reconstruction->DeleteAllPoints2DAndPoints3D();
@@ -455,8 +454,8 @@ int RunPointTriangulatorImpl(
   // Triangulation
   //////////////////////////////////////////////////////////////////////////////
 
-  const auto tri_options = mapper_options.Triangulation();
-  const auto map_options = mapper_options.Mapper();
+  const auto tri_options = options.Triangulation();
+  const auto mapper_options = options.Mapper();
 
   const std::vector<image_t>& reg_image_ids = reconstruction->RegImageIds();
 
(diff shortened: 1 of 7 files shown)


src/colmap/exe/sfm.cc
```diff
@@ -456,6 +456,7 @@ int RunPointTriangulatorImpl(
   //////////////////////////////////////////////////////////////////////////////
 
   const auto tri_options = mapper_options.Triangulation();
+  const auto map_options = mapper_options.Mapper();
```
B1ueber2y, line 459: Should we update the naming here? Now that in this exe/sfm.cc ``mapper_options`` refers to a instance that is in type ``IncrementalMapperOptions``, while inside the controller ``mapper_options`` refers to ``map_options`` here (in type ``IncrementalMapper::Options``). Maybe we should update all the ``mapper_options`` inside the controller to ``map_options`` for consistency, to avoid potential confusion :)

src/colmap/sfm/incremental_mapper.cc
```diff
@@ -686,6 +694,17 @@ bool IncrementalMapper::AdjustGlobalBundle(
                                              "registered for global "
                                              "bundle-adjustment";
 
+  BundleAdjustmentOptions ba_options_copy = ba_options;
```
B1ueber2y, line 697: Maybe we should say ``ba_options_strict`` rather than ``ba_options_copy``.

B1ueber2y:
Thanks. Huge step. Wrapping the logics from the controller into the sfm incremental mapper makes a lot of sense. After the simplification the whole logic looks better than before.

src/colmap/exe/sfm.cc
sarlinpe: I don't think that is matters much. We could instead rename `mapper_options` to `controller_options` in this function only. I anyway plan to move this function elsewhere, for example into the `IncrementalMapperController`.

src/colmap/sfm/incremental_mapper.cc
sarlinpe: It's only strict if the condition below is true?

src/colmap/exe/sfm.cc
B1ueber2y: ``controller_options`` sounds good!

src/colmap/sfm/incremental_mapper.cc
B1ueber2y: Maybe ``ba_options_tmp``, ``ba_options_local``? Since we want to change this it s strange to name it ``copy`` to me. Anyway its just a name.

src/colmap/sfm/incremental_mapper.h
```diff
@@ -231,6 +256,12 @@ class IncrementalMapper {
   // Clear the collection of changed 3D points.
   void ClearModifiedPoints3D();
 
+  // Estimate two view geometry and checks if it is suitable for initialization.
```
ahojnnes, line 259: How about the following interface:
```
 std::optional<TwoViewGeometry> EstimateInitialTwoViewGeometry(const Options& options,
                                      image_t image_id1,
                                      image_t image_id2);
```

src/colmap/sfm/incremental_mapper.h
```diff
@@ -157,6 +159,7 @@ class IncrementalMapper {
   // pairs should be passed to `RegisterInitialImagePair`. This function
   // automatically ignores image pairs that failed to register previously.
   bool FindInitialImagePair(const Options& options,
+                            TwoViewGeometry& two_view_geometry,
                             image_t* image_id1,
```
ahojnnes, line 163: Let's change these from pointers to refs?

src/colmap/sfm/incremental_mapper.h
sarlinpe: `std::optional` is a C++17 feature. Do we want to bump the minimal version to 17? Ceres has done this recently for 2.2.0

ahojnnes: Ah, right. Always forget. Sorry.

src/colmap/sfm/incremental_mapper.h
sarlinpe: Done, PTAL.

src/colmap/sfm/incremental_mapper.cc
sarlinpe: Done.

ahojnnes (approved):
Merci, lgtm

PR #2426, opened 2024-02-16
Define vcpkg manifest

ahojnnes:


vcpkg.json
```diff
@@ -0,0 +1,60 @@
...
+                "lapack",
+                "suitesparse"
+            ]
+        },
+        "cgal",
+        "eigen3",
+        "flann",
+        "freeimage",
+        "gflags",
+        "glew",
```
sarlinpe, line 27: Shouldn't this be required only when the GUI is enabled? We don't build it in the pycolmap CI.

pycolmap/ci/vcpkg-dependencies.txt
```diff
@@ -1,19 +0,0 @@
...
-ceres[lapack,suitesparse]
-eigen3
-flann
-freeimage
-metis
-gflags
-glog
-gtest
-sqlite3
-jasper[core]
```
sarlinpe, line 19: We need to be very careful that the added vcpkg dependencies do not increase the size of the wheels - i.e. that no additional libraries are linked to COLMAP or its dependencies. So far we have
- excluded `cgal` and `glew`
- explicitly set `jasper[core]` (included by libraw - freeimage) to avoid its `opengl` feature (enabled by default), which would link unnecessary librairies (opengl, gmp, etc.)

vcpkg.json
ahojnnes: Good catch. Thx.

pycolmap/ci/vcpkg-dependencies.txt
ahojnnes: What's the issue with larger wheels? Is there a maximum binary size enforced by pypi?

ahojnnes: So, Jasper is needed as a dependency induced from freeimage to handle RAW files? Sorry, not sure I understand that part correctly.

sarlinpe: freeimage requires libraw (we can't disable this), which requires jasper. Vcpkg installs `jasper[core,default-features,opengl]` but we only need `jasper[core]`. Could we add it to the manifest even though it's not a direct dependency?

Re wheel size: we're very far from the limit of PyPI, it's just more considerate for folks with low bandwidth.

ahojnnes: Thanks, understood. Re-added jasper[core]. Also added "cgal" as a vcpkg feature.

sarlinpe:
In manifest mode, dependencies are installed in `colmap/build/vcpkg_installed/$TRIPLET/` instead of `vcpkg/installed/$TRIPLET/`. When building an external library that depends on colmap (pycolmap here), how to tell vcpkg to search for dependencies at the new location?

sarlinpe (approved):

PR #2429, opened 2024-02-19
Perfrom triangulation in the IncrementalMapperController

sarlinpe:
This shouldn't incur any logic change.

### DIFF
--- a/src/colmap/sfm/incremental_mapper.cc
+++ b/src/colmap/sfm/incremental_mapper.cc
@@ -732,15 +732,7 @@ bool IncrementalMapper::AdjustGlobalBundle(
 
   // Run bundle adjustment.
   BundleAdjuster bundle_adjuster(ba_options_tmp, ba_config);
-  if (!bundle_adjuster.Solve(reconstruction_.get())) {
-    return false;
-  }
-
-  // Normalize scene for numerical stability and
-  // to avoid large scale changes in viewer.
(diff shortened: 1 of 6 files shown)


src/colmap/sfm/incremental_mapper.cc
```diff
@@ -798,7 +796,6 @@ void IncrementalMapper::IterativeGlobalRefinement(
       break;
     }
   }
-  FilterImages(options);
```
ahojnnes, line 801: We now lost the ability to remove degenerate images here?

sarlinpe: I moved it to the parent scope so it doesn't get called when triangulating only
https://github.com/colmap/colmap/blob/1ff3857f15c9df5143d90b63bf4f64945ced9375/src/colmap/controllers/incremental_mapper.cc#L47

ahojnnes (approved):

Ahmed-Salama:
I noticed that the old implementation calls `mapper.CompleteAndMergeTracks(tri_options);` however the new refactored code miss that part. Was that intentional or is that a mistake?

sarlinpe:
@Ahmed-Salama `CompleteAndMergeTracks` should be called in `IncrementalMapper::IterativeLocalRefinement`, which is called in the new `IncrementalMapperController::TriangulateReconstruction` function. Did you observe that this does not occur in practice?

PR #2432, opened 2024-02-20
setting latest tag for  `COLMAP_GIT_COMMIT` instead of dev or main

behnamasadi:
If the user doesn't specify any commit id (`COLMAP_GIT_COMMIT`) the Dockerfile set `COLMAP_GIT_COMMIT` to the latest tag, i.e. 3.9.1, otherwise it uses the user provided one 
Example 
`docker build --build-arg CUDA_ARCHITECTURES=75 .`
`docker build --build-arg COLMAP_GIT_COMMIT=main --build-arg CUDA_ARCHITECTURES=75 .`


ahojnnes (approved):
Thx

ahojnnes:
@behnamasadi Seems like there is also still some other issue? The CI build still fails.

behnamasadi:
Sorry for my late reply, will investigate it this weekend, thanks for your patience

PR #2436, opened 2024-02-21
Only push a new Docker image on release

sarlinpe:


.github/workflows/build-docker.yml
```diff
@@ -31,7 +35,7 @@ jobs:
       - name: Build and push
         run: |
           dockertag=$(date +%Y%m%d).${{ github.run_number }}
-          if [[ "${{ github.ref }}" == "refs/heads/main" ]]; then
+          if ${{ env.IS_RELEASE }}; then
```
ahojnnes, line 38: I think this only works with:
```suggestion
          if [[ "${{ env.IS_RELEASE }}" == "true" ]]; then
```

ahojnnes (approved):
Merci.

PR #2439, opened 2024-02-22
Filter unused two_view_geometry

whuaegeanse:
This PR improves the performance of DatabaseCache in certain scenarios by filtering out unused two_veiw_geometry. For example, the user only wants to process some of the images in the entire dataset.

### DIFF
--- a/src/colmap/scene/database_cache.cc
+++ b/src/colmap/scene/database_cache.cc
@@ -71,9 +71,31 @@ std::shared_ptr<DatabaseCache> DatabaseCache::Create(
   timer.Restart();
   LOG(INFO) << "Loading matches...";
 
+  std::unordered_set<image_t> image_ids;
+  std::vector<class Image> images;
+
+  {
+    images = database.ReadAllImages();
+
+    // Determines for which images data should be loaded.
+    if (image_names.empty()) {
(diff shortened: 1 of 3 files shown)


src/colmap/scene/database_cache.cc
```diff
@@ -71,9 +71,31 @@ std::shared_ptr<DatabaseCache> DatabaseCache::Create(
...
+          image_ids.insert(image.ImageId());
+        }
+      }
+    }
+  }
+
   std::vector<image_pair_t> image_pair_ids;
   std::vector<TwoViewGeometry> two_view_geometries;
-  database.ReadTwoViewGeometries(&image_pair_ids, &two_view_geometries);
+  database.ReadTwoViewGeometries(&image_pair_ids,
```
ahojnnes, line 96: Does this really save much runtime or memory? Below in L130, we are already applying that same filter?

whuaegeanse: Sorry for the late reply. This PR filters two_view_geometry based on image_ids, which can reduce the memory occupied by two_view_geometries. In particular, when the user only processes a small subset of all images, such as hierarchical SfM based on incremental SfM, each computing node only reconstructs a small part of all images.

PR #2458, opened 2024-03-07
Expose FindCorrespondences to python

mikoff:
This public function was omitted from pycolmap during refactoring.
The simplest commit to restore it.

### DIFF
--- a/src/pycolmap/scene/correspondence_graph.h
+++ b/src/pycolmap/scene/correspondence_graph.h
@@ -72,7 +72,18 @@ void BindCorrespondenceGraph(py::module& m) {
             }
             self.AddCorrespondences(image_id1, image_id2, matches);
           })
-      .def("find_correspondences", &CorrespondenceGraph::FindCorrespondences)
+      .def("find_correspondences",
+           [](const CorrespondenceGraph& self,
+              const image_t image_id,
+              const point2D_t point2D_idx) {
+             const auto corr_range =
+               self.FindCorrespondences(image_id, point2D_idx);
+             std::vector<CorrespondenceGraph::Correspondence> correspondences;               
+             for (const auto* corr = corr_range.beg; corr < corr_range.end; ++corr) {
+               corrs.push_back(*corr);
+             }
+             return corrs;
+           })
       .def("extract_correspondences",
            &CorrespondenceGraph::ExtractCorrespondences)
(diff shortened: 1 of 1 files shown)


sarlinpe:
This was intentionally changed in https://github.com/colmap/pycolmap/commit/edb859c2c4cfe56a3342b4564682418b4283824a following https://github.com/colmap/colmap/commit/b04b56d8b00209a630b6f6d581ac7d00b41f8e28. `extract_correspondences` should not be used instead. I haven't tested this binding and it might unfortunately not work because it expects a pointer to a vector. We might need to wrap it with a lambda to return the vector instead.

mikoff:
I don't get it - `FindCorrespondences` is a public method in `CorrespondenceGraph`, in addition to `ExtractCorrespondences`. So I believe it should be accessible on the Python side as well, along with `ExtractCorrespondences`, which is already there.
Do you mean that `ExtractCorrespondences` should be used instead? If so, should the visibility of the `FindCorrespondences` method be changed to private?

sarlinpe:
`FindCorrespondences` returns a `CorrespondenceRange` with pointers that cannot be exposed to Python. So this method makes sense in C++ but not in Python.

mikoff:
I have changed the binding, it works as expected in Python:
```python
cg.find_correspondences(1, 9839)
>> [Correspondence(image_id=2, point2D_idx=11705),
>>  Correspondence(image_id=4, point2D_idx=11806)]
```

ahojnnes:
Doesn't this new function do exactly the same as the existing extract_correspondences() function?

src/pycolmap/scene/correspondence_graph.h
```diff
@@ -72,7 +72,18 @@ void BindCorrespondenceGraph(py::module& m) {
...
             self.AddCorrespondences(image_id1, image_id2, matches);
           })
-      .def("find_correspondences", &CorrespondenceGraph::FindCorrespondences)
+      .def("find_correspondences",
+           [](const CorrespondenceGraph& self,
+              const image_t image_id,
+              const point2D_t point2D_idx) {
+             const auto corr_range =
+               self.FindCorrespondences(image_id, point2D_idx);
+             std::vector<CorrespondenceGraph::Correspondence> correspondences;
```
tmnku, line 81: Had to rename `correspondences` to `corrs` to compile.

PR #2474, opened 2024-03-20
Expose Reconstruction::SetUp to pycolmap

tmnku:
Reconstruction will only do the book keeping correctly (e.g. in functions such as AddPoint3D) if a CorrespondenceGraph is set.


tmnku:
Just noticed that this change requires `image_pair_stats_` to be setup properly in order to work.

sarlinpe:
This actually requires updating `Reconstruction::SetUp`, PR https://github.com/colmap/colmap/pull/2477 is a bit cleaner - does it work for you?

PR #2478, opened 2024-03-22
Add bindings for substeps of incremental mapper with a python example

B1ueber2y:
This PR contains the updates on the python bindings for substeps of IncrementalMapper and IncrementalMapperController. As a result, we are able to abstract a python pipeline as in the new example:

https://github.com/B1ueber2y/colmap/blob/features/incremental_mapper_bindings/pycolmap/example_substeps.py

where the users (particularly from the research community) can play around with the logic in python and potentially iterate faster. Up to discussions on the naming of this example file python logic and on where to put it. 

Ideally we can port the logic of BA with pyceres later such that one can construct customized problem, which can be potentially more friendly when integrating new features, sensors or the residuals from global sfm etc. 

cc: @sarlinpe @ahojnnes


B1ueber2y:
The code should be algorithmically equivalent to the C++ implementation with a small time overhead from Python (12s to 14s from my test on the 11-image example sequence). However, the results are not the same across runs although with ``pycolmap.set_random_seed(0)".

src/pycolmap/timer.h
```diff
@@ -0,0 +1,25 @@
...
+      .def("Pause", &Timer::Pause)
+      .def("Resume", &Timer::Resume)
+      .def("Reset", &Timer::Reset)
+      .def("ElapsedMicroSeconds", &Timer::ElapsedMicroSeconds)
+      .def("ElapsedSeconds", &Timer::ElapsedSeconds)
+      .def("ElapsedMinutes", &Timer::ElapsedMinutes)
+      .def("ElapsedHours", &Timer::ElapsedHours)
+      .def("PrintSeconds", &Timer::PrintSeconds)
+      .def("PrintMinutes", &Timer::PrintMinutes)
+      .def("PrintHours", &Timer::PrintHours);
```
sarlinpe, line 24: snake_case for all Python functions and variables

src/pycolmap/scene/reconstruction.cc
```diff
@@ -64,6 +64,35 @@ void BindReconstruction(py::module& m) {
...
+          },
+          "point3D_id"_a)
+      .def(
+          "image_pair",
+          [](Reconstruction& self, image_pair_t pair_id) {
+            const Reconstruction::ImagePairStat& image_pair =
+                self.ImagePair(pair_id);
+            return image_pair;
+          },
+          "pair_id"_a)
```
sarlinpe, line 95: Is this needed? We can already do `reconstruction.images[image_id]`.

src/pycolmap/scene/database_cache.cc
```diff
@@ -0,0 +1,43 @@
...
+            return camera;
+          },
+          "camera_id"_a)
+      .def(
+          "image",
+          [](DatabaseCache& self, image_t image_id) {
+            const struct Image& image = self.Image(image_id);
+            return image;
+          },
+          "image_id"_a)
```
sarlinpe, line 36: Is this needed? `DatabaseCache::{Cameras,Images}` should be sufficient.

src/pycolmap/scene/reconstruction_manager.cc
```diff
@@ -0,0 +1,30 @@
...
+void BindReconstructionManager(py::module& m) {
+  py::class_<ReconstructionManager, std::shared_ptr<ReconstructionManager>>(
+      m, "ReconstructionManager")
+      .def(py::init<>())
+      .def("size", &ReconstructionManager::Size)
+      .def(
+          "get",
+          [](ReconstructionManager& self, size_t idx) {
+            std::shared_ptr<const Reconstruction> rec = self.Get(idx);
+            return rec;
```
sarlinpe, line 22: Is the lambda needed? Isn't it sufficient to directly bind `ReconstructionManager::Get`? Use `py::overload_cast` to return the non-const version.

src/colmap/controllers/incremental_mapper.h
```diff
@@ -165,21 +165,38 @@ class IncrementalMapperController : public BaseController {
   void TriangulateReconstruction(
       const std::shared_ptr<Reconstruction>& reconstruction);
 
- private:
   bool LoadDatabase();
+
+  // getter functions for python pipelines
+  const std::string GetImagePath() const { return image_path_; }
```
sarlinpe, line 171: The `Get*` prefix is a bit unnecessary, the codebase doesn't have it anywhere else. Maybe remove it for consistency and conciseness? Same for the corresponding Python bindings.

src/pycolmap/sfm/incremental_mapper.cc
```diff
@@ -1,27 +1,89 @@
...
+           &IncrementalMapperController::ReconstructSubModel,
+           "core_mapper"_a,
+           "mapper_options"_a,
+           "reconstruction"_a)
+      .def("InitializeReconstruction",
+           &IncrementalMapperController::InitializeReconstruction,
+           "core_mapper"_a,
+           "mapper_options"_a,
+           "reconstruction"_a)
+      .def("Run", &IncrementalMapperController::Run);
```
sarlinpe, line 80: snake_case for Python functions.

src/pycolmap/sfm/incremental_mapper.cc
```diff
@@ -96,3 +158,143 @@ void BindIncrementalMapper(py::module& m) {
...
+      .def(
+          "estimate_initial_two_view_geometry",
+          [](IncrementalMapper& self,
+             const IncrementalMapper::Options& options,
+             const image_t image_id1,
+             const image_t image_id2) {
+            TwoViewGeometry two_view_geometry;
+            bool success = self.EstimateInitialTwoViewGeometry(
+                options, two_view_geometry, image_id1, image_id2);
+            return std::make_pair(success, two_view_geometry);
```
sarlinpe, line 232: Alternatively: return `None` on failure

src/colmap/controllers/incremental_mapper.h
B1ueber2y: The method name GetReconstructionManager and GetDatabaseCache will collapse with the class name without "Get", so here we may need to find a different name. Also, "get_" is used here in Python bindings before: https://github.com/colmap/colmap/blob/main/src/pycolmap/pipeline/sfm.cc#L251. So I would personally prefer to keep it as it is. Or do you prefer to only keep "Get" for "ReconstructionManager" and "DatabaseCache"?

src/pycolmap/scene/reconstruction.cc
B1ueber2y: Indeed. This is resolved. Thanks

src/colmap/controllers/incremental_mapper.h
sarlinpe: Collapsing with the class name is fine, method named are scoped. This is already like this in `Reconstruction`:
https://github.com/colmap/colmap/blob/06a805386afb3db9dc6704c7ba2a80861bb64f7e/src/colmap/scene/reconstruction.h#L85-L87

src/pycolmap/sfm/incremental_mapper.cc
B1ueber2y: Thanks. This is done.

src/pycolmap/scene/database_cache.cc
```diff
@@ -0,0 +1,29 @@
...
+                  "min_num_matches"_a,
+                  "ignore_watermarks"_a,
+                  "image_names"_a)
+      .def("num_cameras", &DatabaseCache::NumCameras)
+      .def("num_images", &DatabaseCache::NumImages)
+      .def("exists_camera", &DatabaseCache::ExistsCamera, "camera_id"_a)
+      .def("exists_image", &DatabaseCache::ExistsImage, "image_id"_a)
+      .def_property_readonly("cameras", &DatabaseCache::Cameras)
+      .def_property_readonly("images", &DatabaseCache::Images)
+      .def("correspondence_graph", &DatabaseCache::CorrespondenceGraph)
```
sarlinpe, line 27: Shouldn't this also be `def_property_readonly` ?

src/pycolmap/sfm/incremental_mapper.cc
```diff
@@ -96,3 +158,146 @@ void BindIncrementalMapper(py::module& m) {
...
+             const IncrementalMapper::Options& options,
+             const image_t image_id1,
+             const image_t image_id2) {
+            TwoViewGeometry two_view_geometry;
+            bool success = self.EstimateInitialTwoViewGeometry(
+                options, two_view_geometry, image_id1, image_id2);
+            if (success)
+              return py::cast(two_view_geometry);
+            else
+              return py::cast<py::object>(Py_None);
```
sarlinpe, line 235: Use `py::none()` instead?

https://github.com/colmap/colmap/blob/06a805386afb3db9dc6704c7ba2a80861bb64f7e/src/pycolmap/estimators/absolute_pose.cc#L38

B1ueber2y: Lambda function needs a consistent type of output. I was using ``py::none()`` before but got this unfortunately:

error: inconsistent types ‘pybind11::object’ and ‘pybind11::none’ deduced for lambda return type.

So I changed into the current design.

sarlinpe: How about if you annotate the return type with `-> py::object`? Then you maybe don’t even need the other `py::cast`

src/pycolmap/scene/database_cache.cc
B1ueber2y: Done.

src/pycolmap/sfm/incremental_mapper.cc
B1ueber2y: Nice idea. This is done. the ``py::cast`` on TwoViewGeometry object is still needed though.

src/pycolmap/scene/reconstruction_manager.cc
B1ueber2y: Done. Thanks

src/colmap/controllers/incremental_mapper.h
B1ueber2y: This is done. Thanks

src/colmap/controllers/incremental_mapper.h
```diff
@@ -158,33 +158,50 @@ class IncrementalMapperController : public BaseController {
...
 
   void TriangulateReconstruction(
       const std::shared_ptr<Reconstruction>& reconstruction);
 
- private:
   bool LoadDatabase();
+
+  // getter functions for python pipelines
+  const std::string ImagePath() const { return image_path_; }
+  const std::string DatabasePath() const { return database_path_; }
```
ahojnnes, line 172: const ref?

ahojnnes: Above and below for all getters. Or does this not work with Pybind?

src/pycolmap/timer.h
```diff
@@ -0,0 +1,25 @@
...
+
+void BindTimer(py::module& m) {
+  py::class_<Timer>(m, "Timer")
+      .def(py::init<>())
+      .def("start", &Timer::Start)
+      .def("restart", &Timer::Restart)
+      .def("pause", &Timer::Pause)
+      .def("resume", &Timer::Resume)
+      .def("reset", &Timer::Reset)
+      .def("elapsed_microSeconds", &Timer::ElapsedMicroSeconds)
```
ahojnnes, line 18:
```suggestion
      .def("elapsed_micro_seconds", &Timer::ElapsedMicroSeconds)
```

pycolmap/custom_incremental_mapping.py
```diff
@@ -0,0 +1,309 @@
+import shutil
```
ahojnnes, line 1: Maybe we also call this "example_incremental_mapping.py" and add a brief `""" Comment """` to the header explaining what the purpose of this example is (i.e., equivalent logic as C++ incremental mapper, etc.)?

ahojnnes (commented):
Thanks, this looks great! Some minor comments.

src/colmap/controllers/incremental_mapper.h
B1ueber2y: It should work with pybind. I have updated this.

src/pycolmap/timer.h
B1ueber2y: Updated. Thanks

pycolmap/custom_incremental_mapping.py
B1ueber2y: I have added a draft description. Since Paul has restructured the file a bit and only use ``example.py`` as the main runner, I am looping him in here @sarlinpe  for potential further changes on this : )

ahojnnes (approved):
LGTM, thank you!

pycolmap/custom_incremental_mapping.py
ahojnnes: Thanks, this looks good.

PR #2497, opened 2024-04-07
Fix missing public link deps, break circular feature-scene dependency

ahojnnes:
* Propagates missing public link dependencies that prevent ceres headers from being found.
* These new dependencies revealed a circular dependency between scene/feature that is broken by a new feature_types library.


sarlinpe:
1) Wouldn't it be simpler to merge `feature/types.h` into `util/types.h`?
2) The pycolmap CI is down because `liblzma` is now down following the vulnerability discovered in xz: https://github.com/microsoft/vcpkg/issues/37839 Best is to wait that this gets fixed in vcpkg.

ahojnnes:
1. I would like to avoid shoving more types into the global util/types to avoid a global knot in the dependency graph.
2. Thanks, I'll make a vcpkg update in a separate PR first.

sarlinpe:
liblzma is now back online but freeimage fails on all VCPKG-based pipelines.

sarlinpe (approved):

PR #2510, opened 2024-04-12
Include pybind eigen header

tmnku:
I needed the header to run this [example](https://github.com/colmap/colmap/tree/main/pycolmap#sift-feature-extraction). Without it pycolmap would complain about passing a np.array.


sarlinpe:
Thank you! This issue was likely introduced when I split pycolmap into separate compilation units in https://github.com/colmap/colmap/commit/06a805386afb3db9dc6704c7ba2a80861bb64f7e. It might affect other files - we could preemptively add this missing include to all files that don't yet have it.

sarlinpe (approved):

Issue #2525, opened 2024-04-19
'std::out_of_range' error when using hierarchical_mapper with CMD


GrayMask:
**Environment:**
 - OS: Ubuntu 22.04.3 LTS
 - COLMAP Version: 05473b1

**Describe the bug**
When I try to run `hierarchical_mapper` on a small dataset (89 images) and it crashed

And the command is like :
`colmap hierarchical_mapper --database_path [PATH_TO]/database.db --image_path  [PATH_TO]/images --output_path [PATH_TO]/sparse`

The error message is:
```
==============================================================================
Partitioning scene
==============================================================================
I0418 19:50:06.968232 11593 hierarchical_mapper.cc:132] Reading images...
I0418 19:50:06.968348 11593 scene_clustering.cc:315] Reading scene graph...
I0418 19:50:06.969987 11593 scene_clustering.cc:320] Partitioning scene graph...
I0418 19:50:06.970058 11593 hierarchical_mapper.cc:148]   Cluster 1 with 89 images
I0418 19:50:06.970065 11593 hierarchical_mapper.cc:153] Clusters have 89 images
I0418 19:50:06.970073 11593 misc.cc:198] 
==============================================================================
Reconstructing clusters
==============================================================================
I0418 19:50:06.970914 11594 incremental_mapper.cc:225] Loading database
I0418 19:50:06.972855 11594 database_cache.cc:65] Loading cameras...
I0418 19:50:06.973109 11594 database_cache.cc:75]  3 in 0.000s
I0418 19:50:06.973121 11594 database_cache.cc:83] Loading matches...
I0418 19:50:06.976661 11594 database_cache.cc:89]  675 in 0.004s
I0418 19:50:06.976677 11594 database_cache.cc:105] Loading images...
I0418 19:50:06.988271 11594 database_cache.cc:155]  90 in 0.012s (connected 89)
I0418 19:50:06.988304 11594 database_cache.cc:166] Building correspondence graph...
I0418 19:50:07.013770 11594 database_cache.cc:203]  in 0.025s (ignored 0)
I0418 19:50:07.013913 11594 timer.cc:91] Elapsed time: 0.001 [minutes]
I0418 19:50:07.017303 11594 incremental_mapper.cc:265] Finding good initial image pair
I0418 19:50:07.100188 11594 incremental_mapper.cc:289] Initializing with image pair #46 and #87
I0418 19:50:07.105177 11594 incremental_mapper.cc:294] Global bundle adjustment
[...]

PR #2526, opened 2024-04-19
Fix 'std::out_of_range' error when using hierarchical_mapper

GrayMask:
Fix a bug described here:
https://github.com/colmap/colmap/issues/2525#issue-2251904473


ahojnnes (approved):
Thanks for the fix. LGTM.

Issue #2530, opened 2024-04-22
The opaque type binding of ListPoint2D is not working


B1ueber2y:
Initially described and discussed here: https://github.com/colmap/colmap/pull/2509

The opaque type binding of ListPoint2D (here: https://github.com/colmap/colmap/blob/main/src/pycolmap/scene/point2D.cc#L35) is not working after the attempt of removing all headers for faster compilation introduced in: https://github.com/colmap/colmap/pull/2443.

To reproduce:

```
import pycolmap
rec = pycolmap.Reconstruction("example/sfm/0")
print(type(rec.images[1].points2D)) # return <class 'list'> rather than <class 'ListPoint2D'> 
```

This issue is crucial to the performance of ``image.points2D`` due to the incurring copying behavior, and is a must-fix case since it will affect the performance of constructing bundle adjustment with pyceres, where the code does not necessarily need to traverse all 2d points.

To add on it, I can confirm that the fix for ``reconstruction.points3D`` employed in https://github.com/colmap/colmap/pull/2502 and https://github.com/colmap/colmap/pull/2523 does not resolve this issue.

cc: @sarlinpe

B1ueber2y:
OK news to this. This feature is never working even before https://github.com/colmap/colmap/pull/2443 (checkout back before the commit it still does not work), so it is a different issue to ``reconstruction.points3D``.

PR #2533, opened 2024-04-22
Fix binding for std::vector<Point2D>

sarlinpe:
The fix introduced in https://github.com/colmap/pycolmap/pull/95 actually disables the binding of `std::vector<Point2D>`, which defaults to a regular Python list and is much slower, as reported in https://github.com/colmap/colmap/issues/2530. To fix this, I had to 1) remove the Eigen aligned operator, and 2) apply the opaque macro to `camera.cc` and `image.cc`.

Run time of `image.points2D[index]`:
- main branch: 1ms
- this PR: 7us
- using `image.point2D(index)`: 4us (introduced in this PR as shortcut)

Fixes https://github.com/colmap/colmap/issues/2530.


B1ueber2y:
Looks nice. Thanks. Any idea why point2D(index) is faster? Should we introduce this to all similar interfaces?

sarlinpe:
Building the opaque object wrapping `std::vector` and calling `__getitem__` for have some overheads. Since you already added `Reconstruction::{Camera,Image,Point3D}` I think that there's nowhere else to apply this.

ahojnnes (approved):

PR #2551, opened 2024-05-03
Add ignore missing files option to geometric patchmatch for PMVS

awarebayes:
If this flag is not included, check fails in https://github.com/colmap/colmap/blob/main/src/colmap/mvs/patch_match.cc#L507. 

It looks for an image that is not in this PMVS option, but in other option. 

With this flag set to true, geometric patchmatch works perfectly fine (after running photometric).


awarebayes:
Well, it seems that the entire geometric mvs with pmvs does not work

PR #2558, opened 2024-05-07
Implicitly convert iterator to ListPoint2D

sarlinpe:
Fix errors when pass an empty list as a `vector<Point2D>`:
```
    pycolmap.Image(points2D=[])
RuntimeError: instance allocation failed: new instance has no pybind11-registered base types
```
```
    camera.cam_from_img([])
TypeError: cam_from_img(): incompatible function arguments.
```
cc @B1ueber2y

### DIFF
--- a/src/pycolmap/scene/bindings.cc
+++ b/src/pycolmap/scene/bindings.cc
@@ -1,3 +1,5 @@
+#include "pycolmap/scene/types.h"
+
 #include <pybind11/pybind11.h>
 
 namespace py = pybind11;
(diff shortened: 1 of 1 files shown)


B1ueber2y:
I confirm that the fix works in my test. Thanks! However, do we need to put it in the ``scene/bindings.cc``? It does not feel like the ideal place : )

sarlinpe:
It really needs to be there. This doesn't work if placed in `point2D.cc`. A working alternative is to place it in both `image.cc` and `camera.cc` , which duplicates code.

ahojnnes (approved):

PR #2561, opened 2024-05-14
Add support for IMU Preintegration and discrete-time visual-inertial optimization

B1ueber2y:

- Base class implementation for IMU calibration and measurements
- On-manifold preintegration with noise propagation
- Preintegration cost factor in ceres
- Time-varying bias modeling and incorporation of bias correction in the preintegration factor
- Pybind for the features above

The current design requires storing a Rigid3d for ``cam_to_imu`` and a scale parameter at optimization, which is visual centric (this cannot work when visual images are not available). This is planned to be fixed in the future with better multi-sensor rig support. 

The implementation is done from scratch with the following paper references and the GTSAM library (https://github.com/borglab/gtsam):

[A] Forster et al. "On-Manifold Preintegration for Real-Time Visual-Inertial Odometry", https://rpg.ifi.uzh.ch/docs/TRO16_forster.pdf
[B] Cioffi et al. "Continuous-time vs. discrete-time vision-based SLAM: A comparative study", https://arxiv.org/pdf/2202.08894
[C] von Stumberg et al. DM-VIO: Delayed Marginalization Visual-Inertial Odometry, https://arxiv.org/pdf/2201.04114

More tests and benchmarks will be done and a minimal example will be uploaded in the following days (worst case next week), to make sure that the implementation is fully correct.

### DIFF
--- a/README.md
+++ b/README.md
@@ -42,7 +42,7 @@ and consider citing relevant third-party dependencies.
 Download
 --------
 
-* Binaries for **Windows** and **Mac** and other resources can be downloaded
+* Binaries for **Windows** and other resources can be downloaded
   from https://github.com/colmap/colmap/releases.
 * Binaries for **Linux/Unix/BSD** are available at
   https://repology.org/metapackage/colmap/versions.
--- a/doc/install.rst
+++ b/doc/install.rst
@@ -156,9 +156,13 @@ Configure and compile COLMAP::
     export PATH="/usr/local/opt/qt@5/bin:$PATH"
     mkdir build
     cd build
-    cmake ..  -GNinja -DQt5_DIR=/usr/local/opt/qt/lib/cmake/Qt5
+    cmake .. -GNinja -DQt5_DIR=/usr/local/opt/qt/lib/cmake/Qt5
     ninja
     sudo ninja install
+
+On Macs with ARM the brew paths are different so you need this
+
+    cmake .. -GNinja -DCMAKE_PREFIX_PATH="/opt/homebrew/opt/flann;/opt/homebrew/opt/metis;/opt/homebrew/opt/suite-sparse;/opt/homebrew/opt/qt@5;/opt/homebrew/opt/freeimage"
 
 If you have Qt 6 installed on your system as well, you might have to temporarily
 link your Qt 5 installation while configuring CMake::


ahojnnes:
This looks exciting, thank you Shaohui. I'll try to carve out some time in the next days to have a look.

src/colmap/sensor/imu.h
```diff
@@ -0,0 +1,145 @@
...
+
+// priority list for Imu measurements based on timestamps
+class ImuMeasurements {
+ public:
+  ImuMeasurements() = default;
+  ~ImuMeasurements() = default;
+  explicit ImuMeasurements(const std::vector<ImuMeasurement>& ms) {
+    insert(ms);
+  }
+  ImuMeasurements(const ImuMeasurements& ms) { insert(ms); }
```
ahojnnes, line 94: Inserting from a list of measurements should be much more efficient by just copying the vector and sorting it by time?

src/colmap/sensor/imu.h
```diff
@@ -0,0 +1,145 @@
...
+    auto it = std::lower_bound(
+        measurements_.begin(),
+        measurements_.end(),
+        m,
+        [](const ImuMeasurement& m1, const ImuMeasurement& m2) {
+          return m1.timestamp < m2.timestamp;
+        });
+    measurements_.insert(it, m);
+  }
+  void insert(const std::vector<ImuMeasurement>& ms) {
```
ahojnnes, line 105: Is there an actual use case where we would insert one by one?

src/colmap/sensor/imu.h
```diff
@@ -0,0 +1,145 @@
...
+  double gyro_saturation_max = 7.8;
+
+  /// Norm of the Gravitational acceleration. [m/s^2]
+  double gravity_magnitude = 9.81007;
+
+  /// Expected IMU Rate [1/s]
+  double imu_rate = 20.0;
+
+  ImuCalibration() = default;
+  ~ImuCalibration() = default;
```
ahojnnes, line 71: Not needed?

ahojnnes: Same for other structs in this file.

src/colmap/scene/imu.h
```diff
@@ -0,0 +1,131 @@
...
+
+  inline const camera_t& ImuId() const;
+  inline camera_t ImuId();
+  inline const image_t& ImageId() const;
+  inline image_t ImageId();
+
+ private:
+  Eigen::Matrix<double, 9, 1> data_ =
+      Eigen::Matrix<double, 9, 1>::Zero();  // 3-DoF speed + 6-DoF biases (acc +
+                                            // gyro)
```
ahojnnes, line 86: Could be nicer to store them here as three independent Vector3d's and then, whenever we need them as a singl state-vector (e.g., during optimization), copy them into a Vector9d ?

Then this whole class can become a simple data struct.

src/colmap/scene/imu.h
```diff
@@ -0,0 +1,131 @@
...
+class Imu {
+ public:
+  Imu() = default;
+  ~Imu() = default;
+  ImuCalibration calib;
+  camera_t imu_id = kInvalidCameraId;
+
+  // information for the associated visual camera. TODO: change to rigs
+  camera_t camera_id = kInvalidCameraId;  // the camera linked to IMU.
+  Rigid3d cam_to_imu;
```
ahojnnes, line 56: Let's consistently use a_from_b notation instead of b_to_a.

src/colmap/geometry/pose.cc
```diff
@@ -153,6 +154,27 @@ Rigid3d InterpolateCameraPoses(const Rigid3d& cam_from_world1,
...
+  double theta = omega.norm();
+  Eigen::Matrix3d J;
+  if (theta < 1e-6)
+    J = Eigen::Matrix3d::Identity() - 0.5 * skew_omega;
+  else {
+    Eigen::Matrix3d M = skew_omega / theta;
+    J = Eigen::Matrix3d::Identity() - ((1.0 - std::cos(theta)) / theta) * M +
+        (1.0 - std::sin(theta) / theta) * M * M;
+  }
+  return J;
```
ahojnnes, line 175:
```suggestion
  const double theta = omega.norm();
  if (theta < 1e-6) {
    return Eigen::Matrix3d::Identity() - 0.5 * skew_omega;
  } else {
    const Eigen::Matrix3d M = skew_omega / theta;
    return Eigen::Matrix3d::Identity() - (((1.0 - std::cos(theta)) / theta) +
        (1.0 - std::sin(theta) / theta) * M) * M;
  }
```

src/colmap/geometry/pose.cc
```diff
@@ -153,6 +154,27 @@ Rigid3d InterpolateCameraPoses(const Rigid3d& cam_from_world1,
                  cam_from_world1.translation + translation12 * t);
 }
 
+Eigen::Quaterniond QuaternionFromAngleAxis(const Eigen::Vector3d& omega) {
+  double half_angle = 0.5 * omega.norm();
+  Eigen::Vector3d axis = omega.normalized();
+  Eigen::Quaterniond q(Eigen::AngleAxisd(half_angle, axis));
+  return q;
+}
```
ahojnnes, line 162: 1. We unnecessarily compute the norm of omega twice.
2. omega.normalized() can lead to NaNs.

src/colmap/geometry/pose.cc
B1ueber2y: Thanks. (1.0 - std::cos(theta)) / theta) is a double so we need to multiple an identity matrix here. This is resolved.

src/colmap/geometry/pose.cc
B1ueber2y: Thanks. This is updated.

src/colmap/sensor/imu.h
B1ueber2y: Thanks. This is updated.

src/colmap/sensor/imu.h
B1ueber2y: I would vote for keeping this incremental insertion feature for a potential use case when multiple IMUs are available (though one could still argue to have all the measurements collected first). 

But I agree that the performance can be optimized by merging two sorted list for the second method.

src/colmap/scene/imu.h
B1ueber2y: In the current plan, it is very likely that the bias goes into the sensor calibration in the end, and the time-varying bias is modeled with the interpolator design. So this ``ImuState`` class will very likely be useless when we add the rigs. 

This being said, it is also logical to first make the rig design available and then adapt this PR accordingly : )

B1ueber2y:
After several updates and fixes I think it is safe to confirm that the implementation is stable and correct. I have just added a test example for vi optimization here: https://github.com/colmap/colmap/pull/2561/commits/f11c24df4aa5a03c8e7b5bf789cda5cdd8e5dc02.

In this test, the data is captured with Meta Aria glasses. The intrinsics and extrinsics are initialized from the [MPS service](https://facebookresearch.github.io/projectaria_tools/docs/data_utilities/core_code_snippets/mps). We input to our system the rectified IMU measurement, using the predicted rectification matrix and biases in the MPS output, and do visual-inertial optimization with COLMAP bundle adjustment and the pycolmap imu preintegration algorithms and factors, with the extrinsics initialized from MPS, velocities initialized with linear approximation, and imu_from_cam transformation initialized as identity. 

The test proves that the optimization converges well to a residual level that is similar to the previous bundle adjustment. The resulting scale is nearly identity and the time-varying bias over the rectified measurement is near zero, which proves that the algorithm does the same thing as the inertial optimization integrated in the MPS service. Another test shows that the residuals with the correct imu_from_cam initialization (from MPS) is generally very small.  

I would consider this feature to be ready for review, while we could wait for the finish of rig development to merge it, such that the bias goes into the calibration in the new design.

src/colmap/geometry/pose.cc
B1ueber2y: This was actually a bug. AngleAxisd should receive the full angle rather than half angle. And this is fixed now.

Richard-coder:
@B1ueber2y hi, I run your example for vi optimization, I get this error:
error

B1ueber2y:
> Richard-coder

Hey, your error is because that your pycolmap installed cannot parse the classes from pyceres. This is very likely due to the fact that your pycolmap and pyceres are using different version of ceres at building. For example, if you install pyceres with pypi and install pycolmap from source. This can be resolved by either building both from source, or installing both with wheels. For pycolmap you can find wheels for this PR from github action here: https://github.com/colmap/colmap/actions/runs/9393336850

Richard-coder:
I install pycolmap wheels form PR, I can run example.py now, thank you.

src/pycolmap/scene/imu.cc
```diff
@@ -0,0 +1,123 @@
...
+          py::overload_cast<>(&ImuState::Data),
+          [](ImuState& self, const Eigen::Matrix<double, 9, 1>& data) {
+            self.Data() = data;
+          })
+      .def_property_readonly("velocity", &ImuState::Velocity)
+      .def_property_readonly("velocity_ptr", &ImuState::VelocityPtr)
+      .def_property_readonly("acc_bias", &ImuState::AccBias)
+      .def_property_readonly("acc_bias_ptr", &ImuState::AccBiasPtr)
+      .def_property_readonly("gyro_bias", &ImuState::GyroBias)
+      .def_property_readonly("gyro_bias_ptr", &ImuState::GyroBiasPtr)
```
sarlinpe, line 114: You can make the properties writable and remove the `set_*` methods.

B1ueber2y: The three properties here are continuous data under Eigen::Vector9d. To be able to do ``set_*`` we require the input to be Eigen::Vector3d to avoid invalid access of data.

src/colmap/scene/imu.h
```diff
@@ -0,0 +1,134 @@
...
+const Eigen::Matrix<double, 9, 1>& ImuState::Data() const { return data_; }
+
+Eigen::Matrix<double, 9, 1>& ImuState::Data() { return data_; }
+
+void ImuState::SetVelocity(const Eigen::Vector3d& vec) {
+  data_.head<3>() = vec;
+}
+
+const Eigen::Vector3d ImuState::Velocity() const {
+  return Eigen::Vector3d(data_.data());
```
sarlinpe, line 101: Wouldn't it be cleaner to return a `Eigen::map<Eigen::Vector3d>` here?

B1ueber2y: We anyway will wait for the rig feature to merge this, where the bias will go to calibration and this will become a single attribute : )

PR #2573, opened 2024-05-22
Split pair generation and matching

sarlinpe:
This is an attempt at splitting the generation of image pairs (exhaustive, spatial, vocabulary tree, sequential) and the matching process, which were deeply entangled. This simplifies the code and makes it possible to generate the pairs from Python for a subsequent Python-based matching.

The design is sometimes convoluted because I make sure to avoid materializing all pairs in memory, and instead generate them by batches.

This currently lives in `controllers/` but could be moved to `features/`.


src/colmap/controllers/feature_matching.cc
```diff
@@ -47,523 +46,75 @@ void PrintElapsedTime(const Timer& timer) {
...
-                                       image_ids.at(image_idx2_quadratic));
-            }
-          }
-        } else {
-          break;
-        }
-      }
-
-      DatabaseTransaction database_transaction(&database_);
+      std::vector<std::pair<image_t, image_t>> image_pairs =
```
ahojnnes, line 87: nit: const

src/colmap/controllers/feature_matching.cc
```diff
@@ -47,523 +46,75 @@ void PrintElapsedTime(const Timer& timer) {
...
-        cache_(5 * options_.block_size, &database_),
-        matcher_(matching_options, geometry_options, &database_, &cache_) {
-    THROW_CHECK(options.Check());
+  GenericFeatureMatcher(
+      const typename DerivedPairGenerator::PairOptions& pair_options,
+      const SiftMatchingOptions& matching_options,
+      const TwoViewGeometryOptions& geometry_options,
+      const std::string& database_path)
+      : pair_options_(pair_options),
+        database_(new Database(database_path)),
```
ahojnnes, line 58: nit: make_shared instead of new

src/colmap/controllers/feature_matching.cc
```diff
@@ -47,523 +46,75 @@ void PrintElapsedTime(const Timer& timer) {
...
-    }
-
-    matcher->Match(image_pairs);
-
-    PrintElapsedTime(timer);
-  }
-}
-
-class ExhaustiveFeatureMatcher : public Thread {
+template <typename DerivedPairGenerator>
```
ahojnnes, line 49: I guess instead of templating we could pass an abstract pair generator and the cache object from the outside. Not necessarily important for here but theoretically has some benefits for injecting a mock/fake for testing purposes of this class.

src/colmap/controllers/feature_matching.cc
```diff
@@ -1088,8 +331,8 @@ class FeaturePairsFeatureMatcher : public Thread {
       : options_(options),
         matching_options_(matching_options),
         geometry_options_(geometry_options),
-        database_(database_path),
-        cache_(kCacheSize, &database_) {
+        database_(new Database(database_path)),
```
ahojnnes, line 334: make_shared

src/colmap/controllers/feature_matching.cc
```diff
@@ -810,9 +186,10 @@ class TransitiveFeatureMatcher : public Thread {
                            const std::string& database_path)
       : options_(options),
         matching_options_(matching_options),
-        database_(database_path),
-        cache_(options_.batch_size, &database_),
-        matcher_(matching_options, geometry_options, &database_, &cache_) {
+        database_(new Database(database_path)),
```
ahojnnes, line 189: make_shared

ahojnnes (approved):
Thanks, looks like a good refactor to me. I am also fine with the proposed move from controllers/ to features/.

ahojnnes (approved):

ahojnnes:
@sarlinpe Mac is unhappy with the following error:
```
colmap::(anonymous namespace)::FeaturePairsFeatureMatcher::kCacheSize, referenced from:
```

PR #2575, opened 2024-05-24
Add ObservationManager

sarlinpe:
This removes the filtering logic in `Reconstruction` as well as the SfM-specific data in `Image`, such as the visibility pyramid. The logic is unchanged except one minor point: I removed the `num_observations` column in the GUI database image viewer because this field is now only available during reconstruction. It is anyway mostly redundant with `num_points2D`.


src/colmap/scene/synthetic_test.cc
```diff
@@ -80,9 +82,39 @@ TEST(SynthesizeDataset, Nominal) {
...
+        proj_center1 = proj_centers.at(image_id1);
+      }
+
+      for (size_t i2 = 0; i2 < i1; ++i2) {
+        const image_t image_id2 = point3D.track.Element(i2).image_id;
+        const Eigen::Vector3d proj_center2 = proj_centers.at(image_id2);
+
+        const double tri_angle = CalculateTriangulationAngle(
+            proj_center1, proj_center2, point3D.xyz);
+        EXPECT_GE(tri_angle, DegToRad(kMinTriAngleDeg));
```
sarlinpe, line 114: This test sometimes fails on mac and Windows with the previous value of 1°. It seems that the sampling of camera poses doesn't constrain them to have a sufficient angle.

src/colmap/sfm/observation_manager.h
```diff
@@ -51,9 +51,9 @@ class ObservationManager {
     size_t num_total_corrs = 0;
   };
 
-  ObservationManager(std::shared_ptr<Reconstruction> reconstruction,
-                     std::shared_ptr<const CorrespondenceGraph>
-                         correspondence_graph = nullptr);
+  explicit ObservationManager(std::shared_ptr<Reconstruction> reconstruction,
```
ahojnnes, line 54: nit: explicit only meaningful for single-arg constructor.

src/colmap/controllers/incremental_mapper.cc
```diff
@@ -552,7 +552,7 @@ void IncrementalMapperController::TriangulateReconstruction(
     LOG(INFO) << StringPrintf("Triangulating image #%d (%d)", image_id, i);
     const size_t num_existing_points3D = image.NumPoints3D();
     LOG(INFO) << "=> Image sees " << num_existing_points3D << " / "
-              << image.NumObservations() << " points";
+              << image.NumPoints2D() << " points";
```
ahojnnes, line 555: The # observations is more meaningful, as it is the upper bound of visible points in the image, but it's not critical if not easy to do here.

src/colmap/estimators/alignment.cc
```diff
@@ -408,9 +408,10 @@ bool AlignReconstructionsViaPoints(const Reconstruction& src_reconstruction,
   return report.success;
 }
 
-bool MergeReconstructions(const double max_reproj_error,
-                          const Reconstruction& src_reconstruction,
-                          Reconstruction* tgt_reconstruction) {
+bool MergeReconstructions(
+    const double max_reproj_error,
+    const Reconstruction& src_reconstruction,
+    const std::shared_ptr<Reconstruction>& tgt_reconstruction) {
```
ahojnnes, line 414: I suggest to make this a ref instead of a shared_ptr. The current shared_ptr interface enforces a particular memory management model on the user side. For example, a reconstruction in a unique_ptr would be difficult to pass here without copying the data.

src/colmap/sfm/incremental_mapper.cc
```diff
@@ -839,6 +852,11 @@ std::shared_ptr<class Reconstruction> IncrementalMapper::Reconstruction()
   return reconstruction_;
 }
 
+std::shared_ptr<class ObservationManager>
```
ahojnnes, line 855: Didn't check all the callsites, but could we instead return a reference here instead of the shared_ptr or alternatively a const ref to the shared_ptr, otherwise each access to the shared_ptr incurs a copy and then a ref counter decrement when that object is destructred.

src/colmap/sfm/observation_manager.cc
```diff
@@ -0,0 +1,495 @@
...
+#include "colmap/scene/projection.h"
+#include "colmap/util/logging.h"
+#include "colmap/util/misc.h"
+
+namespace colmap {
+
+bool MergeAndFilterReconstructions(
+    const double max_reproj_error,
+    const Reconstruction& src_reconstruction,
+    const std::shared_ptr<Reconstruction>& tgt_reconstruction) {
```
ahojnnes, line 44: Same suggestion here: Make it a ref instead of shared_ptr.

src/colmap/sfm/observation_manager.cc
```diff
@@ -0,0 +1,495 @@
...
+  return true;
+}
+
+const int ObservationManager::kNumPoint3DVisibilityPyramidLevels = 6;
+
+ObservationManager::ObservationManager(
+    std::shared_ptr<Reconstruction> reconstruction,
+    std::shared_ptr<const CorrespondenceGraph> correspondence_graph)
+    : reconstruction_(std::move(THROW_CHECK_NOTNULL(reconstruction))),
+      correspondence_graph_(std::move(correspondence_graph)) {
```
ahojnnes, line 60: Also here THROW_CHECK_NOTNULL?

ahojnnes:
Looks great to me. I assume most of the code in the ObservationManager is untouched and just moved from Reconstruction/Image classes? Have you run through a dataset and checked that it produces the exact same number of final observations?

src/colmap/sfm/observation_manager.h
sarlinpe: Done.

src/colmap/controllers/incremental_mapper.cc
sarlinpe: Done.

src/colmap/estimators/alignment.cc
sarlinpe: Good point. Fixed. For consistency, I have also updated `IncrementalTriangulator::reconstruction_` to be a reference.

src/colmap/sfm/incremental_mapper.cc
sarlinpe: Done.

src/colmap/sfm/observation_manager.cc
sarlinpe: Done.

src/colmap/sfm/observation_manager.cc
sarlinpe: `correspondence_graph` is an optional argument since it is not needed to perform filtering only.

sarlinpe:
PTAL. I have indeed mostly moved around code without change of logic. I have checked that the result is identical in terms of reconstruction statistics.

src/colmap/sfm/observation_manager.h
sarlinpe: ClangTidy actually requires it because the second argument is optional.

ahojnnes:
Perfect, thanks a lot.

ahojnnes (approved):

Issue #752, opened 2019-12-04
Clarify license status


rkern:
Your `COPYING.txt` refers to various "optional dependencies" that may have different licenses. However, I do not see any optional dependencies, only required dependencies. Graclus and PBA are GPLed, and LSD is AGPLed. Given the copyleft nature of these licenses, Colmap, when built into a binary, is necessarily GPLed (with AGPL obligations for the LSD library itself). Your source code in `src/` is BSD-licensed, but the assembled binary is not.

The AGPL is quite an extreme version of copyleft and imposes a number of obligations that even the GPL does not. ~There are other BSD-licensed implementations of the LSD out there. [OpenCV has one](https://github.com/opencv/opencv/blob/master/modules/imgproc/src/lsd.cpp).~ (edit: looks like the actual implementation was removed, presumably because it copied from the same sources as yours. Alas.)

But even if you do not replace the LSD code, I do recommend correcting `COPYING.txt` to note that you have required GPL and AGPL dependencies, and that Colmap as a whole is licensed under the GPL even if the parts in `src/` are also licensed as BSD.

Thank you.

ahojnnes:
Thanks, the "optional" is indeed not reflecting the latest state of the system anymore. I pushed a commit to fix the formulation.

cuixinye:
@rkern @ahojnnes Could you please re-open the issue to update LSD part to permission license.
Refer to [opencv_contrib/issues/2524](https://github.com/opencv/opencv_contrib/issues/2524), it's resolved in opencv project. 
I would be great if the whole colmap work could remove AGPL code.
Other reference: https://github.com/rafael-grompone-von-gioi/binomial_nfa/tree/main

ahojnnes:
@cuixinye As of this PR, colmap can be built without LSD: https://github.com/colmap/colmap/pull/2578

PR #2578, opened 2024-05-27
LSD: making the AGPL dependency optional

zap150:
While Colmap itself is licensed with the permissive BSD license, the LSD dependency is AGPL. This basically means that any binary stemming from the codebase is AGPL as well, and thus cannot be used e.g. commercially. Other optional dependencies look okay (please correct me if I am wrong, e.g. CGAL can be turned off).

In this pull request I added the option to exclude LSD (similarly as can be done for CGAL). This affects the `model_orientation_aligner` module by removing the `MANHATTAN-WORLD` option.

The PR is related to issue https://github.com/colmap/colmap/issues/752.  A possible alternative could be to use OpenCV solution (see the issue above and https://github.com/opencv/opencv/blob/2cd330486ec4597eab49c1575fc4a6603f205a6a/modules/imgproc/src/lsd.cpp#L167). OpenCV excluded this part for while, the re-added LSD after its crucial part https://github.com/rafael-grompone-von-gioi/binomial_nfa had been relincesed. I have also contacted the author of LSD (https://github.com/theWorldCreator/LSD) but got no response.


ahojnnes:
Thank you very much for disentangling the LSD dependency through a compile time option. The code changes look good to me.

ahojnnes (approved):

ahojnnes:
@zap150 You would need to run code formatting (`scripts/format/clang_format.sh`) to appease the CI checks.

PR #2595, opened 2024-06-17
Synthesize full two-view geometry and raw matches

ahojnnes:


src/colmap/scene/synthetic.cc
```diff
@@ -73,13 +93,19 @@ void SynthesizeExhaustiveMatches(Reconstruction* reconstruction,
...
+
+      database->WriteMatches(image1.ImageId(), image2.ImageId(), matches);
       database->WriteTwoViewGeometry(
           image1.ImageId(), image2.ImageId(), two_view_geometry);
     }
   }
 }
 
-void SynthesizeChainedMatches(Reconstruction* reconstruction,
+void SynthesizeChainedMatches(const SyntheticDatasetOptions& options,
```
B1ueber2y, line 107: Only ``inlier_match_ratio`` is used in this function. Is it necessary to pass a whole ``SyntheticDatasetOptions`` here?

src/colmap/geometry/essential_matrix.cc
```diff
@@ -142,4 +142,10 @@ Eigen::Matrix3d InvertEssentialMatrix(const Eigen::Matrix3d& E) {
   return E.transpose();
 }
 
+Eigen::Matrix3d FundamentalFromEssentialMatrix(const Eigen::Matrix3d& K2,
```
B1ueber2y, line 145: This function is correct. However, this reminds me that COLMAP only does undistortion at ``CamFromImg`` and ``ImgFromCam``. In such case, should we add a ``IsUndistorted`` check when applying the ``LORANSAC<FundamentalMatrixSevenPointEstimator, FundamentalMatrixEightPointEstimator>`` in ``EstimateCalibratedTwoViewGeometry`` here (https://github.com/colmap/colmap/blob/main/src/colmap/estimators/two_view_geometry.cc#L458).

src/colmap/scene/synthetic.cc
ahojnnes: OK, done.

src/colmap/geometry/essential_matrix.cc
ahojnnes: Yes, we don't estimate any distortion coefficients during two-view estimation but only do so in the later multi-view reconstruction process. If there are reliable solvers to do so, we could incorporate them into the two-view estimation procedure. Otherwise, I would not enforce any specific checks at the two-view stage, otherwise we prevent people from specifying known distortion parameters when reading images. Does that make sense?

ahojnnes: OK, I see what you mean now. I think that makes sense, but I would tackle it in a separate PR. Is this something you could add? Thanks.

PR #2601, opened 2024-06-19
Fix cost functions for pose graph optimization

B1ueber2y:
Just found a long-standing bug from https://github.com/colmap/colmap/pull/2378 in both cost functions in pose graph optimization:

After LLT it should be ``L_matrix.transpose()`` rather than ``L_matrix`` that is left multiplied on the residuals.

Moreover, SVD is faster and more stable compared to ``covariance.inverse().llt()``. I have updated this as well.


B1ueber2y:
This only affects a covariance that is non-diagonal so might be overlooked for ages.

B1ueber2y:
OK an update to this. While the approach from SVD is more robust to ill-conditioned covariance and gives correct residual, the resulting submatrix is not necessarily lower-triangular, resulting in potentially different ordering of the residuals (which will not pass the unit test since we are testing each dimension of residuals there). 

I am reverting the SVD to inverse + llt for now and will leave it to the users to make sure that the covariance is well-conditioned.

ahojnnes (approved):

PR #2603, opened 2024-06-19
Faster homography estimator

ahojnnes:


src/colmap/estimators/homography_matrix_test.cc
```diff
@@ -38,38 +38,114 @@
...
+    HomographyMatrixEstimator estimator;
     std::vector<Eigen::Matrix3d> models;
-    est_tform.Estimate(src, dst, &models);
+    estimator.Estimate(src, dst, &models);
+    ASSERT_EQ(models.size(), 1);
+
+    std::vector<double> residuals;
+    estimator.Residuals(src, dst, models[0], &residuals);
 
+    for (size_t i = 0; i < 4; ++i) {
```
B1ueber2y, line 90: should be kNumPoints here?

src/colmap/estimators/homography_matrix_test.cc
```diff
@@ -38,38 +38,114 @@
...
+    HomographyMatrixEstimator estimator;
+    std::vector<Eigen::Matrix3d> models;
+    estimator.Estimate(src, dst, &models);
     ASSERT_EQ(models.size(), 1);
 
     std::vector<double> residuals;
-    est_tform.Residuals(src, dst, models[0], &residuals);
+    estimator.Residuals(src, dst, models[0], &residuals);
 
     for (size_t i = 0; i < 4; ++i) {
```
B1ueber2y, line 121: kNumPoints?

src/colmap/estimators/homography_matrix_test.cc
```diff
@@ -38,38 +38,114 @@
...
+    for (size_t i = 0; i < 4; ++i) {
+      EXPECT_LT(residuals[i], 1e-6);
+    }
+  }
+}
+
+// Test numerical stability with large coordinates. This is to ensure that the
+// homography matrix estimator is numerically stable despite not using
+// coordinate normalization. We can do this because of double precision.
+TEST(HomographyMatrix, NumericalStability) {
```
B1ueber2y, line 99: Tbh I am not sure about the validity of this test. To be able to test numerical stability should we add a small noise to ``(expected_H * x.homogeneous()).hnormalized()`` and relax the checking threshold? 

Also, I personally dont see a clear reason not to have coordinate normalization, since it is generally not the performance bottleneck, or is it?

B1ueber2y (approved):

src/colmap/estimators/homography_matrix_test.cc
ahojnnes: It's less about noise but the fact that the different columns in the constraint matrix will have largely different magnitudes of values, because of some columns set to constants, some of them to a single coordinate and some of them to the product of two coordinates. When the coordinates are large, they will get huge compared to others. I will add another test with adding some noise as you suggest.

ahojnnes: Added a bunch more tests and also parameterized them to run all of them on different number of points.

PR #2598, opened 2024-06-18
Support Adjoint matrix computation for Rigid3d

B1ueber2y:
Reference: 
- https://gtsam.org/2021/02/23/uncertainties-part3.html
- https://github.com/microsoft/lamar-benchmark/blob/fa6f5409364a11d5e59a57cec2ad1d7f9e6a8888/scantools/capture/pose.py#L99

The adjoint matrix is very useful on propagating uncertainty for Rigid3d. 

A unit test is added as well.

PR #2610, opened 2024-06-21
Covariance estimation for bundle adjustment with Schur elimination

B1ueber2y:
This PR contains the covariance estimation for bundle adjustment with handcrafted Schur elimination, which has been forever missing from the ceres solver. Also, the interface for world_from_cam to cam_from_world conversion for the 6x6 covariance is provided with the adjoint feature from https://github.com/colmap/colmap/pull/2598

This is a practically extremely useful feature since the ceres covariance computation will generally fail for large-scale bundle adjustment problem mainly due to poorly conditioned 3D points. Many of these points generally cannot be easily identified with heuristics such as triangulation angles, making the ceres covariance computation intractable. Here are some examples of this long standing issue in the ceres development thread (seems that they stop developing this feature):

https://groups.google.com/g/ceres-solver/c/719fVUZqOZk
https://groups.google.com/g/ceres-solver/c/1SrLa1Pr8D4

With this new feature we will be able to reliably extract covariance from the bundle adjustment problem as long as the camera poses are well constrained (i.e. without Gauge ambiguity), even with poorly conditioned points. The covariance computation will also be faster with the use of Schur elimination.

Here is a minimal testing example in Python (If we feel necessary I can also try to write a unit test in C++). To be able to run the test one only needs a bundle adjusted COLMAP SfM model without rank deficiency. I am using the SfM model produced by the pycolmap example from here: https://github.com/colmap/colmap/blob/main/pycolmap/example.py

```
import pycolmap
import pyceres
import numpy as np

# setup bundle adjustment
recon = pycolmap.Reconstruction("example/sfm/0")
ba_options = pycolmap.BundleAdjustmentOptions()
ba_config = pycolmap.BundleAdjustmentConfig()
for img_id in recon.images:
    ba_config.add_image(img_id)
bundle_adjuster = pycolmap.BundleAdjuster(ba_options, ba_config)
bundle_adjuster.set_up_problem(recon, pyceres.TrivialLoss())
problem = bundle_adjuster.problem 

# if needed we can add more constraints here (e.g. PGO, control points).

# constrain problem (must-do for visual-only SfM to remove gauge)
image1_id, image2_id = 1, 2
problem.set_parameter_block_constant(recon.images[image1_id].cam_from_world.rotation.quat)
problem.set_parameter_block_constant(recon.images[image1_id].cam_from_world.translation)
problem.set_manifold(recon.images[image2_id].cam_from_world.translation, pyceres.SubsetManifold(3, np.array([0])))

# solve
solver_options = pyceres.SolverOptions()
summary = pyceres.SolverSummary()
pyceres.solve(solver_options, problem, summary)

# [TEST] covariance: direct interface
covs = pycolmap.estimate_pose_covariance_from_ba(problem, recon)
covs_from_ceres = pycolmap.estimate_pose_covariance_from_ba_ceres_backend(bundle_adjuster.problem, recon)
for img_id, _ in covs.items():
    assert covs[img_id].shape[0] == 5 if img_id == image2_id else 6 
    np.testing.assert_allclose(covs[img_id], covs_from_ceres[img_id], atol=1e-5)

# [TEST] cross-image covariance: pybind classes
cov_estimator = pycolmap.BundleAdjustmentCovarianceEstimator(problem, recon)
cov_estimator.compute()
cov_estimator_ceres = pycolmap.BundleAdjustmentCovarianceEstimatorCeresBackend(problem, recon)
cov_estimator_ceres.compute()
for i in np.arange(2, 10): # in this example model image_ids are from 1 to 11.
    cov = cov_estimator.get_pose_covariance(i, i + 1)
    cov_from_ceres = cov_estimator_ceres.get_pose_covariance(i, i + 1)
    np.testing.assert_allclose(cov, cov_from_ceres, atol=1e-5)

# [TEST] multi-image covariance
cov = cov_estimator.get_pose_covariance([3, 5, 7, 8])
cov_from_ceres = cov_estimator.get_pose_covariance([3, 5, 7, 8])
assert cov.shape[0] == 24 # 6 * 4
np.testing.assert_allclose(cov, cov_from_ceres, atol=1e-5)

# [TEST] covariance on intrinsics
cov_estimator.compute_full()
cov_estimator_ceres.compute_full()
for camera_id in recon.cameras:
    params = recon.cameras[camera_id].params
    cov_params = cov_estimator.get_covariance(params)
    assert cov_params.shape[0] == 2 # (f, k1)
    cov_params_from_ceres = cov_estimator.get_covariance(params)
    np.testing.assert_allclose(cov_params, cov_params_from_ceres, atol=1e-5)
```


sarlinpe:
Very nice, thanks! I will review this in more details on Monday.

B1ueber2y:
Let me restructure this now to support cross-pose covariance and also covariance for other parameters.

B1ueber2y:
New feature arised from the new update:
1. cross-image covariance
2. covariance on other variables (except for points)
3. error handling for rank deficient jacobian.

The testing example is updated in the PR description to incorporate tests for the first two features in the new update.

src/colmap/estimators/covariance.cc
```diff
@@ -0,0 +1,477 @@
...
+
+bool BundleAdjustmentCovarianceEstimatorCeresBackend::Compute() {
+  ceres::Covariance::Options options;
+  ceres::Covariance covariance_computer(options);
+  std::vector<const double*> parameter_blocks;
+  for (const auto& block : pose_blocks_) {
+    parameter_blocks.push_back(block);
+  }
+  try {
+    covariance_computer.Compute(parameter_blocks, problem_);
```
ahojnnes, line 301: I thought ceres doesn't use exceptions? Should this instead do `if (!covariance_computer.Compute(...)) { return false };`?

src/colmap/estimators/covariance.cc
```diff
@@ -0,0 +1,477 @@
...
+    std::map<image_t, Eigen::MatrixXd>& res) {
+  BundleAdjustmentCovarianceEstimatorCeresBackend estimator(problem,
+                                                            reconstruction);
+  if (!estimator.Compute()) return false;
+  res.clear();
+  for (const auto& image : reconstruction->Images()) {
+    image_t image_id = image.first;
+    if (!estimator.HasPose(image_id)) continue;
+    Eigen::MatrixXd cov = estimator.GetPoseCovariance(image_id);
+    res.insert(std::make_pair(image_id, cov));
```
ahojnnes, line 455:
```suggestion
    res.emplace(image_id, estimator.GetPoseCovariance(image_id));
```
... to avoid unnecessary copy of matrix. Same below.

src/colmap/estimators/covariance.cc
```diff
@@ -0,0 +1,477 @@
...
+  for (const auto& block : blocks) {
+    THROW_CHECK(HasBlock(block));
+    int index = GetBlockIndex(block);
+    int num_params_pose = GetBlockTangentSize(block);
+    for (int i = 0; i < num_params_pose; ++i) {
+      indices.push_back(index + i);
+    }
+  }
+#if EIGEN_VERSION_AT_LEAST(3, 4, 0)
+  Eigen::MatrixXd output = cov_variables_(indices, indices);
```
ahojnnes, line 242: Is this faster than the manual loop below? If not, then it might not be worth the ifdef?

src/colmap/estimators/covariance.cc
```diff
@@ -0,0 +1,477 @@
...
+bool BundleAdjustmentCovarianceEstimatorCeresBackend::ComputeFull() {
+  ceres::Covariance::Options options;
+  ceres::Covariance covariance_computer(options);
+  std::vector<const double*> parameter_blocks;
+  for (const auto& block : pose_blocks_) {
+    parameter_blocks.push_back(block);
+  }
+  for (const auto& block : other_variables_blocks_) {
+    parameter_blocks.push_back(block);
+  }
```
ahojnnes, line 277:
```suggestion
  std::vector<const double*> parameter_blocks = pose_blocks_;
  parameter_blocks.insert(parameter_blocks.end(), other_variables_blocks_.begin(), other_variables_blocks_.end());
```

src/colmap/estimators/covariance.cc
```diff
@@ -0,0 +1,477 @@
...
+  int index = GetBlockIndex(params);
+  int num_params = GetBlockTangentSize(params);
+  return cov_variables_.block(index, index, num_params, num_params);
+}
+
+Eigen::MatrixXd BundleAdjustmentCovarianceEstimatorBase::GetCovariance(
+    const std::vector<double*>& blocks) const {
+  THROW_CHECK(HasValidFullCovariance());
+  std::vector<int> indices;
+  for (const auto& block : blocks) {
```
ahojnnes, line 233:
```suggestion
  for (const double* block : blocks) {
```

src/colmap/estimators/covariance.cc
```diff
@@ -0,0 +1,477 @@
...
+  Eigen::SparseMatrix<double> J_c =
+      J_full.block(0, 0, num_residuals, num_params - num_params_points_);
+  Eigen::SparseMatrix<double> J_p = J_full.block(
+      0, num_params - num_params_points_, num_residuals, num_params_points_);
+  Eigen::SparseMatrix<double> H_cc = J_c.transpose() * J_c;
+  Eigen::SparseMatrix<double> H_cp = J_c.transpose() * J_p;
+  Eigen::SparseMatrix<double> H_pc = H_cp.transpose();
+  Eigen::SparseMatrix<double> H_pp = J_p.transpose() * J_p;
+  // in-place computation of H_pp_inv
+  Eigen::SparseMatrix<double> H_pp_inv = H_pp;  // shallow copy
```
ahojnnes, line 354: Maybe I misunderstand how Eigen works but isn't this a full copy of the data?

src/colmap/estimators/covariance.cc
```diff
@@ -0,0 +1,477 @@
...
+  Eigen::MatrixXd S_poses = S_aa - S_ab * ldltOfS_bb.solve(S_ba);
+
+  // Step 5: Compute covariance
+  LOG(INFO) << StringPrintf(
+      "Inverting Schur complement for pose parameters (n = %d)",
+      num_params_poses_);
+  Eigen::FullPivLU<Eigen::MatrixXd> luOfS_poses(S_poses);
+  if (luOfS_poses.rank() < S_poses.rows()) {
+    LOG(INFO) << StringPrintf(
+        "Unable to compute covariance. The Schur complement on pose parameters "
```
ahojnnes, line 432: How well does this approach work when we also estimate intrinsics and some of the (shared) intrinsics variables are ill-conditioned?

src/colmap/estimators/covariance.h
```diff
@@ -0,0 +1,174 @@
...
+
+#include <Eigen/Core>
+#include <Eigen/Dense>
+#include <Eigen/Sparse>
+#include <ceres/ceres.h>
+
+namespace colmap {
+
+// Update covariance (n = 6) for rigid3d.inverse()
+Eigen::Matrix6d GetCovarianceForPoseInverse(const Eigen::Matrix6d& covar,
```
ahojnnes, line 43: This could move to the rigid3d file and get a small unit test?

src/colmap/estimators/covariance.h
```diff
@@ -0,0 +1,174 @@
...
+
+  // reconstruction
+  Reconstruction* reconstruction_;
+
+ private:
+  // set up parameter blocks
+  void Setup();
+};
+
+class BundleAdjustmentCovarianceEstimatorCeresBackend
```
ahojnnes, line 127: nit: could move to .cc file in an anonymous namespace. Same for the other implementation below.

ahojnnes: ... and actually also the base interface.

src/colmap/estimators/covariance.h
```diff
@@ -0,0 +1,174 @@
...
+  // Compute the Schur complement for poses and other variables by eliminating
+  // 3D points
+  void ComputeSchurComplement();
+  bool HasValidSchurComplement() const;
+};
+
+bool EstimatePoseCovarianceCeresBackend(
+    ceres::Problem* problem,
+    Reconstruction* reconstruction,
+    std::map<image_t, Eigen::MatrixXd>& res);
```
ahojnnes, line 167: nit
```suggestion
    std::map<image_t, Eigen::MatrixXd>& image_to_cov);
```

src/colmap/estimators/covariance.h
```diff
@@ -0,0 +1,174 @@
...
+
+bool EstimatePoseCovariance(ceres::Problem* problem,
+                            Reconstruction* reconstruction,
+                            std::map<image_t, Eigen::MatrixXd>& res,
```
ahojnnes, line 171: We use MatrixXd to deal with constant rotation/translation components. Would be good to briefly document the ordering of the parameter blocks in the resulting matrix in the header here?

ahojnnes:
Thank you very much. I think it could be great to have a minimal test that cross-checks (+- numerical accuracy) the computed covariance between the ceres implementation and the schur based implementation on a well-conditioned problem? If we want to be diligent, then another small test on a problem with ill-conditioned points could also be tested. The ceres backend would be expected to fail and the schur implementation succeed?

B1ueber2y:
> Thank you very much. I think it could be great to have a minimal test that cross-checks (+- numerical accuracy) the computed covariance between the ceres implementation and the schur based implementation on a well-conditioned problem? If we want to be diligent, then another small test on a problem with ill-conditioned points could also be tested. The ceres backend would be expected to fail and the schur implementation succeed?

Thanks! I will update according to the comments. I have tested both minimal tests you proposed in Python. The first one is attached in the code snippet of the PR description. Do we want a unit test in C++ as well? This may require handcrafted construction of the reconstruction (and particularly challenging for the ill-posed ones since it generally happens for large-scale problems).

src/colmap/estimators/covariance.cc
B1ueber2y: Yes. But generally the issue happens only for the points. The constraints on the intrinsics are generally pretty dense since it comes from all the residuals. Nonetheless, how about we adding damping factors to all the matrices (rather than only the point blocks) to avoid any potential failures? I can do this if agreed.

src/colmap/estimators/covariance.h
B1ueber2y: > We use MatrixXd to deal with constant rotation/translation components. 

Do you mean non-constant?

ahojnnes:
I think it can be good to have this test in C++ so it is executed as part of CI. Could you reuse the new synthetic tooling to create a large problem and then manually insert some I'll conditioned points?

src/colmap/estimators/covariance.h
ahojnnes: Yes, the matrix only contains the non constant parts, I assume?

src/colmap/estimators/covariance.cc
B1ueber2y: This can potentially be problematic for parameters with small scale (e.g., the distortion coefficient k1). Maybe we need a more aggressive damping factor there. But I would vote for keeping it like this and return False if the intrinsincs are rank deficient

src/colmap/estimators/covariance.h
B1ueber2y: Yes. I will add some documentations in the header.

src/colmap/estimators/covariance.cc
B1ueber2y: Yes you are right. Thanks. This is updated.

src/colmap/estimators/covariance.cc
B1ueber2y: Yes. I changed this into reference now.

src/colmap/estimators/covariance.cc
B1ueber2y: I added the damping factor to the blocks of other variables when we only need pose covariances here https://github.com/colmap/colmap/pull/2610/commits/b6691bf7e3c88150a055d957d5a38dbf23a42e83. This does not alter the results much numerically in my Python test.

src/colmap/estimators/covariance.h
B1ueber2y: Done here: https://github.com/colmap/colmap/pull/2610/commits/9473fe689e543724f50b040e6f37f74c1b458a5e

src/colmap/estimators/covariance.h
B1ueber2y: I dont get it here. What could be moved to .cc? This is a derived class?

B1ueber2y:
> I think it can be good to have this test in C++ so it is executed as part of CI. Could you reuse the new synthetic tooling to create a large problem and then manually insert some I'll conditioned points?

Will do tonight or tomorrow.

src/colmap/estimators/covariance.h
ahojnnes: I didn't see that the class is also bound to Python. Only looked at the C++ path. Then it has to be exposed in the header, of course.

src/colmap/estimators/covariance.h
```diff
@@ -0,0 +1,178 @@
...
+};
+
+class BundleAdjustmentCovarianceEstimator
+    : public BundleAdjustmentCovarianceEstimatorBase {
+ public:
+  BundleAdjustmentCovarianceEstimator(ceres::Problem* problem,
+                                      Reconstruction* reconstruction)
+      : BundleAdjustmentCovarianceEstimatorBase(problem, reconstruction) {}
+
+  void SetLambda(double lambda);  // update daming factor. default = 1e-6
```
ahojnnes, line 141: Instead of a separate method, we could also just expect it in the constructor and keep it const?

src/colmap/geometry/rigid3.h
```diff
@@ -76,6 +76,13 @@ inline Rigid3d Inverse(const Rigid3d& b_from_a) {
   return a_from_b;
 }
 
+// Update covariance (6x6) for rigid3d.inverse()
+inline Eigen::Matrix6d GetCovarianceForRigid3dInverse(
+    const Eigen::Matrix6d& covar, const Rigid3d& rigid3) {
+  Eigen::Matrix6d adjoint = rigid3.Adjoint();
+  return adjoint * covar * adjoint.transpose();
+};
```
ahojnnes, line 84:
```suggestion
inline Eigen::Matrix6d GetCovarianceForRigid3dInverse(
    const Eigen::Matrix6d& covar, const Rigid3d& rigid3) {
  const Eigen::Matrix6d adjoint = rigid3.Adjoint();
  return adjoint * covar * adjoint.transpose();
}
```

src/colmap/geometry/rigid3.h
```diff
@@ -76,6 +76,13 @@ inline Rigid3d Inverse(const Rigid3d& b_from_a) {
   return a_from_b;
 }
 
+// Update covariance (6x6) for rigid3d.inverse()
+inline Eigen::Matrix6d GetCovarianceForRigid3dInverse(
```
sarlinpe, line 80: Maybe make this a class method of Rigid3d?

src/pycolmap/geometry/bindings.cc
```diff
@@ -71,6 +71,11 @@ void BindGeometry(py::module& m) {
   py::implicitly_convertible<py::array, Eigen::Quaterniond>();
   MakeDataclass(PyRotation3d);
 
+  m.def("get_covariance_for_rigid3d_inverse",
```
sarlinpe, line 74: Maybe make this a class method of Rigid3d (like `Inverse`) and name it `covariance_of_inverse` or so?

src/pycolmap/estimators/covariance.cc
```diff
@@ -0,0 +1,92 @@
...
+      py::arg("problem"),
+      py::arg("reconstruction"),
+      py::arg("lambda") = 1e-6);
+
+#define REGISTER_ESTIMATOR(NAME)                                               \
+  .def(py::init<ceres::Problem*, Reconstruction*>())                           \
+      .def("compute", &NAME::Compute)                                          \
+      .def("compute_full", &NAME::ComputeFull)                                 \
+      .def("get_pose_covariance",                                              \
+           (Eigen::MatrixXd(NAME::*)() const) & NAME::GetPoseCovariance)       \
```
sarlinpe, line 46: Use `py::overload_cast` instead of C-style cast here and below: https://pybind11.readthedocs.io/en/stable/classes.html#overloaded-methods

src/pycolmap/estimators/covariance.cc
```diff
@@ -0,0 +1,92 @@
...
+        std::map<image_t, Eigen::MatrixXd> res;
+        if (!EstimatePoseCovariance(problem, reconstruction, res, lambda))
+          return py::none();
+        return py::cast(res);
+      },
+      py::arg("problem"),
+      py::arg("reconstruction"),
+      py::arg("lambda") = 1e-6);
+
+#define REGISTER_ESTIMATOR(NAME)                                               \
```
sarlinpe, line 41: Instead of a macro, can't you instead bind the base class, such that the derived ones do not need to bind inherited attributes twice? https://pybind11.readthedocs.io/en/stable/classes.html#inheritance-and-automatic-downcasting

B1ueber2y: I dont like the idea of binding the base class. It should not be exposed as it is virtual.

src/pycolmap/estimators/covariance.cc
```diff
@@ -0,0 +1,92 @@
...
+using namespace colmap;
+using namespace pybind11::literals;
+namespace py = pybind11;
+
+void BindCovarianceEstimator(py::module& m) {
+  m.def(
+      "estimate_pose_covariance_from_ba_ceres_backend",
+      [](ceres::Problem* problem,
+         Reconstruction* reconstruction) -> py::object {
+        std::map<image_t, Eigen::MatrixXd> res;
```
sarlinpe, line 19: Also rename to `image_to_cov`?

src/pycolmap/estimators/covariance.cc
sarlinpe: Why? If you don't bind a constructor for the base class then users won't be able to create any instance of it.

src/colmap/estimators/covariance.h
```diff
@@ -0,0 +1,178 @@
...
+// adjustment. The Schur complement is computed explicitly to eliminate the
+// hessian block for all the 3D points, which is essential to avoid Jacobian
+// rank deficiency for large-scale reconstruction
+class BundleAdjustmentCovarianceEstimatorBase {
+ public:
+  BundleAdjustmentCovarianceEstimatorBase(ceres::Problem* problem,
+                                          Reconstruction* reconstruction);
+  virtual ~BundleAdjustmentCovarianceEstimatorBase() = default;
+
+  // Compute covariance for all paramt\ers (except for 3D points)
```
sarlinpe, line 53:
```suggestion
  // Compute covariance for all parameters (except for 3D points)
```

src/pycolmap/estimators/covariance.cc
B1ueber2y: Ah OK then. Let me do it later today.

src/colmap/estimators/covariance.cc
```diff
@@ -0,0 +1,453 @@
...
+        "is rank deficient. Number of columns: %d, rank: %d. This is likely "
+        "due to the poses being underconstrained with Gauge ambiguity.",
+        S_poses.rows(),
+        luOfS_poses.rank());
+    return false;
+  }
+  cov_poses_ = S_poses.inverse();
```
sarlinpe, line 416: Can't we instead call `luOfS_poses.inverse()`? This would avoid computing the LU factorization twice.

src/colmap/estimators/covariance.cc
```diff
@@ -0,0 +1,453 @@
...
+  S_bb += lambda_ * Eigen::MatrixXd::Identity(S_bb.rows(), S_bb.cols());
+  Eigen::LDLT<Eigen::MatrixXd> ldltOfS_bb(S_bb);
+  Eigen::MatrixXd S_poses = S_aa - S_ab * ldltOfS_bb.solve(S_ba);
+
+  // Step 5: Compute covariance
+  LOG(INFO) << StringPrintf(
+      "Inverting Schur complement for pose parameters (n = %d)",
+      num_params_poses_);
+  Eigen::FullPivLU<Eigen::MatrixXd> luOfS_poses(S_poses);
+  if (luOfS_poses.rank() < S_poses.rows()) {
```
sarlinpe, line 407: Instead check `luOfS_poses.isInvertible()`?

src/colmap/estimators/covariance.cc
```diff
@@ -0,0 +1,453 @@
...
+  if (luOfS_dense.rank() < S_dense.rows()) {
+    LOG(INFO) << StringPrintf(
+        "Unable to compute covariance. The Schur complement on all variables "
+        "(except for 3D points) is rank deficient. Number of columns: %d, "
+        "rank: %d.",
+        S_dense.rows(),
+        luOfS_dense.rank());
+    return false;
+  }
+  cov_variables_ = S_dense.inverse();
```
sarlinpe, line 375: Can we share this block of code between `Compute` and `ComputeFull`?

B1ueber2y: The error logs are slightly different. There is no other cases for using this inverse so I would vote for keeping it as in the current form.

sarlinpe, line 375: IIUC this computes a dense inverse with LU factorization. How well does this scale? Did you try to run this on large reconstruction or long sequences? We might need to explore more efficient options. For example, computing a sparse Cholesky factorization of S that can be easily inverted ([Ceres provides this](https://github.com/ceres-solver/ceres-solver/blob/6a74af202d83cf31811ea17dc66c74d03b89d79e/internal/ceres/covariance_impl.cc#L848-L882)).

sarlinpe: You could handle the logging of the error in the respective parent functions?

src/colmap/estimators/covariance.cc
```diff
@@ -0,0 +1,453 @@
...
+  // in-place computation of H_pp_inv
+  Eigen::SparseMatrix<double>& H_pp_inv = H_pp;  // reference
+  int counter_p = 0;
+  for (const double* block : point_blocks_) {
+    int num_params_point = ParameterBlockTangentSize(problem_, block);
+    Eigen::SparseMatrix<double> subMatrix_sparse =
+        H_pp.block(counter_p, counter_p, num_params_point, num_params_point);
+    Eigen::MatrixXd subMatrix = subMatrix_sparse;
+    subMatrix +=
+        lambda_ * Eigen::MatrixXd::Identity(subMatrix.rows(), subMatrix.cols());
```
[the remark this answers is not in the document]
B1ueber2y: Using pseudo inverse here is potentially incorrect for covariance computation. It is setting the reciprocal of all the near-zero values to be zero, which should be large values instead. This can potentially result in underestimation of the covariance. Instead, damping the matrix is generally used for LM optimization and is considered a general practice to handle the singularity.

src/colmap/estimators/covariance.cc
B1ueber2y: Yes I did. S is already a matrix with all the points eliminated, so the matrix is generally not sparse enough, so there is not much to gain for doing sparse inverse here. That being said, we can indeed change it to S_dense.ldlt().solve(Eigen::MatrixXdIdentity(*, *)) which could exploit the symmetry and should be faster.

src/colmap/estimators/covariance.cc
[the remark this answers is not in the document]
B1ueber2y: OK maybe I am the one who is confused here. This might be valid. Then I have a cleaner proposal: how about we just dropping all the damping factors and apply MP inverse for both here and the 3x3 inversion at the point blocks? : )

sarlinpe:
This is really great work, thanks a lot! 

One comment: COLMAP currently resolves the Gauge ambiguity by fixing q_0, t_0, and t_1_x. This affects the resulting covariances, making them relative to the first image, which is arbitrary. To obtain more representative covariances, we could consider instead fixing the Gauge with a penalty on the centroid of the cameras centers, their standard standard deviation, and the average camera orientation - similarly to `Reconstruction::Normalize`.

src/colmap/estimators/covariance.cc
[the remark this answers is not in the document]
sarlinpe: You are right: in the direction that is unconstrained, the pseudo-inverse would yield a zero variance while the damping simply caps it to a maximum value. Then how about replacing the damping by an SVD-based inverse, where we clip the reciprocal of the singular values to a maximum threshold? (~1/damping)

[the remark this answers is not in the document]
sarlinpe: Regarding the damping of the other parameters: how often is this needed? aren't they most often sufficiently constrained?

src/colmap/estimators/covariance.cc
sarlinpe: I thought that the difference between DENSE_SCHUR and SPARSE_SCHUR in Ceres is precisely that S is inverted densely vs sparsely and that the latter is much more appropriate for large problems.

src/colmap/estimators/covariance.cc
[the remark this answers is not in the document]
ahojnnes: Depending on the camera model and scene geometry, the estimation of intrinsics can often be poorly conditioned even for many residuals in the images.

src/colmap/estimators/covariance.h
B1ueber2y: Sounds good. This has been done.

src/colmap/estimators/covariance.cc
B1ueber2y: I was following here: https://github.com/ceres-solver/ceres-solver/blob/6a74af202d83cf31811ea17dc66c74d03b89d79e/internal/ceres/covariance_impl.cc#L833. I guess it does not matter?

src/colmap/estimators/covariance.cc
B1ueber2y: Makes great sense. This is updated. Thanks

src/colmap/estimators/covariance.cc
B1ueber2y: We print the rank here in the logs, and this will require either exposing some intermediate variables or making the rank output in the inverse function. I feel that this abstraction is a little overcomplication. 

Also, here https://github.com/colmap/colmap/pull/2610/commits/9b3ea5f6618d77fdaf47ae57f54de86d13189652 I have updated the matrix inverse for ``ComputeFull`` to use sparse cholesky while use dense LU for handling only pose blocks at ``Compute``. So there is no point to abstract this part anymore.

B1ueber2y: Here https://github.com/colmap/colmap/pull/2610/commits/9b3ea5f6618d77fdaf47ae57f54de86d13189652 I have added sparse support for other variables both at ``ComputeFull`` and at second round of Schur implementation during ``Compute``.

src/colmap/estimators/covariance.cc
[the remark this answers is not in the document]
B1ueber2y: Although I agree that using SVD to clamp the singular values may be a slightly better solution, it is hard to make efficient. Now we are using Cholesky which is much efficient than SVD. I have changed the default value of lambda to 1e-8 and believe that this should not be a problem numerically.

src/colmap/estimators/covariance.cc
sarlinpe: Nice, thanks!

src/colmap/estimators/covariance.cc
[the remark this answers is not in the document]
sarlinpe: For the 3x3 point blocks, you are currently using LU factorization. For 3x3 matrices, I doubt that there is much difference in efficiency between this and SVD. This can also be trivially parallelized over all points with a thread pool. But eventually I'm unsure whether damping vs s-value clamping really matters.

B1ueber2y:
> > I think it can be good to have this test in C++ so it is executed as part of CI. Could you reuse the new synthetic tooling to create a large problem and then manually insert some I'll conditioned points?
> 
> Will do tonight or tomorrow.

@ahojnnes FYI I have uploaded a unit test as requested. Along the way I also fixed a lifetime issue here https://github.com/colmap/colmap/pull/2610/commits/3af52f3cb1543c7e992e30824696d40ed3449c8d for the original ``Solve`` interface of ``BundleAdjuster``, where the ownership of the loss function is lost after solving.

B1ueber2y:
> One comment: COLMAP currently resolves the Gauge ambiguity by fixing q_0, t_0, and t_1_x. This affects the resulting covariances, making them relative to the first image, which is arbitrary. To obtain more representative covariances, we could consider instead fixing the Gauge with a penalty on the centroid of the cameras centers, their standard standard deviation, and the average camera orientation - similarly to `Reconstruction::Normalize` and as described in [_Gauge Independence in Optimization Algorithms for 3D Vision_](https://citeseerx.ist.psu.edu/document?repid=rep1&type=pdf&doi=249b0a02995f7f4c9244655b076a1e43ef3bd806).

Thanks. This actually looks pretty straightforward. And If I am understanding correctly, it should be mathematically correct to apply the transformation directly on the schur complement of the pose blocks with $V_2$ from Eq. (10). One potential bottleneck for this is that it could be difficult to design unit tests to validate.

However, I guess for now the most practical and interesting use case is to use the additional sensors or constraints to make the poses well constrained, and recover metric uncertainty from the problem. So lets leave this Gauge feature for later when I have more time.

B1ueber2y:
To add on the unit test, in the ``RankDeficientPoints`` test the ceres implementation fails 9 out of 10 times while ours always succeeds. Nonetheless I feel that it does not make sense to assert the failure of ceres implementation in the unit test so I removed the assertion of ceres failure here https://github.com/colmap/colmap/pull/2610/commits/1d387e2ee8e926eae60e1f479e7e3d8f7583678f (otherwise the 1 out of 10 sometimes give us error. It really depends on the randomized reconstruction).

B1ueber2y:
Just added a new feature for customized pose blocks support https://github.com/colmap/colmap/pull/2610/commits/9d7fe4cfcec250e5c1ce481fd61c36ab6135a62e, for users who want to specify pose blocks that are interested themselves. This is also quite useful for extending the covariance estimation to other variants such as rig bundle adjustment where the variables are with more complex structures than the simple COLMAP reocnstruction. Example for this new feature is attached here:

```
import pycolmap
import pyceres
import numpy as np

# setup bundle adjustment
recon = pycolmap.Reconstruction("example/sfm/0")
ba_options = pycolmap.BundleAdjustmentOptions()
ba_config = pycolmap.BundleAdjustmentConfig()
for img_id in recon.images:
    ba_config.add_image(img_id)
bundle_adjuster = pycolmap.BundleAdjuster(ba_options, ba_config)
bundle_adjuster.set_up_problem(recon, pyceres.TrivialLoss())
problem = bundle_adjuster.problem

# constrain problem
image1_id, image2_id = 1, 2
problem.set_parameter_block_constant(recon.images[image1_id].cam_from_world.rotation.quat)
problem.set_parameter_block_constant(recon.images[image1_id].cam_from_world.translation)
problem.set_manifold(recon.images[image2_id].cam_from_world.translation, pyceres.SubsetManifold(3, np.array([0])))

# solve
solver_options = pyceres.SolverOptions()
summary = pyceres.SolverSummary()
pyceres.solve(solver_options, problem, summary)

# [TEST]
pose_blocks = []
for img_id in [2, 3, 4, 5]:
    qvec = recon.images[img_id].cam_from_world.rotation.quat
    if not problem.is_parameter_block_constant(qvec):
        pose_blocks.append(qvec);
    tvec = recon.images[img_id].cam_from_world.translation
    if not problem.is_parameter_block_constant(tvec):
        pose_blocks.append(tvec)
cov_estimator = pycolmap.BundleAdjustmentCovarianceEstimator(problem, recon)
cov_estimator.set_pose_blocks(pose_blocks)
cov_estimator.compute()
cov_estimator_ceres = pycolmap.BundleAdjustmentCovarianceEstimatorCeresBackend(problem, recon)
cov_estimator_ceres.set_pose_blocks(pose_blocks)
cov_estimator_ceres.compute()

# full covariance
cov = cov_estimator.get_pose_covariance()
cov_from_ceres = cov_estimator_ceres.get_pose_covariance()
np.testing.assert_allclose(cov, cov_from_ceres, atol=1e-5)
# covariance by block indexing
cov = cov_estimator.get_pose_covariance(recon.images[2].cam_from_world.rotation.quat, recon.images[5].cam_from_world.translation)
cov_from_ceres = cov_estimator_ceres.get_pose_covariance(recon.images[2].cam_from_world.rotation.quat, recon.images[5].cam_from_world.translation)
np.testing.assert_allclose(cov, cov_from_ceres, atol=1e-5)
```

src/colmap/estimators/covariance.cc
```diff
@@ -0,0 +1,588 @@
...
+    int num_params_point = ParameterBlockTangentSize(problem_, block);
+    point_blocks_.push_back(block);
+    num_params_points_ += num_params_point;
+  }
+
+  // Parse parameter blocks for poses
+  SetPoseBlocks(pose_blocks);
+
+  // Parse parameter blocks for other variables
+  SetUpOtherVariablesBlocks();
```
sarlinpe, line 121: This is also called in `SetPoseBlocks`

src/pycolmap/estimators/covariance.cc
```diff
@@ -42,8 +42,23 @@ void BindCovarianceEstimator(py::module& m) {
...
   py::class_<EstimatorBase>(m, "BundleAdjustmentCovarianceEstimatorBase")
-      .def("has_pose", &EstimatorBase::HasPose, py::arg("image_id"))
+      .def(
+          "set_pose_blocks",
+          [](EstimatorBase& self, std::vector<py::array_t<double>>& pyarrays) {
+            std::vector<const double*> blocks;
+            for (auto it = pyarrays.begin(); it != pyarrays.end(); ++it) {
+              py::buffer_info info = it->request();
+              blocks.push_back((const double*)info.ptr);
+            }
```
sarlinpe, line 52: You could create a helper function to convert a `py::array_y<double>` to `std::vector<const double*>` as it is also used bellow.

src/colmap/estimators/covariance.cc
```diff
@@ -0,0 +1,588 @@
...
+
+namespace colmap {
+
+BundleAdjustmentCovarianceEstimatorBase::
+    BundleAdjustmentCovarianceEstimatorBase(ceres::Problem* problem,
+                                            Reconstruction* reconstruction) {
+  THROW_CHECK_NOTNULL(problem);
+  problem_ = problem;
+  THROW_CHECK_NOTNULL(reconstruction);
+  reconstruction_ = reconstruction;
```
sarlinpe, line 44: THROW_CHECK_NOTNULL is pass-through, so you can use an initializer list here: `_problem(THROW_CHECK_NOTNULL(problem))`

src/pycolmap/estimators/covariance.cc
```diff
@@ -0,0 +1,198 @@
...
+
+using namespace colmap;
+using namespace pybind11::literals;
+namespace py = pybind11;
+
+namespace {
+
+void ConvertListOfPyArraysToConstPointers(
+    const std::vector<py::array_t<double>>& pyarrays,
+    std::vector<const double*>& blocks) {
```
sarlinpe, line 18: You could create `blocks` in this function instead of forcing the parent scope to do so.

src/pycolmap/estimators/covariance.cc
```diff
@@ -0,0 +1,198 @@
...
+  blocks.clear();
```
sarlinpe, line 19: Pre-allocate `blocks` with `resize`?

src/pycolmap/estimators/covariance.cc
B1ueber2y: How do you achieve that without copying the blocks?

sarlinpe: Return value optimization - there shouldn't be any copy if default optimizations are enabled.

src/pycolmap/estimators/covariance.cc
B1ueber2y: I have updated it with ``reserve``.

sarlinpe (approved):

src/colmap/estimators/covariance.cc
```diff
@@ -0,0 +1,582 @@
...
+BundleAdjustmentCovarianceEstimatorBase::
+    BundleAdjustmentCovarianceEstimatorBase(
+        ceres::Problem* problem,
+        const std::vector<const double*>& pose_blocks,
+        const std::vector<const double*>& point_blocks)
+    : problem_(THROW_CHECK_NOTNULL(problem)) {
+  SetUp(pose_blocks, point_blocks);
+}
+
+void BundleAdjustmentCovarianceEstimatorBase::SetUp(
```
ahojnnes, line 55: nit: is there a reason to not directly do this in the constructor instead of a separate method?

src/colmap/estimators/covariance.cc
```diff
@@ -0,0 +1,582 @@
...
+
+bool BundleAdjustmentCovarianceEstimatorBase::HasBlock(
+    const double* params) const {
+  return map_block_to_index_.find(params) != map_block_to_index_.end();
+}
+
+bool BundleAdjustmentCovarianceEstimatorBase::HasPoseBlock(
+    const double* params) const {
+  if (!HasBlock(params)) return false;
+  return map_block_to_index_.at(params) < num_params_poses_;
```
ahojnnes, line 182: Could avoid the redundant lookup and just do:
```
const auto it = map_block_to_index_.find(params);
if (it == map_block_to_index_.end()) {
  return false;
}
return it->second;
```

src/colmap/estimators/covariance.cc
```diff
@@ -0,0 +1,582 @@
...
+  Eigen::Matrix<double, Eigen::Dynamic, Eigen::Dynamic, Eigen::RowMajor> covs(
+      num_params, num_params);
+  covariance_computer.GetCovarianceMatrixInTangentSpace(parameter_blocks,
+                                                        covs.data());
+  cov_variables_ = covs;
+  cov_poses_ = cov_variables_.block(0, 0, num_params_poses_, num_params_poses_);
+  return true;
+}
+
+bool BundleAdjustmentCovarianceEstimatorCeresBackend::Compute() {
```
ahojnnes, line 392: Would it make sense to call the two methods "ComputePoseCov" and "ComputePoseAndPointCov" ?

src/colmap/estimators/covariance.cc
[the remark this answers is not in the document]
ahojnnes: I don't have any intuition on whether damping or SVD is better. I'll leave the decision to the two of you.

src/colmap/estimators/covariance.cc
```diff
@@ -0,0 +1,582 @@
...
+  Eigen::SparseMatrix<double> H_pp = J_p.transpose() * J_p;
+  // in-place computation of H_pp_inv
+  Eigen::SparseMatrix<double>& H_pp_inv = H_pp;  // reference
+  int counter_p = 0;
+  for (const double* block : point_blocks_) {
+    int num_params_point = ParameterBlockTangentSize(problem_, block);
+    Eigen::SparseMatrix<double> subMatrix_sparse =
+        H_pp.block(counter_p, counter_p, num_params_point, num_params_point);
+    Eigen::MatrixXd subMatrix = subMatrix_sparse;
+    subMatrix +=
```
ahojnnes, line 451:
```suggestion
    sub_matrix +=
```
nit

src/colmap/estimators/covariance.cc
```diff
@@ -0,0 +1,582 @@
...
+  Eigen::SparseMatrix<double> H_cc = J_c.transpose() * J_c;
+  Eigen::SparseMatrix<double> H_cp = J_c.transpose() * J_p;
+  Eigen::SparseMatrix<double> H_pc = H_cp.transpose();
+  Eigen::SparseMatrix<double> H_pp = J_p.transpose() * J_p;
+  // in-place computation of H_pp_inv
+  Eigen::SparseMatrix<double>& H_pp_inv = H_pp;  // reference
+  int counter_p = 0;
+  for (const double* block : point_blocks_) {
+    int num_params_point = ParameterBlockTangentSize(problem_, block);
+    Eigen::SparseMatrix<double> subMatrix_sparse =
```
ahojnnes, line 448:
```suggestion
    Eigen::SparseMatrix<double> sub_matrix_sparse =
```

src/colmap/estimators/covariance.cc
B1ueber2y: The first method is to indeed compute pose covariance, but the second is to compute the covariance of all variables except for the points, as described in the header, and I felt that ComputePoseAndOtherVariablesCov is too long here.

src/colmap/estimators/covariance_test.cc
```diff
@@ -0,0 +1,247 @@
...
+TEST(Covariance, PoseCovarianceInterface) {
+  Reconstruction reconstruction;
+  GenerateReconstruction(&reconstruction);
+  std::shared_ptr<BundleAdjuster> bundle_adjuster =
+      BuildBundleAdjuster(&reconstruction);
+  bundle_adjuster->Solve(&reconstruction);
+  std::shared_ptr<ceres::Problem> problem = bundle_adjuster->Problem();
+
+  std::map<image_t, Eigen::MatrixXd> image_id_to_covar_ceres;
+  ASSERT_TRUE(EstimatePoseCovarianceCeresBackend(
```
ahojnnes, line 88:
```suggestion
  ASSERT_TRUE(EstimatePoseCovariance(
```

B1ueber2y: Thanks. This has been fixed.

src/colmap/estimators/covariance.cc
B1ueber2y: Thanks. This is fixed now.

src/colmap/estimators/covariance.cc
B1ueber2y: No reasons anymore. I have updated this.

ahojnnes:
Thanks, great work.

cdcseacave:
@B1ueber2y thank you for this contribution. I am trying to port most of your math into my custom SfM pipeline for estimating cov for poses, and it works great for normal scenes; however for some scenes it either fails or the cov is inf, and I am not sure why. These seems to happen in scenes where there are some poses not well constrained, and I can figure out the exact cause: weak connections (constraints) for these poses, or something else. The failure is in "Simplicial LDLT for computing L_inv failed" with "NumericalIssue if the matrix.appears to be negative."
Any idea how to solve this? I do not know the math for this part well enough.

B1ueber2y:
> @B1ueber2y thank you for this contribution. I am trying to port most of your math into my custom SfM pipeline for estimating cov for poses, and it works great for normal scenes; however for some scenes it either fails or the cov is inf, and I am not sure why. These seems to happen in scenes where there are some poses not well constrained, and I can figure out the exact cause: weak connections (constraints) for these poses, or something else. The failure is in "Simplicial LDLT for computing L_inv failed" with "NumericalIssue if the matrix.appears to be negative." Any idea how to solve this? I do not know the math for this part well enough.

Hey. For those poses that are not well constrained, it is possible that there is rank deficiency (or large condition number) when inverting the Hessian, resulting in failures. This actually makes sense since the covariance on certain direction would be infinite, and is not useful to inspect for downstream applications. You would need to either remove the ambiguity by adding additional constraints, or resort to more advanced techniques to deal with the nullspace (e.g., https://dspace.cvut.cz/bitstream/handle/10467/108057/F3-D-2023-Polic-Michal-Polic%20-%20disertacni%20prace.pdf). Thanks!

cdcseacave:
@B1ueber2y Makes sens, thank you. Is it possible to detect only which poses are not well constrained? That is itself would be a valuable info to have for the SfM

B1ueber2y:
> @B1ueber2y Makes sens, thank you. Is it possible to detect only which poses are not well constrained? That in itself would be a valuable info to have for the SfM

The rank deficiency may not be specific to one pose, but can be a global behavior. For example, if the Gauge is not constrained, all the poses can be ambiguous up to a similarity transformation, but the relative poses are well constrained.

cdcseacave:
Gauge is constrained on all degrees of freedom (first camera pose is constant), except explicit lock on scale (though it is constrained too indirectly by the relative pose loss)

PR #2612, opened 2024-06-23
Faster 7-point fundamental matrix estimator

ahojnnes:
The new implementation is ~4x faster and a little more often produces a valid model.

### DIFF
--- a/src/colmap/estimators/fundamental_matrix.cc
+++ b/src/colmap/estimators/fundamental_matrix.cc
@@ -54,36 +54,22 @@ void FundamentalMatrixSevenPointEstimator::Estimate(
 
   models->clear();
 
-  // Note that no normalization of the points is necessary here.
-
   // Setup system of equations: [points2(i,:), 1]' * F * [points1(i,:), 1]'.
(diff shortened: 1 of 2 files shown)


src/colmap/estimators/fundamental_matrix.cc
```diff
@@ -110,35 +96,17 @@ void FundamentalMatrixSevenPointEstimator::Estimate(
...
-
-    Eigen::MatrixXd F = lambda * f1 + mu * f2;
-
-    F.resize(3, 3);
-
-    const double kEps = 1e-10;
-    if (std::abs(F(2, 2)) < kEps) {
-      continue;
-    }
+  const double inv_c0 = 1.0 / coeffs(0);
```
B1ueber2y, line 99: This can potentially result in inf? In the polynomial solver before there was a ``RemoveLeadingZero`` operation.

ahojnnes: Should be addressed now. Thanks.

B1ueber2y (approved):

PR #2618, opened 2024-06-27
Faster essential matrix estimators

ahojnnes:
* ~20% faster 5-point solver.
* ~5x faster 8-point solver.

### DIFF
--- a/src/colmap/estimators/essential_matrix.cc
+++ b/src/colmap/estimators/essential_matrix.cc
@@ -48,39 +48,36 @@ void EssentialMatrixFivePointEstimator::Estimate(
     const std::vector<Y_t>& points2,
     std::vector<M_t>* models) {
   THROW_CHECK_EQ(points1.size(), points2.size());
+  THROW_CHECK_GE(points1.size(), 5);
   THROW_CHECK(models != nullptr);
 
   models->clear();
 
-  // Step 1: Extraction of the nullspace x, y, z, w.
+  // Setup system of equations: [points2(i,:), 1]' * E * [points1(i,:), 1]'.
 
   Eigen::Matrix<double, Eigen::Dynamic, 9> Q(points1.size(), 9);
   for (size_t i = 0; i < points1.size(); ++i) {
-    const double x1_0 = points1[i](0);
-    const double x1_1 = points1[i](1);
-    const double x2_0 = points2[i](0);
-    const double x2_1 = points2[i](1);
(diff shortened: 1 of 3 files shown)


src/colmap/estimators/essential_matrix.cc
```diff
@@ -176,38 +174,39 @@ void EssentialMatrixEightPointEstimator::Estimate(
...
+  Eigen::Vector3d singular_values = svd.singularValues();
   singular_values(2) = 0.0;
-  const Eigen::Matrix3d E = E_raw_svd.matrixU() * singular_values.asDiagonal() *
-                            E_raw_svd.matrixV().transpose();
+  const Eigen::Matrix3d E =
+      svd.matrixU() * singular_values.asDiagonal() * svd.matrixV().transpose();
 
   models->resize(1);
-  (*models)[0] = E;
+  (*models)[0] = normed_from_orig2.transpose() * E * normed_from_orig1;
```
ahojnnes, line 209: Not sure how this ended up wrongly implemented, but the previous implementation was just wrong.

B1ueber2y: Should we change the matches generation for the unit tests to 3D and compute essential matrix from the relative pose. In this way we can avoid the previous issue of wrong implementation. Now we can only compare with the reference : )

ahojnnes: I probably misunderstand your proposal, but there is a new "EssentialMatrixEightPointEstimatorTests" test that does that.

B1ueber2y: Thanks! I did mean exactly this

src/colmap/estimators/essential_matrix.cc
```diff
@@ -176,38 +174,39 @@ void EssentialMatrixEightPointEstimator::Estimate(
...
+        QQ.col(8).data());
+  } else {
+    Eigen::JacobiSVD<Eigen::Matrix<double, Eigen::Dynamic, 9>> svd(
+        A, Eigen::ComputeFullV);
+    Q = Eigen::Map<const Eigen::Matrix<double, 3, 3, Eigen::RowMajor>>(
+        svd.matrixV().col(8).data());
+  }
 
-  // Enforcing the internal constraint that two singular values must be equal
+  // Enforcing the internal constraint that two singular values must non-zero
```
B1ueber2y, line 199: "be non-zero"? typo

B1ueber2y (approved):

PR #2620, opened 2024-06-28
Remove CamFromWorldPrior and create LocationPrior

sarlinpe:
- Create a new `LocationPrior` object that replaces `Image::CamFromWorldPrior`
  - For now it does not store any rotation, since this is not used anywhere in the code
  - It stores a coordinate system so we can distinguish between WGS84, a Cartesian system like ECEF/ENU/LV95, or any datum that we might want to support in the future
  - It can later be extended to store an accuracy estimate, a timestamp (if different from the corresponding image), etc.
- Since the pose prior is rarely used (only for model alignment or spatial matching), it should not break anyone's workflow - we provide a migration script if this does.


scripts/python/database.py
```diff
@@ -58,19 +58,18 @@
     image_id INTEGER PRIMARY KEY AUTOINCREMENT NOT NULL,
     name TEXT NOT NULL UNIQUE,
     camera_id INTEGER NOT NULL,
-    prior_qw REAL,
```
ahojnnes, line 61: What happens if we open an old database? Does it crash?

ahojnnes: OK, I guess the migration script below answers the question. It will crash but can be migrated with the script?

scripts/python/database.py
```diff
@@ -58,19 +58,18 @@
...
-    prior_tx REAL,
-    prior_ty REAL,
-    prior_tz REAL,
     CONSTRAINT image_id_check CHECK(image_id >= 0 and image_id < {}),
     FOREIGN KEY(camera_id) REFERENCES cameras(camera_id))
 """.format(
     MAX_IMAGE_ID
 )
 
+CREATE_LOCATION_PRIORS_TABLE = """CREATE TABLE IF NOT EXISTS location_priors (
```
ahojnnes, line 67: nit: Shall we call it "PosePrior", so it becomes easier to extend it to also hold a orientation prior in the future? For now, we keep it simple and only store the position.

scripts/python/database.py
sarlinpe: It does not crash, it just ignores the columns. Which is why the `--cleanup` option in the migration script is optional.

ahojnnes: Thanks

scripts/python/database.py
sarlinpe: Ok, do you suggest replacing all instances of LocationPrior/location_prior by PosePrior/pose_prior?

sarlinpe: Done

ahojnnes (approved):

daszhugh:
The prior pose of the image can be used for rigid body transformation, or bundle adjustment with gnss. So it is best to keep the prior pose in the image in the form of Cartesian coordinates.

sephyli:
@sarlinpe Should we also remove the code [here](https://github.com/colmap/colmap/blob/main/scripts/python/database.py#L200)?

sarlinpe:
@sephyli Good catch, thank you!

PR #2621, opened 2024-06-28
Add cost function for 3D alignment (with covariance)

B1ueber2y:
Cost function for 3D alignment with similarity transform, with covariance on the target 3D point/location prior. 

Also with the following side updates:

1. Split ceres manifolds to a stand-alone header
2. update the scale binding for Rigid3d to support numpy array, such that we can fix (i.e. set constant) or parameterize it with pyceres
3. Add a new manifold with exponential mapping to ensure the value to be positive, motivated by the common practice for paraemeterizing the scale parameters. This is installed to ``pycolmap`` when ``pyceres`` is detected in the Python environment.


src/colmap/estimators/cost_functions.h
```diff
@@ -453,6 +453,47 @@ struct MetricRelativePoseErrorCostFunction {
...
+        new ceres::
+            AutoDiffCostFunction<Point3dAlignmentCostFunction, 3, 3, 4, 3, 1>(
+                new Point3dAlignmentCostFunction(ref_point, covariance_point)));
+  }
+
+  template <typename T>
+  bool operator()(const T* const point,
+                  const T* const transform_q,
+                  const T* const transform_t,
+                  const T* const log_scale,
```
sarlinpe, line 478: What would it take to be able to directly pass fields of `Sim3d` object? For example:
```python
tfm = pycolmap.Sim3d()
prob.add_residual_block(pycolmap.Point3dAlignmentCost(...), loss, [pt, tfm.rotation, tfm.translation, tfm.scale])
```
1) A custom manifold parameterization that enforces a strictly positive scale?
2) `Sim3d.scale` be a numpy array instead of a double?

B1ueber2y: While this has just been done in the new commit, there is no easy way to bind it to Python since we dont have access to the base binding of PyManifold here: https://github.com/cvg/pyceres/blob/main/_pyceres/core/manifold.h#L62

B1ueber2y: > 2. `Sim3d.scale` be a numpy array instead of a double?

This is done by adding a new interface ``scale_ptr``. Now we can do
```
prob.add_residual_block(cost, loss, [pt, tfm.rotation, tfm.translation, tfm.scale_ptr])
```
The only thing that needs to be resolved is the binding of manifold. The easy way is to add that manifold to pyceres.

src/colmap/estimators/manifold.h
```diff
@@ -0,0 +1,156 @@
...
+class PositiveExponentialManifold : public ceres::Manifold {
+ public:
+  PositiveExponentialManifold() {}
+  ~PositiveExponentialManifold() {}
+
+  bool Plus(const double* x,
+            const double* delta,
+            double* x_plus_delta) const override {
+    for (size_t i = 0; i < size; ++i) {
+      x_plus_delta[i] = std::exp(std::log(x[i]) + delta[i]);
```
sarlinpe, line 90: Isn't this equivalent to `x[i] * std::exp(delta[i])` ?

B1ueber2y: Yes it is. Let me update.

src/colmap/estimators/cost_functions.h
sarlinpe: Try this out
```c++
      .def_property(
          "scale",
          [](Sim3d& self) { return py::array({}, {}, &self.scale, py::cast(self)); },
          [](Sim3d& self, double scale) { self.scale = scale; })
```

sarlinpe: Regarding the binding of the manifold object: all bindings in pyceres (SubsetManifold, etc.) only reference the base `ceres::Manifold`, not PyManifold. Shouldn't this work here too if pyceres has been imported? does it then fail if pyceres hasn't been imported?

sarlinpe: We could provide such bindings only if pyceres is available:
```c++
bool pyceres_is_available = true;
try {
  py::module_::import("pyceres");
} catch (const py::import_error&) {
  pyceres_is_available = false;
}
if (pyceres_is_available) {
  // bind custom manifold
}
```

B1ueber2y: > Try this out
> 
> ```c++
>       .def_property(
>           "scale",
>           [](Sim3d& self) { return py::array({}, {}, &self.scale, py::cast(self)); },
>           [](Sim3d& self, double scale) { self.scale = scale; })
> ```

Thanks. This works in my test.

B1ueber2y:
Python minimal test for the ``PositiveExponentialManifold``:
```
import numpy as np
import pyceres
import pycolmap

def optimize_scale(sim3, scale_prior):
    prob = pyceres.Problem()
    prob.add_residual_block(pyceres.factors.NormalPrior(scale_prior, np.eye(1)), pyceres.TrivialLoss(), [sim3.scale])
    prob.set_manifold(sim3.scale, pycolmap.manifold.PositiveExponentialManifold(1))
    options = pyceres.SolverOptions()
    summary = pyceres.SolverSummary()
    pyceres.solve(options, prob, summary)
    return sim3

sim3d = pycolmap.Sim3d()
sim3d = optimize_scale(sim3d, np.array([5.]))
np.testing.assert_allclose(sim3d.scale, np.array([5.]), atol=1e-5)
sim3d = optimize_scale(sim3d, np.array([-1.]))
np.testing.assert_allclose(sim3d.scale, np.array([0.]), atol=1e-5)
```

Python minimal test for ``Point3dAlignmentCost`` on top of some variable points and ``pycolmap.Sim3d``:
```
import numpy as np
import pyceres
import pycolmap
import copy

sim3 = pycolmap.Sim3d()
target_sim3 = pycolmap.Sim3d(2., np.array([0., 0., 0., 1.]), np.array([2., 3., 4.]))
ref_points = [np.array([k, k, k]).astype(float) for k in range(10)]

prob = pyceres.Problem()
points = []
for i, ref_point in enumerate(ref_points):
    if i < 3:
        point = ref_point + np.ones(3) * 0.5 # add noise
    else:
        point = ref_point
    points.append(point)
    target_point = target_sim3 * ref_point
    cost = pycolmap.cost_functions.Point3dAlignmentCost(target_point, np.eye(3))
    prob.add_residual_block(cost, pyceres.TrivialLoss(), [points[i], sim3.rotation.quat, sim3.translation, sim3.scale])
    if i >= 3:
        prob.set_parameter_block_constant(ref_points[i])
prob.set_manifold(sim3.rotation.quat, pyceres.QuaternionManifold())
prob.set_manifold(sim3.scale, pycolmap.manifold.PositiveExponentialManifold(1))
options = pyceres.SolverOptions()
summary = pyceres.SolverSummary()
pyceres.solve(options, prob, summary)

np.testing.assert_allclose(points[0], ref_points[0], atol=1e-5)
np.testing.assert_allclose(points[1], ref_points[1], atol=1e-5)
np.testing.assert_allclose(points[2], ref_points[2], atol=1e-5)
np.testing.assert_allclose(sim3.scale, target_sim3.scale, atol=1e-5)
np.testing.assert_allclose(sim3.rotation.quat, target_sim3.rotation.quat, atol=1e-5)
np.testing.assert_allclose(sim3.translation, target_sim3.translation, atol=1e-5)
```

src/colmap/estimators/cost_functions.h
B1ueber2y: > We could provide such bindings only if pyceres is available:
> 
> ```c++
> bool pyceres_is_available = true;
> try {
>   py::module_::import("pyceres");
> } catch (const py::import_error&) {
>   pyceres_is_available = false;
> }
> if (pyceres_is_available) {
>   // bind custom manifold
> }
> ```

While this could work, this has one clear drawback. Assume the users have pyceres installed and this part of code is thus binded, they will be forced to import pyceres before importing pycolmap. Otherwise the import of pycolmap will raise an error because the ceres::Manifold binding is not found. This does not feel ideal to the pycolmap package.

B1ueber2y: > > We could provide such bindings only if pyceres is available:
> > ```c++
> > bool pyceres_is_available = true;
> > try {
> >   py::module_::import("pyceres");
> > } catch (const py::import_error&) {
> >   pyceres_is_available = false;
> > }
> > if (pyceres_is_available) {
> >   // bind custom manifold
> > }
> > ```
> 
> While this could work, this has one clear drawback. Assume the users have pyceres installed and this part of code is thus binded, they will be forced to import pyceres before importing pycolmap. Otherwise the import of pycolmap will raise an error because the ceres::Manifold binding is not found. This does not feel ideal to the pycolmap package.

Also, seems that the CI does not accept try catch on the py::import_error: https://github.com/colmap/colmap/actions/runs/9727459725/job/26847035961?pr=2621

src/pycolmap/estimators/manifold.cc
```diff
@@ -0,0 +1,34 @@
...
+  py::class_<PositiveExponentialManifold<ceres::DYNAMIC>, ceres::Manifold>(
+      m, "PositiveExponentialManifold")
+      .def(py::init<int>());
+#endif
+}
+
+void BindManifold(py::module& m_parent) {
+  py::module_ m = m_parent.def_submodule("manifold");
+  bool is_pyceres_available = IsPyceresAvailable();
+  if (is_pyceres_available) {
```
sarlinpe, line 31:
```suggestion
  if (IsPyceresAvailable()) {
```

src/colmap/estimators/cost_functions.h
B1ueber2y: > We could provide such bindings only if pyceres is available:
> 
> ```c++
> bool pyceres_is_available = true;
> try {
>   py::module_::import("pyceres");
> } catch (const py::import_error&) {
>   pyceres_is_available = false;
> }
> if (pyceres_is_available) {
>   // bind custom manifold
> }
> ```

py::import_error cannot be caught when pyceres is not installed. Changing it to the general py::error_already_set resolves the issue.

sarlinpe (approved):

B1ueber2y:
I have updated the description for this PR according to the new changes. PTAL when time permits : ) @ahojnnes

src/colmap/estimators/cost_functions.h
```diff
@@ -453,6 +453,47 @@ struct MetricRelativePoseErrorCostFunction {
...
+                  const T* const transform_q,
+                  const T* const transform_t,
+                  const T* const scale,
+                  T* residuals_ptr) const {
+    const Eigen::Quaternion<T> T_q = EigenQuaternionMap<T>(transform_q);
+    const Eigen::Matrix<T, 3, 1> transform_point =
+        T_q * EigenVector3Map<T>(point) * scale[0] +
+        EigenVector3Map<T>(transform_t);
+    for (size_t i = 0; i < 3; ++i) {
+      residuals_ptr[i] = transform_point[i] - T(ref_point_[i]);
```
ahojnnes, line 485: Using EigenVector3Map this could be an eigen expression.

src/colmap/estimators/cost_functions.h
```diff
@@ -453,6 +453,47 @@ struct MetricRelativePoseErrorCostFunction {
   const EigenMatrix6d sqrt_information_j_;
 };
 
+// Cost function for aligning one 3D point with a reference 3D point with
+// covariance. Convention is similar to colmap::Sim3d
+// r = scale * R * point + t - ref_point
+struct Point3dAlignmentCostFunction {
```
ahojnnes, line 459: nit: I recommend to be more explicit about the coordinate frames, for example, use the following variable names: `b_from_a`, `ref_point_in_b`, `cov_in_b`, etc.

ahojnnes (approved):

B1ueber2y:
This feature is actually broken with wheel installation (only works with build from source) after https://github.com/colmap/colmap/pull/2621/commits/42ed7bc419fbae482f7bf72fa82f902dfddadcdc, potentially due to https://github.com/colmap/colmap/pull/2632 and https://github.com/colmap/colmap/pull/2634. PR redirected here: https://github.com/cvg/pyceres/pull/52

PR #2625, opened 2024-06-30
Add support for IMU Preintegration and discrete-time visual-inertial optimization

B1ueber2y:
Reopening for https://github.com/colmap/colmap/pull/2561 (due to some technical issues during rebase)


robinlindh:
Hello! Very nice work! Would be very exciting to get this kind visual-inertial optimization into colmap.

I'm right now trying this out, as it fits a refinement pipeline I work on. I currently use only visual feature matching, so I wanted to try picking the bundle adjustment parts from this PR for adding IMU too.

So far I got it to run similar to in your example.py, but I'm getting a very large cost (some trillions initially and then decreasing to billions after like 50 iterations of ceres solver). Might be that I'm passing IMU readings in the wrong coordinate convention somehow.

Which coordinate system convention, units etc does your IMU code expect? It was a bit hard to judge just from the example code (I did download that np pickle which gives some insights but not sure still). It seems that downloaded file in the example has also done some kind of pre-processing already?

Right now I'm capturing raw Accelerometer and Gyro from ARKit for my dataset, but flipping axes a bit to fit the colmap coordinate system conventions. And some interpolation logic to match timestamps between the gyro and accelerometer.
I was also a bit unsure whether gyro values can be used as is, or if the values need to somehow be converted into "angle axis" which your code seems to expect.

Thankful for any help, and happy to report back here with my results later.

B1ueber2y:
Hey. Thanks for your interest.

> So far I got it to run similar to in your example.py, but I'm getting a very large cost (some trillions initially and then decreasing to billions after like 50 iterations of ceres solver). Might be that I'm passing IMU readings in the wrong coordinate convention somehow.

The noise parameters in the example is picked from the aria doc and are known to be with significant underestimation, which will result in a large cost in general. Using the right noise parameters corresponding to your IMU is crucial for incorporating the preintegration cost into your pipeline.  

> Right now I'm capturing raw Accelerometer and Gyro from ARKit for my dataset, but flipping axes a bit to fit the colmap coordinate system conventions. And some interpolation logic to match timestamps between the gyro and accelerometer. I was also a bit unsure whether gyro values can be used as is, or if the values need to somehow be converted into "angle axis" which your code seems to expect.

The unit of the gyro values in the example is rad/s. 

This feature will likely not be merged very soon. But feel free to give more valuable feedback here. Thanks! : )

robinlindh:
Thanks for the reply!
I managed to find out the correct coordinate conversion I think between ARKit raw IMU capture and what you use here.
In case it's of help to anyone else, I ended up just swapping the order of X and Y for both the gyro and IMU.
That on its own brought the cost down a lot and seems to give better results too.

As for the IMU noise and random walk, I just now changed the values to more appropriate ones for iPhone, and indeed the costs are much more reasonable now, similar order of magnitude as my total reprojection cost from visual features.

> This feature will likely not be merged very soon. But feel free to give more valuable feedback here. Thanks! : )

No worries at all! I know the PR is still in draft stage too and that there was more work you wanted to do for full integration. For me I can already use it and hopefully just more feedback here can help others and eventually help this PR to get merged :)

Cheers!

davidwatkins-bdai:
@B1ueber2y what is the status on this PR? Is there ongoing work that needs to be done in order to finalize this? The example currently does not run as is with your most recent changes.

mvictora-bdai:
I think the main problem with the current example is that it doesn't use the changes introduced in https://github.com/colmap/colmap/pull/2896/files . However, I have a change that does work. I don't have permissions to push my branch but I basically just changed one function:

```python
def solve_bundle_adjustment(
    reconstruction, ba_options, ba_config, preintegrated_measurements, variables
):
    bundle_adjuster = pycolmap.create_default_bundle_adjuster(ba_options, ba_config, reconstruction)

    summary = bundle_adjuster.solve()

    print(summary.BriefReport())
    return summary
```

B1ueber2y:
> @B1ueber2y what is the status on this PR? Is there ongoing work that needs to be done in order to finalize this? The example currently does not run as is with your most recent changes.

Hellooo. Thanks for the interest. This has been fixed just now. This PR likely wont be merged very soon since we plan to only support IMU after supporting rig data structure in COLMAP https://github.com/colmap/colmap/pull/2698. I will try to find some time to push this in the following weeks.

mvictora-bdai:
> Hellooo. Thanks for the interest. The example has been fixed just now. This PR likely wont be merged very soon since we plan to only support IMU preintegration after supporting rig data structure in COLMAP #2698. I will try to find some time to push this in the following weeks.

Asking for a bit of clarification here -- does the code work as-is (at least if you're able to adapt your data to fit to something like the example)? You're waiting for rig data structure work to be merged, and then you will rework this code to be built on top of a multi-sensor rig data structure before merging it to main? Will this new data structure help with being able to achieve better state estimation?

B1ueber2y:
> > Hellooo. Thanks for the interest. The example has been fixed just now. This PR likely wont be merged very soon since we plan to only support IMU preintegration after supporting rig data structure in COLMAP #2698. I will try to find some time to push this in the following weeks.
> 
> Asking for a bit of clarification here -- does the code work as-is (at least if you're able to adapt your data to fit to something like the example)? You're waiting for rig data structure work to be merged, and then you will rework this code to be built on top of a multi-sensor rig data structure before merging it to main? Will this new data structure help with being able to achieve better state estimation?

Yes your understanding is correct. The feature is completed and we are waiting for the rig support to be completed. There is no point to support IMU preintegration in COLMAP without rig support, since this is only useful for visual-inertial optimization. Thanks!

mvictora-bdai:
I want to probe a bit deeper on what rig support will provide -- right now, I want to use IMU data from a GoPro to improve localization I achieve using glomap. However, to do that I need to:

1. Interpolate/align my IMU timestamps with my RGB frames
2. Handle the transform from my IMU sensor to my camera

Will https://github.com/colmap/colmap/pull/2698 help me handle those needs?

B1ueber2y:
> I want to probe a bit deeper on what rig support will provide -- right now, I want to use IMU data from a GoPro to improve localization I achieve using glomap. However, to do that I need to:
> 
>     1. Interpolate/align my IMU timestamps with my RGB frames
> 
>     2. Handle the transform from my IMU sensor to my camera
> 
> 
> Will #2698 help me handle those needs?

The rig support can help you handle the transformation from your IMU sensor to the camera. However, we dont plan to support synchronization optimization in the very near future, so you will have to provide synchronization yourself. If the timestamps of your IMU measurements and RGB images are well synchronized we could help handle it in the visual-inertial optimization.

B1ueber2y:
Want to make an update to this PR that there is this [project](https://github.com/cvg/lamaria) (cc: @nushakrishnan @sarlinpe ) just released dependent on the wheels built from here: https://github.com/cvg/lamaria/blob/main/requirements.txt#L4-L6.

And there is a working example of visual-inertial optimization available in the repo: https://github.com/cvg/lamaria?tab=readme-ov-file#example-visual-inertial-optimization

The optimization setup can be found here: https://github.com/cvg/lamaria/blob/main/lamaria/pipeline/optim/vi_optimization.py

The optimization is only a byproduct of the groundtruthing pipeline of the dataset, and would need more rigorous testing and benchmarking. But I believe its a good reference point when we start moving forward on this PR to merge the IMU utility into COLMAP.

pycolmap/tests/vi_optimization/example.py
```diff
@@ -0,0 +1,312 @@
+"""
```
sarlinpe, line 1: Move this file to `python/examples/vi_optimization.py`?

pycolmap/tests/vi_optimization/example.py
```diff
@@ -0,0 +1,312 @@
...
+def add_imu_residuals(
+    prob,
+    reconstruction,
+    preintegrated_measurements,
+    variables,
+    optimize_scale=True,
+    optimize_gravity=True,
+    optimize_imu_from_cam=True,
+    optimize_bias=True,
+):
```
sarlinpe: Having type hints everywhere would be great to help follow the example.

src/colmap/estimators/cost_functions.h
```diff
@@ -340,6 +340,17 @@ inline void EigenQuaternionToAngleAxis(const T* eigen_quaternion,
   ceres::QuaternionToAngleAxis(quaternion, angle_axis);
 }
 
+template <typename T>
+inline void EigenAngleAxisToQuaternion(const T* angle_axis,
```
sarlinpe: Rename to AngleAxisToEigenQuaternion?

src/colmap/estimators/imu_preintegration.h
```diff
@@ -0,0 +1,304 @@
...
+namespace colmap {
+
+struct ImuPreintegrationOptions {
+  // integration noise density
+  bool use_integration_noise = false;
+  double integration_noise_density = 0.2;  // [m/s*1/sqrt(Hz)]
+
+  // check whether to reintegrate
+  double reintegrate_vel_norm_thres = 0.0001;
+  double reintegrate_angle_norm_thres = 0.0001;
```
sarlinpe: Can we better document this options with more descriptive comments? Including the missing units.

src/colmap/estimators/imu_preintegration.h
```diff
@@ -0,0 +1,304 @@
...
+  const Eigen::Vector6d& Biases() const;
+  const Eigen::Matrix<double, 15, 15> Covariance() const;
+  const Eigen::Matrix<double, 15, 15> SqrtInformation() const;
+  const double GravityMagnitude() const;
+  const ImuMeasurements Measurements() const;
+
+ private:
+  // Clock
+  double t_start_ = 0.0;  // from what time to start the integration.
+  double t_end_ = 0.0;    // until what time to do the integration.
```
sarlinpe: Are these in seconds?

src/colmap/estimators/imu_preintegration.h
```diff
@@ -0,0 +1,304 @@
...
+
+  // Measurements
+  ImuMeasurements measurements_;
+
+  // Options
+  ImuPreintegrationOptions options_;
+
+  // IMU Calibration.
+  ImuCalibration calib_;
+  // TODO: the rectification matrix and bias should go into calibration
```
sarlinpe: Should we address the TODOs before merging this PR?

src/colmap/estimators/imu_preintegration.h
```diff
@@ -0,0 +1,304 @@
...
+      Eigen::Matrix<double, 9, 6>::Zero();  // jacobian of rotation,
+                                            // translation, velocity over biases
+
+  // Covariance propagation
+  Eigen::Matrix<double, 15, 15> covs_ =
+      Eigen::Matrix<double, 15, 15>::Zero();  // covariances (rotation +
+                                              // translation + velocity + acc
+                                              // bias + gyro bias)
+  Eigen::Matrix<double, 15, 15> sqrt_information_ =
+      Eigen::Matrix<double, 15, 15>::Zero();
```
sarlinpe: Consider better documenting the shape of these matrices, with comments that are more nicely formatted.

src/colmap/geometry/pose.h
```diff
@@ -86,6 +86,20 @@ Eigen::Quaterniond AverageQuaternions(
     const std::vector<Eigen::Quaterniond>& quats,
     const std::vector<double>& weights);
 
+// Compute a quaternion from lie algebra
```
sarlinpe: nit: Comments should preferably end with a period. This applies to other comments as well.

src/colmap/scene/imu.h
```diff
@@ -0,0 +1,134 @@
...
+
+#include <vector>
+
+#include <Eigen/Dense>
+#include <Eigen/Geometry>
+
+namespace colmap {
+
+// An Imu class storing the sensor information and a linked visual camera
+// TODO: support rigs + unify sensors
```
sarlinpe: Should we address this comment or leave for later?

src/colmap/scene/imu.h
```diff
@@ -0,0 +1,134 @@
...
+
+  inline const camera_t& ImuId() const;
+  inline camera_t ImuId();
+  inline const image_t& ImageId() const;
+  inline image_t ImageId();
+
+ private:
+  Eigen::Matrix<double, 9, 1> data_ =
+      Eigen::Matrix<double, 9, 1>::Zero();  // 3-DoF speed + 6-DoF biases (acc +
+                                            // gyro)
```
sarlinpe: Consider improving the formatting of the comment and making it more informative.

src/colmap/sensor/imu.h
```diff
@@ -0,0 +1,154 @@
...
+    } else {
+      for (auto it = ms.begin(); it != ms.end(); ++it) insert(*it);
+    }
+  }
+  void insert(const ImuMeasurements& ms) {
+    if (empty()) {
+      measurements_ = ms.Data();
+    } else {
+      for (auto it = ms.begin(); it != ms.end(); ++it) insert(*it);
+      // TODO: merge two sorted list could be made faster
```
sarlinpe: At least throw an error?

src/colmap/sensor/imu.h
```diff
@@ -0,0 +1,154 @@
...
+    auto it = std::lower_bound(
+        measurements_.begin(),
+        measurements_.end(),
+        m,
+        [](const ImuMeasurement& m1, const ImuMeasurement& m2) {
+          return m1.timestamp < m2.timestamp;
+        });
+    measurements_.insert(it, m);
+  }
+  void insert(const std::vector<ImuMeasurement>& ms) {
```
sarlinpe: Would we benefit from a method that inserts measurements that are already sorted?

src/colmap/estimators/imu_preintegration.h
```diff
@@ -0,0 +1,304 @@
...
+  void SetBiases(const Eigen::Vector6d& biases);
+
+  // Add measurements: measurements need to be added in chronological order
+  void AddMeasurement(const ImuMeasurement& m);
+  void AddMeasurements(const ImuMeasurements& ms);
+  void Finish();
+  bool HasFinished() const;
+
+  // Reintegrate
+  bool CheckReintegrate(const Eigen::Vector6d& biases) const;
```
sarlinpe: ShouldReintegrate might be more informative.

src/pycolmap/scene/imu.cc
```diff
@@ -0,0 +1,121 @@
...
+                     &ImuMeasurement::linear_acceleration)
+      .def_readwrite("angular_velocity", &ImuMeasurement::angular_velocity)
+      .def("__repr__", [](const ImuMeasurement& m) {
+        std::stringstream ss;
+        ss << "ImuMeasurement("
+           << "t=" << m.timestamp << ", "
+           << "acc=[" << m.linear_acceleration.format(vec_fmt) << "], "
+           << "gyro=[" << m.angular_velocity.format(vec_fmt) << "])";
+        return ss.str();
+      });
```
sarlinpe, line 55: Define `operator<<` in the header files instead.

src/colmap/estimators/imu_preintegration.cc
```diff
@@ -0,0 +1,412 @@
...
+  }
+  Finish();
+}
+
+void PreintegratedImuMeasurement::Reintegrate(const Eigen::Vector6d& biases) {
+  SetBiases(biases);
+  Reintegrate();
+}
+
+// data interfaces
```
sarlinpe: Should all of these be inlined?

sarlinpe (commented):
I haven't yet checked the maths 😬

B1ueber2y:
Thanks for the review. However, I dont expect this PR to be merged anytime soon though. A better redesign of this PR with the new colmap rig abstraction would be:

- Step 1. add timestamp to each measurement, e.g., colmap::Image
- Step 2. support time-varying calibration, since imu bias parameters should conceptually go to ``IMUCalibration`` rather than ``IMUState``. IMUState should ideally only include speed and potentially angular velocity with the rig modeling (can be a bit tricky with multiple IMUs).
- Step 3. make IMU a class of the sensor alongside ``colmap::Camera``, and IMUMeasurement a class of measurements alongside ``colmap::Image``, and add to the Reconstruction. 

Now in the other repo, all these are achieved with a temporary solution in the Python side, with the abstractions of ``TimedReconstruction`` (corresponding to step 1): https://github.com/cvg/lamaria/blob/main/lamaria/structs/timed_reconstruction.py and ``VIReconstruction`` (corresponding to step 3): https://github.com/cvg/lamaria/blob/main/lamaria/structs/vi_reconstruction.py.

TannerGilbert:
> Thanks for the review. However, I dont expect this PR to be merged anytime soon though. A better redesign of this PR with the new colmap rig abstraction would be:
> 
> * Step 1. add timestamp to each measurement, e.g., colmap::Image
> * Step 2. support time-varying calibration, since imu bias parameters should conceptually go to `IMUCalibration` rather than `IMUState`. IMUState should ideally only include speed and potentially angular velocity with the rig modeling (can be a bit tricky with multiple IMUs).
> * Step 3. make IMU a class of the sensor alongside `colmap::Camera`, and IMUMeasurement a class of measurements alongside `colmap::Image`, and add to the Reconstruction.
> 
> Now in the other repo, all these are achieved with a temporary solution in the Python side, with the abstractions of `TimedReconstruction` (corresponding to step 1): https://github.com/cvg/lamaria/blob/main/lamaria/structs/timed_reconstruction.py and `VIReconstruction` (corresponding to step 3): https://github.com/cvg/lamaria/blob/main/lamaria/structs/vi_reconstruction.py.

@B1ueber2y as far as I understand rig support was added with [3.12.0](https://github.com/colmap/colmap/releases/tag/3.12.0) and extended in the following versions. Would it now be possible to redesign the PR with the changes you mentioned above?

B1ueber2y:
This gigantic PR is ready to review. However, the testing of the python example was currently blocked by https://github.com/colmap/colmap/issues/4285. 

Feel free to have a pass if time permits @ahojnnes @sarlinpe

B1ueber2y:
To make a summary, there has been a deep rework on this PR, including:
* Separation of preintegration and cost functor.
* int64_t timestamp with utilities.
* Reintegration callback to avoid mutate variables.
* Converted Forster et al. TRO'16 fully to left convention with a rework on maths.
* Added an integration variant: RK4 described in Eckenhoff et al. "Closed-form Preintegration Methods for Graph-based Visual-Inertial Navigation", which is used in OpenVINS.
* Two different costs, one IMU-centric, one visual centric (with additional scale, Rigid3d, and gravity).
* Analytical Jacobians that gets significantly faster than the autodiff for these two costs.
* Extensive test coverage.

src/colmap/scene/imu.h
```diff
@@ -0,0 +1,117 @@
...
+
+namespace colmap {
+
+// An Imu class storing the sensor information and a linked visual camera.
+// TODO: Integrate with the Rig abstraction (sensor/rig.h) by making IMU a
+// proper sensor in the rig and using sensor_from_rig transforms.
+class Imu {
+ public:
+  Imu() = default;
+  ~Imu() = default;
```
ahojnnes: Can this just be a struct?

src/colmap/scene/imu.h
```diff
@@ -0,0 +1,117 @@
...
+
+  inline Eigen::Map<Eigen::Vector3d> bias_accel() {
+    return Eigen::Map<Eigen::Vector3d>(params.data() + 6);
+  }
+  inline Eigen::Map<const Eigen::Vector3d> bias_accel() const {
+    return Eigen::Map<const Eigen::Vector3d>(params.data() + 6);
+  }
+
+  camera_t imu_id = kInvalidCameraId;
+  image_t image_id = kInvalidImageId;
```
ahojnnes: move to the top?

ahojnnes: Should these measurements more generally be associated as sensor_t and data_t instead of camera/image?

src/colmap/geometry/pose.h
```diff
@@ -115,6 +115,29 @@ Eigen::Quaterniond AverageQuaternions(
...
+//
+// @param omega         The angles in so(3).
+//
+// @return              The quaternion corresponding to omega in so(3).
+Eigen::Quaterniond QuaternionFromAngleAxis(const Eigen::Vector3d& omega);
+
+// Compute the left Jacobian J_l from lie algebra.
+// Used for bias Jacobian propagation in IMU preintegration.
+//
+// @param omega         The angles in so(3)
```
ahojnnes, line 128: nit:
```suggestion
// @param omega         The angles in SO(3).
```

src/colmap/estimators/imu_preintegration_test.cc
```diff
@@ -0,0 +1,348 @@
...
+    PreintegratedImuData data_minus =
+        IntegrateWithBiases(accel, gyro, N, dt, biases_minus, method);
+
+    dp_dba_numeric.col(k) =
+        (data_plus.delta_p - data_minus.delta_p) / (2 * eps);
+    dv_dba_numeric.col(k) =
+        (data_plus.delta_v - data_minus.delta_v) / (2 * eps);
+  }
+
+  const double tol = 1e-4;
```
ahojnnes: nit constexpr kTol

src/colmap/estimators/imu_preintegration.h
```diff
@@ -0,0 +1,243 @@
...
+
+#include <Eigen/Core>
+#include <Eigen/Geometry>
+#include <ceres/ceres.h>
+
+namespace colmap {
+
+// Midpoint: trapezoidal integration, sufficient for typical 200-1000 Hz IMUs.
+// RK4: closed-form rotation integrals with analytical bias Jacobians and
+//      RK4 covariance propagation, more accurate for lower IMU rates.
```
ahojnnes: I guess midpoint is faster. Maybe worth calling out the difference and when to use which?

src/colmap/estimators/imu_preintegration.h
```diff
@@ -0,0 +1,243 @@
...
+// Algorithm class that performs IMU preintegration.
+// Owns the raw measurements, calibration, and integration options.
+// Produces a PreintegratedImuData data struct via Extract().
+class ImuPreintegrator {
+ public:
+  ImuPreintegrator(const ImuPreintegrationOptions& options,
+                   const ImuCalibration& calib,
+                   timestamp_t t_start,
+                   timestamp_t t_end);
+  ~ImuPreintegrator() = default;
```
ahojnnes: Not needed?

src/colmap/estimators/imu_preintegration.h
```diff
@@ -0,0 +1,243 @@
...
+
+  // Accumulated preintegrated data.
+  PreintegratedImuData data_;
+
+  // Raw measurements, sorted by timestamp. Chronological order is enforced
+  // by FeedImu() via THROW_CHECK_GT on consecutive timestamps.
+  ImuMeasurements measurements_;
+
+  // Options
+  ImuPreintegrationOptions options_;
```
ahojnnes: const?

src/colmap/estimators/imu_preintegration.cc
```diff
@@ -0,0 +1,587 @@
...
+  // so that sqrt_info^T * sqrt_info = V * D^{-1} * V^T = information.
+  sqrt_information = D_inv_sqrt.asDiagonal() * saes.eigenvectors().transpose();
+}
+
+ImuPreintegrator::ImuPreintegrator(const ImuPreintegrationOptions& options,
+                                   const ImuCalibration& calib,
+                                   timestamp_t t_start,
+                                   timestamp_t t_end) {
+  options_ = options;
+  calib_ = calib;
```
ahojnnes: Initialize with member initializer lists?

src/colmap/estimators/imu_preintegration.h
```diff
@@ -0,0 +1,243 @@
...
+                         const Eigen::Vector3d& gyro_true,
+                         double dt,
+                         double accel_noise_density,
+                         double gyro_noise_density);
+
+  void IntegrateRK4(const Eigen::Vector3d& accel_true,
+                    const Eigen::Vector3d& gyro_true,
+                    double dt,
+                    double accel_noise_density,
+                    double gyro_noise_density);
```
ahojnnes: In terms of design, would it make sense to have a base class with shared functionality and a static `Create(ImuIntegrationMethod)` factory method that returns different implementations of a `virtual Integrate()` method instead of implementing this with a switch/case table? The current design is also fine, just an idea.

src/colmap/estimators/imu_preintegration.h
```diff
@@ -0,0 +1,243 @@
...
+//      RK4 covariance propagation, more accurate for lower IMU rates.
+MAKE_ENUM_CLASS(ImuIntegrationMethod, 0, MIDPOINT, RK4);
+
+struct ImuPreintegrationOptions {
+  ImuIntegrationMethod method = ImuIntegrationMethod::RK4;
+
+  // Whether to add integration noise to the covariance propagation.
+  bool use_integration_noise = false;
+  // Integration noise density. [m/s*1/sqrt(Hz)]
+  double integration_noise_density = 0.2;
```
ahojnnes: A brief comment on how to use these in practice? When would I enable this and why and how does the density affect the results?

src/colmap/estimators/imu_preintegration.cc
```diff
@@ -0,0 +1,587 @@
...
+  double vars_bg = pow(calib_.bias_gyro_random_walk_sigma, 2) * dt;
+  data_.covariance.block<3, 3>(0, 0) +=
+      Eigen::Matrix3d::Identity() * vars_omega;
+  data_.covariance.block<3, 3>(3, 3) += Eigen::Matrix3d::Identity() * vars_p;
+  data_.covariance.block<3, 3>(6, 6) += Eigen::Matrix3d::Identity() * vars_v;
+  data_.covariance.block<3, 3>(9, 9) += Eigen::Matrix3d::Identity() * vars_bg;
+  data_.covariance.block<3, 3>(12, 12) += Eigen::Matrix3d::Identity() * vars_ba;
+}
+
+void ImuPreintegrator::IntegrateRK4(const Eigen::Vector3d& accel_true,
```
ahojnnes: Could apply some stricter const correctness here.

src/colmap/estimators/cost_functions/utils.h
```diff
@@ -41,6 +42,27 @@ using EigenVector3Map = Eigen::Map<const Eigen::Matrix<T, 3, 1>>;
 template <typename T>
 using EigenQuaternionMap = Eigen::Map<const Eigen::Quaternion<T>>;
 
+template <typename T>
```
ahojnnes: Should go into quaternion_utils.h ?

src/colmap/estimators/cost_functions/quaternion_utils_test.cc
```diff
@@ -0,0 +1,138 @@
...
+                          Eigen::Vector4d(rm.x(), rm.y(), rm.z(), rm.w())) /
+                         (2.0 * eps);
+    }
+    EXPECT_THAT(QuaternionRightMultMatrix(p), EigenMatrixNear(J_numeric, 1e-5));
+  }
+}
+
+TEST(QuaternionRotatePointWithJac, Nominal) {
+  SetPRNGSeed(42);
+  const double eps = 1e-7;
```
ahojnnes: constexpr kEps

ahojnnes: Same elsewhere.

src/colmap/estimators/imu_preintegration_cost.h
```diff
@@ -0,0 +1,856 @@
...
+    Eigen::Map<Eigen::Matrix<T, 15, 1>> residuals_data(residuals);
+    residuals_data.applyOnTheLeft(data_->sqrt_information.cast<T>());
+    return true;
+  }
+
+ private:
+  const PreintegratedImuData* data_;
+};
+
+// Analytical-Jacobian version of VisualCentricImuPreintegrationCostFunctor.
```
ahojnnes, line 558: How important is it to have this analytically derived vs. using auto-diff?

ahojnnes (commented):
Looks fantastic. Made a first (very superficial) pass and need to carve out some dedicated time to look at this in detail. Happy holidays in the meanwhile.

src/colmap/geometry/pose.h
B1ueber2y: Here i think the original comment is actually correct. The parameter omega is a 3-vector in so(3) (the Lie algebra, i.e., an axis-angle/rotation vector), not SO(3) (the Lie group, i.e., a rotation matrix or quaternion).

src/colmap/estimators/imu_preintegration.cc
B1ueber2y: I did not understand this comment. Could you help clarify? Thanks.

src/colmap/estimators/imu_preintegration_cost.h
B1ueber2y: Will do some proper benchmarking with the google benchmark tooling. IIRC on Monday switching to analytics improved the speed by some margin.

src/colmap/estimators/imu_preintegration.cc
ahojnnes: F1, ... P_dot_1, ... could all be const.

src/colmap/sensor/imu.h
```diff
@@ -0,0 +1,108 @@
...
+                         const ImuMeasurement& measurement);
+
+// IMU measurements stored as a plain vector. Callers must ensure
+// measurements are sorted by timestamp (chronological order).
+using ImuMeasurements = std::vector<ImuMeasurement>;
+
+// Extract measurements spanning the edge [t1, t2] from a sorted vector.
+// Returns measurements from the sample just before t1 through the sample at
+// or just after t2.
+ImuMeasurements GetMeasurementsContainEdge(const ImuMeasurements& measurements,
```
ahojnnes: nit: ExtractMeasurementsInTimeRange?

B1ueber2y: Its actually ContainEdge in the sense that we need the measurements before and after as well, to be able to integrate through the full range. I updated the comments.

TannerGilbert:
@B1ueber2y I tried running this today but first I ran into #4285. PR #4320 should fix this. 

Afterwards though I ran into the following issue with `pyceres`. I'm currently using `https://github.com/cvg/pyceres.git@v2.6`. I didn't find `ProductManifold` in pyceres even when looking at main though.
```
Traceback (most recent call last):
  File "/workspace/demo/reconstruction/reconstruct_cubemap.py", line 450, in <module>
    run_vi_refinement(rec, database_path, args.imu_csv, args.output)
  File "/workspace/demo/reconstruction/reconstruct_360.py", line 425, in run_vi_refinement
    summary = solve_vi_bundle_adjustment(
              ^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/workspace/demo/reconstruction/reconstruct_360.py", line 274, in solve_vi_bundle_adjustment
    add_imu_residuals(problem, reconstruction, imu_data, variables)
  File "/workspace/demo/reconstruction/reconstruct_360.py", line 255, in add_imu_residuals
    pyceres.ProductManifold(
    ^^^^^^^^^^^^^^^^^^^^^^^
AttributeError: module 'pyceres' has no attribute 'ProductManifold
```

src/colmap/estimators/imu_preintegration.h
B1ueber2y: We only have two methods and I do not expect us to support more in the future. Both methods do not have any persistent method-specific config/state. So I would vote for keeping it as it currently is.

src/colmap/estimators/imu_preintegration_cost.h
B1ueber2y: The analytical jacobian on the cost functors itself have large speed up factors.


| case | AutoDiff | Analytical | speedup |
  |----------------|----------|------------|---------|
  | body-centric | ~18.4 µs | ~2.29 µs | ~8× |
  | visual-centric | ~35.2 µs | ~3.39 µs | ~10× |

src/colmap/sensor/imu.h
```diff
@@ -0,0 +1,234 @@
...
+
+namespace colmap {
+
+// References:
+// [1] https://github.com/ethz-asl/kalibr/wiki/IMU-Noise-Model-and-Intrinsics
+// [2]
+// https://github.com/uzh-rpg/rpg_svo_pro_open/blob/master/svo_common/include/svo/common/imu_calibration.h
+// Default parameters are for ADIS16448 IMU.
+struct ImuCalibration {
+  /// Gyro noise density (sigma). [rad/s*1/sqrt(Hz)]
```
ahojnnes: nit: Applies also to other comments below.
```suggestion
  // Gyro noise density (sigma). [rad/s*1/sqrt(Hz)]
```

src/colmap/sensor/imu.h
```diff
@@ -0,0 +1,234 @@
...
+  double accel_saturation_max = 150;
+
+  /// Norm of the Gravitational acceleration. [m/s^2]
+  double gravity_magnitude = 9.81007;
+
+  /// Expected IMU Rate [1/s]
+  double imu_rate = 20.0;
+
+  /// Rectification matrices correcting axis misalignment and scale.
+  /// m_true = M^{-1}(m - b). Identity if data is already rectified.
```
ahojnnes: What is b and m here?

src/colmap/sensor/imu.h
```diff
@@ -0,0 +1,234 @@
...
+  /// TODO: Support online calibration by making these optimizable
+  /// as parameter blocks in the IMU cost function.
+  Eigen::Matrix3d gyro_rectification = Eigen::Matrix3d::Identity();
```
ahojnnes, line 80: Can we be more explicit about the transformation? For example, `gyro_true = true_from_measured * gyro_measured` or similar. Just making these up, as I have not read the whole code and starting here with the review.

src/colmap/sensor/imu.h
```diff
@@ -0,0 +1,234 @@
...
+  Eigen::Matrix3d accel_rectification = Eigen::Matrix3d::Identity();
+};
+
+struct ImuMeasurement {
+  timestamp_t timestamp = kInvalidTimestamp;  // [nanoseconds]
+  Eigen::Vector3d gyro = Eigen::Vector3d::Zero();
+  Eigen::Vector3d accel = Eigen::Vector3d::Zero();
```
ahojnnes, line 87: Is 0 a good default or use quiet_NaN similar to pose prior?

src/colmap/sensor/imu.h
```diff
@@ -0,0 +1,234 @@
...
+  void Insert(const ImuMeasurement& m) {
+    // Fast path: append if empty or new measurement comes after all existing.
+    if (Empty() || m.timestamp > measurements_.back().timestamp) {
+      measurements_.push_back(m);
+      return;
+    }
+    if (m.timestamp == measurements_.back().timestamp) {
+      throw std::invalid_argument("Duplicate timestamp in ImuMeasurements: " +
+                                  std::to_string(m.timestamp));
+    }
```
ahojnnes: Wouldn't this already be handled by the below check?

src/colmap/sensor/imu.h
```diff
@@ -0,0 +1,234 @@
...
+    if (Empty()) {
+      measurements_ = ms;
+      std::sort(measurements_.begin(),
+                measurements_.end(),
+                [](const ImuMeasurement& m1, const ImuMeasurement& m2) {
+                  return m1.timestamp < m2.timestamp;
+                });
+      ThrowIfHasDuplicates(measurements_);
+    } else {
+      for (auto it = ms.begin(); it != ms.end(); ++it) Insert(*it);
```
ahojnnes: This loop has quite bad performance. Wouldn't it be more efficient to just append all measurements at the end of the vector and then sort + ThrowIfDuplicates? Alternatively, take the vector by copy, then sort and use InsertSorted?

src/colmap/sensor/imu.h
```diff
@@ -0,0 +1,234 @@
...
+
+// Sorted list of IMU measurements ordered by timestamp.
+class ImuMeasurements {
+ public:
+  ImuMeasurements() = default;
+  explicit ImuMeasurements(const std::vector<ImuMeasurement>& ms) {
+    Insert(ms);
+  }
+
+  void Insert(const ImuMeasurement& m) {
```
ahojnnes: Do we really need this interface? It has bad complexity.

src/colmap/sensor/imu.h
```diff
@@ -0,0 +1,234 @@
...
+};
+
+std::ostream& operator<<(std::ostream& stream,
+                         const ImuCalibration& calibration);
+
+std::ostream& operator<<(std::ostream& stream,
+                         const ImuMeasurement& measurement);
+
+// Sorted list of IMU measurements ordered by timestamp.
+class ImuMeasurements {
```
ahojnnes, line 103: * The methods in the header of this class probably all need inlining otherwise we'll end up with duplicate symbols?
* I'd suggest moving most of the heavier methods into the .cc file.

src/colmap/sensor/imu.h
```diff
@@ -0,0 +1,234 @@
...
+  /// Gyroscope saturation. [rad/s]
+  double gyro_saturation_max = 7.8;
+
+  /// Accelerometer saturation. [m/s^2]
+  double accel_saturation_max = 150;
+
+  /// Norm of the Gravitational acceleration. [m/s^2]
+  double gravity_magnitude = 9.81007;
+
+  /// Expected IMU Rate [1/s]
```
ahojnnes: nit
```suggestion
  /// Expected IMU rate. [1/s]
```

src/colmap/sensor/imu.cc
```diff
@@ -0,0 +1,78 @@
...
+// ARISING IN ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE
+// POSSIBILITY OF SUCH DAMAGE.
+
+#include "colmap/sensor/imu.h"
+
+#include "colmap/util/logging.h"
+
+namespace colmap {
+
+ImuMeasurements ImuMeasurements::ExtractMeasurementsContainEdge(
```
ahojnnes: nit: Maybe better?
```suggestion
ImuMeasurements ImuMeasurements::ExtractMeasurementsInRange(
```

src/colmap/sensor/imu.cc
```diff
@@ -0,0 +1,78 @@
...
+  // The edge cannot be bracketed if it extends beyond the available samples
+  // (e.g. the first/last image edge or across an IMU gap). Return empty so
+  // callers can skip this edge rather than treating it as a fatal error.
+  if (t1 < front().timestamp || t2 > back().timestamp) {
+    return ImuMeasurements();
+  }
+  auto cmp = [](const ImuMeasurement& m1, const ImuMeasurement& m2) {
+    return m1.timestamp < m2.timestamp;
+  };
+  ImuMeasurement dummy;
```
ahojnnes:
```suggestion
  ImuMeasurement range;
```

src/colmap/sensor/imu_test.cc
```diff
@@ -0,0 +1,224 @@
...
+#include "colmap/sensor/imu.h"
+
+#include <gtest/gtest.h>
+
+namespace colmap {
+namespace {
+
+const Eigen::Vector3d kZero = Eigen::Vector3d::Zero();
+
+ImuMeasurement M(timestamp_t t) { return ImuMeasurement(t, kZero, kZero); }
```
ahojnnes: CreateMeasurement?

src/colmap/sensor/imu_test.cc
```diff
@@ -0,0 +1,224 @@
...
+
+TEST(ImuMeasurements, Nominal) {
```
ahojnnes: Suggest decomposing these into dedicated tests for each "ms".

src/pycolmap/estimators/imu_preintegration.cc
```diff
@@ -0,0 +1,100 @@
+#include "colmap/estimators/imu_preintegration.h"
```
ahojnnes, line 1: Let's cover these new bindings in our bindings tests.

src/colmap/scene/imu.h
```diff
@@ -0,0 +1,111 @@
...
+#include <ostream>
+#include <vector>
+
+#include <Eigen/Dense>
+#include <Eigen/Geometry>
+
+namespace colmap {
+
+// An Imu class storing the sensor information and a linked visual camera.
+// TODO: Integrate with the Rig abstraction (sensor/rig.h) by making IMU a
```
ahojnnes, line 47: Why not now? Feels like we will otherwise create a lot of technical debt?

src/colmap/scene/imu.h
```diff
@@ -0,0 +1,111 @@
...
+
+  inline Eigen::Map<Eigen::Vector3d> bias_accel() {
+    return Eigen::Map<Eigen::Vector3d>(params.data() + 6);
+  }
+  inline Eigen::Map<const Eigen::Vector3d> bias_accel() const {
+    return Eigen::Map<const Eigen::Vector3d>(params.data() + 6);
+  }
+};
+
+inline std::ostream& operator<<(std::ostream& stream, const Imu& imu) {
```
ahojnnes: Move to .cc

src/colmap/scene/imu.h
```diff
@@ -0,0 +1,111 @@
...
+  stream << "Imu("
+         << "imu_id=" << imu.imu_id << ", "
+         << "camera_id=" << imu.camera_id << ")";
+  return stream;
+}
+
+inline std::ostream& operator<<(std::ostream& stream, const ImuState& state) {
```
ahojnnes: Move to .cc

src/colmap/scene/CMakeLists.txt
```diff
@@ -36,6 +36,7 @@ COLMAP_ADD_LIBRARY(
         camera.h camera.cc
         frame.h frame.cc
         image.h image.cc
+        imu.h
```
ahojnnes: Add tests.

src/colmap/sensor/imu_test.cc
```diff
@@ -0,0 +1,224 @@
...
+// INTERRUPTION) HOWEVER CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN
+// CONTRACT, STRICT LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE)
+// ARISING IN ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE
+// POSSIBILITY OF SUCH DAMAGE.
+
+#include "colmap/sensor/imu.h"
+
+#include <gtest/gtest.h>
+
+namespace colmap {
```
ahojnnes, line 37: Add test for ostream. See all other structs/classes with equivalent tests.

src/colmap/estimators/imu_preintegration.h
```diff
@@ -0,0 +1,259 @@
...
+// Left-multiply integration convention:
+//   delta_R_{k+1} = Exp(-w*dt) * delta_R_k
+//   delta_R = body_from_world_j * world_from_body_i
+//   delta_p = body_from_world_i * (p_j - p_i - v_i*dt - 0.5*g*dt^2)
+//   delta_v = body_from_world_i * (v_j - v_i - g*dt)
+//
+// Right-perturbation model for bias Jacobians:
+//   delta_R(bg + dbg) ≈ delta_R * Exp(dR_dbg * dbg)
+//   The bias correction is always right-multiplied in the cost function.
+struct PreintegratedImuData {
```
ahojnnes, line 87: nit: Maybe PreintegratedImuState ?

src/colmap/estimators/imu_preintegration.h
ahojnnes: The benefit of having an abstract interface is that we can have parameterized tests or mock/fake the implementation when the pre-integrator is part of another pipeline. Unless it's difficult to implement (and probably Claude can one-shot the change?) or there is some other conceptual blocker, I'd suggest to make the change.

src/colmap/estimators/imu_preintegration.h
```diff
@@ -0,0 +1,259 @@
...
+  // Raw measurements, sorted by timestamp. Chronological order is enforced
+  // by FeedImu() via THROW_CHECK_GT on consecutive timestamps.
+  ImuMeasurements measurements_;
+
+  // Options
+  const ImuPreintegrationOptions options_;
+
+  // IMU Calibration.
+  const ImuCalibration calib_;
+  Eigen::Matrix3d accel_rect_mat_inv_ = Eigen::Matrix3d::Identity();
```
ahojnnes: Same comment as in sensor/imu.h about using more explicit notation for the transformation. In this case, maybe `measured_from_true` ?

src/colmap/estimators/imu_preintegration.h
```diff
@@ -0,0 +1,259 @@
...
+struct ImuReintegrationOptions {
+  // Threshold on gyro bias change (angle norm = ||delta_bg|| * delta_t).
+  // [rad/s]
+  double reintegrate_angle_norm_thres = 1e-4;
+  // Threshold on accel bias change (velocity norm = ||delta_ba|| * delta_t).
+  // [m/s]
+  double reintegrate_vel_norm_thres = 1e-4;
+};
+
+class ImuReintegrationCallback : public ceres::IterationCallback {
```
ahojnnes, line 247: Should this go into another file (e.g., imu_reintragration.h) or follow-up PR?

src/colmap/estimators/imu_preintegration.cc
```diff
@@ -0,0 +1,589 @@
...
+
+void PreintegratedImuData::Finalize(double max_condition_number) {
+  THROW_CHECK(max_condition_number > 0.0 || max_condition_number == -1.0)
+      << "max_condition_number must be positive or -1 (disabled)";
+
+  // Enforce symmetry.
+  covariance = (covariance + covariance.transpose()) / 2.0;
+
+  // Eigendecomposition for robust sqrt-information.
+  Eigen::SelfAdjointEigenSolver<Eigen::Matrix<double, 15, 15>> saes(covariance);
```
ahojnnes: const

src/colmap/estimators/imu_preintegration.cc
```diff
@@ -0,0 +1,589 @@
...
+      options_(options),
+      calib_(calib),
+      accel_rect_mat_inv_(calib.accel_rectification.inverse()),
+      gyro_rect_mat_inv_(calib.gyro_rectification.inverse()) {
+  THROW_CHECK_LT(t_start, t_end);
+  data_.gravity_magnitude = calib.gravity_magnitude;
+  Reset();
+}
+
+void ImuPreintegrator::Reset() {
```
ahojnnes, line 96: Maybe instead just do `data_ = ImuPreintegrationData()` ?

src/colmap/estimators/imu_preintegration.cc
```diff
@@ -0,0 +1,589 @@
...
+  // Left convention midpoint (trapezoidal) integration.
+  // Based on Forster et al. "On-Manifold Preintegration for Real-Time
+  // Visual-Inertial Odometry", TRO 2016. The original paper uses right-multiply
+  // integration; here we use left-multiply integration with right-perturbation
+  // bias Jacobians. delta_R^T rotates accel from body_k to body_i frame.
+
+  // Integration step.
+  // dR = Exp(-omega * dt): body_from_world evolves as R_BW_{k+1} = Exp(-w*dt) *
+  // R_BW_k.
+  Eigen::Quaterniond dq = QuaternionFromAngleAxis(-gyro_true * dt);
```
ahojnnes: nit: some of the variables here can benefit from const's

src/colmap/estimators/imu_preintegration.cc
```diff
@@ -0,0 +1,589 @@
...
+  Eigen::Matrix3d Jl = LeftJacobianFromAngleAxis(gyro_true * dt);
+  Eigen::Matrix3d dR_dbg_updated = data_.dR_dbg + Rs_T * Jl * dt;
+  A.block<3, 3>(0, 9) = dR_dbg_updated - data_.dR_dbg;
+  data_.dR_dbg = dR_dbg_updated;
+
+  // propagate
+  data_.covariance = A * data_.covariance * A.transpose();
+
+  // Step 2: add noise.
+  double vars_v = pow(accel_noise_density, 2) * dt;
```
ahojnnes: pow() is slow, replace with a * a.

src/colmap/estimators/imu_preintegration.cc
```diff
@@ -0,0 +1,589 @@
...
+  double vars_omega = pow(gyro_noise_density, 2) * dt;
+  double vars_p =
+      0.5 * vars_v * dt * dt + pow(options_.integration_noise_density, 2) * dt;
+  double vars_ba = pow(calib_.bias_accel_random_walk_sigma, 2) * dt;
+  double vars_bg = pow(calib_.bias_gyro_random_walk_sigma, 2) * dt;
+  data_.covariance.block<3, 3>(0, 0) +=
+      Eigen::Matrix3d::Identity() * vars_omega;
```
ahojnnes: I think you can do `.diagonal() += vars_omega`. Same for below.

src/colmap/estimators/imu_preintegration.cc
```diff
@@ -0,0 +1,589 @@
...
+    f_1 = (w_dt * cos_wt - sin_wt) / (mag_w2 * mag_w);
+    f_2 = (w_dt * w_dt - 2 * cos_wt - 2 * w_dt * sin_wt + 2) /
+          (2 * mag_w2 * mag_w2);
+    f_3 = -(1 - cos_wt) / mag_w2;
+    f_4 = (w_dt - sin_wt) / (mag_w2 * mag_w);
+  }
+
+  // Integration matrices for position (H_p) and velocity (H_v).
+  // Closed-form integral coefficients (same signs as Eckenhoff IJRR 2018).
+  const Eigen::Matrix3d alpha_arg = (dt2 / 2.0) * I3 + f_1 * w_x + f_2 * w_x_2;
```
ahojnnes: nit: `0.5 * dt2` is more efficient.

src/colmap/estimators/imu_preintegration.cc
```diff
@@ -0,0 +1,589 @@
...
+  // Based on Forster et al. "On-Manifold Preintegration for Real-Time
+  // Visual-Inertial Odometry", TRO 2016. The original paper uses right-multiply
+  // integration; here we use left-multiply integration with right-perturbation
+  // bias Jacobians. delta_R^T rotates accel from body_k to body_i frame.
+
+  // Integration step.
+  // dR = Exp(-omega * dt): body_from_world evolves as R_BW_{k+1} = Exp(-w*dt) *
+  // R_BW_k.
+  Eigen::Quaterniond dq = QuaternionFromAngleAxis(-gyro_true * dt);
+  Eigen::Matrix3d Rs = data_.delta_R.toRotationMatrix();
```
ahojnnes: Should we store delta_R as a rotation matrix? It seems like we convert back and forth without ever needing the quaternion during integration?

src/colmap/estimators/imu_preintegration.cc
```diff
@@ -0,0 +1,589 @@
...
+                                    double dt,
+                                    double accel_noise_density,
+                                    double gyro_noise_density) {
+  // Left convention with closed-form rotation integrals, analytical bias
+  // Jacobians, and RK4 covariance propagation.
+  //
+  // Based on Eckenhoff et al. "Closed-form Preintegration Methods for
+  // Graph-based Visual-Inertial Navigation", IJRR 2018 (left convention).
+
+  const Eigen::Matrix3d I3 = Eigen::Matrix3d::Identity();
```
ahojnnes: Can be a static constexpr in anonymous namespace?

src/colmap/estimators/imu_preintegration.cc
```diff
@@ -0,0 +1,589 @@
...
+  // The Eckenhoff d_R_bw formula expects the multiplicative-transport J_q
+  // (≈ Jr(w*T)*T), which is the transpose of our data_.dR_dbg (≈ Jl(w*T)*T).
+  const Eigen::Matrix3d J_q = data_.dR_dbg.transpose();
+
+  const Eigen::Vector3d e_1(1, 0, 0);
+  const Eigen::Vector3d e_2(0, 1, 0);
+  const Eigen::Vector3d e_3(0, 0, 1);
+  const Eigen::Matrix3d e_1x = CrossProductMatrix(e_1);
+  const Eigen::Matrix3d e_2x = CrossProductMatrix(e_2);
+  const Eigen::Matrix3d e_3x = CrossProductMatrix(e_3);
```
ahojnnes: Could statically precompute these.

src/colmap/estimators/imu_preintegration.cc
```diff
@@ -0,0 +1,589 @@
...
+    }
+    for (int k = 0; k < 3; ++k) {
+      for (int j = 0; j < 4; ++j) {
+        df_dw[j][k] = gyro_true(k) * g[j];
+      }
+    }
+  }
+
+  const Eigen::Matrix3d* d_R_bw[3] = {&d_R_bw_1, &d_R_bw_2, &d_R_bw_3};
+  const Eigen::Matrix3d* e_kx[3] = {&e_1x, &e_2x, &e_3x};
```
ahojnnes: Alternatively could store these as directly in an std::array above instead of the extra pointer indirection.

src/colmap/estimators/imu_preintegration.cc
```diff
@@ -0,0 +1,589 @@
...
+  }
+
+  //==========================================================================
+  // Step 3: Covariance propagation (RK4).
+  //==========================================================================
+
+  // Continuous-time noise matrix Q_c (12x12):
+  // [gyro_noise(3), gyro_walk(3), accel_noise(3), accel_walk(3)].
+  Eigen::Matrix<double, 12, 12> Q_c = Eigen::Matrix<double, 12, 12>::Zero();
+  Q_c.block<3, 3>(0, 0) = I3 * (gyro_noise_density * gyro_noise_density);
```
ahojnnes: Can use .diagonal() here as well?

src/colmap/estimators/imu_preintegration.cc
```diff
@@ -0,0 +1,589 @@
...
+                    "t_end. Ignore the current measurement.";
+    return;
+  }
+
+  // Append measurements
+  measurements_.Insert(m);
+  IntegrateOneMeasurement(last_measurement, m);
+}
+
+void ImuPreintegrator::FeedImu(const ImuMeasurements& ms) {
```
ahojnnes: nit: Integrate() ?

ahojnnes (commented):
Thanks, looks great already. Left an initial round of comments.

src/colmap/estimators/imu_preintegration.h
B1ueber2y: This makes great sense to me. I have added the abstraction and changed to parameterized tests.

src/colmap/sensor/imu.h
B1ueber2y: I can see its usability in the stream setup. I would vote for keeping it. I updated the other suboptimal insertion methods.

src/colmap/estimators/imu_preintegration.h
B1ueber2y: State is already taken by ImuState and it means something different. After preintegration it s not really a state anymore, so I would prefer the original one "data" out of the two.

src/colmap/estimators/imu_preintegration.cc
B1ueber2y: quaternion is used in the cost functions and bias multiplication, where we need the performance the most. So keeping quaternion over matrix makes sense to me.

src/colmap/sensor/imu.h
B1ueber2y: These two values are always available for every Imu measurement, so is never optional as in the case of pose prior. So I personally think having the default as NaN is not as good as zero.

src/colmap/estimators/imu_preintegration.h
B1ueber2y: It belongs to this PR because we use it in the python example. And I dont think its worth splitting to another file. 

To be clear on what this was trying to achieve, in the imu preintegration convention (e.g., in VI.C of Forster et al. https://rpg.ifi.uzh.ch/docs/TRO16_forster.pdf), the bias was handled as constant with a linear jacobian approximation, which is oftentimes accurate enough. However, if the bias drifts too much its common practice to reintegrate to get an updated linearized jacobian, which is in practice quite crucial. Here we achieve the same behavior with a ceres callback.

src/colmap/scene/imu.h
B1ueber2y: This PR focuses on discrete-time VI optimization with IMU preintegration, which is a cost level update for post optimization that is not coupled into the sfm pipeline. To achieve this we do not need to necessarily benefit from the rig abstraction for now. 

Integrating it with the rig abstraction requires some additional efforts on updating data structure in the reconstruction and incremental pipeline to include ImuMeasurement and ImuState (still need to think of how to make it memory efficient), which would make this already gigantic PR even harder to review. So I suggest we keep this TODO out of the scope of this PR.

B1ueber2y:
Thanks for the detailed review! All comments addressed for this round : )

PR #2629, opened 2024-07-02
fix: Add check for data existence when extracting data from CorrespondenceGraph

ChenYaoR:
The `CorrespondenceGraph` functions `NumObservationsForImage` and `NumCorrespondencesForImage` do not verify if the image ID exists in the correspondence image map, potentially leading to errors.

### DIFF
--- a/src/colmap/sfm/observation_manager.cc
+++ b/src/colmap/sfm/observation_manager.cc
@@ -78,7 +78,7 @@ ObservationManager::ObservationManager(
         kNumPoint3DVisibilityPyramidLevels, camera.width, camera.height);
     image_stat.num_correspondences_have_point3D.resize(image.NumPoints2D(), 0);
     image_stat.num_visible_points3D = 0;
-    if (correspondence_graph_) {
+    if (correspondence_graph_ && correspondence_graph_->ExistsImage(id_image.first)) {
       image_stat.num_observations =
           correspondence_graph_->NumObservationsForImage(id_image.first);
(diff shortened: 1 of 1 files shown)


ahojnnes:
Could you please explain under which circumstances this causes a problem? The image should always exist, otherwise this hints at another problem? Thank you.

ChenYaoR:
> Could you please explain under which circumstances this causes a problem? The image should always exist, otherwise this hints at another problem? Thank you.

@ahojnnes Hi, I encountered this problem while reproducing the [3dgs paper](https://github.com/zju3dv/street_gaussians). When processing their prepared data using **colmap point_triangulator**, there is an obstacle that disappears for a short period of time, resulting in it not having a corresponding match image. This leads to an abnormal termination. Below are screenshots of the problem I encountered:

no match img:
no match

exception throw img:
exception

ahojnnes:
That sounds like the provided database is inconsistent with the reconstruction. In other words, the provided data is inconsistent. I suggest to fix the problem on the data side instead of silently ignoring the issue in the code. It's actually useful if this crashes, because it will otherwise hide issues with inconsistent input data.

ChenYaoR:
> That sounds like the provided database is inconsistent with the reconstruction. In other words, the provided data is inconsistent. I suggest to fix the problem on the data side instead of silently ignoring the issue in the code. It's actually useful if this crashes, because it will otherwise hide issues with inconsistent input data.

Indeed, thank you for the explanation

PR #2633, opened 2024-07-02
Faster covariance computation for small blocks

B1ueber2y:
This PR includes faster covariance computation for small blocks, as a follow up for https://github.com/colmap/colmap/pull/2610. This is achieved by storing the inversion of the triangular matrix L from sparse Cholesky rather than storing the full covariance matrix. 

This brings ~5x constant speed up on the computation of per-image pose covariance while the asymptotic complexity is still cubic (as in the table below, the sparse cholesky operation is already cubic).

Below are some benchmark timing results for per-pose covariance estimation. Control points were used to fix the Gauge ambiguity.

| # images   | # pose parameters | sparse cholesky (in both methods)  | full time before from https://github.com/colmap/colmap/pull/2610 | full time in this PR | 
| -------- | -------- | -------- | -------- | -------- |
| 400 | 2400 | 1.9s | 13.3s | 3.8s |
| 600 | 3600 | 6.9s | 52.9s | 13.4s |
| 800 | 4800 | 16.9s | 147.8s | 31.9s |
| 1200 | 7200 | 56.6s | 487.4s | 105.7s |
| 1800 | 10800 | 189.1s | 1644.7s | 374.6s |

The asymptotic complexity is O(n^3) with a very stable constant so one can estimate the time needed for larger reconstruction using the numbers from this table. The time may vary a bit depending on the level of connection (shared 3D points across images) in the reconstruction. Same unit tests from https://github.com/colmap/colmap/pull/2610 were done here to verify the correctness.

To be able to use this feature in Python one just needs to substitute ``estimator.compute()`` into ``estimator.factorize()`` for the colmap estimator ``pycolmap.BundleAdjustmentCovarianceEstimator``. Note that the method ``Compute()`` is still preferred if one wants the full covariance matrix (``n_parameters`` by ``n_parameters`` ).

### DIFF
--- a/src/colmap/estimators/covariance.cc
+++ b/src/colmap/estimators/covariance.cc
@@ -585,6 +585,8 @@ bool BundleAdjustmentCovarianceEstimator::FactorizeFull() {
   Eigen::MatrixXd L_dense = L;
   L_matrix_variables_inv_ = L_dense.triangularView<Eigen::Lower>().solve(
       Eigen::MatrixXd::Identity(L_dense.rows(), L_dense.cols()));
+  LOG(INFO) << "Finish factorization by having the lower triangular matrix L "
+               "inverted.";
   return true;
 }
 
@@ -677,6 +679,8 @@ bool BundleAdjustmentCovarianceEstimator::Factorize() {
   Eigen::MatrixXd L_poses_dense = L_poses;
   L_matrix_poses_inv_ = L_poses_dense.triangularView<Eigen::Lower>().solve(
       Eigen::MatrixXd::Identity(L_poses_dense.rows(), L_poses_dense.cols()));
+  LOG(INFO) << "Finish factorization by having the lower triangular matrix L "
+               "inverted.";
   return true;
 }
 
--- a/src/colmap/estimators/covariance.cc
+++ b/src/colmap/estimators/covariance.cc
@@ -218,6 +218,39 @@ bool BundleAdjustmentCovarianceEstimatorBase::HasValidFullCovariance() const {
   return cov_variables_.size() != 0;
 }
 
+double BundleAdjustmentCovarianceEstimatorBase::GetPoseCovarianceByIndex(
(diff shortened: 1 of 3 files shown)


src/colmap/estimators/covariance.cc
```diff
@@ -452,6 +479,115 @@ bool BundleAdjustmentCovarianceEstimator::HasValidSchurComplement() const {
...
+
+bool BundleAdjustmentCovarianceEstimator::FactorizeFull() {
+  if (!HasValidSchurComplement()) {
+    ComputeSchurComplement();
+  }
+  LOG(INFO) << StringPrintf(
+      "Inverting Schur complement for all variables except for 3D points (n = "
+      "%d)",
+      num_params_poses_ + num_params_other_variables_);
+  LOG(INFO) << StringPrintf("Start sparse Choleskey decomposition (n = %d)",
```
ahojnnes, line 561:
```suggestion
  LOG(INFO) << StringPrintf("Start sparse Cholesky decomposition (n = %d)",
```

src/colmap/estimators/covariance.cc
```diff
@@ -460,6 +596,8 @@ bool BundleAdjustmentCovarianceEstimator::ComputeFull() {
       "Inverting Schur complement for all variables except for 3D points (n = "
       "%d)",
       num_params_poses_ + num_params_other_variables_);
+  LOG(INFO) << StringPrintf("Start sparse Choleskey decomposition (n = %d)",
```
ahojnnes, line 599:
```suggestion
  LOG(INFO) << StringPrintf("Start sparse Cholesky decomposition (n = %d)",
```

ahojnnes (approved):
Looks good to me, (on a high level) thank you. Maybe let Paul also have a look, since I didn't really have time to review the math thoroughly.

ahojnnes (approved):

Issue #115, opened 2017-05-03
error on running the colmap functions


dineshreddy91:
successfully compiled but keep getting the above error

The output of cmake

-- Boost version: 1.64.0
-- Found the following Boost libraries:
--   program_options
--   filesystem
--   regex
--   system
--   unit_test_framework
-- Found Eigen3
--   Includes : /usr/include/eigen3
-- Found FreeImage
--   Includes : /usr/include
--   Libraries : /usr/lib/libfreeimage.so
-- Found Glog
--   Includes : /usr/include
--   Libraries : /usr/lib/x86_64-linux-gnu/libglog.so
-- Found required Ceres dependency: Eigen version 3.2.0 in /usr/include/eigen3
-- Found required Ceres dependency: Glog in /usr/include
-- Found Ceres version: 1.11.0 installed in: /usr/local
-- Found Glew
--   Includes : /usr/include
--   Libraries : /usr/lib/x86_64-linux-gnu/libGLEW.so
-- Found Qt 5.2.1
-- Build type not specified, using RelWithDebInfo
-- Enabling OpenMP support
-- Disabling link-time optimization (-flto)
-- Enabling CUDA support (Version 8.0)
-- Disabling profiling support
-- Configuring done
-- Generating done

QOpenGLFunctions created with non-current context 
*** Aborted at 1493778755 (unix time) try "date -d @1493778755" if you are using GNU date ***
PC: @     0x7f82888aafa1 (unknown)
*** SIGSEGV (@0x8) received by PID 48425 (TID 0x7f828b15a800) from PID 8; stack trace: ***
    @     0x7f828a606330 (unknown)
    @     0x7f82888aafa1 (unknown)
    @     0x7f8288ab525a (unknown)
    @     0x7f8288ab8493 (unknown)
    @           0x5d4c1f (unknown)
    @           0x599350 (unknown)
    @           0x599798 (unknown)
    @           0x59eb8c (unknown)
    @           0x58a5ec (unknown)
    @           0x591ce8 (unknown)
    @           0x57f9e9 (unknown)
    @     0x7f8286cc2f45 (unknown)
    @           0x587537 (unknown)
    @                0x0 (unknown)
Segmentation fault (core dumped)

ahojnnes:
Are you running the GUI without an attached display?

ahojnnes:
Closing this for now due to lack of feedback. Feel free to reopen if you still have problems.

Issue #113, opened 2017-04-27
Is it possible to get the odometry result from the reconstruction result of a sequence of images


xiaofeiwu:
I reconstructed a 3D model for a sequence of images such as Kitti Mono. The reconstruction looks nice. However, when I tried to get the odometry of the cameras from the tvec of the images, it is totally wrong. Could anyone give me some hint why?

ahojnnes:
tvec defines the translation from world to camera coordinate system. To obtain the camera trajectory you need to compute ``R^T * T`` while `R` is the 3x3 rotation matrix composed from the `qvec` and `T` is the `tvec`.

Issue #112, opened 2017-04-27
Blending map


tmac1002:
Hi, I used some UAV images to generate sparse point. However, the result map seemed blending. I am very confused. I just have no idea why this happened. Could you help me explain this situation or what exactly reasons caused this results. Thank you very much!
![blend_image](https://cloud.githubusercontent.com/assets/11924213/25466076/c09953ba-2b38-11e7-9469-f3ee65cf524b.JPG)

ahojnnes:
I assume you do not have a calibration for your images? This problem is, for example, described in "Critical Configurations for Radial Distortion Self-Calibration" by Wu C.

You should be able to solve the problem by fixing the intrinsic parameters to values obtained from an offline calibration.

Issue #111, opened 2017-04-25
Error "403 Forbidden" for Mac pre-built binary


konstantinkobs:
When I want to download the pre-built development version of COLMAP on [https://demuc.de/colmap/](https://demuc.de/colmap/) I only get a "403 Forbidden" page. Are there no pre-built binaries yet or is this a server issue?

ahojnnes:
Thanks, this should be fixed now.

PR #2660, opened 2024-07-17
Pose Prior based Incremental Mapper

ferreram:
This PR adds the possibility to apply the Incremental Mapper pipeline while using position priors taken from the database.  The prior position can be either in geographic coordinates or in cartesian coordinates.  The position priors are used during the global BA only.  For now, this prior based pipeline is only accessible through command line.   A unit test has been added.

More is to come (add python bindings, prior based BA controller, possibility to setup priors from a file instead of through the database, python scripts to populate the database with priors, etc...)  but I'd like to ensure that you are already okay with the current state of this PR before diving further.


src/colmap/controllers/incremental_mapper.h
```diff
@@ -114,6 +114,24 @@ struct IncrementalMapperOptions {
...
+  // Whether to use prior camera positions
+  bool use_prior_position = false;
+
+  // Whether to setup the prior positions from database
+  bool set_prior_position_from_database = true;
+
+  // Standard deviation on the position priors
+  double prior_position_std_x = 1.;
+  double prior_position_std_y = 1.;
+  double prior_position_std_z = 1.;
```
sarlinpe, line 126: Would it make sense to extend `PosePrior` with a full covariance such has each measurement can have a different noise?

src/colmap/controllers/incremental_mapper.cc
```diff
@@ -248,6 +262,78 @@ bool IncrementalMapperController::LoadDatabase() {
...
+  for (const auto& image : database_cache_->Images()) {
+    if (image.second.HasPosePrior()) {
+      ++nb_prior_positions;
+      if (image.second.WorldFromCamPrior().coordinate_system !=
+          PosePrior::CoordinateSystem::WGS84) {
+        prior_is_gps = false;
+        // break;
+      } else {
+        v_gps_prior.push_back(image.second.WorldFromCamPrior().position);
+        // Image with the lowest id is to be used as the origin for prior
```
sarlinpe, line 296: To minimize the approximation error of ENU, you might want to use the average position instead of an arbitrary one.

src/colmap/scene/image.h
```diff
@@ -83,6 +84,12 @@ class Image {
   inline const Rigid3d& CamFromWorld() const;
   inline Rigid3d& CamFromWorld();
 
+  // Camera to world pose prior.
+  inline const PosePrior& WorldFromCamPrior() const;
+  inline PosePrior& WorldFromCamPrior();
+  // Check if PosePrior is valid
+  inline bool HasPosePrior() const;
```
sarlinpe, line 91: Instead of adding a new field to `Image`, how about extending `DatabaseCache` with pose priors? `SetupPriorPoseFromDatabase` could then be part of `DatabaseCache` (update all priors in-place).

src/colmap/estimators/bundle_adjustment.cc
```diff
@@ -810,6 +813,257 @@ void RigBundleAdjuster::ParameterizeCameraRigs(Reconstruction* reconstruction) {
...
+
+    if (prior_options_.use_prior_position && image.HasPosePrior()) {
+      problem_->AddResidualBlock(PositionPriorErrorCostFunction::Create(
+                                     image.WorldFromCamPrior().position,
+                                     prior_options_.prior_position_covariance),
+                                 prior_loss_function,
+                                 cam_from_world_rotation,
+                                 cam_from_world_translation);
+    }
+  }
```
sarlinpe, line 1064: Only this section is different from `BundleAdjuster::AddImageToProblem`, could we reuse the code of this this parent function by calling it first?

src/colmap/estimators/bundle_adjustment.cc
```diff
@@ -810,6 +813,257 @@ void RigBundleAdjuster::ParameterizeCameraRigs(Reconstruction* reconstruction) {
...
+  for (const auto point3D_id : config_.VariablePoints()) {
+    AddPointToProblem(point3D_id, reconstruction, loss_function);
+  }
+  for (const auto point3D_id : config_.ConstantPoints()) {
+    AddPointToProblem(point3D_id, reconstruction, loss_function);
+  }
+
+  ParameterizeCameras(reconstruction);
+  ParameterizePoints(reconstruction);
+}
```
sarlinpe, line 988: Mostly identical to `BundleAdjuster::SetUpProblem`, can we reuse it?

src/colmap/estimators/bundle_adjustment.h
```diff
@@ -270,6 +282,46 @@ class RigBundleAdjuster : public BundleAdjuster {
...
+                         ceres::LossFunction* prior_loss_function);
+
+  Options prior_options_;
+
+  std::unique_ptr<ceres::LossFunction> prior_loss_function_;
+
+  // Rotation from camera frame to world frame (i.e. prior pos frame)
+  // to be used to allow the alignment w.r.t. to the position priors
+  // through a common rotation even when some poses are set as const
+  Eigen::Vector4d qwc0_ = Eigen::Vector4d(1.0, 0.0, 0.0, 0.0);
```
sarlinpe, line 322: This seems to not be used anywhere?

sarlinpe (commented):
Thanks a lot for this contribution and sorry that it took me so long to look at it. This is very exciting! Did you test this on real data and could you share a sample of such data?

ferreram:
@sarlinpe thanks for your reply!  I have a sample dataset of real data here acquired with a UAV with photos having GPS in the EXIF metadata here: https://filesender.renater.fr/?s=download&token=0418e1d0-ed8d-4bd4-89ec-e9f0b7772fe7.

src/colmap/estimators/bundle_adjustment.h
ferreram: You are right.  This part is not used right now as prior positions are only used in the global BA with no constant pose.  I just removed it.

src/colmap/estimators/bundle_adjustment.cc
ferreram: I have just improved the code to better re-use functions from the BundleAdjuster class.

src/colmap/estimators/bundle_adjustment.cc
ferreram: Done.  We are now calling a AddPosePriorToProblem function after a call to setUpProblem.

src/colmap/controllers/incremental_mapper.h
ferreram: Good idea.  I will take care of it.

src/colmap/controllers/incremental_mapper.cc
ferreram: I agree that using the average position could be slightly more accurate.  However, using the prior of the image with the lowest id in the database as the reference allows users to easily know which GPS measurement has been used as reference.  This might be valuable to users that need to convert back the estimated position into GPS (lat/lon) data.

ferreram: Moreover, the model_aligner function already uses the prior of the image with the lowest id in the database as the reference.  So doing the same here allows to have coherent results between both features.

src/colmap/controllers/incremental_mapper.cc
```diff
@@ -253,6 +267,86 @@ bool IncrementalPipeline::LoadDatabase() {
...
+  double ref_lat = std::numeric_limits<double>::max();
+  double ref_lon = std::numeric_limits<double>::max();
+
+  // Get ordered image_ids to use image with lowest ID as the GPS ref.
+  std::set<image_t> image_ids_with_prior;
+  for (const auto& image : database_cache_->Images()) {
+    if (image.second.HasPosePrior()) {
+      image_ids_with_prior.insert(image.first);
+    }
+  }
```
sarlinpe, line 298: Images in the `DatabaseCache` are not ordered so you need to explicitly sort them as in https://github.com/colmap/colmap/blob/bb6a67990fd069baddac56d3e4604ee10d658b29/src/colmap/feature/matcher.cc#L127-L130

ferreram: Yes that's why I have used a set instead of a vector here, so that the image_ids are automatically sorted.  Would you prefer that I use a vector and sort it afterward?

src/colmap/scene/database.cc
```diff
@@ -441,6 +449,12 @@ std::vector<Image> Database::ReadAllImages() const {
 
   SQLITE3_CALL(sqlite3_reset(sql_stmt_read_images_));
 
+  for (auto& image : images) {
```
ahojnnes, line 452: nit: Alternatively, this could be more efficiently done by implementing a ReadAllPosePriors() method?

src/colmap/estimators/cost_functions.h
```diff
@@ -493,6 +493,43 @@ struct Point3dAlignmentCostFunction {
...
+  }
+
+  template <typename T>
+  bool operator()(const T* const cam_from_world_q,
+                  const T* const cam_from_world_t,
+                  T* residuals_ptr) const {
+    Eigen::Map<Eigen::Matrix<T, 3, 1>> residuals(residuals_ptr);
+    residuals = (EigenQuaternionMap<T>(cam_from_world_q).inverse() *
+                 -EigenVector3Map<T>(cam_from_world_t)) -
+                world_from_cam_position_prior_.cast<T>();
```
ahojnnes, line 521: nit: the following could save a few instructions?

```suggestion
    residuals = world_from_cam_position_prior_.cast<T>() + (EigenQuaternionMap<T>(cam_from_world_q).inverse() *
                 EigenVector3Map<T>(cam_from_world_t));
```

src/colmap/controllers/incremental_mapper.h
ahojnnes: Yes, +1

src/colmap/exe/colmap.cc
```diff
@@ -128,6 +128,8 @@ int main(int argc, char** argv) {
   commands.emplace_back("patch_match_stereo", &colmap::RunPatchMatchStereo);
   commands.emplace_back("point_filtering", &colmap::RunPointFiltering);
   commands.emplace_back("point_triangulator", &colmap::RunPointTriangulator);
+  commands.emplace_back("position_prior_mapper",
```
ahojnnes, line 131: I suggest calling this `pose_prior_mapper`, so we can generalize it to also consider rotations later without renaming it.

src/colmap/estimators/bundle_adjustment.cc
```diff
@@ -825,6 +828,189 @@ void RigBundleAdjuster::ParameterizeCameraRigs(Reconstruction* reconstruction) {
...
+  // Fix 7-DOFs of BA problem for classic BA
+  if (!prior_options_.use_prior_position) {
+    const std::vector<image_t>& reg_image_ids = reconstruction->RegImageIds();
+    config_.SetConstantCamPose(reg_image_ids[0]);
+    config_.SetConstantCamPositions(reg_image_ids[1], {0});
+  }
+
+  // Apply RANSAC-based Sim3 Alignment
+  if (prior_options_.use_prior_position) {
+    LOG(INFO) << "Initial alignment error w.r.t. prior position:";
```
ahojnnes, line 882: nit: Shall we change these LOG(INFO) to VLOG(2) here and below in the same scope?

src/colmap/estimators/bundle_adjustment.cc
```diff
@@ -825,6 +828,189 @@ void RigBundleAdjuster::ParameterizeCameraRigs(Reconstruction* reconstruction) {
...
+      LOG(INFO) << "  - median: " << std::sqrt(Median(verr2_wrt_prior)) << " m";
+
+    } else {
+      LOG(WARNING) << "Sim3 alignment w.r.t. prior position failed!";
+    }
+  }
+
+  // Put the centroid of the reconstruction to (0.,0.,0.)
+  // to avoid any numerical instability
+  Sim3d sim_to_center;
```
ahojnnes, line 928: Reconstruction::Normalize() could achieve the same but maybe a bit more robust in case of outliers. What do you think about changing this method to the following interface:
```
  // Normalize scene by scaling and translation to avoid degenerate
  // visualization after bundle adjustment and to improve numerical
  // stability of algorithms.
  //
  // Translates scene such that the mean of the camera centers or point
  // locations are at the origin of the coordinate system.
  //
  // Scales scene such that the minimum and maximum camera centers are at the
  // given `extent`, whereas `p0` and `p1` determine the minimum and
  // maximum percentiles of the camera centers considered.
  Sim3d Normalize(bool fixed_scale = true,
                  double extent = 10.0,
                  double p0 = 0.1,
                  double p1 = 0.9,
                  bool use_images = true);
```

ahojnnes:
On the high level, this looks great to me, thank you for your efforts to implement this feature and the diligence on implementing tests, etc. I left a few suggestions as comments.

src/colmap/exe/colmap.cc
sarlinpe: +1

src/colmap/estimators/bundle_adjustment.cc
sarlinpe: +1

src/colmap/controllers/incremental_mapper.cc
sarlinpe: Sorry, I had missed this. It's probably fine.

src/colmap/scene/image.h
sarlinpe: I don't think so. I suggest that `DatabaseCache` exposes an `std::unordered_map<image_t, PosePrior>`, which the `IncrementalMapper` can forward to `PositionPriorBundleAdjuster::SetUpProblem`. This would also greatly simplify `SetupPriorPoseFromDatabase`.

src/colmap/estimators/bundle_adjustment.cc
sarlinpe: I think that `AddPosePriorToProblem` should be called from `SetUpProblem`. You could simply call the parent `BundleAdjuster::SetUpProblem` at the beginning of `PositionPriorBundleAdjuster::SetUpProblem`.

src/colmap/estimators/bundle_adjustment.cc
```diff
@@ -825,6 +828,189 @@ void RigBundleAdjuster::ParameterizeCameraRigs(Reconstruction* reconstruction) {
...
+bool PositionPriorBundleAdjuster::Solve(Reconstruction* reconstruction) {
+  loss_function_ =
+      std::unique_ptr<ceres::LossFunction>(options_.CreateLossFunction());
+
+  if (options_.use_robust_loss_on_prior_position) {
+    prior_loss_function_ =
+        std::make_unique<ceres::CauchyLoss>(options_.prior_position_loss_scale);
+  }
+
+  // Compute initial squared error between position priors & current images
```
sarlinpe, line 852: Instead of aligning again before each BA, wouldn't it be easier to somewhere store the `Sim3d` transformation between normalized and metric reconstructions? @ahojnnes @B1ueber2y as discussed previously, this could be stored in the `Reconstruction` object. Something like a Gage object:
```C++
struct Gage {
  std::optional<double> scale;  // metric scale of the reconstruction
  std::optional<Eigen::Vector3d> gravity;  // [0,0,-g] in the world coordinate frame
  std::optional<double> heading; // w.r.t East or North axis
  std::optional<Eigen::Vector3d> origin; // in ECEF or WGS84
}
```
This will be useful for IMU optimization too and would simplify managing the gage freedom of the BA (e.g. whether to fix the first pose or not).

src/colmap/controllers/incremental_mapper.h
ferreram: :+1:

src/colmap/estimators/bundle_adjustment.cc
ferreram: :+1: I have also added a few more factorization which makes the PosePriorBundleAdjuster looks great IMO.

src/colmap/scene/database.cc
ferreram: We now access separately the pose priors in the database :+1:

src/colmap/estimators/cost_functions.h
ferreram: Great idea! Done :+1:

src/colmap/exe/colmap.cc
ferreram: Done :+1:

src/colmap/estimators/bundle_adjustment.cc
ferreram: Done :+1:

src/colmap/estimators/bundle_adjustment.cc
ferreram: Indeed something like this would work!  I have modified the PosePriorBundleAdjuster to hold the current sim3d transformation that has been applied to the reconstruction.  This actually led to a cleaner and more efficient code as the position priors are now transformed only when creating the cost function.  I think it would not be much work to set this sim3d object to the result of Normalize().  I am leaving it as it is for the moment due to @sarlinpe idea about storing a "gauge" object that would hold the transformation between the normalized and metric reconstruction.

src/colmap/estimators/bundle_adjustment.h
```diff
@@ -287,6 +299,51 @@ class RigBundleAdjuster : public BundleAdjuster {
...
+  struct Options {
+    // Whether to use prior camera positions
+    bool use_prior_position = true;
+
+    // Maximum RANSAC error for Sim3D alignment
+    double ransac_max_error = 0.;
+
+    // Sim3d transformation that project reconstruction's centroid to (0.,0.,0.)
+    Sim3d sim_to_center;
+  };
```
sarlinpe, line 313: Unlike `RigBundleAdjuster::Options`, this looks like an internal state of the object - is this correct? then 1) should this really be named `Options`? 2) shouldn't the definition be private since there is no public getter/setter for `prior_options_`?

src/colmap/estimators/bundle_adjustment.h
```diff
@@ -86,6 +86,19 @@ struct BundleAdjustmentOptions {
   int max_num_images_direct_dense_gpu_solver = 200;
   int max_num_images_direct_sparse_gpu_solver = 4000;
 
+  // Whether to use prior camera positions
```
sarlinpe, line 89: These options are specific to the `PosePriorBundleAdjuster`. Shouldn't they be part of `PosePriorBundleAdjuster::Options` instead? would this make propagating them from the CLI through the parent call stack more difficult?

src/colmap/estimators/bundle_adjustment.h
```diff
@@ -287,6 +299,51 @@ class RigBundleAdjuster : public BundleAdjuster {
...
+ public:
+  struct Options {
+    // Whether to use prior camera positions
+    bool use_prior_position = true;
+
+    // Maximum RANSAC error for Sim3D alignment
+    double ransac_max_error = 0.;
+
+    // Sim3d transformation that project reconstruction's centroid to (0.,0.,0.)
+    Sim3d sim_to_center;
```
sarlinpe, line 312: I suggest calling this `normalized_from_metric`

src/colmap/controllers/incremental_mapper.h
sarlinpe: Thank you!

src/colmap/controllers/incremental_mapper.cc
sarlinpe: This makes sense, thank you.

src/colmap/scene/image.h
sarlinpe: This looks great, thank you!

src/colmap/estimators/bundle_adjustment.cc
sarlinpe: I suggest relying on `Reconstruction::Normalize` for now. I can look into adding the gauge object later - it requires changes beyond the scope of this PR.

src/colmap/scene/database.cc
```diff
@@ -451,8 +451,10 @@ PosePrior Database::ReadPosePrior(const image_t image_id) const {
   if (rc == SQLITE_ROW) {
     prior.position =
         ReadStaticMatrixBlob<Eigen::Vector3d>(sql_stmt_read_pose_prior_, rc, 1);
+    prior.position_covariance =
+        ReadStaticMatrixBlob<Eigen::Matrix3d>(sql_stmt_read_pose_prior_, rc, 2);
     prior.coordinate_system = static_cast<PosePrior::CoordinateSystem>(
-        sqlite3_column_int64(sql_stmt_read_pose_prior_, 2));
+        sqlite3_column_int64(sql_stmt_read_pose_prior_, 3));
```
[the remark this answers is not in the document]
ferreram: Unfortunately it is not so reading a database with a different data structure for its pose_priors table will lead to a SQL logic error.  Any idea on how to make it backward compatible?  Otherwise we could probably update the database migration python scripts.

HernandoR:
it is a fantastic feature. But i'm a little confusing now about the PosePrior class in this repo. As to my understanding, this class can only hold the *Position prior* that is only the Translation Vector prior known is used. and (do point out if i'm wrong) i understand your initial idea is to use the full prior pose, so where the Quaternion is hiding?

src/colmap/estimators/bundle_adjustment.h
```diff
@@ -86,6 +86,19 @@ struct BundleAdjustmentOptions {
   int max_num_images_direct_dense_gpu_solver = 200;
   int max_num_images_direct_sparse_gpu_solver = 4000;
 
+  // Whether to use prior camera positions
+  bool use_prior_position = false;
+
+  // Standard deviation on the position priors
+  // Eigen::Vector3d prior_position_std = Eigen::Vector3d::Ones();
```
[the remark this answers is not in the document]
ferreram: :+1:

src/colmap/scene/database.cc
```diff
@@ -1212,8 +1238,9 @@ void Database::PrepareSQLStatements() {
   // write_*
   //////////////////////////////////////////////////////////////////////////////
   sql =
-      "INSERT INTO pose_priors(image_id, position, coordinate_system) "
-      "VALUES(?, ?, ?);";
+      "INSERT INTO pose_priors(image_id, position, covariance, "
```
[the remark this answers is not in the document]
ferreram: Done :+1:

src/colmap/scene/database_cache.cc
```diff
@@ -158,6 +159,29 @@ std::shared_ptr<DatabaseCache> DatabaseCache::Create(
...
+
+  timer.Restart();
+  LOG(INFO) << "Loading pose priors...";
+
+  {
+    cache->pose_priors_.reserve(database.NumPosePriors());
+
+    for (const auto& id_and_image : cache->images_) {
+      if (database.ExistsPosePrior(id_and_image.first)) {
+        // Should we ensure that PosePrior isValid() == true?
```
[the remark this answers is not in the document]
ferreram: :+1:

src/colmap/estimators/bundle_adjustment.h
ferreram: The struct Options is now used to store the use of a robust loss on the priors & the related scale as well as the ransac threshold error.  Other parameters have been moved into the class directly.

src/colmap/estimators/bundle_adjustment.h
ferreram: You were right, it is cleaner to have it directly into PosePriorBundleAdjuster::Options.  I have updated the sfm IncrementalMapper options to propagate those through the parent call stack.  Doing it like this feels right as many options are already propagated from the IncrementalPipelineOptions to the IncrementalMapper options.
Let me know what you think of it.

src/colmap/estimators/bundle_adjustment.cc
ferreram: I have changed the Reconstruction::Normalize() method with the interface you suggested.  I have only set fixed_scale to false by default instead of true in order for the method to have the same behavior as it used to by default.  I am now relying on it in the PosePriorBundleAdjuster to normalize the reconstruction before solving the ceres problem and transforming it back to the metric coordinates after.

src/colmap/estimators/bundle_adjustment.cc
```diff
@@ -839,6 +842,211 @@ void RigBundleAdjuster::ParameterizeCameraRigs(Reconstruction* reconstruction) {
...
+
+    VLOG(2) << "Rigid Sim3 alignment w.r.t. prior position:\n"
+            << "  - scale : " << sim3_tform.scale << "\n"
+            << "  - trans : " << sim3_tform.translation.transpose() << "\n";
+
+    VLOG(2) << "Sim3 alignment error w.r.t. prior position:\n"
+            << "  - rmse:   " << std::sqrt(Mean(verr2_wrt_prior)) << " m\n"
+            << "  - median: " << std::sqrt(Median(verr2_wrt_prior)) << " m\n";
+  } else {
+    LOG(WARNING) << "Sim3 alignment w.r.t. prior position failed!";
```
sarlinpe, line 1030: Shouldn't we set `use_prior_position_=false` in case the alignment failed?

src/colmap/estimators/bundle_adjustment.cc
```diff
@@ -839,6 +842,211 @@ void RigBundleAdjuster::ParameterizeCameraRigs(Reconstruction* reconstruction) {
...
+  loss_function_ =
+      std::unique_ptr<ceres::LossFunction>(options_.CreateLossFunction());
+
+  if (prior_options_.use_robust_loss_on_prior_position) {
+    prior_loss_function_ = std::make_unique<ceres::CauchyLoss>(
+        prior_options_.prior_position_loss_scale);
+  }
+
+  // Initialize images' position w.r.t. priors with a rigid sim3D alignment
+  Sim3DAlignment(reconstruction);
```
sarlinpe, line 870: I suggest making `Sim3DAlignment` return the success:
```suggestion
  use_prior_position_ = Sim3DAlignment(reconstruction);
```
This is cleaner in case `Solve` is called multiple times (currently it relies on default values of `use_prior_position_`).

src/colmap/sfm/incremental_mapper.h
```diff
@@ -118,6 +118,16 @@ class IncrementalMapper {
...
+    // Whether to use prior camera positions
+    bool use_prior_position = false;
+
+    // Whether to use a robust loss on prior locations
+    bool use_robust_loss_on_prior_position = false;
+
+    // Threshold on the residual for the robust loss
+    // (chi2 for 3DOF at 95% = 7.815)
+    double prior_position_loss_scale = 7.815;
+
```
sarlinpe, line 130: @ahojnnes Do you see a cleaner way to pass these options from the CLI to the PosePriorBA? The set of options will likely grow in the future with additional sensors...

src/colmap/estimators/bundle_adjustment.cc
sarlinpe: Nice, thank you for all this great work!

src/colmap/scene/database.cc
```diff
@@ -1339,9 +1367,10 @@ void Database::CreateImageTable() const {
 void Database::CreatePosePriorTable() const {
   const std::string sql =
       "CREATE TABLE IF NOT EXISTS pose_priors"
-      "   (image_id          INTEGER  PRIMARY KEY  NOT NULL,"
-      "    position          BLOB,"
-      "    coordinate_system INTEGER               NOT NULL,"
+      "   (image_id                   INTEGER  PRIMARY KEY  NOT NULL,"
+      "    position                   BLOB,"
+      "    position_covariance        BLOB,"
```
sarlinpe, line 1372: Would you mind updating the Python database interface to add the new column? https://github.com/colmap/colmap/blob/9c1ad104d2d54dc18996d1ac7a0d4befb9e91b9b/scripts/python/database.py#L211

sarlinpe:
@HernandoR Very good point. The PosePrior and PosePriorBA only handle position priors for now but can be extended to orientation priors in the future. We don't have good data to develop this for now - I guess that such priors are most likely to be obtained with compass?

src/colmap/scene/database.cc
ferreram: Done :+1: 
I have also added some tests in the example_usage() function of the scripts.

src/colmap/estimators/bundle_adjustment.cc
ferreram: :+1:

src/colmap/scene/database.cc
sarlinpe: Nice, thank you!

src/colmap/scene/database.cc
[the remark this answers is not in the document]
sarlinpe: This makes sense, thank you!

src/colmap/scene/database.cc
```diff
@@ -453,6 +453,13 @@ PosePrior Database::ReadPosePrior(const image_t image_id) const {
         ReadStaticMatrixBlob<Eigen::Vector3d>(sql_stmt_read_pose_prior_, rc, 1);
     prior.coordinate_system = static_cast<PosePrior::CoordinateSystem>(
         sqlite3_column_int64(sql_stmt_read_pose_prior_, 2));
+    prior.position_covariance =
+        ReadStaticMatrixBlob<Eigen::Matrix3d>(sql_stmt_read_pose_prior_, rc, 3);
+
+    if (prior.position_covariance.isZero()) {
```
sarlinpe, line 459: I am curious why this is needed - if NaNs are written with `WriteStaticMatrixBlob`, isn't the matrix returned by `ReadStaticMatrixBlob` NaN too?

src/colmap/geometry/gps.h
```diff
@@ -93,14 +93,27 @@ struct PosePrior {
...
+  PosePrior(const Eigen::Vector3d& position, const Eigen::Matrix3d& covariance)
+      : position(position), position_covariance(covariance) {}
+  PosePrior(const Eigen::Vector3d& position,
+            const Eigen::Matrix3d& covariance,
+            const CoordinateSystem system)
+      : position(position),
+        position_covariance(covariance),
+        coordinate_system(system) {}
+
+  inline bool IsValid() const { return position.allFinite(); }
```
sarlinpe, line 113: Last minor comment: we need to add bindings for the new constructors, attribute, and methods in pycolmap/geometry/bindings.cc - thank you!

src/colmap/scene/database.cc
ferreram: I did this because when the position_covariance column is added to databases that were not created this new table structure, the covariance matrix is set to zero by default.

src/colmap/geometry/gps.h
ferreram: :+1:

src/colmap/scene/database.cc
sarlinpe: You might be able to pass a default value when adding the column, something like this:
```c++
const std::string sql = "ALTER TABLE pose_priors ADD COLUMN position_covariance BLOB DEFAULT (?);";
sqlite3_stmt* sql_stmt;
SQLITE3_CALL(sqlite3_prepare_v2(database_, sql.c_str(), -1, &sql_stmt, 0));
WriteStaticMatrixBlob(sql_stmt, PosePrior().position_covariance, 1);
SQLITE3_CALL(sqlite3_step(sql_stmt));
SQLITE3_CALL(sqlite3_finalize(sql_stmt));
```

ferreram: Done :+1: 

It seems it is not possible to set a dynamic default parameter but I solved it by directly applying an update procedure on the column.

src/colmap/controllers/incremental_mapper.cc
```diff
@@ -255,6 +259,12 @@ bool IncrementalPipeline::LoadDatabase() {
     return false;
   }
 
+  // If prior positions are to be used and setup from the database, convert
+  // geographic coords. to cartesian ones
+  if (options_->use_prior_position) {
+    return database_cache_->setupPosePriors();
```
ahojnnes, line 265: nit
```suggestion
    return database_cache_->SetupPosePriors();
```

src/colmap/estimators/bundle_adjustment.cc
```diff
@@ -848,6 +851,212 @@ void RigBundleAdjuster::ParameterizeCameraRigs(Reconstruction* reconstruction) {
...
+
+PosePriorBundleAdjuster::PosePriorBundleAdjuster(
+    const BundleAdjustmentOptions& options,
+    const Options& prior_options,
+    const BundleAdjustmentConfig& config,
+    const std::unordered_map<image_t, PosePrior>& image_id_to_pose_prior)
+    : BundleAdjuster(options, config),
+      prior_options_(prior_options),
+      image_id_to_pose_prior_(image_id_to_pose_prior) {
+  setRansacMaxErrorFromPriorsCovariance();
```
ahojnnes, line 866:
```suggestion
  SetRansacMaxErrorFromPriorsCovariance();
```

src/colmap/estimators/bundle_adjustment.cc
```diff
@@ -848,6 +851,212 @@ void RigBundleAdjuster::ParameterizeCameraRigs(Reconstruction* reconstruction) {
...
+  }
+}
+
+void PosePriorBundleAdjuster::AddPosePriorToProblem(
+    image_t image_id,
+    const PosePrior& prior,
+    Reconstruction* reconstruction,
+    ceres::LossFunction* prior_loss_function) {
+  if (!prior.IsValid() || !prior.IsCovarianceValid()) {
+    LOG(ERROR) << "Could not add prior for image #" << image_id << "\n";
```
ahojnnes, line 939: The newline is added by glog automatically.
```suggestion
    LOG(ERROR) << "Could not add prior for image #" << image_id;
```

src/colmap/estimators/bundle_adjustment.cc
```diff
@@ -848,6 +851,212 @@ void RigBundleAdjuster::ParameterizeCameraRigs(Reconstruction* reconstruction) {
...
+
+    const auto report = ransac.Estimate(v_src, v_tgt);
+
+    if (report.success) {
+      // Apply sim3 transform
+      sim3_tform = Sim3d::FromMatrix(report.model);
+      success = true;
+    }
+  } else {
+    success = colmap::EstimateSim3d(v_src, v_tgt, sim3_tform);
```
ahojnnes, line 1013:
```suggestion
    success = EstimateSim3d(v_src, v_tgt, sim3_tform);
```

src/colmap/exe/sfm.cc
```diff
@@ -45,6 +45,24 @@
 
 namespace colmap {
 
+void updateDatabasePosePriorsCovariance(const std::string& _database_path,
```
ahojnnes, line 48:
```suggestion
void UpdateDatabasePosePriorsCovariance(const std::string& _database_path,
```

src/colmap/exe/sfm.cc
```diff
@@ -45,6 +45,24 @@
...
+                                        const Eigen::Matrix3d& _covariance) {
```
ahojnnes, line 49:
```suggestion
                                        const Eigen::Matrix3d& covariance) {
```

src/colmap/scene/database_cache.cc
```diff
@@ -209,4 +231,69 @@ const class Image* DatabaseCache::FindImageWithName(
   return nullptr;
 }
 
+bool DatabaseCache::setupPosePriors() {
```
ahojnnes, line 234:
```suggestion
bool DatabaseCache::SetupPosePriors() {
```

src/colmap/scene/database_cache.cc
```diff
@@ -209,4 +231,69 @@ const class Image* DatabaseCache::FindImageWithName(
...
+  LOG(INFO) << "Setting up prior positions...";
+
+  Timer timer;
+  timer.Start();
+
+  if (NumPosePriors() < 3) {
```
ahojnnes, line 240: I guess this is only valid for when only having position based priors.

Conceptually, it feels like the database cache should at most warn but not produce an error in this case. I would have expected the incremental mapper to error out in this case.

src/colmap/exe/sfm.cc
```diff
@@ -339,6 +357,143 @@ int RunHierarchicalMapper(int argc, char** argv) {
...
+  OptionManager options;
+  options.AddDatabaseOptions();
+  options.AddImageOptions();
+  options.AddDefaultOption("input_path", &input_path);
+  options.AddRequiredOption("output_path", &output_path);
+  options.AddMapperOptions();
+
+  options.mapper->use_prior_position = true;
+
+  options.AddDefaultOption("set_database_priors_covariance",
```
ahojnnes, line 378: Is the following maybe a more intuitive option name? If I understand correctly, if this is false, then the covariance is read from the database?
```suggestion
  options.AddDefaultOption("overwrite_priors_covariance",
```

src/colmap/scene/synthetic.cc
```diff
@@ -277,6 +278,38 @@ void SynthesizeDataset(const SyntheticDatasetOptions& options,
...
+        static const double lat0 = 47.37851943807808;
+        static const double lon0 = 8.549099927632087;
+        static const double alt0 = 451.5;
+
+        noisy_prior.position =
+            gps_trans.ENUToEll({noisy_prior.position}, lat0, lon0, alt0)[0];
+        noisy_prior.coordinate_system = PosePrior::CoordinateSystem::WGS84;
+      }
+
+      database->WritePosePrior(image_id, noisy_prior);
```
ahojnnes, line 310: I guess we could extend the synthetic_test.cc with a small check on consistent pairs of projection centers and pose priors (if the stddev is zero)?

src/colmap/sfm/incremental_mapper.h
ahojnnes: The structure is indeed getting a little bit out of control. I don't think we want to block this PR here, as this was an issue already before. We can probably group them more meaningfully or reuse the underlying options structs. Not sure.

ahojnnes (approved):
Great effort. I left another round of mostly "nit" comments. After this, the PR looks good to me and we can merge from my perspective. Thank you very much!

src/colmap/scene/database_cache.cc
ferreram: You are completely right.  I have modified it so that it only returns false if there is no prior in the database.

src/colmap/exe/sfm.cc
ferreram: Sounds good to me :+1: 

Yes for now, the covariance is always read from the database but enabling the option overwrites the current state of the priors covariance.
I am planning to later add more ways for setting up the priors covariance (python / c++ scripts to set it from a file and maybe also a button in the database_widget).

src/colmap/scene/synthetic.cc
ferreram: Done :+1:

sarlinpe (approved):

B1ueber2y:
No tests were added for ``PositionPriorErrorCostFunction``. Do you plan to add it in the future? : )

ferreram:
@B1ueber2y I wanted to do it and then forgot... I have just made a PR with test for PositionPriorErrorCostFunctor!

ahojnnes:
@ferreram Thank you!

mattiasmar:
@ahojnnes Seems like great work. Are there instructions or examples for the usage of this feature? I'm explicitly looking for how to do COLMAP reconstruction with pose priors from rtabmap. rtabmap would give me the absolute coordinates and direction of my baselink (robot position).

mattiasmar:
Is there a pycolmap interface for the pose_prior_mapper functionality?

sarlinpe:
Yes
https://github.com/colmap/colmap/blob/1a497f023153cf7f7074e692e066137342ab645e/src/pycolmap/estimators/bundle_adjustment.cc#L195-L201

mattiasmar:
May I ask for example on how to use this interface?

Until recently my flow look like this:
```

1. rec = pycolmap.Reconstruction()
2. define camera instances: pycolmap.Camera 
3. define images (camera poses + image paths): 
  * im = pycolmap.Image(image_name, [], cam_from_worldinv, camera_id, image_id)
  * rec.add_image(im)  (repeat for all images...)
4. Extract and Match features (with HLOC):    
     4.a extract_features.main(...)
     4.b match_features.main(...)
5. Perform triangulation using the existing camera poses
    triangulation.main(
        reference_model=reconstruction_path,
        sfm_dir = sfm_path,
        image_dir=Path(output_dir),
        pairs=sfm_pairs,
        features=features_path,
        matches=matches_path,
        skip_geometric_verification = True
    )
 final_reconstruction = pycolmap.Reconstruction(sfm_path)
 **pycolmap.bundle_adjustment(final_reconstruction, options)**
```

A) With  the pycolmap.create_pose_prior_bundle_adjuster how would the above flow change? 
B) Is this a correct definition of a PosePrior instance?
```
pose_prior = pycolmap.PosePrior()
*Set the position and orientation priors
pose_prior.position = np.array([1.0, 2.0, 3.0], dtype=np.float64)
pose_prior.orientation = np.array([0.0, 0.0, 0.0, 1.0], dtype=np.float64)  # Quaternion, scalar first

```
C) Would the pose_prior_bundle_adjuster  be executed by calling "solve" and where would results be stored?
```
pose_prior_bundle_adjuster = pycolmap.create_pose_prior_bundle_adjuster(
    options=bundle_adjustment_options,
    prior_options=pose_prior_bundle_adjustment_options,
    config=bundle_adjustment_config,
    pose_priors=pose_priors,
    reconstruction=reconstruction
)
pose_prior_bundle_adjuster.solve()
```

ahojnnes:
A) You would like perform a pose prior based bundle adjustment after your call to triangulation. You may want to alternative a few rounds of (re-)triangulation and bundle adjustment until convergence.
B) The pose prior can currently only contain position information. You also have to define the coordinate system convention for each pose prior.
C) The results will be stored directly in the reconstruction.

shuta-ochiai:
@ferreram
Thanks for your great code!
I tried the `pose_prior_mapper` cmd, but I got an error saying `No pose priors in database...`

I performed the following processes:
1. get images and poses from ARKit app (without points clouds)
2. convert ARKit fmt to colmap fmt (images.bin/txt, cameras.bin/txt, points3D.bin/txt(empty), database.db)
 ```
directory configuration:
arkit_data_path/
       |--- input/
             |--- frame_00001.jpg
             |--- frame_....jpg
       |--- sparse/
             |--- 0/
                |--- images.bin/txt
                |--- cameras.bin/txt
                |--- points3D.bin/txt
                |--- database.db
```
3. extract feature 
```
colmap feature_extractor 
--database_path arkit_data_path/sparse/0/database.db 
--image_path arkit_data_path/input 
--SiftExtraction.use_gpu 1 
--ImageReader.single_camera 1
```
4. feature matching
```
colmap sequential_mathcer
--database_path arkit_data_path/sparse/0/databse.db
--SiftExtraction.use_gpu 1
```
5. pose prior mapper
```
colmap pose_prior_mapper
--image_path arkit_data_path/input
--database_path arkit_data_path/sparse/0/database.db
--output_path sfm_processed_data_path
--Mapper.multiple_models 0
--Mapper.ba_refine_focal_length 0
--Mapper.ba_refine_extra_params 0
--overwrite_priors_covariance 1
--prior_position_std_x 0.05
--prior_position_std_y 0.05
--prior_position_std_z 0.01
``` 
6. bundle adjuster
```
colmap bundle_adjuster
--input_path sfm_processed_data_path
--output_path sfm_processed_data_path
--BundleAdjustment.refine_focal_length 0
--BundleAdjustment.refine_principal_point 0
--BundleAdjustment.refine_extra_params 0
```


Error Encountered: 
```
I1202 01:23:10.560431   948 sfm.cc:54] Setting up database pose priors with the same covariance matrix:                                    
0.0025      0      0                                                                                                                       
     0 0.0025      0                                                                                                                       
     0      0 0.0001                                                                                                                       
I1202 01:23:10.574162   948 incremental_pipeline.cc:237] Loading database                                                                  
I1202 01:23:10.624039   948 database_cache.cc:66] Loading cameras...                                                                       
I1202 01:23:10.628621   948 database_cache.cc:76]  1 in 0.005s                                                                             
I1202 01:23:10.628628   948 database_cache.cc:84] Loading matches...                                                                       
I1202 01:23:11.159878   948 database_cache.cc:89]  1082 in 0.531s                                                                          
I1202 01:23:11.159904   948 database_cache.cc:105] Loading images...                                                                       
I1202 01:23:16.291671   948 database_cache.cc:153]  200 in 5.132s (connected 200)                                                          
I1202 01:23:16.291709   948 database_cache.cc:164] Loading pose priors...                                                                  
I1202 01:23:16.612479   948 database_cache.cc:175]  0 in 0.321s                                                                            
I1202 01:23:16.612496   948 database_cache.cc:184] Building correspondence graph...                                                        
I1202 01:23:16.776397   948 database_cache.cc:210]  in 0.164s (ignored 0)                                                                  
I1202 01:23:16.776595   948 timer.cc:91] Elapsed time: 0.103                                                                               
I1202 01:23:16.776602   948 database_cache.cc:228] Setting up prior positions...                                                           
E1202 01:23:16.776602   948 database_cache.cc:234] No pose priors in database...                                                           
E1202 01:23:16.950085   948 sfm.cc:465] failed to create sparse model     
```

If you have any solutions., please let me know!

**Comment by HernandoR** [comment]
To my knowledge, you have to insert the pose prior to the database, ie. the database.db. Colmap itself can only read and convert that info from exif info of the images. Please check if your images have corresponding information. Otherwise you have to insert manually.

I think there is a py interface named writeposeprior, please checkout on the pycolmap directory.

Liu Zhen
________________________________

**Comment by mattiasmar** [comment]
@ahojnnes I tried a basic call to create_pose_prior_bundle_adjuster, but get an exception:

```
    rec = pycolmap.Reconstruction('output/sfm')  #Folder containing images.bin, cameras.bin and points3D.bin
    pose_prior_bundle_adjuster = pycolmap.create_pose_prior_bundle_adjuster(
        pose_priors=pose_priors,
        reconstruction= rec
    )
```

`pose_priors` is a dict of int (image id) VS pycolmap.PosePrior objects

```
create_pose_prior_bundle_adjuster(): incompatible function arguments. The following argument types are supported:
    1. (options: pycolmap._core.BundleAdjustmentOptions, prior_options: pycolmap._core.PosePriorBundleAdjustmentOptions, config: pycolmap._core.BundleAdjustmentConfig, pose_priors: dict[int, pycolmap._core.PosePrior], reconstruction: pycolmap._core.Reconstruction) -> pycolmap._core.BundleAdjuster

Invoked with: kwargs: options=BundleAdjustmentOptions(loss_function_type=LossFunctionType.TRIVIAL, loss_function_scale=1.0, refine_focal_length=False, refine_principal_point=False, refine_extra_params=False, refine_extrinsics=True, print_summary=True, use_gpu=True, gpu_index='-1', min_num_residuals_for_cpu_multi_threading=50000, min_num_images_gpu_solver=50, max_num_images_direct_dense_cpu_solver=50, max_num_images_direct_sparse_cpu_solver=1000, max_num_images_direct_dense_gpu_solver=200, max_num_images_direct_sparse_gpu_solver=4000), pose_priors={1: PosePrior(position=[43.502, -30.3787, -114.512], position_covariance=[1, 0, 0, 0, 1, 0, 0, 0, 1], coordinate_system=CARTESIAN), 2: PosePrior(position=[43.4313, -30.4905, -114.934], position_covariance=[1, 0, 0, 0, 1, 0, 0, 0, 1], coordinate_system=CARTESIAN), 3: PosePrior(position=[43.5459, -30.594, -115.323], position_covariance=[1, 0, 0, 0, 1, 0, 0, 0, 1], ....
```

Could you share a mini example that could serve me and the community as a guide?

**Comment by bbzh** [comment]
Is there an official instruction or example to use the GPS prior? That will be very useful for the community.

**Comment by shuta-ochiai** [comment]
[HernandoR](https://github.com/HernandoR) cc [ferreram](https://github.com/ferreram)
Thanks your reply!

I succeeded in running the pose_prior_mapper by writing pose_prior information to the database. 
However, I encountered errors "Could not register, trying another image." 
If you have any solutions, I would appreciate it if you could share them with me.

The following is the log when executing pose_prior_mapper:
```
I1215 08:52:39.800783  1462 sfm.cc:54] Setting up database pose priors with the same covariance matrix: 
0.0025      0      0
     0 0.0025      0
     0      0 0.0001
I1215 08:52:39.998967  1462 incremental_pipeline.cc:237] Loading database
I1215 08:52:40.071281  1462 database_cache.cc:66] Loading cameras...
I1215 08:52:40.076885  1462 database_cache.cc:76]  1 in 0.006s
I1215 08:52:40.076900  1462 database_cache.cc:84] Loading matches...
I1215 08:52:40.188160  1462 database_cache.cc:89]  716 in 0.111s
I1215 08:52:40.188179  1462 database_cache.cc:105] Loading images...
I1215 08:52:44.664470  1462 database_cache.cc:153]  200 in 4.476s (connected 199)
I1215 08:52:44.664518  1462 database_cache.cc:164] Loading pose priors...
I1215 08:52:45.635279  1462 database_cache.cc:175]  199 in 0.971s
I1215 08:52:45.635304  1462 database_cache.cc:184] Building correspondence graph...
I1215 08:52:45.714617  1462 database_cache.cc:210]  in 0.079s (ignored 0)
I1215 08:52:45.714818  1462 timer.cc:91] Elapsed time: 0.094 [minutes]
I1215 08:52:45.714829  1462 database_cache.cc:228] Setting up prior positions...
I1215 08:52:45.714917  1462 timer.cc:91] Elapsed time: 0.000 [minutes]
I1215 08:52:45.760480  1462 incremental_pipeline.cc:282] Finding good initial image pair
I1215 08:52:46.130198  1462 incremental_pipeline.cc:306] Initializing with image pair #71 and #67
I1215 08:52:46.131283  1462 incremental_pipeline.cc:311] Global bundle adjustment
I1215 08:52:46.211699  1462 incremental_pipeline.cc:390] Registering image #68 (3)
I1215 08:52:46.211710  1462 incremental_pipeline.cc:393] => Image sees 155 / 2012 points
I1215 08:52:46.326551  1462 incremental_pipeline.cc:42] Retriangulation and Global bundle adjustment
I1215 08:52:46.536549  1462 incremental_pipeline.cc:390] Registering image #69 (4)
.....
I1215 08:55:29.570765  1462 incremental_pipeline.cc:393] => Image sees 256 / 966 points
I1215 08:55:29.825865  1462 incremental_pipeline.cc:390] Registering image #5 (103)
I1215 08:55:29.825887  1462 incremental_pipeline.cc:393] => Image sees 46 / 534 points
I1215 08:55:29.913379  1462 incremental_pipeline.cc:42] Retriangulation and Global bundle adjustment
I1215 08:55:37.116593  1462 incremental_pipeline.cc:390] Registering image #6 (104)
I1215 08:55:37.116622  1462 incremental_pipeline.cc:393] => Image sees 38 / 170 points
I1215 08:55:37.118126  1462 incremental_pipeline.cc:404] => Could not register, trying another image.
I1215 08:55:37.118134  1462 incremental_pipeline.cc:390] Registering image #110 (104)
I1215 08:55:37.118137  1462 incremental_pipeline.cc:393] => Image sees 41 / 501 points
I1215 08:55:37.196182  1462 incremental_pipeline.cc:390] Registering image #111 (105)
I1215 08:55:37.196203  1462 incremental_pipeline.cc:393] => Image sees 75 / 524 points
I1215 08:55:37.291985  1462 incremental_pipeline.cc:390] Registering image #112 (106)
I1215 08:55:37.292003  1462 incremental_pipeline.cc:393] => Image sees 85 / 777 points
I1215 08:55:37.382576  1462 incremental_pipeline.cc:390] Registering image #113 (107)
I1215 08:55:37.382594  1462 incremental_pipeline.cc:393] => Image sees 158 / 1019 points
I1215 08:55:37.518312  1462 incremental_pipeline.cc:390] Registering image #114 (108)
I1215 08:55:37.518332  1462 incremental_pipeline.cc:393] => Image sees 134 / 525 points
I1215 08:55:37.651600  1462 incremental_pipeline.cc:390] Registering image #115 (109)
I1215 08:55:37.651618  1462 incremental_pipeline.cc:393] => Image sees 41 / 109 points
I1215 08:55:37.713141  1462 incremental_pipeline.cc:390] Registering image #88 (110)
I1215 08:55:37.713160  1462 incremental_pipeline.cc:393] => Image sees 34 / 274 points
I1215 08:55:37.842927  1462 incremental_pipeline.cc:390] Registering image #6 (111)
I1215 08:55:37.842949  1462 incremental_pipeline.cc:393] => Image sees 38 / 170 points
I1215 08:55:37.843928  1462 incremental_pipeline.cc:404] => Could not register, trying another image.
I1215 08:55:37.843935  1462 incremental_pipeline.cc:42] Retriangulation and Global bundle adjustment
I1215 08:55:44.583489  1462 incremental_pipeline.cc:390] Registering image #6 (111)
I1215 08:55:44.583528  1462 incremental_pipeline.cc:393] => Image sees 38 / 170 points
I1215 08:55:44.584636  1462 incremental_pipeline.cc:404] => Could not register, trying another image.
I1215 08:55:44.584645  1462 incremental_pipeline.cc:42] Retriangulation and Global bundle adjustment
I1215 08:55:51.800009  1462 timer.cc:91] Elapsed time: 3.197 [minutes]
```

PR #2675, opened 2024-07-26
Build and deploy docs from CI

ahojnnes:


ahojnnes:
... cannot push from one repo to another one without personal access tokens, it seems. Let's keep it manual.

PR #2145, opened 2023-09-05
Incremental Model Refiner

tsattler:
A variant of the point_triangulator (the code is based on the point_triangulator). For a given number of iterations, it, in each iteration, runs bundle adjustment and then completes and merges tracks. The bundle adjustment optimizes the camera parameters as well, allowing the refiner to handle inaccurate input poses better than the point_triangulator.

PR #2685, opened 2024-07-31
Incremental model refiner for DetectorFreeSfM

Dawars:
This feature helps refine cameras and models.

- Adds an optional parameter to TriangulateReconstruction for the fixed image ids
- Adds a setter function and call it after BeginReconstruction.

Do I need to add tests/python bindings?

Based on https://github.com/colmap/colmap/pull/2145 and https://github.com/hxy-123/colmap

### DIFF
--- a/src/colmap/controllers/incremental_mapper.cc
+++ b/src/colmap/controllers/incremental_mapper.cc
@@ -538,10 +538,12 @@ void IncrementalMapperController::Reconstruct(
 }
 
 void IncrementalMapperController::TriangulateReconstruction(
-    const std::shared_ptr<Reconstruction>& reconstruction) {
+    const std::shared_ptr<Reconstruction>& reconstruction,
+    const std::unordered_set<image_t> fixed_image_ids) {
   THROW_CHECK(LoadDatabase());
   IncrementalMapper mapper(database_cache_);
   mapper.BeginReconstruction(reconstruction);
+  mapper.SetFixedImageIds(fixed_image_ids);
 
   LOG(INFO) << "Iterative triangulation";
   const std::vector<image_t>& reg_image_ids = reconstruction->RegImageIds();
--- a/src/colmap/controllers/incremental_mapper.cc
+++ b/src/colmap/controllers/incremental_mapper.cc
@@ -543,8 +543,10 @@ void IncrementalMapperController::TriangulateReconstruction(
   THROW_CHECK(LoadDatabase());
   IncrementalMapper mapper(database_cache_);
   mapper.BeginReconstruction(reconstruction);
-  mapper.SetFixedImageIds(fixed_image_ids);
-
+  
+  if (!fixed_image_ids.empty()) {
+    mapper.SetFixedImageIds(fixed_image_ids);
+  }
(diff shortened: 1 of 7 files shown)


src/colmap/controllers/incremental_mapper.cc
```diff
@@ -538,11 +538,15 @@ void IncrementalMapperController::Reconstruct(
 }
 
 void IncrementalMapperController::TriangulateReconstruction(
-    const std::shared_ptr<Reconstruction>& reconstruction) {
+    const std::shared_ptr<Reconstruction>& reconstruction,
+    const std::unordered_set<image_t> fixed_image_ids) {
```
sarlinpe, line 542: const ref?

src/colmap/exe/sfm.cc
```diff
@@ -183,6 +183,66 @@ int RunColorExtractor(int argc, char** argv) {
...
+
+  if (!image_list_path.empty()) {
+    const auto image_names = ReadTextFileLines(image_list_path);
+    options.mapper->image_names =
+        std::unordered_set<std::string>(image_names.begin(), image_names.end());
+  }
+
+  // Loads the list of images for which the camera pose will be fixed.
+  auto reconstruction = reconstruction_manager->Get(0);
+  std::unordered_set<image_t> image_ids_fixed_poses;
```
sarlinpe, line 224: nit: call this `fixed_image_ids` to be consistent with `TriangulateReconstruction`

src/colmap/sfm/incremental_mapper.cc
```diff
@@ -108,6 +108,11 @@ IncrementalMapper::IncrementalMapper(
       num_total_reg_images_(0),
       num_shared_reg_images_(0) {}
 
+void IncrementalMapper::SetFixedImageIds(
+    std::unordered_set<image_t> image_ids) {
```
sarlinpe, line 112: const ref?

src/colmap/sfm/incremental_mapper.cc
```diff
@@ -108,6 +108,11 @@ IncrementalMapper::IncrementalMapper(
       num_total_reg_images_(0),
       num_shared_reg_images_(0) {}
 
+void IncrementalMapper::SetFixedImageIds(
```
sarlinpe, line 111: Call this `SetExistingImageIds` and move the declaration+definition closer to `ExistingImageIds`

src/colmap/exe/sfm.cc
```diff
@@ -183,6 +183,66 @@ int RunColorExtractor(int argc, char** argv) {
   return EXIT_SUCCESS;
 }
 
+int RunIncrementalModelRefiner(int argc, char** argv) {
```
sarlinpe, line 186: A lot of this is very similar to `RunPointTriangulator`. Would it make sense to instead augment `RunPointTriangulator` with this new feature? can we minimally allow the user the control whether to 1) fix all images (the default of `RunPointTriangulator`), 2) fix no image, 3) fix a subset of images?

sarlinpe (changes requested):

PR #2697, opened 2024-08-10
Add camera pointer in the image class

B1ueber2y:
This PR adds support for directly indexing cameras in the image class with backward compatibility. As one of the steps for enabling better rig design, an optional camera pointer is added into the image class, and get initialized with ``SyncCameraPointers`` in the reconstruction class. The IO is unchanged in this feature. 

As a byproduct, one can directly call image.ProjectPoint(point3d) when the camera pointer is initialized (will return None when the point is behind the camera). Relevant pybind is also supported.

```
import pycolmap
image = recon.images[1]
print(image.camera) 
```

We could also think of many more interesting features to be added in the future (e.g. fundamental matrix from two image instances etc.) Also, as a debatable point, we could potentially require the pointer to be always initialized by deprecating the SetCameraId() interface.

### DIFF
--- a/src/colmap/controllers/feature_extraction.cc
+++ b/src/colmap/controllers/feature_extraction.cc
@@ -87,7 +87,7 @@ struct ImageData {
   ImageReader::Status status = ImageReader::Status::FAILURE;
 
   Camera camera;
-  Image image;
+  BaseImage image;
   PosePrior pose_prior;
   Bitmap bitmap;
   Bitmap mask;
@@ -544,7 +544,7 @@ class FeatureImporterController : public Thread {
 
       // Load image data and possibly save camera to database.
       Camera camera;
-      Image image;
+      BaseImage image;
       PosePrior pose_prior;
       Bitmap bitmap;
       if (image_reader.Next(&camera, &image, &pose_prior, &bitmap, nullptr) !=
--- a/src/colmap/controllers/image_reader.cc
+++ b/src/colmap/controllers/image_reader.cc
@@ -86,7 +86,7 @@ ImageReader::ImageReader(const ImageReaderOptions& options, Database* database)
 }
 
 ImageReader::Status ImageReader::Next(Camera* camera,
-                                      Image* image,
+                                      BaseImage* image,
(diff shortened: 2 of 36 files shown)


ahojnnes:
As an alternative, and I don't know if it's easily possible, we could require the camera to exist upon construction of an image. In other words, make it a required argument to the Image constructor.

sarlinpe:
> As an alternative, and I don't know if it's easily possible, we could require the camera to exist upon construction of an image. In other words, make it a required argument to the Image constructor.

I agree that this would be nicer, but then the Camera object needs to be instantiated via a `shared_ptr` all throughout the codebase - is this reasonable? how about declaring `camera_` as `Camera*` instead of `shared_ptr<Camera>` to let the calling scope manage the lifetime of the object?

ahojnnes:
> I agree that this would be nicer, but then the Camera object needs to be instantiated via a `shared_ptr` all throughout the codebase - is this reasonable? how about declaring `camera_` as `Camera*` instead of `shared_ptr<Camera>` to let the calling scope manage the lifetime of the object?

Agreed that the raw pointer can be better and thus not enforce a particular ownership model on the caller side.

B1ueber2y:
I agree that this sounds nice. However, requiring a raw camera pointer at image construction will potentially make the design of the interfaces here much uglier then before: https://github.com/colmap/colmap/blob/main/src/colmap/scene/database.h#L156-L158.

B1ueber2y:
> I agree that this sounds nice. However, requiring a raw camera pointer at image construction will potentially make the design of the interfaces here much uglier then before: https://github.com/colmap/colmap/blob/main/src/colmap/scene/database.h#L156-L158.

A random proposal to resolve this: we could do one level abstraction further by having class ``Image`` inheriting from something like ``ImageBase``, where the base abstraction is completely similar as the ``Image`` class in the COLMAP before, such that it does not hold the raw pointer of the camera and is only used in the ``Database`` readers.

B1ueber2y:
Closed due to totally different design as it started. Reopening in a new PR.

PR #2698, opened 2024-08-10
Add frame impl for future rig support

B1ueber2y:
A draft implementation of the frame class for the rig support, with backward compatibility. However, with the new design CamFromWorld() wont make sense as a reference anymore when the frame is nontrivial, where we need to composite cam_from_world from the rigs rather than directly using the member variable. I currently throw error on this, which will potentially break a lot of implementations with CamFromWorld() when we really use the rig as input (which should break since they dont currently support rigs). 

Feedbacks on this will be helpful. Marked as draft for now. @ahojnnes @sarlinpe


src/colmap/util/types.h
```diff
@@ -81,6 +81,12 @@ typedef uint32_t camera_t;
 // Unique identifier for images.
 typedef uint32_t image_t;
 
+// Unique identifier for frames.
+typedef uint64_t frame_t;
+
+// Unique identiier for rigs.
```
ahojnnes, line 87: nit: typo

src/colmap/scene/image.h
```diff
@@ -164,9 +185,46 @@ point2D_t Image::NumPoints2D() const {
...
+
+void Image::SetFrame(std::shared_ptr<class Frame> frame) {
+  frame_ = std::move(frame);
+}
+
+bool Image::HasFrame() const { return frame_ != nullptr; }
+
+bool Image::HasNonTrivialFrame() const {
+  return HasFrame() && frame_->HasRigCalibration() &&
+         !frame_->RigCalibration()->IsReference(
```
ahojnnes, line 202: This assumes that one of the cameras always coincides with the rigs origin? This may not be a convenient choice in case when there is an IMU, where it might be easiest to model everything in the IMU frame?

src/colmap/scene/image.h
```diff
@@ -124,8 +140,13 @@ class Image {
...
   point2D_t num_points3D_;
 
+  // [Optional] The corresponding frame (rig) of the image
+  frame_t frame_id_;
+  std::shared_ptr<class Frame> frame_ = nullptr;
+
   // The pose of the image, defined as the transformation from world to camera.
-  Rigid3d cam_from_world_;
+  // Only useful when the corresponding frame (rig) does not exist.
+  mutable Rigid3d cam_from_world_;
```
ahojnnes, line 149: Let's avoid mutable here?

src/colmap/scene/frame.h
```diff
@@ -0,0 +1,264 @@
...
+#include <set>
+#include <vector>
+
+namespace colmap {
+
+// Sensor type
+enum class SensorType {
+  Camera = 0,
+  IMU = 1,
+  Location = 2,  // include GNSS, radios, compass, etc.
```
ahojnnes, line 45: Do we need to decompose the different location sensor types here already?

src/colmap/scene/frame.h
```diff
@@ -0,0 +1,264 @@
...
+  std::map<sensor_t, bool> is_fixed_sensor_from_rig_;  // for optimization
+};
+
+class Frame {
+ public:
+  // Access the unique identifier of the frame
+  inline frame_t FrameId() const;
+  inline void SetFrameId(frame_t frame_id);
+
+  // Access data ids
```
ahojnnes, line 106: Sorry, I probably forgot and we discussed this some months ago. Not sure I fully understand what the data_t is, in particular the uint64_t refers to what?

ahojnnes (commented):
Thank you. This seems to match what we discussed some months ago, but I also don't have everything still in memory :-) I left a few comments and suggestions.

src/colmap/scene/image.h
B1ueber2y: There might be some misunderstanding here. We can set the IMU to be the reference frame in the current ``Frame`` class to support this?

src/colmap/scene/image.h
B1ueber2y: My initial thought is to do this: https://github.com/colmap/colmap/pull/2698/files#diff-665386c3e6e1ebf0f2b3f383a2925c7a970b2e256f2f889f27fb090065158fbdR210, but indeed this feels unnecessary. Then we may want to explicitly mention in the comments somewhere that the member ``cam_from_world_`` is not used when ``frame_`` is available. 

Note that with the current rig support we cannot do anything as easy as ``image.CamfromWorld() = cam_from_world_prior`` anymore when ``frame_`` is available, which actually makes sense but can be concerning.

src/colmap/scene/image.h
ahojnnes: I see, got it. I read through the code top to bottom and assumed the "Frame" was only used for cameras. This makes sense. The only scenario that is not possible is to have an origin that doesn't coincide with any of the sensors. Not sure if this is useful but I currently also don't see why we would need to have one of them assigned as the reference? Maybe you can explain your motivation here. Thanks.

src/colmap/scene/image.h
ahojnnes: Thanks, I would avoid such magic and make it more explicit. I only very rarely use "mutable" and in 99% of those rare cases, I use it for std::mutex members.

src/colmap/scene/frame.h
ahojnnes: Sorry, still don't fully understand what this translates to in practice. So, in case of an image, what exactly would be the uint64_t?

B1ueber2y: The uint64_t is ``image_id``. ``data_t`` is ``std::pair<std::pairSensorType, sensor_t>, uint64_t>``, where the ``sensor_t`` field corresponds to ``camera_id``, and ``uint64_t`` corresponds to ``image_id`` before.

ahojnnes: Got it, thanks.

src/colmap/scene/image.h
ahojnnes: Thanks. It could make sense to add this motivation somewhere in the header?

ahojnnes:
Thanks @B1ueber2y. On the high-level, this looks good to me. I suggest to also gain @sarlinpe's feedback before we move ahead. Cheers.

src/colmap/scene/image.h
B1ueber2y: Its unclear how to return a const reference here without mutuble: https://github.com/colmap/colmap/pull/2698/files#diff-665386c3e6e1ebf0f2b3f383a2925c7a970b2e256f2f889f27fb090065158fbdR211. Any ideas?

B1ueber2y:
> Thanks @B1ueber2y. On the high-level, this looks good to me. I suggest to also gain @sarlinpe's feedback before we move ahead. Cheers.

Thanks. @sarlinpe PTAL when time permits : )

src/colmap/scene/frame.h
```diff
@@ -0,0 +1,270 @@
...
+
+  // Reference sensor id which has the identity transformation to the rig.
+  sensor_t ref_sensor_id_;
+
+  // list of sensors
+  std::set<sensor_t> sensor_ids_;
+
+  // sensor_from_rig transformation.
+  std::map<sensor_t, Rigid3d> map_sensor_from_rig_;
+  std::map<sensor_t, bool> is_fixed_sensor_from_rig_;  // for optimization
```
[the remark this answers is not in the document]
B1ueber2y: This is planned for optimization. Now I just add the setter to allow the input. Will add the getter once we move on to the rig bundle adjuster.

src/colmap/scene/frame.h
```diff
@@ -0,0 +1,270 @@
...
+
+// Sensor type
+enum class SensorType {
+  Camera = 0,
+  IMU = 1,
+  Location = 2,  // include GNSS, radios, compass, etc.
+};
+
+// Unique identifier of the sensor
+typedef std::pair<SensorType, uint32_t> sensor_t;
```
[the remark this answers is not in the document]
B1ueber2y: struct is also fine, just that we need to make it hashable manually.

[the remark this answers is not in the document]
ahojnnes: I would make it a struct and overload the std::hash/operator<. It makes the usage side more readable if one can do sensor_id.type and sensor_id.id vs. .first/.second.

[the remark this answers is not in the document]
ahojnnes: Same goes for data_t

src/colmap/scene/frame.h
```diff
@@ -0,0 +1,270 @@
...
+
+ private:
+  // Unique identifier of the device.
+  rig_t rig_id_;
+
+  // Reference sensor id which has the identity transformation to the rig.
+  sensor_t ref_sensor_id_;
+
+  // list of sensors
+  std::set<sensor_t> sensor_ids_;
```
ahojnnes, line 99: Alternative to set/map we could just sorted vectors and explicit std::sort on each AddSensor(). It may result in more efficient access, which will happen a lot whenever the rig transformation is needed?

src/colmap/scene/frame.h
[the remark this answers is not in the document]
ahojnnes: I suggest to remove anything that is unused for now and add it when it is actually needed.

src/colmap/scene/frame.h
```diff
@@ -0,0 +1,270 @@
...
+}
+
+void RigCalibration::AddSensor(sensor_t sensor_id,
+                               const Rigid3d& sensor_from_rig,
+                               bool is_fixed) {
+  if (NumSensors() == 0)
+    LOG(FATAL_THROW) << "The reference sensor needs to added first before any "
+                        "sensor being added.";
+  else {
+    if (HasSensor(sensor_id))
```
ahojnnes, line 178: THROW_CHECK(!HasSensor(sensor_id)) << "..."; ?

src/colmap/scene/frame.h
```diff
@@ -0,0 +1,270 @@
...
+  rig_t rig_id_;
+
+  // Reference sensor id which has the identity transformation to the rig.
+  sensor_t ref_sensor_id_;
+
+  // list of sensors
+  std::set<sensor_t> sensor_ids_;
+
+  // sensor_from_rig transformation.
+  std::map<sensor_t, Rigid3d> map_sensor_from_rig_;
```
ahojnnes, line 102:
```suggestion
  std::map<sensor_t, Rigid3d> sensors_from_rig_;
```

src/colmap/scene/frame.h
```diff
@@ -0,0 +1,270 @@
...
+                        bool is_fixed = false);
+
+  // Check whether the sensor exists in the rig
+  inline bool HasSensor(sensor_t sensor_id) const;
+
+  // Count the number of sensors available in the rig
+  inline size_t NumSensors() const;
+
+  // Access the reference sensor id (default to be the first added sensor)
+  inline sensor_t RefSensorId() const;
```
ahojnnes, line 82: Would it make sense to remove the need for a reference sensor and treat all of them the same?

src/colmap/scene/frame.h
```diff
@@ -0,0 +1,270 @@
...
+  // Access the unique identifier of the rig
+  inline rig_t RigId() const;
+  inline void SetRigId(rig_t rig_id);
+
+  // Add sensor into the rig
+  // ``AddReferenceSensor`` needs to called first before all the ``AddSensor``
+  // operation
+  inline void AddReferenceSensor(sensor_t ref_sensor_id);
+  inline void AddSensor(sensor_t sensor_id,
+                        const Rigid3d& sensor_from_rig = Rigid3d(),
```
ahojnnes, line 72: I suggest removing the default parameters here.

src/colmap/util/types.h
```diff
@@ -82,6 +82,12 @@ typedef uint32_t camera_t;
 // Unique identifier for images.
 typedef uint32_t image_t;
 
+// Unique identifier for frames.
+typedef uint64_t frame_t;
```
ahojnnes, line 86: uint32_t will be enough?

src/colmap/scene/image.h
```diff
@@ -200,9 +221,46 @@ point2D_t Image::NumPoints2D() const {
 
 point2D_t Image::NumPoints3D() const { return num_points3D_; }
 
-const Rigid3d& Image::CamFromWorld() const { return cam_from_world_; }
+frame_t Image::FrameId() const { return frame_id_; }
+
+void Image::SetFrameId(frame_t frame_id) { frame_id_ = frame_id; }
+
+const std::shared_ptr<class Frame> Image::Frame() const { return frame_; }
```
ahojnnes, line 228:
```suggestion
const std::shared_ptr<class Frame>& Image::Frame() const { return frame_; }
```

src/colmap/scene/frame.h
```diff
@@ -0,0 +1,270 @@
...
+
+  // Check whether the data id is existent in the frame
+  inline bool HasData(data_t data_id) const;
+
+  // Access the unique identifier of the rig
+  inline rig_t RigId() const;
+  inline void SetRigId(rig_t rig_id);
+
+  // Check if the frame has a non-trivial rig calibration
+  inline const std::shared_ptr<class RigCalibration> RigCalibration() const;
```
ahojnnes, line 125:
```suggestion
  inline const std::shared_ptr<class RigCalibration>& RigCalibration() const;
```

src/colmap/scene/frame.h
```diff
@@ -0,0 +1,270 @@
...
+#include "colmap/util/types.h"
+
+#include <map>
+#include <set>
+#include <vector>
+
+namespace colmap {
+
+// Sensor type
+enum class SensorType {
```
ahojnnes, line 42: Let's use our new enum class macro.

src/colmap/scene/frame.h
B1ueber2y: I think the abstraction of a reference sensor is useful for the following two reasons: 1) It enables minimal storage of the rig parameters, since we anyway will need a ``frame_from_world_`` variable. 2) In visual-inertial optimization it s always ideal to select the IMU as the reference sensor since they do not have the scale ambiguity.

src/colmap/scene/image.h
B1ueber2y: I ve made an update to resolve this: https://github.com/colmap/colmap/pull/2698/commits/885593b3f56675cc3fde032cc4987d25dd720fb2. Specifically, I added a ``CamFromWorldValue`` interface to support return by value rather than reference, where the cam from world can be composited from rig support. 

For unification in the future, we could potentially deprecate the ``cam_from_world`` member and always use the frame support, while using a trivial frame for a single image. However, this would require storing the pose information inside the frames that are stored independent from the images. I added a TODO [here](https://github.com/colmap/colmap/pull/2698/commits/885593b3f56675cc3fde032cc4987d25dd720fb2#diff-665386c3e6e1ebf0f2b3f383a2925c7a970b2e256f2f889f27fb090065158fbdR166)

src/colmap/scene/frame.h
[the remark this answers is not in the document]
B1ueber2y: Thanks. This is done: https://github.com/colmap/colmap/pull/2698/commits/63055e92e794a62a63ade40408fa2a355fd60ce2

src/colmap/scene/frame.h
B1ueber2y: I am not sure if I fully understand the proposal here. Does this mean that we do binary search (e.g., std::lower_bound) at indexing?

B1ueber2y:
I have tried to address all the comments, and will move forward once we are happy with the current design. The missing steps for this PR are:

- [ ] Adapt rig bundle adjuster
- [ ] Support easier IO from the user & store the Frame data and RigCalibration data somewhere (probably in the reconstruction)
- [ ] Support the new structs/classes from this PR in pycolmap

After this PR:
* Integrate IMU preintegration PR: https://github.com/colmap/colmap/pull/2625
* Rig-based incremental/global mapper
* Better data format prototyping

src/colmap/scene/frame.h
```diff
@@ -0,0 +1,337 @@
+// Copyright (c) 2023, ETH Zurich and UNC Chapel Hill.
```
ahojnnes, line 1: nit: remove 2023

src/colmap/scene/frame.h
```diff
@@ -0,0 +1,337 @@
...
+#include <vector>
+
+namespace colmap {
+
+// Sensor type
+MAKE_ENUM_CLASS_OVERLOAD_STREAM(SensorType,
+                                0,
+                                CAMERA,
+                                IMU,
+                                LOCATION  // include GNSS, radios, compass, etc.
```
ahojnnes, line 52: How is location different from GNSS?

src/colmap/scene/frame.h
```diff
@@ -0,0 +1,337 @@
...
+struct data_t {
+  sensor_t sensor_id;
+  uint64_t id;
+  data_t(const sensor_t& sensor_id, uint64_t id)
+      : sensor_id(sensor_id), id(id) {}
+
+  bool operator<(const data_t& other) const {
+    return std::tie(sensor_id, id) < std::tie(other.sensor_id, other.id);
+  }
+  bool operator==(const data_t& other) const {
```
ahojnnes, line 77: Does this need an "inline" to avoid duplicate symbol issues?

ahojnnes: Same for other methods.

src/colmap/scene/frame.h
```diff
@@ -0,0 +1,337 @@
...
+class RigCalibration {
+ public:
+  // Access the unique identifier of the rig
+  inline rig_t RigId() const;
+  inline void SetRigId(rig_t rig_id);
+
+  // Add sensor into the rig
+  // ``AddReferenceSensor`` needs to called first before all the ``AddSensor``
+  // operation
+  inline void AddReferenceSensor(sensor_t ref_sensor_id);
```
ahojnnes, line 98: nit
```suggestion
  inline void AddRefSensor(sensor_t ref_sensor_id);
```

src/colmap/scene/frame.h
ahojnnes: Since you already have std::map, the std::set seems redundant?

src/colmap/scene/frame.h
```diff
@@ -0,0 +1,337 @@
...
+    return std::tie(type, id) < std::tie(other.type, other.id);
+  }
+  bool operator==(const sensor_t& other) const {
+    return type == other.type && id == other.id;
+  }
+};
+
+struct data_t {
+  sensor_t sensor_id;
+  uint64_t id;
```
ahojnnes, line 70: This can be any of image_t, ... ? Will be good to add some comments above the fields to explain what they mean.

src/colmap/scene/frame.h
```diff
@@ -0,0 +1,337 @@
...
+void RigCalibration::AddReferenceSensor(sensor_t ref_sensor_id) {
+  THROW_CHECK(sensor_ids_.empty());  // The reference sensor must be added first
+  ref_sensor_id_ = ref_sensor_id;
+  sensor_ids_.insert(ref_sensor_id);
+}
+
+void RigCalibration::AddSensor(sensor_t sensor_id,
+                               const Rigid3d& sensor_from_rig) {
+  if (NumSensors() == 0)
+    LOG(FATAL_THROW) << "The reference sensor needs to added first before any "
```
ahojnnes, line 203: Let's throw an exception instead for consistency.

src/colmap/scene/frame.h
```diff
@@ -0,0 +1,337 @@
...
+                        "sensor being added.";
+  else {
+    THROW_CHECK(!HasSensor(sensor_id))
+        << StringPrintf("Sensor id (%d, %d) is inserted twice into the rig",
+                        sensor_id.type,
+                        sensor_id.id);
+    sensors_from_rig_.emplace(sensor_id, sensor_from_rig);
```
ahojnnes, line 210: Things would be easier if we added an identity transform for the ref sensor here?

B1ueber2y: With the identity transformation we always need to composite the cam_from_world and wont be able to get the reference anymore. See the other discussions on "Would it make sense to remove the need for a reference sensor and treat all of them the same?" There are quite some benefits to have a reference sensor.

src/colmap/scene/frame.h
B1ueber2y: Nice catch. This list is removed.

src/colmap/scene/frame.h
B1ueber2y: Yes. I ve added the comments.

src/colmap/scene/frame.h
ahojnnes: Thanks, makes sense.

src/colmap/scene/image.h
```diff
@@ -202,33 +225,115 @@ point2D_t Image::NumPoints2D() const {
...
+
+bool Image::HasFrame() const { return frame_ != nullptr; }
+
+bool Image::HasNonTrivialFrame() const {
+  return HasFrame() && frame_->HasRigCalibration() &&
+         !frame_->RigCalibration()->IsReference(
+             sensor_t(SensorType::CAMERA, CameraId()));
+}
+
+Rigid3d Image::CamFromWorldValue() const {
```
ahojnnes, line 246: nit: maybe better?
```suggestion
Rigid3d Image::CopyCamFromWorld() const {
```

src/colmap/scene/image.h
```diff
@@ -202,33 +225,115 @@ point2D_t Image::NumPoints2D() const {
...
+  if (HasFrame()) {
+    frame_->SetFrameFromWorld(cam_from_world);
+  } else {
+    cam_from_world_ = cam_from_world;
+  }
 }
 
-bool Image::HasPose() const { return cam_from_world_.has_value(); }
+bool Image::HasPose() const {
+  if (HasFrame())
```
ahojnnes, line 325: Instead of switching here at runtime, can we just introduce a frame for every image?

B1ueber2y: Sounds good. We could initialize a trivial frame whenever an image is created. And then switch the frame by ``SetFrame`` if we have the rig setup.

B1ueber2y: Done here: https://github.com/colmap/colmap/pull/2698/commits/05f739763620d4c6183ec605a72f5a80b4bc7f0a. Also I could remove the TODO I added exactly for this. Now we dont need to switch and there is no cam_from_world_ member in the image anymore. All the interfaces remain unchanged and the tests have passed. I have very good feelings with the current design.

B1ueber2y: PTAL

src/colmap/scene/frame.h
B1ueber2y: I am closing this as this was addressed in the other discussion.

src/colmap/scene/frame.cc
```diff
@@ -0,0 +1,56 @@
...
+      << "The reference sensor needs to added first before any "
+         "sensor being added.";
+  THROW_CHECK(!HasSensor(sensor_id))
+      << StringPrintf("Sensor id (%d, %d) is inserted twice into the rig",
+                      sensor_id.type,
+                      sensor_id.id);
+  sensors_from_rig_.emplace(sensor_id, sensor_from_rig);
+}
+
+Frame::Frame() {}
```
ahojnnes, line 54: nit: could be moved to header `Frame() = default;`. Same for RigCalibration above.

src/colmap/scene/frame.h
```diff
@@ -0,0 +1,338 @@
...
+    return std::tie(type, id) < std::tie(other.type, other.id);
+  }
+  inline bool operator==(const sensor_t& other) const {
+    return type == other.type && id == other.id;
+  }
+  inline bool operator!=(const sensor_t& other) const {
+    return !(*this == other);
+  }
+};
+const sensor_t kInvalidSensorId =
```
ahojnnes, line 68: nit: static constexpr ?

src/colmap/scene/frame.h
```diff
@@ -0,0 +1,338 @@
...
+    data_t(kInvalidSensorId, std::numeric_limits<uint32_t>::max());
+
+// Rig calibration storing the sensor from rig transformation.
+// The reference sensor shares identity poses with the device.
+// This design is mainly for two purposes:
+// 1) In the visual-inertial optimization one of the IMUs is generally used as
+// the reference frame since it is metric.
+// 2) Not having a reference frame brings a 6 DoF Gauge for each rig, which is
+// not ideal particularly when it comes to covariance estimation.
+class RigCalibration {
```
ahojnnes, line 100: Some basic unit tests for this and the frame class will be great.

ahojnnes (approved):
Looks great to me, some final comments from my side.

In terms of next steps, I was hoping that, in addition to the IMU/frame BA, we can also adjust database and reconstruction I/O and then also incremental mapper to support rigs. This will be a longer term project. It may make sense to catch up in person before embarking on this journey. :-)

src/colmap/scene/image.h
ahojnnes: No strong feelings but slight preference for existing "copy".

src/colmap/scene/frame.h
B1ueber2y: I added some basic tests. Thanks!

src/colmap/scene/image.h
B1ueber2y: This is not really a copy. When rig (non-trivial frame) is available ``CamFromWorldValue()`` will return a composited pose with ``frame_from_world`` and ``sensor_from_rig``. So I dont feel like using the word copy. I am keeping the ``CamFromWorldValue()`` now (until we come up with a better one).

B1ueber2y:
> Looks great to me, some final comments from my side.
> 
> In terms of next steps, I was hoping that, in addition to the IMU/frame BA, we can also adjust database and reconstruction I/O and then also incremental mapper to support rigs. This will be a longer term project. It may make sense to catch up in person before embarking on this journey. :-)

Thanks for all the comments. I have just addressed them all and added some tests. 

Actually after thinking about it more, the next step would actually already be to adjust database and reconstruction I/O rather than to support the BA. We would love to store the rig calibration (``std::map<rig_t, std::shared_ptr<RigCalibration>>``)  and all the frames (``std::map<frame_t, std::shared_ptr<Frame>>``) in the reconstruction. This will also result in full unification of rig and normal setup on the data structure level. However, this I/O will be made easier when the proto is available. What s your take on this?

On the other hand, do you think we should first merge this PR as a quote and then develop the I/O and BA, or do it in separate PRs one after another. 

Yes. Supporting incremental mapper and global mapper on top of rigs will be very exciting but effort demanding. Lets find some time to catch up in person and discuss in more details : )

ahojnnes:
As for next steps, I suggest decoupling changes to I/O format from rig support, unless we think that it's fundamentally not possible to do this with the existing custom format. I am all in for protobuf serialization but want to be very careful here in the sense that I don't want to break existing users. Ideally, we can be fully backwards compatible to existing models. I guess we could create a new frames.bin file and a new rigs.bin file. The existing cameras/images/points3D could stay identical?

src/colmap/scene/image.h
ahojnnes: How about `ComposeCamFromWorld` ?

src/colmap/scene/frame.h
```diff
@@ -0,0 +1,419 @@
...
+    return sensor_id == other.sensor_id && id == other.id;
+  }
+  inline bool operator!=(const data_t& other) const {
+    return !(*this == other);
+  }
+};
+constexpr data_t kInvalidDataId =
+    data_t(kInvalidSensorId, std::numeric_limits<uint32_t>::max());
+
+// Rig calibration storing the sensor from rig transformation.
```
ahojnnes, line 94: Last nit: Does it make sense to split this into a separate rig_calib.h/.cc file?

B1ueber2y: Thanks. This makes sense. I put all implementations regarding ``sensor_t`` and ``RigCalibration`` into a separate file at ``sensor`` folder instead.

src/colmap/scene/image.h
B1ueber2y: > How about `ComposeCamFromWorld` ?

Sounds good. This is updated.

B1ueber2y:
> As for next steps, I suggest decoupling changes to I/O format from rig support, unless we think that it's fundamentally not possible to do this with the existing custom format. I am all in for protobuf serialization but want to be very careful here in the sense that I don't want to break existing users. Ideally, we can be fully backwards compatible to existing models. I guess we could create a new frames.bin file and a new rigs.bin file. The existing cameras/images/points3D could stay identical?

Sounds good. We can do either of the following two regarding the frame association:

1) keep the cameras/images/points3D fully identical as before, and write into frames.bin and rigs.bin only when there are non-trivial frame available. At reading, we check if there are frames.bin and rigs.bin available in the same directory. If so we read in the frames/rigs and update the corresponding images. The benefit for this is that it will be fully backwards compatible, but the drawback is that from images.bin itself we will have no idea about the rig/frame information.

2) store an additional frame id field in the image. At reading, when the frame id is valid we look for frames.bin and rigs.bin. Contrary to 1), the benefit is that from images.bin we already have the frame correspondence, while the drawback is slightly different file format from before and broken reading when users use the previous version of COLMAP to read into the new format.

ahojnnes:
I think it is fine to let images have no idea about their association to rigs/frames. Frames are the concept that ties together sensor measurements (ie images) and rig calibrations. Do you see a problem with that other than having to read frame and rig files first, so we can correctly populate the data structures?

B1ueber2y:
> I think it is fine to let images have no idea about their association to rigs/frames. Frames are the concept that ties together sensor measurements (ie images) and rig calibrations. Do you see a problem with that other than having to read frame and rig files first, so we can correctly populate the data structures?

Sounds good. This will the next step, and should not be that difficult. I will find time for this but probably only next week. 

The short-term roadmap for this will be:
* add frame and rig calib to Reconstruction class and support I/O with custom format
* support rig bundle adjuster with the new Reconstruction class and deprecate RigBundleAdjuster
* integrate the other PR on IMU sensor and preintegration factors

We could discuss about long-term rig sfm offline : )

PR #2700, opened 2024-08-10
Export mac binaries in CI

ahojnnes:


.github/workflows/build-mac.yml
```diff
@@ -21,6 +21,7 @@ jobs:
             os: macos-14,
             arch: arm64,
```
sarlinpe, line 22: Very cool! This creates a binary for arm64 only, would it make sense to also create a binary for intel machines? We can have another config with a `macos-13` runner.

sarlinpe (approved):

.github/workflows/build-mac.yml
ahojnnes: One problem I have is that the binary doesn't seem to work under arm64. The changes in this PR work under x86 on my 10 year old macbook pro. If you have some minutes to investigate what's going wrong under arm64, that would be great. I don't have access to an arm64 machine.

sarlinpe: I also don't have an arm64 machine at hand for now. Could we instead test the binary in a fresh arm64 runner in the CI?

PR #2702, opened 2024-08-11
Abstract BaseImage and add raw camera pointer to the Image class

B1ueber2y:
Reopening https://github.com/colmap/colmap/pull/2697.

This PR adds support for directly indexing cameras in the image class with backward compatibility, as one of the steps for enabling better rig design. 

Specifically, the PR is structured as follows:

- Abstract a ``BaseImage`` class similar to the ``Image`` class before, make the ``Image`` class a child class of ``BaseImage`` with a raw camera pointer. 

- Force the raw camera pointer when constructing an ``Image`` object. The ``Image`` object can also be constructed from a ``BaseImage`` object and a camera pointer, with the check that the camera id is consistent. Delete the method ``SetCameraId`` in the ``Image`` class. 

- Adapt the changes across COLMAP. In particular, the use of ``BaseImage`` is kept throughout database and ui, while ``Image`` is used in ``DatabaseCache`` and ``Reconstruction``. In the reconstruction, checks are made to make sure that: 1) images must be read after cameras. 2) ``AddImage`` must be done with the camera present in the reconstruction. Interfaces are kept unchanged. Tests are updated as well.

- As in https://github.com/colmap/colmap/pull/2697, we can add more interesting interfaces on the ``Image`` class (e.g., projecting a 3D point onto pixels), now without the need of any additional checks since the raw camera pointer is required to be initialized at the construction. 

- Update pybind accordingly. Note that only the ``BaseImage`` class can be made a data class since the ``Image`` class does not support IO without the cameras available.

pybind test example:
```
import pycolmap
image = recon.images[1]

print(image)
# Image(image_id=1, camera=Camera(camera_id=1), name="0000.png", triangulated=3633/9946)
print(isinstance(image, pycolmap.BaseImage)) # inheritance
# True
print(image.cast_to_base())
# BaseImage(image_id=1, camera_id=1, name="0000.png", triangulated=3633/9946)

print(image.camera) 

print(image.camera_id)
# 1
image.camera_id = 2
# E20240811 14:02:57.401376 2366510 image.cc:206] Error! The ``camera_id`` property can no longer be directly set in the Image class since it now comes from ``image.camera``. Update the ``camera`` property instead.
*** ValueError: [image.cc:206] Error! The ``camera_id`` property can no longer be directly set in the Image class since it now comes from ``image.camera``. Update the ``camera`` property instead.

image.camera = recon.cameras[2]
print(image.camera_id)
# 2
print(recon.images[1].camera_id)
# 2
```

### DIFF
--- a/src/colmap/scene/image.h
+++ b/src/colmap/scene/image.h
@@ -68,8 +68,8 @@ class Image {
   // Check whether identifier of camera has been set.
   inline bool HasCamera() const;
   // [Optional]: Access the shared pointer of the camera
-  inline std::shared_ptr<struct Camera> Camera() const;
-  inline void SetCamera(const std::shared_ptr<struct Camera>& camera);
+  inline std::shared_ptr<struct Camera> CameraPtr() const;
+  inline void SetCameraPtr(const std::shared_ptr<struct Camera>& camera);
   // check whether the camera pointer has been set
   inline bool HasCameraPtr() const;
 
@@ -110,6 +110,11 @@ class Image {
 
   // Extract the viewing direction of the image.
   Eigen::Vector3d ViewingDirection() const;
+
+  // Reproject the 3D point onto the image in pixels (only when the camera
+  // pointer is available) Return false if the 3D point is behind the camera.
+  std::pair<bool, Eigen::Vector2d> ProjectPoint3D(
+      const Eigen::Vector3d& point3D) const;
 
  private:
   // Identifier of the image, if not specified `kInvalidImageId`.
@@ -161,9 +166,11 @@ inline void Image::SetCameraId(const camera_t camera_id) {
 
 inline bool Image::HasCamera() const { return camera_id_ != kInvalidCameraId; }
 
-inline std::shared_ptr<struct Camera> Image::Camera() const { return camera_; }
+inline std::shared_ptr<struct Camera> Image::CameraPtr() const {
+  return camera_;
+}
 
-inline void Image::SetCamera(const std::shared_ptr<struct Camera>& camera) {
+inline void Image::SetCameraPtr(const std::shared_ptr<struct Camera>& camera) {
   THROW_CHECK_NE(camera->camera_id, kInvalidCameraId);
   camera_ = camera;
   camera_id_ = camera->camera_id;
--- a/src/colmap/scene/image.h
+++ b/src/colmap/scene/image.h
@@ -113,7 +113,7 @@ class Image {
 
   // Reproject the 3D point onto the image in pixels (only when the camera
   // pointer is available) Return false if the 3D point is behind the camera.
-  std::pair<bool, Eigen::Vector2d> ProjectPoint3D(
+  std::pair<bool, Eigen::Vector2d> ProjectPoint(
       const Eigen::Vector3d& point3D) const;
 
  private:
--- a/src/colmap/scene/image.h
+++ b/src/colmap/scene/image.h
@@ -67,6 +67,11 @@ class Image {
   inline void SetCameraId(camera_t camera_id);
   // Check whether identifier of camera has been set.
   inline bool HasCamera() const;
+  // [Optional]: Access the shared pointer of the camera
+  inline std::shared_ptr<struct Camera> Camera() const;
+  inline void SetCamera(const std::shared_ptr<struct Camera>& camera);
+  // check whether the camera pointer has been set
+  inline bool HasCameraPtr() const;
 
   // Check if image is registered.
   inline bool IsRegistered() const;
@@ -116,6 +121,8 @@ class Image {
   // The identifier of the associated camera. Note that multiple images might
   // share the same camera. If not specified `kInvalidCameraId`.
   camera_t camera_id_;
+  // [Optional] the shared pointer of the camera
+  std::shared_ptr<struct Camera> camera_ = nullptr;
 
   // Whether the image is successfully registered in the reconstruction.
   bool registered_;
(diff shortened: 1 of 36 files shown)


B1ueber2y:
nit: If I can choose I would personally rather call ``BaseImage`` and ``Image`` to be ``Image`` and ``View`` instead (as I did in limap). But I totally understand that this will be a much too huge name changes that can affect everything : )

src/colmap/scene/reconstruction.cc
```diff
@@ -116,7 +116,20 @@ void Reconstruction::AddCamera(struct Camera camera) {
   THROW_CHECK(cameras_.emplace(camera_id, std::move(camera)).second);
 }
 
+void Reconstruction::AddImage(const BaseImage& base_image) {
```
ahojnnes, line 119: For the move to work, the const needs to disappear.

src/colmap/scene/reconstruction.h
```diff
@@ -102,7 +102,9 @@ class Reconstruction {
   // might be taken by the same camera.
   void AddCamera(struct Camera camera);
 
-  // Add new image.
+  // Add new image. The corresponding camera needs to be existent in the
```
ahojnnes, line 105:
```suggestion
  // Add new image. The corresponding camera needs to exist in the
```
nit

src/pycolmap/scene/image.cc
```diff
@@ -52,80 +65,94 @@ std::shared_ptr<Image> MakeImage(const std::string& name,
...
+                                 const std::vector<T>& points2D,
+                                 const Rigid3d& cam_from_world,
+                                 image_t image_id) {
+  auto image = std::make_shared<Image>(camera);
+  image->SetName(name);
+  image->SetPoints2D(points2D);
+  image->CamFromWorld() = cam_from_world;
+  image->SetImageId(image_id);
+  return image;
+}
```
ahojnnes, line 80:
```suggestion
std::shared_ptr<Image> MakeImage(Camera* camera,
                                 std::string name,
                                 std::vector<T> points2D,
                                 Rigid3d cam_from_world,
                                 image_t image_id) {
  auto image = std::make_shared<Image>(camera);
  image->SetName(std::move(name));
  image->SetPoints2D(std::move(points2D));
  image->CamFromWorld() = std::move(cam_from_world);
  image->SetImageId(image_id);
  return image;
}
```

src/colmap/scene/image_test.cc
```diff
@@ -35,7 +35,7 @@ namespace colmap {
 namespace {
```
ahojnnes, line 36: It will be great to have some minimal tests for the new Image class.

src/colmap/scene/image.h
```diff
@@ -131,53 +131,92 @@ class Image {
...
+  void SetCameraId(camera_t camera_id) = delete;
+
+  // Reproject the 3D point onto the image in pixels (only when the camera
+  // pointer is available) Return false if the 3D point is behind the camera.
+  std::pair<bool, Eigen::Vector2d> ProjectPoint(
+      const Eigen::Vector3d& point3D) const;
+
+ private:
+  // The address of the corresponding camera
+  struct Camera* camera_;
```
ahojnnes, line 157: Hm, after looking at the code, I am not sure I would still go with the raw pointer but rather with the shared_ptr. There is no callsite where we would not use a shared_ptr for the camera?

sarlinpe: @ahojnnes How would you pass an element of `Reconstruction::cameras_` (unordered_map) as a `shared_ptr`?

@B1ueber2y is it legal to pass a NULLPTR camera? if not, 1) do you check for it? and 2) shouldn't `camera_` be a reference instead of a pointer?

sarlinpe:
Thanks for all this work. I am concerned by the increase in complexity that this update brings.
1. What was the issue with making the camera optional, as originally proposed in https://github.com/colmap/colmap/pull/2697 ? wouldn't this still be simpler than the added abstraction?
2. Alternative, can't we always ensure that a camera is available when creating an image? for example, in `Database::ReadImage`, also read and return the corresponding camera? would this be too inefficient?

src/colmap/scene/image.h
B1ueber2y: I agree that shared_ptr is not working well here. Since across COLMAP cameras are not initialized as shared pointers.

I didnt check for nullptr and could add it. And also, I agree that putting ``camera_`` as a reference is much better, and will make the updates. Thanks.

src/pycolmap/scene/image.cc
B1ueber2y: This implementation was kept consistent with https://github.com/colmap/colmap/blob/main/src/pycolmap/scene/image.cc#L39. Also May I ask what the issue is for the const reference here?

src/colmap/scene/image_test.cc
B1ueber2y: Added here: https://github.com/colmap/colmap/pull/2702/commits/9a4c478a51a1e772fe3ac511b8ea9e69a955e3f7

ahojnnes:
> What was the issue with making the camera optional, as originally proposed in https://github.com/colmap/colmap/pull/2697 ? wouldn't this still be simpler than the added abstraction?

I appreciate the effort but also liked the original approach better for its simplicity.

> Alternative, can't we always ensure that a camera is available when creating an image? for example, in Database::ReadImage, also read and return the corresponding camera? would this be too inefficient?

I think the overhead would not be measurable. If really needed, it could even be combined into a single efficient SQL query.

PR #2710, opened 2024-08-14
Fix table mangled by clang-format

sweber1:
Noticed a table in exe/feature.h was mangled. Happened when clang-format options where simplified in https://github.com/colmap/colmap/pull/1950.

This reverts it to what it used to look like: https://github.com/colmap/colmap/blob/3947f67243e4924683e14481b4dbb44593950c52/src/colmap/exe/feature.h#L44-L52 and turns off clang-format for those lines.


ahojnnes (approved):

ahojnnes:
Thanks!

Issue #2703, opened 2024-08-12
Why does COLMAP set the Ceres solver tolerance value to 0 in the bundle adjustment option?


juu-k:
Hello, I was looking at the COLMAP options and noticed that the function_tolerance, gradient_tolerance, and parameter_tolerance values for the Ceres solver used in the Bundle Adjustment step are all set to zero.

![image](https://github.com/user-attachments/assets/de62e0ac-1a9d-4d8a-82d4-a8da2542e1d2)

However, the initial values in the link below have a very small value close to 0 (ex. 1e-6, 1e-10). Why is COLMAP set to 0?

http://ceres-solver.org/nnls_solving.html#_CPPv4N5ceres6Solver7Options18function_toleranceE

sarlinpe:
The gradient and parameter tolerances are overridden for the local and global BA: 
https://github.com/colmap/colmap/blob/6247ca0673b28554cd10b89cbac4dfe0b9ec9837/src/colmap/controllers/incremental_mapper.cc#L104-L105
https://github.com/colmap/colmap/blob/6247ca0673b28554cd10b89cbac4dfe0b9ec9837/src/colmap/controllers/incremental_mapper.cc#L129-L130
Function tolerances are indeed set to zero:
https://github.com/colmap/colmap/blob/6247ca0673b28554cd10b89cbac4dfe0b9ec9837/src/colmap/controllers/incremental_mapper.h#L94
https://github.com/colmap/colmap/blob/6247ca0673b28554cd10b89cbac4dfe0b9ec9837/src/colmap/controllers/incremental_mapper.h#L106

@ahojnnes  I think that we really need a more extensive suite of benchmarks to tune these parameters and make sure that increasing their values reduces the time and doesn't degrade the performance. I suspect that the parameter tolerance could help most.

ahojnnes:
@sarlinpe Agreed that we need such a benchmark. These have all been tuned in a rather empirical manner many years ago, but meanwhile many of the underlying algorithms and implementations have also changed quite drastically. It will be good to revisit these.

PR #2713, opened 2024-08-15
Write out options to ini in full precision, relax bundle adjuster convergence

ahojnnes:
Addresses https://github.com/colmap/colmap/issues/2703

* Make sure we write out parameters in full precision, so small values do not get truncated to 0.
* Conservatively relax the bundle adjustment convergence thresholds. Before this PR, bundle adjustment would always run for the full iterations because of 0 tolerances.

### DIFF
--- a/src/colmap/estimators/bundle_adjustment.h
+++ b/src/colmap/estimators/bundle_adjustment.h
@@ -74,7 +74,7 @@ struct BundleAdjustmentOptions {
 
   BundleAdjustmentOptions() {
     solver_options.function_tolerance = 0.0;
-    solver_options.gradient_tolerance = 0.0;
+    solver_options.gradient_tolerance = 1e-4;
     solver_options.parameter_tolerance = 0.0;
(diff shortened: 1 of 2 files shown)


src/colmap/estimators/bundle_adjustment.h
```diff
@@ -74,7 +74,7 @@ struct BundleAdjustmentOptions {
 
   BundleAdjustmentOptions() {
     solver_options.function_tolerance = 0.0;
-    solver_options.gradient_tolerance = 0.0;
+    solver_options.gradient_tolerance = 1e-4;
```
ahojnnes, line 77: The ceres default is 1e-10. colmap default for incremental global BA is 1.0 and for local BA is 10.0. For custom global BA convergence, the 1e-4 should be a conservative value. I tested this on a few scenes and it converges before reaching the full iterations.

B1ueber2y (approved):

PR #2715, opened 2024-08-15
Add optional raw camera pointer to image class

B1ueber2y:
Reverting to the design choice of https://github.com/colmap/colmap/pull/2697 while dropping the design from https://github.com/colmap/colmap/pull/2702.

With the camera pointer being optional, the best design IMO for the reconstruction would be still be keeping the ``SyncCameraPointers`` method. So the overall design from https://github.com/colmap/colmap/pull/2697 is kept here while only updating the shared pointer to raw pointer. 

Example following https://github.com/colmap/colmap/pull/2697:
```
import pycolmap
import numpy as np
image = recon.images[1]
print(image.camera) 

print(image)
# Image(image_id=1, camera=Camera(camera_id=1), name="0000.png", triangulated=3620/9946)
print(image.has_camera(), image.has_camera_ptr())
# True True
image.reset_camera_ptr()
print(image)
# Image(image_id=1, camera_id=1, name="0000.png", triangulated=3620/9946)
print(image.has_camera(), image.has_camera_ptr())
# True False
```

### DIFF
--- a/src/colmap/scene/image.h
+++ b/src/colmap/scene/image.h
@@ -67,6 +67,11 @@ class Image {
   inline void SetCameraId(camera_t camera_id);
   // Check whether identifier of camera has been set.
   inline bool HasCamera() const;
+  // [Optional]: Access the shared pointer of the camera
+  inline std::shared_ptr<struct Camera> CameraPtr() const;
+  inline void SetCameraPtr(const std::shared_ptr<struct Camera>& camera);
+  // check whether the camera pointer has been set
+  inline bool HasCameraPtr() const;
 
   // Check if image is registered.
   inline bool IsRegistered() const;
@@ -106,6 +111,11 @@ class Image {
   // Extract the viewing direction of the image.
   Eigen::Vector3d ViewingDirection() const;
 
+  // Reproject the 3D point onto the image in pixels (only when the camera
+  // pointer is available) Return false if the 3D point is behind the camera.
+  std::pair<bool, Eigen::Vector2d> ProjectPoint(
+      const Eigen::Vector3d& point3D) const;
+
  private:
   // Identifier of the image, if not specified `kInvalidImageId`.
   image_t image_id_;
@@ -116,6 +126,8 @@ class Image {
   // The identifier of the associated camera. Note that multiple images might
   // share the same camera. If not specified `kInvalidCameraId`.
   camera_t camera_id_;
+  // [Optional] the shared pointer of the camera
+  std::shared_ptr<struct Camera> camera_ = nullptr;
 
   // Whether the image is successfully registered in the reconstruction.
   bool registered_;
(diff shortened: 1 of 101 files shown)


src/colmap/scene/image.h
```diff
@@ -149,11 +163,29 @@ inline camera_t Image::CameraId() const { return camera_id_; }
...
 
 inline bool Image::HasCamera() const { return camera_id_ != kInvalidCameraId; }
 
+inline struct Camera* Image::CameraPtr() const {
+  THROW_CHECK(HasCameraPtr());
+  return camera_;
+}
+
+inline void Image::SetCameraPtr(struct Camera* camera) {
+  THROW_CHECK_NE(camera->camera_id, kInvalidCameraId);
```
ahojnnes, line 180: THROW_CHECK_NOTNULL(camera);

src/colmap/scene/image.h
```diff
@@ -149,11 +163,29 @@ inline camera_t Image::CameraId() const { return camera_id_; }
...
+  camera_ = camera;
+  camera_id_ = camera->camera_id;
```
ahojnnes, line 182: This is technically not needed anymore now?

src/colmap/scene/image.h
```diff
@@ -149,11 +163,29 @@ inline camera_t Image::CameraId() const { return camera_id_; }
...
+}
+
+inline void Image::UnsetCameraPtr() { camera_ = nullptr; }
```
ahojnnes, line 185: nit: ResetCameraPtr ?

src/colmap/scene/image.h
ahojnnes, line 182: THROW_CHECK_EQ(camera_id_, camera->camera_id) ?

ahojnnes: Similarly, in the SetCameraId() method, we probably want to do `if (HasCameraPtr()) { THROW_CHECK_EQ(camera_id_, camera->camera_id)` ?

src/colmap/scene/reconstruction.h
```diff
@@ -102,8 +102,8 @@ class Reconstruction {
   // might be taken by the same camera.
   void AddCamera(struct Camera camera);
 
-  // Add new image.
-  void AddImage(class Image image);
+  // Add new image. optionally load camera address in the image
```
ahojnnes, line 105:
```suggestion
  // Add new image. Optionally load camera address in the image.
```
nit

src/pycolmap/scene/image.cc
```diff
@@ -26,10 +26,15 @@ std::string PrintImage(const Image& image) {
   std::stringstream ss;
   ss << "Image(image_id="
      << (image.ImageId() != kInvalidImageId ? std::to_string(image.ImageId())
-                                            : "Invalid")
-     << ", camera_id="
-     << (image.HasCamera() ? std::to_string(image.CameraId()) : "Invalid")
-     << ", name=\"" << image.Name() << "\""
+                                            : "Invalid");
+  if (!image.HasCameraPtr()) {
```
ahojnnes, line 30: Assuming that the camera ptr and the id are always consistent, we would not need this branch here?

B1ueber2y:
Once we are happy with the class design, I could try to start removing the use of ``colmap::CameraRig`` and visual-only ``RigBundleAdjuster`` with the new ``Frame`` class, as a better starting point for multi-sensor rig development.

src/colmap/scene/image.h
B1ueber2y: There might be some confusion here.

The idea of the design is that ``camera_`` always has higher priority than ``camera_id_``, so the ``camera_id`` should change accordingly to the corresponding camera. Thus, ``SetCameraId()`` is disabled when ``camera_`` is present (see here): https://github.com/colmap/colmap/pull/2715/files#diff-665386c3e6e1ebf0f2b3f383a2925c7a970b2e256f2f889f27fb090065158fbdR166

src/pycolmap/scene/image.cc
B1ueber2y: Here I update the printing to show to the user if the optional camera ptr is set or not. Check the python example in the PR description (the printing is slightly different)

Issue #2643, opened 2024-07-09
CUDA-accelerated BA / sparse solvers


pwais:
**Is your feature request related to a problem? Please describe.**
An early experiment showed that there was large speed-up using _dense_ solvers https://github.com/colmap/colmap/pull/2161

The goal is to leverage GPU to speed up BA solving in general.

**Describe the solution you'd like**
Recently Ceres merged cuDSS support for sparse Cholesky https://github.com/ceres-solver/ceres-solver/commit/6fb3dae4eeef855568e47ebbb29a8ba4f3c9153f   and the unit tests show an example of CUDA sparse Bundle Adjustment: https://github.com/ceres-solver/ceres-solver/commit/6fb3dae4eeef855568e47ebbb29a8ba4f3c9153f#diff-dc637bf6d41e3df3a877146387833e8147c045b9b0476a1c55b0f3f0ffc12a0fR49 

Perhaps these new changes could be adopted into COLMAP

**Additional context**
Another PR aiming at faster BA: https://github.com/colmap/colmap/pull/2567

sarlinpe:
I agree that this would be a valuable addition, feel free to give it a try if you have time. We could start from https://github.com/colmap/colmap/pull/2161 and support the sparse solvers as well (which will be available only in the next version of Ceres). I would automatically fall back to the CPU version is the installed Ceres version is not sufficiently high.

ahojnnes:
Thank you @dkargin. There is also quite a bit of literature on the topic, including the previously in colmap supported PBA backend: https://grail.cs.washington.edu/projects/mcba/pba.pdf

It could be interesting to see the speedup of just plugging in the new ceres sparse cuda solver and evaluate if and what speedup we could gain with such a simple change without any other customization.

whuaegeanse:
> 1. All calculations must be done on GPU. GPU-CPU transfers are very expensive compared to calculation speed. Both linear solver (PCG, Schur+PCG, ILU+PCG, whatever) and evaluation of residuals and jacobians and hessian stuff must be done exclusively on GPU.

For very large problems, such as more than 10,000 images, is the memory on GPU large enough?

[...]

PR #2717, opened 2024-08-16
Add support for CUDA sparse BA solver

ahojnnes:
This PR implements enhancement: https://github.com/colmap/colmap/issues/2643. It builds upon Ceres' recent CUDA_SPARSE solver type.

Initial experiments show significant runtime improvements. On my machine with an Intel Core i9 10920X and an NVidia RTX 2070, I see a consistent 3x speedup for reconstructions with ~500-5000 images. For smaller problems with ~100 images, the runtime is roughly equivalent. These experiments were done using CUDA 12.5 and cudss 0.3.0.

For now, the feature is disabled by default and requires explicit enabling of the option. This is because no robustness is implemented against situations where the GPU does not have enough memory and some of the thresholds that determine the usage of sparse direct vs. indirect solvers need to be tuned in this new scenario.

### DIFF
--- a/src/colmap/estimators/bundle_adjustment.cc
+++ b/src/colmap/estimators/bundle_adjustment.cc
@@ -346,7 +346,7 @@ ceres::Solver::Options BundleAdjuster::SetUpSolverOptions(
     solver_options.linear_solver_type = ceres::SPARSE_SCHUR;
 #if (CERES_VERSION_MAJOR >= 3 ||                                \
      (CERES_VERSION_MAJOR == 2 && CERES_VERSION_MINOR >= 2)) && \
-    !defined(CERES_NO_CUDSS)
+    !defined(CERES_NO_CUDSS) && defined(CUDA_ENABLED)
     if (options_.use_gpu) {
       const std::vector<int> gpu_indices = CSVToVector<int>(options_.gpu_index);
       THROW_CHECK_GT(gpu_indices.size(), 0);
--- a/src/colmap/estimators/bundle_adjustment.cc
+++ b/src/colmap/estimators/bundle_adjustment.cc
@@ -344,7 +344,9 @@ ceres::Solver::Options BundleAdjuster::SetUpSolverOptions(
   } else if (num_images <= options_.max_num_images_direct_sparse_solver &&
              has_sparse) {
     solver_options.linear_solver_type = ceres::SPARSE_SCHUR;
-#if !defined(CERES_NO_CUDSS)
+#if (CERES_VERSION_MAJOR >= 3 ||                                \
+     (CERES_VERSION_MAJOR == 2 && CERES_VERSION_MINOR >= 2)) && \
+    !defined(CERES_NO_CUDSS)
     if (options_.use_gpu) {
       const std::vector<int> gpu_indices = CSVToVector<int>(options_.gpu_index);
       THROW_CHECK_GT(gpu_indices.size(), 0);
--- a/src/colmap/estimators/bundle_adjustment.cc
+++ b/src/colmap/estimators/bundle_adjustment.cc
@@ -33,6 +33,7 @@
 #include "colmap/estimators/manifold.h"
 #include "colmap/scene/projection.h"
 #include "colmap/sensor/models.h"
+#include "colmap/util/cuda.h"
 #include "colmap/util/misc.h"
 #include "colmap/util/threading.h"
 #include "colmap/util/timer.h"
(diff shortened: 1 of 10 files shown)


ahojnnes:
@S-o-T FYI, since I believe you added this new feature to Ceres.

src/colmap/estimators/bundle_adjustment.cc
```diff
@@ -329,17 +330,30 @@ ceres::Solver::Options BundleAdjuster::SetUpSolverOptions(
...
+  if (num_images <= options_.max_num_images_direct_dense_solver) {
     solver_options.linear_solver_type = ceres::DENSE_SCHUR;
-  } else if (num_images <= kMaxNumImagesDirectSparseSolver && has_sparse) {
+  } else if (num_images <= options_.max_num_images_direct_sparse_solver &&
+             has_sparse) {
     solver_options.linear_solver_type = ceres::SPARSE_SCHUR;
+#if (CERES_VERSION_MAJOR >= 3 ||                                \
+     (CERES_VERSION_MAJOR == 2 && CERES_VERSION_MINOR >= 2)) && \
+    !defined(CERES_NO_CUDSS) && defined(CUDA_ENABLED)
+    if (options_.use_gpu) {
```
S-o-T, line 350: I think that it might be worth experimentation/tuning to find problem size threshold to switch from CPU to GPU sparse backend, although it might be too sensitive to choice between suitesparse/eigen (until mkl sparse backend will be merged to ceres-solver).

ahojnnes: Yes, the current parameters are not tuned. It will be difficult to find universal thresholds here that generalize across different CPU/GPU models. It was a little easier before, as we switched between algorithms that all run on the CPU. As such, I decided to expose the parameters now explicitly through the option manager for the bundle adjuster and we might need to do the same for the bundle adjustment that runs as part of the mapper. For now, this feature is disabled by default, as I want to gain some more experience with it over the next weeks.

mihaidusmanu (approved):

tineras:
I'm getting the following when trying to use the ba_use_gpu flag:

`W20241128 14:23:50.326298 19296 bundle_adjustment.cc:301] Requested to use GPU for bundle adjustment, but Ceres was compiled without CUDA support. Falling back to CPU-based dense solvers.
W20241128 14:23:50.326341 19296 bundle_adjustment.cc:320] Requested to use GPU for bundle adjustment, but Ceres was compiled without cuDSS support. Falling back to CPU-based sparse solvers.`

The only instance of ceres I know of on my machine is the ceres.dll inside of the colmap/bin folder. I am using CUDA 11.8.

Am I missing something or using this incorrectly?

ahojnnes:
You need to custom compile the latest Ceres development version with CUDA support. This is not part of the current set of pre compiled colmap binaries.

tineras:
> You need to custom compile the latest Ceres development version with CUDA support. This is not part of the current set of pre compiled colmap binaries.

Okay, I can certainly do that. I updated my original message, but I'll include it in this reply as well. Sorry to hijack this topic. I'm happy to open a new issues, but was trying to avoid that.

"I am only trying this because I'm getting sudden crashes during bundle adjustment and have tried numerous, known working datasets that never crashed before and now everything I process crashes during BA and I am trying to pinpoint what dependencies on my system might have changed."

tineras:
> You need to custom compile the latest Ceres development version with CUDA support. This is not part of the current set of pre compiled colmap binaries.

Looks like ceres is causing the problem for me. I am doing another run now with a ceres dll (with CUDA support) that I built using vcpkg.

`Faulting application name: colmap.exe, version: 0.0.0.0, time stamp: 0x67486f74
Faulting module name: ceres.dll, version: 0.0.0.0, time stamp: 0x6739347e
Exception code: 0xc0000005
Fault offset: 0x000000000001748e
Faulting process id: 0x0x39A0
Faulting application start time: 0x0x1DB41FC694ECFF6
Faulting application path: C:\proj\colmap_3_11_0\bin\colmap.exe
Faulting module path: C:\proj\colmap_3_11_0\bin\ceres.dll
Report Id: 699ead97-3e8e-4237-bbe0-3d4161d2b5e8
Faulting package full name: 
Faulting package-relative application ID: `

ahojnnes:
You'll have to recompile colmap from scratch. You cannot just recompile the ceres.dll and replace it. First of all, the colmap build system will detect CUDA support in ceres at compile and not at runtime. Second, C++ does not have a stable ABI, so you cannot mix and match different compiler or standard library versions. It will be difficult to match the exact combination of OS/compiler/stdlib/etc. on your system.

PR #2723, opened 2024-08-20
Make VisualIndex::Quantize const to improve readability

IshitaTakeshi:
`VisualIndex::Quantize` is called from only `VisualIndex::Build` and modifies the member variable `visual_words_`. 

In the current implementation, we cannot easily guess what `Quantize` actually modifies when reading the `Build` function. 

I made the `VisualIndex::Quantize` const and made it explicitly return the output.

### DIFF
--- a/src/colmap/retrieval/visual_index.h
+++ b/src/colmap/retrieval/visual_index.h
@@ -155,7 +155,8 @@ class VisualIndex {
 
  private:
   // Quantize the descriptor space into visual words.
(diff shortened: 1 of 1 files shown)


ahojnnes (approved):
Thanks, this is indeed a nicer interface.

ahojnnes:
@IshitaTakeshi You will need to run `scripts/format/clang_format.sh` for fix the failing CI check. Thanks.

PR #2725, opened 2024-08-21
Fix CUDA_ENABLED macro in new bundle adjustment code

drkoller:
I think that the macro CUDA_ENABLED should be COLMAP_CUDA_ENABLED instead, in the new bundle adjustment code that directs Ceres to use the GPU.


ahojnnes:
Thanks for the fix. LGTM.

ahojnnes (approved):

PR #2733, opened 2024-08-25
Fix colmap namespace in the macro support of logging.

B1ueber2y:
To be able to use the macro when COLMAP is used as a third party


ahojnnes (approved):
Great, thank you!

PR #2735, opened 2024-08-26
Implement transitive matcher with pair generator + tests

ahojnnes:
Consistently implements transitive matching through pair generation interface. This allows to implement tests against the pair generation. No change in behavior.


sarlinpe (approved):
Nice, thanks a lot!

PR #2742, opened 2024-09-01
Fix abseil that ceres 2.3.0 depends on

whuaegeanse:
When building colmap using ceres 2.3.0 (ceres master branch), the cmake output log is as follows:
```
CMake Error at D:/xxx1/scripts/buildsystems/vcpkg.cmake:859 (_find_package):
  Found package configuration file:

    D:/xxx2/lib/cmake/Ceres/CeresConfig.cmake

  but it set Ceres_FOUND to FALSE so package "Ceres" is considered to be NOT
  FOUND.  Reason given by package:

  The following imported targets are referenced, but are missing: absl::log
  absl::check absl::fixed_array

Call Stack (most recent call first):
  cmake/FindDependencies.cmake:49 (find_package)
  CMakeLists.txt:106 (include)


-- Configuring incomplete, errors occurred!
```

This PR fixes the issue by adding an option called ABSL_ENABLED.
When building colmap with ceres 2.3.0, turn on ABSL_ENABLED.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -47,6 +47,7 @@ option(CCACHE_ENABLED "Whether to enable compiler caching, if available" ON)
 option(CGAL_ENABLED "Whether to enable the CGAL library" ON)
 option(LSD_ENABLED "Whether to enable the LSD library" ON)
 option(UNINSTALL_ENABLED "Whether to create a target to 'uninstall' colmap" ON)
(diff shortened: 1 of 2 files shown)


ahojnnes:
I appreciate the fix but isn't this something that should be fixed on the Ceres side? The ceres cmake config should find absl if it depends on it?

whuaegeanse:
Yes, you are right. I will push another PR to Ceres.

Jiatutu:
Has the problem been solved?

PR #2721, opened 2024-08-19
Automatically generate stub files

sarlinpe:
- [x] Cleanup the stub files
- [x] Figure out how to generate them in the CI for Unix
- [ ] will do Windows in a subsequent PR
- [x] Test them in the CI
- [ ] Merge upstream the changes of https://github.com/sarlinpe/pybind11-stubgen

PR #2745, opened 2024-09-03
Provide reasonable defaults for some estimator options

sarlinpe:
https://github.com/colmap/colmap/pull/2721 inadvertently broke the default options for `pycolmap.estimate_triangulation`, which I then realized were completely wrong. To fix this, I think that `EstimateTriangulationOptions` and `AbsolutePose{Estimation,Refinement}Options` should provide sensible defaults instead of, as we have been doing it so far, monkey-patching some values in Python. This PR doesn't change the default behavior of COLMAP and only slightly affects PyCOLMAP's `estimate_triangulation`/`absolute_pose_estimation`/`rig_absolute_pose_estimation` to be consistent with the corresponding COLMAP options.

### DIFF
--- a/src/colmap/estimators/pose.h
+++ b/src/colmap/estimators/pose.h
@@ -66,6 +66,15 @@ struct AbsolutePoseEstimationOptions {
   // Options used for P3P RANSAC.
   RANSACOptions ransac_options;
 
+  AbsolutePoseEstimationOptions() {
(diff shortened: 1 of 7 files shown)


B1ueber2y:
Crucial fix. Thanks!

ahojnnes (approved):

PR #2747, opened 2024-09-04
Add mixed Python-C++ PyCOLMAP package

sarlinpe:
Move the C++ extension to `pycolmap._core` such that we can augment PyCOLMAP with pure Python modules. Notes:
- Other projects would often place the extension in a second package `_pycolmap`, but this pollutes the `site-packages` directory
- In `help(pycolmap)`, this shows types like `pycolmap._core.Reconstruction` instead of `pycolmap.Reconstruction`... @B1ueber2y  Did you manage to fix this in limap?


B1ueber2y:
I did it in a quite dirty way here: https://github.com/cvg/limap/blob/main/limap/__init__.py. I am sure there are better alternatives (at least it should be switched to explicit import)

sarlinpe:
Types in `help(pycolmap)` still appear as `pycolmap._core` but this doesn't matter much because we now have the doc pages. Overall this new package structure follows the one of pytorch.

B1ueber2y:
Thanks! This worked in my test. 

However, are you sure that this is the only way to achieve this? "import *" looks quite aggressive and is widely suggested to be avoided in all costs (that s the reason why I considered what I did in limap ugly)

sarlinpe:
This is a common solution (for example [as done in torch](https://github.com/pytorch/pytorch/blob/deb820602aa255bb899defaca1f8d26ee8e949e1/torch/__init__.py#L357-L368)) and the only alternative to manually listing the symbols to be imported (which is impractical for us).

ahojnnes:
LGTM, also given that other notable libraries follow the same approach.

ahojnnes (approved):

pycolmap/pycolmap/__init__.py
```diff
@@ -0,0 +1,15 @@
+from ._core import *  # noqa F403
```
pablospe, line 1: Maybe something to consider. To avoid importing internal symbols, the following code snippet reduces the risk of name conflicts and keeps the `pycolmap` namespace clean:

```python
from . import _core as _core_module

# Loop over all the members of the module, adding the symbols to the current
# module's namespace (i.e., pycolmap package's), and skipping internal symbols
# (names starting with '_').
for _name, _symbol in vars(_core_module).items():
    if not _name.startswith("_"):
        globals()[_name] = _symbol

# Cleanup.
del _core_module, _name, _symbol
```

This solution ensures that only the relevant symbols from `_core` are shown, without explicitly listing `pycolmap._core` when using `help(pycolmap)`. Please note that I haven't tested this. I haven't tested.

PR #2758, opened 2024-09-05
Fix options for CUDA sparse BA solver

### LINKED ISSUE #2746: COLMAP with CUDA-enabled BA
**Describe the bug**
When COLMAP tries to do the BA using CUDA (with _--Mapper.ba_use_gpu option_) it raises a ceres error on `detect_structure.cc` at line 111, indicating empty rows on the matrix.

**To Reproduce**
1. Execute the usual colmap pipeline: extract the features, do the matching.
2. At mapper step, use the option flag `--Mapper.ba_use_gpu on`
> colmap mapper --Mapper.ba_use_gpu on --database_path ~/data/sfm/street_pole1/sparse/database.db --image_path ~/data/sfm/street_pole1/images --output_path ~/data/sfm/street_pole1/sparse
3. When the number of registered files reach 50 files, the retriangulation and Global BA happens.
4. The error triggers with message:
> F0903 18:59:56.927320 76973 detect_structure.cc:111] Check failed: *row_block_size != 0 (0 vs. 0) No rows found

**Expected behavior**
It was expected to run without erros, given that the same dataset without GPU flags runs smooth.

**Screenshots**
error
error_gdb

**Environment:**
 - OS: Ubuntu 22.04
 - COLMAP Version: 4ccfeb55e6548a2d15636796e42e1df292c073c5
 - Capture Device: Samsung S23
 - Ceres version: 2.2.0

**Related issue**: https://github.com/colmap/colmap/issues/2643
**Related PR:** https://github.com/colmap/colmap/pull/2717

whuaegeanse:
This PR attempts to fix #2746 .

cudss is available in ceres since 2.3.0.
In [ceres 2.2.0](https://github.com/ceres-solver/ceres-solver/releases/tag/2.2.0), CUDA_SPARSE is used for the CGNR solver instead of the SPARSE_SCHUR solver.

### DIFF
--- a/src/colmap/estimators/bundle_adjustment.cc
+++ b/src/colmap/estimators/bundle_adjustment.cc
@@ -347,20 +347,29 @@ ceres::Solver::Options BundleAdjuster::SetUpSolverOptions(
       options_.max_num_images_direct_dense_cpu_solver;
   int max_num_images_direct_sparse_solver =
       options_.max_num_images_direct_sparse_cpu_solver;
-#if (CERES_VERSION_MAJOR >= 3 ||                                \
-     (CERES_VERSION_MAJOR == 2 && CERES_VERSION_MINOR >= 2)) && \
-    !defined(CERES_NO_CUDSS) && defined(COLMAP_CUDA_ENABLED)
+
+#ifdef COLMAP_CUDA_ENABLED
+#if (CERES_VERSION_MAJOR >= 3 || \
+     (CERES_VERSION_MAJOR == 2 && CERES_VERSION_MINOR >= 2))
   if (options_.use_gpu && num_images >= options_.min_num_images_gpu_solver) {
     const std::vector<int> gpu_indices = CSVToVector<int>(options_.gpu_index);
     THROW_CHECK_GT(gpu_indices.size(), 0);
     SetBestCudaDevice(gpu_indices[0]);
     solver_options.dense_linear_algebra_library_type = ceres::CUDA;
-    solver_options.sparse_linear_algebra_library_type = ceres::CUDA_SPARSE;
     max_num_images_direct_dense_solver =
         options_.max_num_images_direct_dense_gpu_solver;
+  }
+#endif
+
+#if (CERES_VERSION_MAJOR >= 3 ||                                \
+     (CERES_VERSION_MAJOR == 2 && CERES_VERSION_MINOR >= 3)) && \
+    !defined(CERES_NO_CUDSS)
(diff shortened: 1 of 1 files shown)


ahojnnes:
Thank you @whuaegeanse. CGNR is technically not a sparse direct solver, so the current option description is not correct. I am still wondering whether CUDA_SPARSE would result in a speedup even without cudss? Is there a way to switch between the sparse direct and indirect CUDA solvers in ceres >=2.3?

S-o-T:
> Is there a way to switch between the sparse direct and indirect CUDA solvers in ceres >=2.3?

ceres-solver still can't leverage CUDA-capable GPU in context of pcg with implicit schur complement. There is an [issue](https://github.com/ceres-solver/ceres-solver/issues/871) available to be tracked.

@DmitriyKorchemkin [did](https://ceres-solver-review.googlesource.com/q/status:open+-is:wip+owner:dmitriy.korchemkin@gmail.com) some work in that direction, although it staled a while ago.

ahojnnes:
Thank you @S-o-T for the answer. To me the fix here looks correct. Would you be able to also confirm? Thanks.

src/colmap/estimators/bundle_adjustment.cc
```diff
@@ -347,20 +347,29 @@ ceres::Solver::Options BundleAdjuster::SetUpSolverOptions(
...
     SetBestCudaDevice(gpu_indices[0]);
     solver_options.dense_linear_algebra_library_type = ceres::CUDA;
-    solver_options.sparse_linear_algebra_library_type = ceres::CUDA_SPARSE;
     max_num_images_direct_dense_solver =
         options_.max_num_images_direct_dense_gpu_solver;
+  }
+#endif
+
+#if (CERES_VERSION_MAJOR >= 3 ||                                \
+     (CERES_VERSION_MAJOR == 2 && CERES_VERSION_MINOR >= 3)) && \
```
S-o-T, line 365: There are might be users that would prefer to use ceres-solver version without dependency on abseil, but with support of cudss (which was merged before introduction of abseil dep, which, in turn, was merged before  CERES_VERSION_MINOR was incremented). But i believe that this is a marginal case that can be solved by local patches.

S-o-T (approved):

src/colmap/estimators/bundle_adjustment.cc
ahojnnes: Thank you. Let's keep that edge case to the expert user. As far as I can see, there is active development on the Ceres side to make the absl dependency more easily consumable.

ahojnnes (approved):
Thank you @whuaegeanse for the fix.

PR #2762, opened 2024-09-06
Store shared camera ptr for reconstruction images

ahojnnes:


src/colmap/scene/reconstruction.cc
```diff
@@ -117,6 +117,13 @@ void Reconstruction::AddCamera(struct Camera camera) {
 }
 
 void Reconstruction::AddImage(class Image image) {
+  THROW_CHECK(image.HasCameraId());
+  const auto camera_it = cameras_.find(image.CameraId());
```
B1ueber2y, line 121: Should we throw assertion error here on that the camera must exist

src/colmap/scene/reconstruction.h
```diff
@@ -102,7 +102,9 @@ class Reconstruction {
   // might be taken by the same camera.
   void AddCamera(struct Camera camera);
 
-  // Add new image.
```
B1ueber2y, line 105: We should also add comments at the two reconstruction IO functions like: https://github.com/colmap/colmap/pull/2702/files#diff-6ff1c633d8187e2cb2de82d2b9217f65046013c4edfd3c8eaf53e0a95b91461cR40.

B1ueber2y:
This looks great to me. Should we also add the ProjectPoint function we had in the other PR before?

src/colmap/scene/reconstruction.cc
ahojnnes: Done.

src/colmap/scene/reconstruction.h
ahojnnes: Done

ahojnnes:
> This looks great to me. Should we also add the ProjectPoint function we had in the other PR before?

It's already there.

B1ueber2y (approved):

sarlinpe (approved):

PR #2763, opened 2024-09-09
Use shared camera pointer in more places

ahojnnes:


src/colmap/estimators/alignment.cc
```diff
@@ -441,8 +439,7 @@ bool MergeReconstructions(const double max_reproj_error,
     src_image.CamFromWorld() =
         TransformCameraWorld(tgt_from_src, src_image.CamFromWorld());
     if (!tgt_reconstruction.ExistsCamera(src_image.CameraId())) {
-      tgt_reconstruction.AddCamera(
-          src_reconstruction.Camera(src_image.CameraId()));
+      tgt_reconstruction.AddCamera(*src_image.CameraPtr());
     }
     tgt_reconstruction.AddImage(src_image);
```
B1ueber2y, line 444: Is this correct? The camera was copied in AddCamera so the address has changed?

B1ueber2y: Actually thinking of this, the ``AddImage`` method from https://github.com/colmap/colmap/pull/2762 is potentially dangerous. It should reset the camera pointer to the one for the camera in the reconstruction. The current check does not handle the life of the camera in a safe way.

B1ueber2y: Ah OK I missed the update of the implementation. My design before was buggy but the new is correct. Never mind : )

B1ueber2y (approved):

src/colmap/estimators/alignment.cc
ahojnnes: Thanks for catching this bug. Also, the function should have a unit test, though it's beyond the scope of this PR now. I added a TODO to keep track of it.

B1ueber2y (approved):

src/colmap/estimators/alignment.cc
B1ueber2y: Added here: https://github.com/colmap/colmap/pull/2766.

PR #2764, opened 2024-09-09
Avoid unnecessary copy of queue in IncrementalTriangulator::Complete()

ahojnnes:
The order in which the points are completed does not matter, so the change of FIFO to LIFO queue produces the same results.

### DIFF
--- a/src/colmap/sfm/incremental_triangulator.cc
+++ b/src/colmap/sfm/incremental_triangulator.cc
@@ -688,15 +688,11 @@ size_t IncrementalTriangulator::Complete(const Options& options,
   std::vector<TrackElement> queue = point3D.track.Elements();
 
   const int max_transitivity = options.complete_max_transitivity;
-  for (int transitivity = 0; transitivity < max_transitivity; ++transitivity) {
-    if (queue.empty()) {
-      break;
-    }
+  for (int transitivity = 1; transitivity <= max_transitivity; ++transitivity) {
+    while (!queue.empty()) {
+      const TrackElement queue_elem = queue.back();
+      queue.pop_back();
 
-    const std::vector<TrackElement> prev_queue = queue;
-    queue.clear();
-
-    for (const TrackElement& queue_elem : prev_queue) {
       const auto corr_range = correspondence_graph_->FindCorrespondences(
           queue_elem.image_id, queue_elem.point2D_idx);
       for (const auto* corr = corr_range.beg; corr < corr_range.end; ++corr) {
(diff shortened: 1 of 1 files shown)


B1ueber2y (approved):

B1ueber2y:
Not sure how we overlooked this. But this update does not seem correct? The second level of transitivitiy will never get executed since the queue is already empty.

ahojnnes:
I'll have a look tonight. How did you notice it? Different behavior observed?

ahojnnes:
Please double check: https://github.com/colmap/colmap/pull/3094

I just evaluated this fix against the main branch with the new benchmark suite and the A/B comparison looks like this. It's still a bug that should be fixed.
```
I20250103 17:19:16.339394 3018444 compare.py:main:58] Results A - B:

botanical_garden   0.00   0.00   0.00   0.00      0     0    0       0
boulders           0.00   0.00   0.00   0.00      0     0    0       0
bridge            -0.00  -0.00  -0.00  -0.00      0     0    0       0
courtyard          0.00   0.00   0.00   0.00      0     0    0       0
delivery_area      0.00   0.00   0.00   0.00      0     0    0       0
door               0.00   0.00   0.00   0.00      0     0    0       0
electro           -0.00  -0.00  -0.00  -0.00      0     0    0       0
exhibition_hall    0.00   0.00   0.00   0.00      0     0    0       0
facade            -0.01  -0.01  -0.00  -0.00      0     0    0       0
kicker             0.00   0.00   0.00   0.00      0     0    0       0
lecture_room      -0.00  -0.00  -0.00  -0.00      0     0    0       0
living_room        0.00   0.00   0.00   0.00      0     0    0       0
lounge             0.00   0.00   0.00   0.00      0     0    0       0
meadow             0.00   0.00   0.00   0.00      0     0    0       0
observatory       -0.00  -0.00  -0.00  -0.00      0     0    0       0
office             0.00   0.00   0.00   0.00      0     0    0       0
old_computer       0.00   0.00   0.00   0.00      0     0    0       0
pipes              0.00   0.00   0.00   0.00      0     0    0       0
playground        -0.00   0.00   0.00   0.00      0     0    0       0
relief            -0.00  -0.00  -0.00  -0.00      0     0    0       0
relief_2          -0.00  -0.00  -0.00  -0.00      0     0    0       0
statue             0.00   0.00   0.00   0.00      0     0    0       0
terrace           -0.00  -0.00  -0.00  -0.00      0     0    0       0
terrace_2         -0.00  -0.00  -0.00  -0.00      0     0    0       0
terrains          -0.00  -0.00  -0.00  -0.00      0     0    0       0
overall           -0.00  -0.00  -0.00  -0.00      0     0    0       0
average           -0.00  -0.00  -0.00  -0.00      0     0    0       0
```

B1ueber2y:
> I'll have a look tonight. How did you notice it? Different behavior observed?

No different behavior observed. Noticed since another project was looking into the ``Complete`` operation.

> I just evaluated this fix against the main branch with the new benchmark suite and the A/B comparison looks like this.

I think the previous logic was basically infinite depth transitivity since the transitivity was not updated. So even if there is different in A/B test, fixing this bug can only make the results worse but can potentially lead to faster results (which is not reflected in the test). The bug should be fixed in the sense that the ``complete_max_transitivity`` was not useful before.

PR #2769, opened 2024-09-10
print RMS reprojection error in solver summary

tmnku:
According to ceres e.g. here https://github.com/ceres-solver/ceres-solver/blob/ea4400cd57f17699077ec8905b503b9320aa1d6d/internal/ceres/solver_test.cc#L334 the cost in the summary is 0.5 * sum of squared residuals.

### DIFF
--- a/src/colmap/estimators/bundle_adjustment.cc
+++ b/src/colmap/estimators/bundle_adjustment.cc
@@ -862,12 +862,12 @@ void PrintSolverSummary(const ceres::Solver::Summary& summary,
 
   log << std::right << std::setw(16) << "Initial cost : ";
   log << std::right << std::setprecision(6)
-      << std::sqrt(summary.initial_cost / summary.num_residuals_reduced)
+      << std::sqrt(2 * summary.initial_cost / summary.num_residuals_reduced)
       << " [px]\n";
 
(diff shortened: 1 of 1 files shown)


ahojnnes:
Thank you. If I understand well, this change is to make the cost reported in Ceres' Brief/Full summary consistent with colmap's reported cost?

tmnku:
I just noticed that I might have misinterpreted the printed summary, so feel free to close the Issue. I thought of the cost as some kind of reprojection error and was confused why it deviated so much from `Reconstruction::ComputeMeanReprojectionError()`. I also just noticed that my pull request is more like the RMS residual. The RMS reprojection error would require to divide by `0.5 * summary.num_residuals_reduced`. I think the latter is more intuitive.

ahojnnes:
@tmnku Notice that the initial/final_cost values computed by Ceres also include the robustifier, so these are not directly comparable. Furthermore, ComputeMeanReprojectionError() computes `mean(sqrt(dx_1^2 + dy_1^2) + sqrt(dx_2^2 + dy_2^2) + ... + sqrt(dx_N^2 + dy_N^2))` while the initial/final_cost are computed as the RMS over all dx/dy residuals.

PR #2779, opened 2024-09-11
Fix missing camera ptr for Reconstruction.DeleteAllPoints2DAndPoints3D()

B1ueber2y:
Addresses https://github.com/colmap/colmap/issues/2775. Missed from https://github.com/colmap/colmap/pull/2762

### DIFF
--- a/src/colmap/scene/reconstruction.cc
+++ b/src/colmap/scene/reconstruction.cc
@@ -237,6 +237,7 @@ void Reconstruction::DeleteAllPoints2DAndPoints3D() {
     new_image.SetImageId(image.second.ImageId());
     new_image.SetName(image.second.Name());
     new_image.SetCameraId(image.second.CameraId());
+    new_image.SetCameraPtr(image.second.CameraPtr());
     new_image.SetRegistered(image.second.IsRegistered());
     new_image.CamFromWorld() = image.second.CamFromWorld();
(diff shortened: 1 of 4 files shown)


ahojnnes:
Is it possible to construct a test case that covers the issue? Thanks for the fix.

B1ueber2y:
> Is it possible to construct a test case that covers the issue? Thanks for the fix.

Added some tests in the reconstruction class (we need to add some pipeline tests in the future). PTAL when time permits.  The point_triangulator breaks in the current head so we should merge this PR ASAP : )

src/colmap/scene/reconstruction.cc
```diff
@@ -43,6 +43,15 @@ namespace colmap {
 
 Reconstruction::Reconstruction() : max_point3D_id_(0) {}
 
+bool Reconstruction::Check() const {
+  for (const auto& image : images_) {
```
ahojnnes, line 47: I recommend moving this check to the test file, where its only use case is. No need to clutter the public interface of the reconstruction.

ahojnnes (commented):
Thanks, small suggestion. Otherwise, LGTM.

src/colmap/scene/reconstruction.cc

ahojnnes (approved):

ahojnnes (approved):

PR #2780, opened 2024-09-12
Rename remaining proj_matrix instances to cam_from_world

ahojnnes:
No change in behavior. Simple rename.


B1ueber2y:
Could it be confusing that the type of cam_from_world is sometimes Rigid3d and sometimes Eigen::Matrix3x4d

B1ueber2y (approved):

ahojnnes:
> Could it be confusing that the type of cam_from_world is sometimes Rigid3d and sometimes Eigen::Matrix3x4d

Yes, I was going to follow up in a separate PR but it needs some logic change in various functions related to decomposition of H, E, F, etc.

PR #2783, opened 2024-09-13
Add support for keypoint uncertainty and its integration in bundle adjustment

B1ueber2y:
- 2d covariance in keypoint class and Point2D class (which are set to identity since being unsupported in the current detectors)
- make covariance optional in all reprojection error costs and update pybind
- adapt to the new cost in bundle adjustment

Now all the keypoint uncertainty is initialized as identity and the covariance is made optional in the reprojection error costs **so there should be no logic changes and broken interfaces**. Just that there can be expected slight performance drop on the bundle adjustment efficiency. From the pycolmap example (11 images) the efficiency drop is innegligible.


src/colmap/scene/point2d.h
```diff
@@ -42,6 +42,9 @@ struct Point2D {
   // The image coordinates in pixels, starting at upper left corner with 0.
   Eigen::Vector2d xy = Eigen::Vector2d::Zero();
 
+  // The 2D covariance of the keypoint
+  Eigen::Matrix2d covar = Eigen::Matrix2d::Identity();
```
ahojnnes, line 46: What's the practical use case where we have a full 2D covariance? AFAIK, none of the existing keypoint detectors provides a meaningful 2D covariance? If we find a convincing use case, I suggest to parameterize this minimally with 3 floats and probably single-precision will be enough.

src/colmap/estimators/cost_functions.h
```diff
@@ -84,16 +90,19 @@ class ReprojErrorCostFunction {
...
                             point3D_in_cam[2],
-                            &residuals[0],
-                            &residuals[1]);
-    residuals[0] -= T(observed_x_);
-    residuals[1] -= T(observed_y_);
+                            &residuals_ptr[0],
+                            &residuals_ptr[1]);
+    residuals_ptr[0] -= T(observed_x_);
+    residuals_ptr[1] -= T(observed_y_);
+    Eigen::Map<Eigen::Matrix<T, 2, 1>> residuals(residuals_ptr);
```
ahojnnes, line 97: What's the performance impact of this change?

src/colmap/scene/point2d.h
sarlinpe: Agree. There are different levels of detail possible:
1. identical for all points in an image (this would account for image resizing and different types of detectors)
2. isotropic (single scalar), 
3. anisotropic (2x2 matrix)

To me, it's not obvious that we need 3) directly. 2) has less impact on performance and is less invasive (using the `IsotropicNoiseCostFunctionWrapper`).

If 3), how would we obtain such covariances? In the case of SIFT, can we extract them from the affine shape? are they absolute in scale? For other detectors, I know about https://arxiv.org/pdf/2305.12250 but they are up to a scale factor.

src/colmap/estimators/cost_functions.h
```diff
@@ -56,26 +56,32 @@ inline Eigen::MatrixXd SqrtInformation(const Eigen::MatrixXd& covariance) {
...
-  explicit ReprojErrorCostFunction(const Eigen::Vector2d& point2D)
-      : observed_x_(point2D(0)), observed_y_(point2D(1)) {}
-
-  static ceres::CostFunction* Create(const Eigen::Vector2d& point2D) {
+  explicit ReprojErrorCostFunction(
+      const Eigen::Vector2d& point2D,
+      const Eigen::Matrix2d& point2D_covar = Eigen::Matrix2d::Identity())
+      : observed_x_(point2D(0)),
+        observed_y_(point2D(1)),
+        sqrt_information_point_(SqrtInformation(point2D_covar)) {}
```
sarlinpe, line 64: It would be much cleaner to have a wrapper cost function that can be bypassed if the covariance is identity, such that the performance is degraded only for non-default covariances. I looked into this a while ago and the [WeightedCostFunction](https://groups.google.com/g/ceres-solver/c/OobKeuuvXHE) appears to be only solution. This might be easier to implement now than in 2013... maybe worth reviving that thread? or use the `IsotropicNoiseCostFunctionWrapper` for now.

src/colmap/estimators/cost_functions.h
sarlinpe: +1, we might need to benchmark it on a large BA problem.

B1ueber2y: Made a new update here where the cost adaptively decides how to use the given covariance (identity / diagonal / general). With this update I believe we can officially drop all the IsotropicCovariance implementation. We could switch to the ``WeightedCostFunction`` when ceres has this supported. 

Will post the performance benchmarking on a large BA problem with the wheels built in this PR when it is available (i found that the wheels are much faster than the build from source in terms of BA performance)

src/colmap/scene/point2d.h
B1ueber2y: Yes. There are no current off-the-shelf detector that supports full 2D covariance. However, the current COLMAP does not support diagonal anisotropic covariance (case 1) from @sarlinpe) which is quite common from image resizing. Also, I have an internal development of 2D uncertainty in hand so would benefit from this update, and this will enable further development on such feature.

B1ueber2y: Also, I have switched to 3-dim Vector3f in the Point2D class. PTAL : )

ahojnnes: For case 1, a single scalar per image would be sufficient, which has negligible storage overhead. For case 2, a single float per image point or, if using suitable quantization, a single uint8/uint16. For case 3, I would like to avoid supporting such experimental features in the main branch until they are proven to benefit the entire community. For classic blob/corner-based detectors (handcrafted and learned), I would expect an isotropic covariance to work quite well. They rarely ever detect features on elongated structures where the covariance is anisotropic? In this case, the viewing geometry of the supporting cameras to be the dominating factor for determining the covariance of 3D points?

B1ueber2y: How about an image resized without ration kept? This could happen commonly in practice, which requires two parameters? I understand the point that full covariance might be too experimental to be supported, but this diagonal covariance would be quite practical (also generally falls into case 1))

B1ueber2y: How about we revert the data flow but proceed with the support on the cost function only. This won't hurt anything but only give more flexible support : )

src/colmap/estimators/cost_functions.h
```diff
@@ -84,16 +107,25 @@ class ReprojErrorCostFunction {
...
                             point3D_in_cam[2],
-                            &residuals[0],
-                            &residuals[1]);
-    residuals[0] -= T(observed_x_);
-    residuals[1] -= T(observed_y_);
+                            &residuals_ptr[0],
+                            &residuals_ptr[1]);
+    residuals_ptr[0] -= T(observed_x_);
+    residuals_ptr[1] -= T(observed_y_);
+    if (covariance_type_ == CovarianceType::GENERAL) {
```
ahojnnes, line 114: It feels like we can perform this check once in the Create(). Then, using templating/constexpr, we can return a cost function that has zero computational/storage overhead during the optimization.

B1ueber2y:
Closed for now. Will open in a separate PR with only the cost functions updated (and templating stuff).

PR #2786, opened 2024-09-16
Implementation of Aria Fisheye camera model

nushakrishnan:
This pull request implements a new camera model called the Aria Fisheye model. The Aria Fisheye model includes:

- Radial distortion (6 distortion coefficients)
- Tangential distortion (2 distortion coefficients)
- Thin-prism distortion (4 distortion coefficients)

The camera model is described here -
https://facebookresearch.github.io/projectaria_tools/docs/tech_insights/camera_intrinsic_models#the-fisheyeradtanthinprism-fisheye624-model

For more details, one can also refer to the Project Aria implementation here - 
https://github.com/facebookresearch/projectaria_tools/blob/main/core/calibration/camera_projections/FisheyeRadTanThinPrism.h

Thanks!


nushakrishnan:
Converted to draft to debug

nushakrishnan:
To test the camera model implementation - 

- Selected two images and obtained the common 3d points visible in both these images using [MPS](https://facebookresearch.github.io/projectaria_tools/docs/data_formats/mps/slam). An example of such a point - 


image

- Compared the projection of the 3d point in the Aria Fisheye COLMAP camera to the projection using Project Aria's implementation [here](https://github.com/facebookresearch/projectaria_tools/blob/35a8e7911116197d5e5f00dfddffa9a2bc297431/core/python/DeviceCalibrationPyBind.h#L173).
- Ensure that the projections are equivalent.

I have also uploaded the script of this test for further clarity.
[test_aria_camera_model.zip](https://github.com/user-attachments/files/17017704/test_aria_camera_model.zip)

src/colmap/sensor/models.h
```diff
@@ -372,6 +374,28 @@ struct ThinPrismFisheyeCameraModel
...
+//
+// See
+// https://facebookresearch.github.io/projectaria_tools/docs/tech_insights/camera_intrinsic_models#the-fisheyeradtanthinprism-fisheye624-model
+//
+// Parameter list is expected in the following order:
+//
+//    fx, fy, cx, cy, k1, k2, k3, k4, k5, k6, p1, p2, sx1, sx2, sy1, sy2
+//
+// Note:
+//    Tangential coefficients p1, p2 correspond to p1, p0 in the Fisheye624 Aria
```
ahojnnes, line 390: Why rename these as compared to the original model? Thanks.

src/colmap/sensor/models_test.cc
```diff
@@ -228,5 +228,26 @@ TEST(ThinPrismFisheye, Nominal) {
                                           0.001});
 }
 
+TEST(AriaFisheye, Nominal) {
+  std::vector<double> params = {241.996,
+                                241.996,
```
ahojnnes, line 233: I suggest using a different value for fx/fy so we can detect possible bugs when accidentally swapping coordinates. Thanks.

B1ueber2y (approved):
I can confirm that the code is correct from my pass.

src/colmap/sensor/models.h
```diff
@@ -372,6 +374,28 @@ struct ThinPrismFisheyeCameraModel
...
+// model. Thin prism coefficients sx1, sx2, sy1, sy2 correspond to s0, s1, s2,
+// s3 in the Fisheye624 Aria model.
+//
+struct AriaFisheyeCameraModel : public BaseCameraModel<AriaFisheyeCameraModel> {
```
ahojnnes, line 394: I suggest calling this model "FisheyeRadTanThinPrismModel". This is also how the Aria project calls it.

src/colmap/sensor/models.h
B1ueber2y: I think the Aria doc switches the convention of tangential distortion as written in "Camera Calibration with Distortion Models and Accuracy Evaluation", J Weng et al., TPAMI, 1992. We can also follow the Aria convention here if the documentation is stable enough.

ahojnnes (approved):
Thank you, looks great.

src/colmap/sensor/models.h
```diff
@@ -1538,6 +1561,128 @@ void ThinPrismFisheyeCameraModel::Distortion(
...
+  T theta_power = T(1);
+  for (int i = 0; i < numK; ++i) {
+    theta_power *= theta2;
+    th_radial += radial_coeffs[i] * theta_power;
+  }
+
+  const T theta_divr =
+      (r < std::numeric_limits<T>::epsilon()) ? static_cast<T>(1.0) : theta / r;
+
+  const T x = th_radial * theta_divr * u;
```
B1ueber2y, line 1665: Should we move the theta_dive mapping outside the distortion function, such that the distortion returns zero offset for extra params all zero

src/colmap/sensor/models.h
```diff
@@ -1538,6 +1561,128 @@ void ThinPrismFisheyeCameraModel::Distortion(
...
+  const T y = th_radial * theta_divr * v;
+
+  const T x2 = x * x;
+  const T y2 = y * y;
+  const T xy = x * y;
+  const T r2 = x2 + y2;
+  const T r4 = r2 * r2;
+
+  const T dx_tang = T(2) * p1 * xy + p0 * (r2 + T(2) * x2);
+  const T dy_tang = T(2) * p0 * xy + p1 * (r2 + T(2) * y2);
```
B1ueber2y, line 1675: This seems to be buggy from the new change?

src/colmap/sensor/models.h
B1ueber2y: Ok we could also do it a separate PR by also updating OpenCVFisheye. Lets keep it in the current way for now

PR #2788, opened 2024-09-17
Fixes and refactoring for bundle adjustment covariance estimation

ahojnnes:
* Refactoring for simplified interface and implementation. Fewer malloc's and thus likely some perf improvement (not profiled).
* Important bug fix for problems where L matrix on pose/other parameters is not dense.


src/pycolmap/estimators/covariance.cc
```diff
@@ -57,14 +57,6 @@ void BindCovarianceEstimator(py::module& m) {
 
   using EstimatorBase = BundleAdjustmentCovarianceEstimatorBase;
   py::class_<EstimatorBase>(m, "BundleAdjustmentCovarianceEstimatorBase")
-      .def(
```
ahojnnes, line 60: I removed this custom method, because the same can be achieved by using one of the constructors.

src/colmap/estimators/covariance_test.cc
```diff
@@ -59,6 +59,7 @@ std::shared_ptr<BundleAdjuster> BuildBundleAdjuster(
   config.SetConstantCamPose(image_ids[0]);
   config.SetConstantCamPositions(image_ids[1], {0});
   BundleAdjustmentOptions options;
+  options.solver_options.num_threads = 1;
   return std::make_shared<BundleAdjuster>(options, config);
```
ahojnnes, line 63: Changed for better determinism of test results.

src/pycolmap/estimators/covariance.cc
B1ueber2y: This feature was added in practical for the rig bundle adjuster. In such case one would like to initialize the estimator with a reconstruction and then set the pose blocks to different ones. Getting point blocks from reconstruction by iteration would be very slow in python unless we add a method in the Reconstruction class to support it.

ahojnnes: Why can you not use the constructor for this? I am not sure I see why this has to be done after creating the covariance class as opposed to during construction?

B1ueber2y: Yes we can. Just that "Getting all point blocks from reconstruction by iteration would be very slow in python unless we add a method in the Reconstruction class to support it."

ahojnnes: OK, I see. But currently estimating covariance for points is anyway not supported, unless I misread the code?

B1ueber2y: The point blocks is to set the points to be eliminated in Schur elimination, and it generally just solely comes from one reconstruction object.

ahojnnes: Sorry, I am probably not seeing the obvious. The SetPoseBlocks() only takes a vector of pose blocks but not point blocks. Why would one need to iterate over points of the reconstruction object in Python? I would expect that one would iterate over all images and extract the qvec/tvec arrays and pass them here. I would expect Python to be fast enough to iterate over a few hundreds or even thousands of images?

B1ueber2y: Because if one uses the constructor, it will be Object(pose_blocks, point_blocks), where one needs to get the point_blocks. But with the SetPoseBlocks, one can do obj = Object(reconstruction); obj.set_pose_blocks(pose_blocks), which is fast in Python.

B1ueber2y: > Sorry, I am probably not seeing the obvious. The SetPoseBlocks() only takes a vector of pose blocks but not point blocks. Why would one need to iterate over points of the reconstruction object in Python? I would expect that one would iterate over all images and extract the qvec/tvec arrays and pass them here. I would expect Python to be fast enough to iterate over a few hundreds or even thousands of images?

I m trying to say that SetPoseBlocks is useful here since from what I understand this PR attempted to remove this interface.

ahojnnes: Thanks for the patience. Got it finally :-)

How about the following: Only expose a single constructor and accepting an "image_ids" vector. If it is non-empty, it will only consider the specified images, otherwise it will extract all.

B1ueber2y: For the rig case (especially without our new frame feature), generally we store the rig_from_world outside the reconstruction. And this was the reason why I developed this (pose_blocks, point_blocks) constructor and the ``SetPoseBlocks`` method, since there is no corresponding image ids for those data.

B1ueber2y: In practical one can create a rig class and construct the bundle adjustment with pycolmap + pyceres, e.g. in lamar: https://github.com/microsoft/lamar-benchmark/blob/main/scantools/capture/session.py#L36. And this interface can perfectly support acquiring pose covariances in such case.

src/colmap/estimators/covariance.cc
```diff
@@ -31,741 +31,400 @@
...
+  const Eigen::SparseMatrix<double> H_aa = J_a.transpose() * J_a;
+  const Eigen::SparseMatrix<double> H_ap = J_a.transpose() * J_p;
+  const Eigen::SparseMatrix<double> H_pa = H_ap.transpose();
+  Eigen::SparseMatrix<double> H_pp = J_p.transpose() * J_p;
+  // In-place computation of H_pp_inv.
+  Eigen::SparseMatrix<double>& H_pp_inv = H_pp;
+  int point_param_idx = 0;
+  for (const detail::PointParam& point : points) {
+    const Eigen::Matrix3d H_pp_idx =
+        H_pp.block(point_param_idx, point_param_idx, 3, 3) +
```
B1ueber2y, line 114: We should use the corresponding parameter size instead of 3 here. This will break if the points are partially fixed, which is actually practical if one has a ground control points with x and y.

sarlinpe (approved):
Nice improvements, thank you!

B1ueber2y (approved):
Looks great. Thanks a lot!

PR #2789, opened 2024-09-17
Move Fisheye coordinate mapping outside Distortion. Abstract BaseFisheyeCameraModel

B1ueber2y:
To unblock https://github.com/colmap/colmap/pull/2786

The Fisheye coordinate mapping should be moved outside the distortion, to ensure that the distortion does not change the coordinate with extra parameters all zero. This is important for IterativeUndistortion to work with zero initialization.

Abstract a BaseFisheyeCameraModel class and adapt the convention to all four existing Fisheye models. There should be no change on the logics.


sarlinpe:
Very neat simplification, LGTM.

B1ueber2y:
Closed. Will be directly merged together in https://github.com/colmap/colmap/pull/2790

PR #2790, opened 2024-09-17
Add validity check for pixel coordinate in the Fisheye camera. Fix tests. 

B1ueber2y:
This PR is a follow-up for https://github.com/colmap/colmap/pull/2789 (should be reviewed on top of that PR being merged), which should resolve the CI issue from https://github.com/colmap/colmap/pull/2786.

The pixel coordinate is constrained in the Fisheye image, where theta needs to be smaller than pi / 2. To be able to support this I did some further abstraction by supporting the conversion from image to Fisheye coordinates.


src/colmap/sensor/models.h
```diff
@@ -960,41 +1031,60 @@ std::vector<double> OpenCVFisheyeCameraModel::InitializeParams(
...
-  *v = (y - c2) / f2;
+  T uu, vv;
+  FisheyeCoordinateFromImg(params, x, y, &uu, &vv);
   *w = 1;
 
-  IterativeUndistortion(&params[4], u, v);
+  // Undistortion
+  IterativeUndistortion(&params[4], &uu, &vv);
+
+  // Back to pixel
```
ahojnnes, line 1086: This comment is wrong?

src/colmap/sensor/models.h
```diff
@@ -960,41 +1031,60 @@ std::vector<double> OpenCVFisheyeCameraModel::InitializeParams(
...
-
   // Lift points to normalized plane
-  *u = (x - c1) / f1;
-  *v = (y - c2) / f2;
+  T uu, vv;
+  FisheyeCoordinateFromImg(params, x, y, &uu, &vv);
   *w = 1;
 
-  IterativeUndistortion(&params[4], u, v);
+  // Undistortion
```
ahojnnes, line 1083: nit: This comment does not add any useful info on top of the code.

src/colmap/sensor/models.h
```diff
@@ -1005,22 +1095,13 @@ void OpenCVFisheyeCameraModel::Distortion(
...
-    *dv = v * thetad / r - v;
-  } else {
-    *du = T(0);
-    *dv = T(0);
-  }
+  const T theta2 = u * u + v * v;
+  const T theta4 = theta2 * theta2;
+  const T theta6 = theta4 * theta2;
+  const T theta8 = theta4 * theta4;
+  T radial = k1 * theta2 + k2 * theta4 + k3 * theta6 + k4 * theta8;
```
ahojnnes, line 1102: const

src/colmap/sensor/models.h
```diff
@@ -960,41 +1031,60 @@ std::vector<double> OpenCVFisheyeCameraModel::InitializeParams(
...
-  Distortion(&params[4], u, v, &du, &dv);
-  *x = u + du;
-  *y = v + dv;
+  T duu, dvv;
+  Distortion(&params[4], uu, vv, &duu, &dvv);
 
   // Transform to image coordinates
-  *x = f1 * *x + c1;
-  *y = f2 * *y + c2;
+  ImgFromFisheyeCoordinate(params, uu + duu, vv + dvv, x, y);
```
ahojnnes, line 1072: nit: ImgFromFisheye and FisheyeFromImg ? To be consistent with NormalFromFisheye and FisheyeFromNormal.

src/colmap/sensor/models.h
```diff
@@ -1404,19 +1515,11 @@ void RadialFisheyeCameraModel::Distortion(
...
-    const T thetad = theta * (T(1) + k1 * theta2 + k2 * theta4);
-    *du = u * thetad / r - u;
-    *dv = v * thetad / r - v;
-  } else {
-    *du = T(0);
-    *dv = T(0);
-  }
+  const T theta2 = u * u + v * v;
+  const T theta4 = theta2 * theta2;
+  T radial = k1 * theta2 + k2 * theta4;
```
ahojnnes, line 1520: const

src/colmap/sensor/models.h
```diff
@@ -1592,4 +1689,49 @@ double CameraModelCamFromImgThreshold(const CameraModelId model_id,
   return -1;
 }
 
+bool CameraModelIsFisheye(const CameraModelId model_id) {
+  switch (model_id) {
+#define CAMERA_MODEL_CASE(CameraModel) case CameraModel::model_id:
+
+    FISHEYE_CAMERA_MODEL_CASES
+    return true;
+    break;
```
ahojnnes, line 1698: break after return not necessary. Also in the default case.

src/colmap/sensor/models.h
```diff
@@ -1592,4 +1689,49 @@ double CameraModelCamFromImgThreshold(const CameraModelId model_id,
...
+                                    const std::vector<double>& params,
+                                    const Eigen::Vector2d& xy) {
+  switch (model_id) {
+#define CAMERA_MODEL_CASE(CameraModel)                 \
+  case CameraModel::model_id: {                        \
+    double uu, vv;                                     \
+    CameraModel::FisheyeCoordinateFromImg(             \
+        params.data(), xy.x(), xy.y(), &uu, &vv);      \
+    const double theta = std::sqrt(uu * uu + vv * vv); \
+    if (theta < M_PI / 2.0)                            \
```
ahojnnes, line 1719: Please add parentheses around if statement bodies.

src/colmap/sensor/models.h
```diff
@@ -1592,4 +1689,49 @@ double CameraModelCamFromImgThreshold(const CameraModelId model_id,
...
+      return true;                                     \
+    else                                               \
+      return false;                                    \
+  } break;
```
ahojnnes, line 1723: break unnecessary

src/colmap/sensor/models.h
```diff
@@ -1592,4 +1689,49 @@ double CameraModelCamFromImgThreshold(const CameraModelId model_id,
...
+
+    FISHEYE_CAMERA_MODEL_CASES
+    default:
+      throw std::domain_error(
+          "Camera model does not exist or is not a fisheye camera");
+      break;
```
ahojnnes, line 1729: unnecessary

src/colmap/sensor/models_test.cc
```diff
@@ -136,6 +136,11 @@ void TestModel(const std::vector<double>& params) {
 
   for (int x = 0; x <= 800; x += 50) {
     for (int y = 0; y <= 800; y += 50) {
+      if (CameraModelIsFisheye(CameraModel::model_id)) {
+        if (!FisheyeCameraModelIsValidPixel(
```
ahojnnes, line 140: can be a single if statement using &&

src/colmap/sensor/models.h
B1ueber2y: Haha thanks. I was following here: https://github.com/colmap/colmap/blob/main/src/colmap/sensor/models.h#L1585. Will update that as well tonight : )

src/colmap/sensor/models.h
```diff
@@ -193,6 +213,35 @@ struct BaseCameraModel {
...
+      *vv *= theta / r;
+    }
+  }
+
+  template <typename T>
+  static inline void NormalFromFisheye(const T uu, const T vv, T* u, T* v) {
+    *u = uu;
+    *v = vv;
+    const T theta = ceres::sqrt(uu * uu + vv * vv);
+    const T theta_cos_theta = theta * ceres::cos(theta);
```
ahojnnes, line 236: Can we not handle the theta > PI/2 case here in the if-condition instead? theta_cos_theta would be zero in this case and then negative?

B1ueber2y: It will be positive again when it becomes large enough. But yes we can do the check here on theta. 

However, how should we do it here? Since we don't have a boolean flag in this function to indicate whether it is viable. And since this is also used in Ceres context, I did not see a good way to integrate here.

ahojnnes: We could just clamp it to the range [EPS, PI - EPS] ?

src/colmap/sensor/models.h
```diff
@@ -1284,58 +1359,64 @@ std::vector<double> SimpleRadialFisheyeCameraModel::InitializeParams(
...
+  *vv = (y - c2) / f;
+}
+
+template <typename T>
+void SimpleRadialFisheyeCameraModel::ImgFromCam(
+    const T* params, T u, T v, T w, T* x, T* y) {
   u /= w;
   v /= w;
 
+  // Fisheye coordinates
```
ahojnnes, line 1389: nit: few more comments that are redundant to the code. Thanks.

src/colmap/sensor/models.h
B1ueber2y: No. The problem is that for Fisheye cameras, there is only a range of pixel coordinates being valid. That is to say, there is only a constrained set of pixels that can be projected onto because of the Fisheye mapping. A distance of pi/2 to the image center indicates an infinity in the normal coordinate, and a distance larger than pi/2 can never be a valid pixel in the Fisheye cameras.

ahojnnes: Yes, I understand. But clamping to the suggested range above would solve that problem? Points outside that range would just be moved to the boundary of the valid pixels using this approach?

ahojnnes: Let's keep it as is. It seems like nobody hit this issue in practice before. Thanks.

ahojnnes (approved):

src/colmap/sensor/models.h
```diff
@@ -1592,4 +1665,46 @@ double CameraModelCamFromImgThreshold(const CameraModelId model_id,
...
+    default:
+      return false;
+
+#undef CAMERA_MODEL_CASE
+  }
+
+  return false;
+}
+
+bool FisheyeCameraModelIsValidPixel(const CameraModelId model_id,
```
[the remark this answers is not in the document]
B1ueber2y: I know that there have not been any cases that hit this, but I could see the potential use case of this feature outside the tests. Should we just keep it as exposed since this is a natural feature for Fisheye cameras?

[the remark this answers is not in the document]
B1ueber2y: Sure sounds good. This is done now.

PR #2796, opened 2024-09-20
Use branch prediction in PRNG functions

ahojnnes:


B1ueber2y (approved):

B1ueber2y:
I don't expect the performance improvement to be significant here? Or where does it matter the most?

ahojnnes:
No, it does not matter much in my profiling except for tight loops.

PR #2805, opened 2024-09-25
Expose UpdatePoint3DErrors to pycolmap

theartful:
I'm working on a SLAM system, and writing the results in colmap format using pycolmap. Computing reprojection error is trivial, but it would be nice of pycolmap can do it as well.


ahojnnes:
Thanks.

PR #2820, opened 2024-10-02
Better minimal generalized relative pose estimator using poselib

ahojnnes:
Towards implementation of "structure-less" resectioning, so colmap works with two-view only overlap scenarios.


B1ueber2y:
Thanks. I thought the idea is to add PoseLib as a dependency rather than copying the function, such that the development flow is healthier? I assume we could just include this: https://github.com/PoseLib/PoseLib/blob/0439b2d361125915b8821043fca9376e6cc575b9/PoseLib/solvers/gen_relpose_6pt.h?

ahojnnes:
You are right but I ran into some blockers in my branch. I got lazy here now... :-| I'll see to resurrect my poselib PR and see if the blocker can be resolved in reasonable time.

PR #2824, opened 2024-10-06
Make Image::cam_from_world_ optional

sarlinpe:
- `Image::cam_from_world_` becomes optional, making the `registered_` flag unnecessary and greatly simplifying the API (this was a common source of confusion)
- Calling `Image::CamFromWorld` is illegal if the pose hasn't been previously set.
- This helped catch a bug in the `PosePriorBA` - all images (even unregistered ones) were used in the optimization.
- `Reconstruction::RegImageIds` returns a set instead of a vector (and not an unordered set because the order should be deterministic across platforms).

### DIFF
--- a/src/colmap/estimators/alignment_test.cc
+++ b/src/colmap/estimators/alignment_test.cc
@@ -126,8 +126,8 @@ TEST(Alignment, MergeReconstructions) {
   Reconstruction tgt_reconstruction = src_reconstruction;
 
   // Remove the camera of the first image from the target reconstruction
-  const std::vector<image_t>& image_ids = tgt_reconstruction.RegImageIds();
-  camera_t camera_id = tgt_reconstruction.Image(image_ids[0]).CameraId();
+  const std::set<image_t> image_ids = tgt_reconstruction.RegImageIds();
+  camera_t camera_id = tgt_reconstruction.Image(*image_ids.begin()).CameraId();
   for (const auto& image_id : image_ids) {
     if (tgt_reconstruction.Image(image_id).CameraId() == camera_id) {
       tgt_reconstruction.DeRegisterImage(image_id);
--- a/src/colmap/estimators/alignment_test.cc
+++ b/src/colmap/estimators/alignment_test.cc
@@ -127,7 +127,8 @@ TEST(Alignment, MergeReconstructions) {
 
   // Remove the camera of the first image from the target reconstruction
   const std::set<image_t> image_ids = tgt_reconstruction.RegImageIds();
-  camera_t camera_id = tgt_reconstruction.Image(*image_ids.begin()).CameraId();
+  const camera_t camera_id =
+      tgt_reconstruction.Image(*image_ids.begin()).CameraId();
   for (const auto& image_id : image_ids) {
     if (tgt_reconstruction.Image(image_id).CameraId() == camera_id) {
       tgt_reconstruction.DeRegisterImage(image_id);
--- a/src/colmap/estimators/bundle_adjustment.cc
+++ b/src/colmap/estimators/bundle_adjustment.cc
@@ -412,7 +412,7 @@ void BundleAdjuster::AddImageToProblem(const image_t image_id,
   Camera& camera = *image.CameraPtr();
 
   // CostFunction assumes unit quaternions.
-  THROW_CHECK(image.IsRegistered());
+  THROW_CHECK(image.HasPose());
   image.CamFromWorld().rotation.normalize();
 
   double* cam_from_world_rotation =
@@ -945,7 +945,7 @@ void PosePriorBundleAdjuster::AddPosePriorToProblem(
     return;
   }
   Image& image = reconstruction->Image(image_id);
-  THROW_CHECK(image.IsRegistered());
+  THROW_CHECK(image.HasPose());
   double* cam_from_world_translation = image.CamFromWorld().translation.data();
 
   // If image has not been added to the problem do not use it
--- a/src/colmap/estimators/bundle_adjustment.cc
+++ b/src/colmap/estimators/bundle_adjustment.cc
@@ -882,9 +882,9 @@ bool PosePriorBundleAdjuster::Solve(Reconstruction* reconstruction) {
 
   // Fix 7-DOFs of BA problem if not enough valid pose priors
   if (!use_prior_position_) {
-    auto it_reg_image_ids = reconstruction->RegImageIds().begin();
-    config_.SetConstantCamPose(*it_reg_image_ids);
-    config_.SetConstantCamPositions(*(++it_reg_image_ids), {0});
+    auto reg_image_ids_it = reconstruction->RegImageIds().begin();
+    config_.SetConstantCamPose(*reg_image_ids_it);
+    config_.SetConstantCamPositions(*(++reg_image_ids_it), {0});
   }
 
   // Normalize the reconstruction to avoid any numerical instability BUT do not
--- a/src/colmap/estimators/bundle_adjustment.cc
+++ b/src/colmap/estimators/bundle_adjustment.cc
@@ -1021,7 +1021,11 @@ bool PosePriorBundleAdjuster::Sim3DAlignment(Reconstruction* reconstruction) {
 
   if (success) {
     reconstruction->Transform(sim3_tform);
+  } else {
+    LOG(WARNING) << "Sim3 alignment w.r.t. prior position failed!";
+  }
 
+  if (VLOG_IS_ON(2) && success) {
     std::vector<double> verr2_wrt_prior;
     verr2_wrt_prior.reserve(vini_err2_wrt_prior.size());
     for (const image_t& image_id : reconstruction->RegImageIds()) {
@@ -1042,8 +1046,6 @@ bool PosePriorBundleAdjuster::Sim3DAlignment(Reconstruction* reconstruction) {
     VLOG(2) << "Sim3 alignment error w.r.t. prior position:\n"
             << "  - rmse:   " << std::sqrt(Mean(verr2_wrt_prior)) << " m\n"
             << "  - median: " << std::sqrt(Median(verr2_wrt_prior)) << " m\n";
-  } else {
-    LOG(WARNING) << "Sim3 alignment w.r.t. prior position failed!";
   }
 
   return success;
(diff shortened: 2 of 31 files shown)


src/colmap/scene/reconstruction.cc
```diff
@@ -262,11 +257,8 @@ void Reconstruction::DeRegisterImage(const image_t image_id) {
     }
   }
 
-  image.SetRegistered(false);
-
-  reg_image_ids_.erase(
-      std::remove(reg_image_ids_.begin(), reg_image_ids_.end(), image_id),
-      reg_image_ids_.end());
+  image.DeRegister();
```
B1ueber2y, line 260: First I think the current logistics is a bit confusing: there are two levels of "registration", and they are not always consistent: 1) whether an image has a cam from world set 2) whether an image has been registered in the reconstruction with Reconstruction::RegisterImage.

Personally I don't think these two levels should be made equivalent, so would be better if we use a different word from "register" in the image class.

And above all, the image should anyway not be deregistered in Reconstruction::DeregisterImage. Only the removal from the reg_image_ids is needed here. Here exactly the confusion of different levels comes to life: with the current logic if we call Reconstruction::RegisterImage and then Reconstruction::DeRegisterImage, the image pose will be lost.

sarlinpe: This PR avoids breaking changes as much as possible. The fact that `Reconstruction::DeregisterImage` de-registers the image object as well is consistent with the current behavior and avoids silently breaking any user code. IMHO whether the image state should be called "registered" or something else is a question beyond this PR. Maybe @ahojnnes can weigh in there.

B1ueber2y: Only with the change of this PR the image pose is lost after RegisterImage and DeRegisterImage, since before only the registered_ flag is changed. I think the registered_ flag before was conceptually different from whether the image has a pose or not in the design, but maybe I misunderstood here @ahojnnes 

P.S. (I agree that the following part might be beyond the scope of this PR) On the other hand, if the registered_ was meant to be equivalent to whether the image has a pose or not, it's a bit strange to have Reconstruction::RegisterImage and Reconstruction::DeregisterImage as all the operations should be on the image objects.

ahojnnes: I don't have a strong opinion but I would suggest removing the "registered" terminology completely from the image now and isolate it to the reconstruction, and simply have a "HasCamFromWorld()" for the image. This is me writing on the phone without being able to look at all usages, so this may not be an ideal suggestion.

src/colmap/estimators/alignment_test.cc
```diff
@@ -126,8 +126,8 @@ TEST(Alignment, MergeReconstructions) {
   Reconstruction tgt_reconstruction = src_reconstruction;
 
   // Remove the camera of the first image from the target reconstruction
-  const std::vector<image_t>& image_ids = tgt_reconstruction.RegImageIds();
-  camera_t camera_id = tgt_reconstruction.Image(image_ids[0]).CameraId();
+  const std::set<image_t> image_ids = tgt_reconstruction.RegImageIds();
```
ahojnnes, line 129: Could keep the reference to avoid the copy?

src/colmap/scene/image.h
```diff
@@ -78,7 +79,7 @@ class Image {
 
   // Check if image is registered.
   inline bool IsRegistered() const;
-  inline void SetRegistered(bool registered);
+  inline void DeRegister();
```
ahojnnes, line 82: As per existing discussion, this would become ResetCamFromWorld() ?

src/colmap/estimators/bundle_adjustment.cc
```diff
@@ -881,9 +882,9 @@ bool PosePriorBundleAdjuster::Solve(Reconstruction* reconstruction) {
 
   // Fix 7-DOFs of BA problem if not enough valid pose priors
   if (!use_prior_position_) {
-    const std::vector<image_t>& reg_image_ids = reconstruction->RegImageIds();
-    config_.SetConstantCamPose(reg_image_ids[0]);
-    config_.SetConstantCamPositions(reg_image_ids[1], {0});
+    auto it_reg_image_ids = reconstruction->RegImageIds().begin();
```
ahojnnes, line 885: nit: The codebase generates uses x_it vs it_x notation.

src/colmap/estimators/alignment_test.cc
sarlinpe: No because we cannot modify the set (via DeRegisterImage below) while iterating it.

ahojnnes: Got it. Didn't see the context below on my phone.

src/colmap/scene/reconstruction.cc
sarlinpe: I have updated `{IsRegistered,DeRegister}` to `{Has,Reset}Pose` (less verbose than `CamFromWorld`). I have also updated `Reconstruction::IsImageRegistered` to use `reg_image_ids_`.  One last point: I still think that removing `Image::ResetPose` from `Reconstruction::DeRegisterImage` is a bad idea since we never do one without the other and it might silently break user code.

src/colmap/scene/reconstruction.cc
```diff
@@ -435,10 +430,10 @@ const class Image* Reconstruction::FindImageWithName(
 std::vector<std::pair<image_t, image_t>> Reconstruction::FindCommonRegImageIds(
     const Reconstruction& other) const {
   std::vector<std::pair<image_t, image_t>> common_reg_image_ids;
-  for (const auto image_id : reg_image_ids_) {
+  for (const auto image_id : RegImageIds()) {
     const auto& image = Image(image_id);
     const auto* other_image = other.FindImageWithName(image.Name());
-    if (other_image != nullptr && other_image->IsRegistered()) {
+    if (other_image != nullptr && other_image->HasPose()) {
```
ahojnnes, line 436: Shall we use other.IsImageRegistered() here?

src/colmap/scene/image.h
```diff
@@ -90,6 +87,11 @@ class Image {
   // World to camera pose.
   inline const Rigid3d& CamFromWorld() const;
   inline Rigid3d& CamFromWorld();
+  inline std::optional<Rigid3d>& MaybeCamFromWorld();
```
ahojnnes, line 90: Is this method actually useful for something?

ahojnnes (approved):

src/colmap/scene/reconstruction.cc
```diff
@@ -87,7 +87,7 @@ void Reconstruction::TearDown() {
   // Remove all not yet registered images.
   std::unordered_set<camera_t> keep_camera_ids;
   for (auto it = images_.begin(); it != images_.end();) {
-    if (it->second.IsRegistered()) {
+    if (it->second.HasPose()) {
```
B1ueber2y, line 90: Should we use IsImageRegistered?

src/colmap/scene/reconstruction.cc
```diff
@@ -125,11 +125,11 @@ void Reconstruction::AddImage(class Image image) {
     image.SetCameraPtr(&camera);
   }
   const image_t image_id = image.ImageId();
-  const bool is_registered = image.IsRegistered();
+  const bool is_registered = image.HasPose();
```
B1ueber2y, line 128: same here

src/colmap/scene/reconstruction.cc
B1ueber2y, line 436: same here. Not sure there are more of such cases

src/colmap/scene/reconstruction_io.cc
```diff
@@ -421,7 +420,7 @@ void WriteImagesText(const Reconstruction& reconstruction,
        << reconstruction.ComputeMeanObservationsPerRegImage() << std::endl;
 
   for (const auto& image : reconstruction.Images()) {
-    if (!image.second.IsRegistered()) {
+    if (!image.second.HasPose()) {
```
B1ueber2y, line 423: IsImageRegistered?

src/colmap/scene/reconstruction_io.cc
```diff
@@ -533,7 +532,7 @@ void WriteImagesBinary(const Reconstruction& reconstruction,
   WriteBinaryLittleEndian<uint64_t>(&file, reconstruction.NumRegImages());
 
   for (const auto& image : reconstruction.Images()) {
-    if (!image.second.IsRegistered()) {
+    if (!image.second.HasPose()) {
```
B1ueber2y, line 535: same

src/colmap/scene/reconstruction_io.cc
```diff
@@ -988,7 +987,7 @@ void ExportVRML(const Reconstruction& reconstruction,
   points.emplace_back(-six / 3.0, +siy / 3.0, six * 1.0 * 2.0);
 
   for (const auto& image : reconstruction.Images()) {
-    if (!image.second.IsRegistered()) {
+    if (!image.second.HasPose()) {
```
B1ueber2y, line 990: same

src/colmap/scene/reconstruction_test.cc
```diff
@@ -104,7 +104,7 @@ TEST(Reconstruction, AddImage) {
   reconstruction.AddImage(image);
   EXPECT_TRUE(reconstruction.ExistsImage(1));
   EXPECT_EQ(reconstruction.Image(1).ImageId(), 1);
-  EXPECT_FALSE(reconstruction.Image(1).IsRegistered());
+  EXPECT_FALSE(reconstruction.Image(1).HasPose());
```
B1ueber2y, line 107: same. I will not add more comments regarding this : )

src/colmap/scene/reconstruction.cc
B1ueber2y: Since we have completely disentangled `ResetPose` and `DeRegisterImage` in the new design, I don't think `ResetPose` is needed here. With the new optional design `ResetPose` will also lose the pose information while the previous code will not, so we will anyway need some efforts from the users if they hit this.

src/colmap/scene/reconstruction.cc
sarlinpe: No, here we specifically check whether the new image should be set as registered.

src/colmap/scene/image.h
sarlinpe: Yes, when copying an existing image into a new one without having to check whether it has a pose.

PR #2825, opened 2024-10-06
More informative exception if invalid access of image/camera/point3D

sarlinpe:
- Calling `reconstruction.Image(image_id)` for a nonexistent image ID raises an exception with the uninformative message `_Map_base::at`. This is especially confusing for Python users. This PR makes the exception message more informative.
- This PR also shares the implementations of the const and non-const getters, as suggested here: https://stackoverflow.com/a/856839 Does it still make sense to keep these implementations inlined?

### DIFF
--- a/src/colmap/scene/reconstruction.h
+++ b/src/colmap/scene/reconstruction.h
@@ -266,28 +266,46 @@ size_t Reconstruction::NumRegImages() const { return reg_image_ids_.size(); }
 size_t Reconstruction::NumPoints3D() const { return points3D_.size(); }
 
 const struct Camera& Reconstruction::Camera(const camera_t camera_id) const {
-  return cameras_.at(camera_id);
+  try {
+    return cameras_.at(camera_id);
+  } catch (const std::out_of_range& e) {
+    throw std::invalid_argument(
+        StringPrintf("Camera with ID %d does not exist", camera_id));
+  }
 }
 
 const class Image& Reconstruction::Image(const image_t image_id) const {
-  return images_.at(image_id);
+  try {
+    return images_.at(image_id);
+  } catch (const std::out_of_range& e) {
+    throw std::invalid_argument(
+        StringPrintf("Image with ID %d does not exist", image_id));
+  }
 }
 
 const struct Point3D& Reconstruction::Point3D(
     const point3D_t point3D_id) const {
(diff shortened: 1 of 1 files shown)


B1ueber2y (approved):

src/colmap/scene/reconstruction.h
```diff
@@ -266,28 +266,46 @@ size_t Reconstruction::NumRegImages() const { return reg_image_ids_.size(); }
...
+    return points3D_.at(point3D_id);
+  } catch (const std::out_of_range& e) {
+    throw std::invalid_argument(
+        StringPrintf("Point3D with ID %d does not exist", point3D_id));
+  }
 }
 
 struct Camera& Reconstruction::Camera(const camera_t camera_id) {
-  return cameras_.at(camera_id);
+  return const_cast<struct Camera&>(
```
ahojnnes, line 297: I'd suggest replicating the code here instead of the const cast. I am not sure if the compiler makes some optimizations with the const that then get violated by this const cast. Code replication is not all that bad and here the savings are also marginal.

ahojnnes (commented):
Thanks, lgtm.

src/colmap/scene/reconstruction.h
```diff
@@ -266,28 +266,46 @@ size_t Reconstruction::NumRegImages() const { return reg_image_ids_.size(); }
 size_t Reconstruction::NumPoints3D() const { return points3D_.size(); }
 
 const struct Camera& Reconstruction::Camera(const camera_t camera_id) const {
-  return cameras_.at(camera_id);
+  try {
+    return cameras_.at(camera_id);
+  } catch (const std::out_of_range& e) {
+    throw std::invalid_argument(
```
ahojnnes, line 272: The .at method throws an out of range exception. If we want full backwards compatibility on all call sites that may catch, we should rethrow the same.

src/colmap/scene/reconstruction.h
sarlinpe: Ok, done.

src/colmap/scene/reconstruction.h
sarlinpe: We don't catch it anywhere in the COLMAP codebase. But I agree that we should remain backward compatible with user code - I've switched to out-of-range, PTAL.

PR #2829, opened 2024-10-10
Structure-less image registration using generalized relative pose estimator

ahojnnes:
Enables to register images without 3 view overlap by using 2d-2d constraints instead of 3d-3d constraints. This was proposed in "Structure from Motion Using Structure-less Resection" originally. Different from the original paper, the implementation in this PR relies on the generalized relative pose solver, which can deal with more generic setups instead of the presented customized 5+1/2 and 6+1/2 solvers.


src/colmap/sfm/observation_manager.cc
```diff
@@ -455,10 +454,28 @@ size_t ObservationManager::FilterPoints3DWithLargeReprojectionError(
   return num_filtered;
 }
 
+void ObservationManager::RegisterImage(const image_t image_id) {
```
ahojnnes, line 457: Needs unit test.

src/colmap/sfm/observation_manager.cc
```diff
@@ -451,12 +450,37 @@ size_t ObservationManager::FilterPoints3DWithLargeReprojectionError(
   return num_filtered_observations;
 }
 
+void ObservationManager::RegisterFrame(const frame_t frame_id) {
```
B1ueber2y, line 460: I think we should also update the logic to support existing reconstruction here: https://github.com/colmap/colmap/blob/main/src/colmap/sfm/observation_manager.cc#L92.

src/colmap/sfm/incremental_mapper_impl.h
```diff
@@ -75,6 +75,7 @@ class IncrementalMapperImpl {
 
   // Implement IncrementalMapper::FindNextImages
   static std::vector<image_t> FindNextImages(
+      bool structure_less,
```
B1ueber2y: should we make it an optional argument at the end with false as default?

python/examples/custom_incremental_pipeline.py
```diff
@@ -120,35 +129,60 @@ def reconstruct_sub_model(controller, mapper, mapper_options, reconstruction):
...
+                if structure_less:
+                    logging.info(
+                        "Registering image with structure-less fallback"
+                    )
+                    num_vis = (
+                        mapper.observation_manager.num_visible_correspondences(
+                            next_image_id
+                        )
+                    )
+                    num_obs = mapper.observation_manager.num_correspondences(
```
B1ueber2y: num_obs -> num_corrs? Since we have both num_observations and num_corrspondences variables that have different meanings.

src/colmap/sfm/incremental_mapper.cc
```diff
@@ -602,11 +604,278 @@ bool IncrementalMapper::RegisterNextGeneralFrame(const Options& options,
...
+  }
+
+  //////////////////////////////////////////////////////////////////////////////
+  // Structure-less resectioning
+  //////////////////////////////////////////////////////////////////////////////
+
+  StructureLessAbsolutePoseEstimationOptions abs_pose_options;
+  // Structure-less resectioning uses epipolar Sampson error, so we are stricter
+  // in accepting 2D-2D correspondences inliers.
+  abs_pose_options.ransac_options.max_error = 0.5 * options.abs_pose_max_error;
```
B1ueber2y, line 697: Does it make sense to change the default values directly in ``StructureLessAbsolutePoseEstimationOptions``

B1ueber2y (commented):
Thank you!

src/colmap/sfm/incremental_mapper.cc
ahojnnes: Done.

src/colmap/sfm/incremental_mapper.cc
```diff
@@ -602,11 +605,278 @@ bool IncrementalMapper::RegisterNextGeneralFrame(const Options& options,
...
```
B1ueber2y, line 697: Just a reminder: this 0.5 needs to be removed after the new change : )

ahojnnes: I don't think so, why do you say so? The only thing I changed is the default value in the options struct.

B1ueber2y: Oh sorry i misread.

B1ueber2y (approved):

PR #2834, opened 2024-10-13
Return success status in low-level triangulation functions

ahojnnes:
More robustly handle degenerate configurations. Noticed a rare edge case during some tests. Should barely make a difference in practice but avoids issues when using the low-level functions in new places.

### DIFF
--- a/src/colmap/estimators/pose.cc
+++ b/src/colmap/estimators/pose.cc
@@ -145,8 +145,8 @@ bool RefineAbsolutePose(const AbsolutePoseRefinementOptions& options,
       std::make_unique<ceres::CauchyLoss>(options.loss_function_scale);
 
   double* camera_params = camera->params.data();
-  double* rig_from_world_rotation = cam_from_world->rotation.coeffs().data();
-  double* rig_from_world_translation = cam_from_world->translation.data();
+  double* cam_from_world_rotation = cam_from_world->rotation.coeffs().data();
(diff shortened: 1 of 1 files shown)


src/colmap/geometry/triangulation.cc
```diff
@@ -37,63 +37,58 @@
...
-    const std::vector<Eigen::Vector2d>& points2) {
-  THROW_CHECK_EQ(points1.size(), points2.size());
-
-  std::vector<Eigen::Vector3d> points3D(points1.size());
-
-  for (size_t i = 0; i < points3D.size(); ++i) {
-    points3D[i] = TriangulatePoint(
-        cam1_from_world, cam2_from_world, points1[i], points2[i]);
+  const Eigen::JacobiSVD<Eigen::Matrix4d> svd(A, Eigen::ComputeFullV);
+  if (svd.info() != Eigen::Success || svd.matrixV()(3, 3) == 0) {
```
sarlinpe, line 54: info() is only available in Eigen >= 3.4.0: https://gitlab.com/libeigen/eigen/-/commit/5bbc9cea93ef29cee2b8ffb2084d4ebca32600ba

ahojnnes: Thanks, didn't know. Will fix.

sarlinpe (approved):

PR #2841, opened 2024-10-16
Fix copying behaviors of Reconstruction regarding camera pointers

B1ueber2y:
Previous copying behavior is very dangerous after the introduction of camera pointer.

### DIFF
--- a/src/colmap/scene/reconstruction.cc
+++ b/src/colmap/scene/reconstruction.cc
@@ -43,6 +43,18 @@ namespace colmap {
 
 Reconstruction::Reconstruction() : max_point3D_id_(0) {}
 
+Reconstruction::Reconstruction(const Reconstruction& recon)
+    : cameras_(recon.cameras_),
+      images_(recon.images_),
+      points3D_(recon.points3D_),
+      reg_image_ids_(recon.reg_image_ids_),
+      max_point3D_id_(recon.max_point3D_id_) {
+  for (const auto& [image_id, image] : Images()) {
+    Image(image_id).ResetCameraPtr();
+    Image(image_id).SetCameraPtr(&Camera(image.CameraId()));
+  }
+}
+
 std::unordered_set<point3D_t> Reconstruction::Point3DIds() const {
   std::unordered_set<point3D_t> point3D_ids;
   point3D_ids.reserve(points3D_.size());
--- a/src/colmap/scene/reconstruction.cc
+++ b/src/colmap/scene/reconstruction.cc
@@ -43,28 +43,28 @@ namespace colmap {
 
 Reconstruction::Reconstruction() : max_point3D_id_(0) {}
 
-Reconstruction::Reconstruction(const Reconstruction& recon)
-    : cameras_(recon.cameras_),
-      images_(recon.images_),
-      points3D_(recon.points3D_),
-      reg_image_ids_(recon.reg_image_ids_),
-      max_point3D_id_(recon.max_point3D_id_) {
-  for (const auto& [image_id, image] : Images()) {
-    Image(image_id).ResetCameraPtr();
-    Image(image_id).SetCameraPtr(&Camera(image.CameraId()));
+Reconstruction::Reconstruction(const Reconstruction& other)
+    : cameras_(other.cameras_),
+      images_(other.images_),
+      points3D_(other.points3D_),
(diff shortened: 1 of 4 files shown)


src/colmap/scene/reconstruction.h
```diff
@@ -59,6 +59,8 @@ class DatabaseCache;
 class Reconstruction {
  public:
   Reconstruction();
+  // Update the camera pointers for each image at copy constructor.
+  inline Reconstruction(const Reconstruction& recon);
```
ahojnnes, line 63: I suggest to not inline but move the implementation to the .cc file.

ahojnnes:
The same is true for copy assignment?

src/colmap/scene/reconstruction.h
B1ueber2y: Done.

B1ueber2y:
> The same is true for copy assignment?

Yes. This is added now.

src/colmap/scene/reconstruction.cc
```diff
@@ -43,6 +43,33 @@ namespace colmap {
 
 Reconstruction::Reconstruction() : max_point3D_id_(0) {}
 
+Reconstruction::Reconstruction(const Reconstruction& recon)
```
ahojnnes, line 46: Would it make sense to add a test for these two new methods and assert equality between the two objects (or their serialized states)?

B1ueber2y: Sorry maybe I am missing something. The Reconstruction class in colmap does not naturally support serialization? Do we write a new method for this?

ahojnnes: You can read and write the reconstruction using reconstruction_io?

B1ueber2y: Thanks. This has been added. Wrote a new function for comparing binary files. I hope I did not miss any existing utilities in the codebase. PTAL

ahojnnes: I created a new PR that should hopefully simplify the tests in this PR: https://github.com/colmap/colmap/pull/2842

src/pycolmap/scene/reconstruction.cc
```diff
@@ -249,6 +250,10 @@ void BindReconstruction(py::module& m) {
            [](const Reconstruction& self, const py::dict&) {
              return Reconstruction(self);
            })
+      .def("__assign__",
```
ahojnnes, line 253: Copying the reconstruction on each assignment in Python is not very idiomatic. I suggest to instead create a "copy()" method and leave the `__assign__` to be "copy-by-ref".

src/colmap/scene/reconstruction_test.cc
```diff
@@ -49,6 +55,61 @@ bool CheckReconstruction(const Reconstruction& reconstruction) {
...
+    if (byte1 != byte2) {
+      return false;  // Files are different
+    }
+  }
+  return true;
+}
+
+// Compare two reconstructions
+// The binary files are compared after serialized to disk
+bool CompareReconstructions(const std::string& test_dir,
```
ahojnnes, line 91: After the new reconstruction_io changes, this can now simply use stringstreams and directly compare the stream contents.

src/pycolmap/scene/reconstruction.cc
sarlinpe: Indeed, I cannot find a reference to `__assign__` anywhere. Dunder names `__*__` are reserved by the Python language, let's not create a new one please. `pycolmap.Reconstruction` already implements `__copy__` and `__deepcopy__`, I think that an extra method is not needed.

B1ueber2y: I dont think this method is necessary. I will remove it.

ahojnnes: Ok, I got tricked into thinking that this is the way to overload the assignment operator in Python. Agreed that we shouldn't invent a new `__*__` operator but instead reuse the `__copy__` method or introduce an explicit copy() method. For example, numpy provides a similar method for arrays.

src/colmap/scene/reconstruction_test.cc
B1ueber2y: Thanks for the PR. This is updated.

ahojnnes (approved):

PR #2846, opened 2024-10-22
Select CUDA device when bundle adjustment uses GPU

ahojnnes:
Small fix for the (very unlikely?) case where CUDA sparse solvers are available but the dense solvers are not.

### DIFF
--- a/src/colmap/estimators/bundle_adjustment.cc
+++ b/src/colmap/estimators/bundle_adjustment.cc
@@ -363,6 +363,12 @@ ceres::Solver::Options BundleAdjuster::SetUpSolverOptions(
     max_num_images_direct_dense_solver =
         options_.max_num_images_direct_dense_gpu_solver;
   }
+#else
+  if (options_.use_gpu) {
+    LOG_FIRST_N(WARNING, 1)
+        << "Requested to use GPU for bundle adjustment, but Ceres was "
+           "compiled without CUDA support. Falling back to CPU-based solvers.";
+  }
 #endif
 
 #if (CERES_VERSION_MAJOR >= 3 ||                                \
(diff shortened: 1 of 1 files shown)


src/colmap/estimators/bundle_adjustment.cc
```diff
@@ -352,13 +352,13 @@ ceres::Solver::Options BundleAdjuster::SetUpSolverOptions(
       options_.max_num_images_direct_sparse_cpu_solver;
 
 #ifdef COLMAP_CUDA_ENABLED
```
sarlinpe, line 354: It would be nice to log a warning when the user selects `use_gpu=true` but either COLMAP or Ceres are not built with Cuda support. Right now it's silent and a bit confusing.

ahojnnes: Good idea. Will do.

src/colmap/estimators/bundle_adjustment.cc
```diff
@@ -352,29 +352,54 @@ ceres::Solver::Options BundleAdjuster::SetUpSolverOptions(
...
+    cuda_solver_enabled = true;
     solver_options.sparse_linear_algebra_library_type = ceres::CUDA_SPARSE;
     max_num_images_direct_sparse_solver =
         options_.max_num_images_direct_sparse_gpu_solver;
   }
+#else
+  if (options_.use_gpu) {
+    LOG_FIRST_N(WARNING, 1)
+        << "Requested to use GPU for bundle adjustment, but Ceres was "
+           "compiled without cuDSS support. Falling back to CPU-based solvers.";
```
sarlinpe, line 387: If Ceres has CUDA support but no cuDSS, the solving can still use the GPU if there are few images and we thus select the dense solver, right? So shouldn't it be instead "Falling back to CPU-based solvers **if many images are optimized**."

ahojnnes: Did something similar. Thanks.

sarlinpe (approved):

PR #2848, opened 2024-10-22
fix: suppress warning about CMP0167 when compile with cmake over 3.30

HernandoR:
CMP1067 is about find_boost module, contents are available [here](https://cmake.org/cmake/help/latest/policy/CMP0167.html).

boost 1.70 was established at Apr 10, 2019, and we should respect the community

Challenges may have of configuring cmake above 3.30, but using boost lib below 1.70. I cannot confirm if this situation can happen, nor can reproduce.

To avoid that challenge, we should set the policy to OLD explicitly align with the default.


ahojnnes:
Thank you for the suggested fix. This needs to be placed in a different file, so downstream consumers of the colmap library also benefit from the fix: https://github.com/colmap/colmap/pull/2853

PR #2855, opened 2024-10-24
Use the C++ string representation for Python dataclass objects

sarlinpe:
https://github.com/colmap/colmap/pull/2842 removed the custom `__repr__` for some objects like `Image`. Pycolmap then falls back to enumerating all the attributes, which is suboptimal - for example for `Image` we don't want to print all 2D points to void bloating the console output. This PR recovers the previous behavior for objects for which `stringstream << obj` is defined.

### DIFF
--- a/src/pycolmap/helpers.h
+++ b/src/pycolmap/helpers.h
@@ -211,7 +211,7 @@ std::string CreateSummary(const T& self, bool write_type) {
 }
 
 template <typename T>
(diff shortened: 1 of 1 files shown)


B1ueber2y:
Indeed. Thanks for the catch on the ``Image`` class __repr__

B1ueber2y (approved):
The new type traits update works in my test.

ahojnnes (approved):

Issue #2802, opened 2024-09-24
Issue with TestCamFromImgToImg in sensor/models_test.cc for Fisheye models


nushakrishnan:
**Describe the bug**
In models_test.cc, [line 171,](https://github.com/colmap/colmap/blob/d70edd784808d29c855fff0b4ffc2bdc3ea290f1/src/colmap/sensor/models_test.cc#L171) we test the unprojection and projection of pixels for various camera models. However, this test fails under certain conditions for Fisheye camera models. For pixel coordinates, whose theta values are greater than pi/4, the test fails.  

**To Reproduce**
Consider the following code snippet. I have selected some camera params for the OpenCVFisheye model. This mini test generates a grid - the pixel coordinates in the white area represent a match between original value and reprojected value. The black region represents a mismatch after reprojection. It also prints the minimum 10 theta outliers where the forward and inverse transformations of the camera model do not match.

```python

import pycolmap
import math
import numpy as np
import matplotlib
import matplotlib.pyplot as plt
from tqdm import tqdm

def test_fisheye_model():
    camera_id = 1
    width = 640
    height = 480

    opencv_fisheye_params = [
        241.996, 252.123, 323.024, 240.497, -0.0232, 0.0924, -0.0591, 0.003
    ]

    fx, fy = opencv_fisheye_params[0], opencv_fisheye_params[1]
    cx, cy = opencv_fisheye_params[2], opencv_fisheye_params[3]

    ocv_camera = pycolmap.Camera(model='OPENCV_FISHEYE',
                                 width=width,
                                 height=height,
                                 params=opencv_fisheye_params,
                                 camera_id=camera_id)

    xs, ys = np.meshgrid(np.arange(0, 800), np.arange(0, 800))
    dists = np.stack([(xs - cx) / fx, (ys - cy) / fy], axis=2)
    thetas = np.linalg.norm(dists, axis=2)
    bw_grid = np.zeros((800, 800)).astype(bool)

    for i in tqdm(range(0, 800, 1)):
        for j in range(0, 800, 1):
            if thetas[i, j] >= math.pi / 2:
                continue
            pixel = np.array([i, j])
            point_colmap = ocv_camera.cam_from_img(pixel)
            reproj_point = ocv_camera.img_from_cam(point_colmap)
            if np.allclose(pixel, reproj_point):
                bw_grid[i, j] = True
            else:
                bw_grid[i, j] = False

[...]

PR #2857, opened 2024-10-26
Use analytical Jacobian in IterativeUndistortion. Add trust region

B1ueber2y:
Attempt on https://github.com/colmap/colmap/issues/2802 (improvement but did not manage to fully fix as discussed in the post below)

Remove templating of CamFromImg as it is never used in the optimization (and is tricky to be used in the optimization due to the unrolling in IterativeUndistortion). 

Use Jet from ceres to enable analytical Jacobian in the IterativeUndistortion.

### DIFF
--- a/src/colmap/sensor/models.h
+++ b/src/colmap/sensor/models.h
@@ -691,7 +691,7 @@ void BaseCameraModel<CameraModel>::IterativeUndistortion(const double* params,
 
     // Update
     Eigen::Vector2d step_x = J.partialPivLu().solve(x + dx - x0);
-    double radius = std::max(x.norm() * kRelStepRadius, kStepRadius);
+    const double radius = std::max(x.norm() * kRelStepRadius, kStepRadius);
     if (step_x.norm() > radius) {
       step_x *= (radius / step_x.norm());
     }
--- a/src/colmap/sensor/models.h
+++ b/src/colmap/sensor/models.h
@@ -686,7 +686,7 @@ void BaseCameraModel<CameraModel>::IterativeUndistortion(const double* params,
     CameraModel::Distortion(
         params_jet, x_jet[0], x_jet[1], &dx_jet[0], &dx_jet[1]);
     dx[0] = dx_jet[0].a;
-    dx[1] = dx_jet[0].a;
+    dx[1] = dx_jet[1].a;
     J(0, 0) = dx_jet[0].v[0];
     J(0, 1) = dx_jet[0].v[1];
     J(1, 0) = dx_jet[1].v[0];
--- a/src/colmap/sensor/models.h
+++ b/src/colmap/sensor/models.h
@@ -665,10 +665,6 @@ void BaseCameraModel<CameraModel>::IterativeUndistortion(const double* params,
   const Eigen::Vector2d x0(*u, *v);
   Eigen::Vector2d x(*u, *v);
   Eigen::Vector2d dx;
-  Eigen::Vector2d dx_0b;
-  Eigen::Vector2d dx_0f;
-  Eigen::Vector2d dx_1b;
-  Eigen::Vector2d dx_1f;
 
   ceres::Jet<double, 2> params_jet[CameraModel::num_extra_params];
   for (size_t i = 0; i < CameraModel::num_extra_params; ++i) {
@@ -676,7 +672,6 @@ void BaseCameraModel<CameraModel>::IterativeUndistortion(const double* params,
   }
   for (size_t i = 0; i < kNumIterations; ++i) {
     // Get Jacobian
-    CameraModel::Distortion(params, x(0), x(1), &dx(0), &dx(1));
     ceres::Jet<double, 2> x_jet[2];
     x_jet[0] = ceres::Jet<double, 2>(x(0));
     x_jet[1] = ceres::Jet<double, 2>(x(1));
@@ -687,10 +682,10 @@ void BaseCameraModel<CameraModel>::IterativeUndistortion(const double* params,
         params_jet, x_jet[0], x_jet[1], &dx_jet[0], &dx_jet[1]);
     dx[0] = dx_jet[0].a;
     dx[1] = dx_jet[1].a;
-    J(0, 0) = dx_jet[0].v[0];
+    J(0, 0) = dx_jet[0].v[0] + 1;
     J(0, 1) = dx_jet[0].v[1];
     J(1, 0) = dx_jet[1].v[0];
-    J(1, 1) = dx_jet[1].v[1];
+    J(1, 1) = dx_jet[1].v[1] + 1;
 
     // Update
     const Eigen::Vector2d step_x = J.partialPivLu().solve(x + dx - x0);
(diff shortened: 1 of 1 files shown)


B1ueber2y:
Making the Jacobian analytical brings some gain. However, Newton does not converge when the theta is larger than around 1.25 for the parameters in https://github.com/colmap/colmap/issues/2802. The loss landscape is very bad for Newton convergence, that leads to extreme oscillations. 

With the previous ``IterativeUndistortion`` the oscillation can lead to a different root potentially with norm larger than pi/2, which essentially invalids the tangent function, leading to > 500 pixel error. Adding a trust region would reduce the forward-backward error to always less than 75 pixels (extreme case happens for thetas close to pi/2), but does not resolve the convergence issue. The fact that k3, k4 are large enough for the radial distortion with an exponentially growing theta^8 makes it hard to converge with theta that is larger than 1.25.

I dont think we can do better with Newton here. I also tried adding damping factors that does not give much gain. I believe we need to resort to other optimization approach in order to fully resolve the issue.

src/colmap/sensor/models.h
```diff
@@ -647,41 +652,49 @@ T BaseCameraModel<CameraModel>::CamFromImgThreshold(const T* params,
...
+    J(0, 1) = dx_jet[0].v[1];
+    J(1, 0) = dx_jet[1].v[0];
+    J(1, 1) = dx_jet[1].v[1] + 1;
+
+    // Update
+    Eigen::Vector2d step_x = J.partialPivLu().solve(x + dx - x0);
+    const double radius = std::max(x.norm() * kRelStepRadius, kStepRadius);
+    if (step_x.norm() > radius) {
+      step_x *= (radius / step_x.norm());
+    }
```
ahojnnes, line 697:
```suggestion
    const double radius_sqr = std::max(x.squaredNorm() * kRelStepRadius * kRelStepRadius, kStepRadius * kStepRadius);
    const double step_x_norm_sqr = step_x.squaredNorm();
    if (step_x_norm_sqr > radius_sqr) {
      step_x *= std::sqrt(radius_sqr / step_x_norm_sqr);
    }
```

src/colmap/sensor/models.h
```diff
@@ -127,8 +127,12 @@ enum class CameraModelId {
       double focal_length, size_t width, size_t height);                      \
   template <typename T>                                                       \
   static void ImgFromCam(const T* params, T u, T v, T w, T* x, T* y);         \
-  template <typename T>                                                       \
-  static void CamFromImg(const T* params, T x, T y, T* u, T* v, T* w);        \
+  static inline void CamFromImg(const double* params,                         \
```
ahojnnes, line 130: Rather than inlining these, would it make sense to move the implementation to the .cc file now? Same goes for the IterativeUndistortion.

ahojnnes (commented):
Thank you, this is a good improvement. I assume this now converges in fewer iterations than before and the autodiff Jacobian is faster to evaluate than numeric differentiation?

For the remaining failure cases, I am wondering whether we can benefit from @vlarsson's approach here: https://github.com/PoseLib/PoseLib/blob/master/PoseLib/misc/colmap_models.cc#L643

src/colmap/sensor/models.h
```diff
@@ -647,41 +652,49 @@ T BaseCameraModel<CameraModel>::CamFromImgThreshold(const T* params,
...
-    J(0, 1) = (dx_1f(0) - dx_1b(0)) / (2 * step1);
-    J(1, 0) = (dx_0f(1) - dx_0b(1)) / (2 * step0);
-    J(1, 1) = 1 + (dx_1f(1) - dx_1b(1)) / (2 * step1);
-    const Eigen::Vector2d step_x = J.partialPivLu().solve(x + dx - x0);
+    // Get Jacobian
+    ceres::Jet<double, 2> x_jet[2];
+    x_jet[0] = ceres::Jet<double, 2>(x(0));
+    x_jet[1] = ceres::Jet<double, 2>(x(1));
+    x_jet[0].v[0] = 1.0;
+    x_jet[1].v[1] = 1.0;
```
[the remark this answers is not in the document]
B1ueber2y: This wont work. We will need to do:
```
x_jet[0] = ceres::Jet<double, 2>(x(0), std::array<double, 2>({1.0, 0.0}))
x_jet[1] = ceres::Jet<double, 2>(x(1), std::array<double, 2>({0.0, 0.1}))
```
I am not sure if this gives us better clarify. I would prefer keeping the original one.

src/colmap/sensor/models.h
B1ueber2y: May I ask what s the benefit of moving ``CamFromImg`` to cc? This will require moving the implementation of ``CamFromImg`` for all classes to cc. Does it make the code structure cleaner? If so I could do it. 

I can see a clear benefit of moving ``IterativeUndistortion`` to cc for faster development, but we unfortunately can not do it in an elegant way as it is a method of a template class with CameraModel as its template.

B1ueber2y:
> Thank you, this is a good improvement. I assume this now converges in fewer iterations than before and the autodiff Jacobian is faster to evaluate than numeric differentiation?
> 
> For the remaining failure cases, I am wondering whether we can benefit from @vlarsson's approach here: https://github.com/PoseLib/PoseLib/blob/master/PoseLib/misc/colmap_models.cc#L643

Thanks. I think the one from PoseLib (optimization on the radius first) only works with radial distortion, but does not work this tangential and ThinPrism. We could definitely try the idea of multiple initialization, but not sure if it can resolve the convergence issue in our case.

src/colmap/sensor/models.h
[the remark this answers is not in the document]
ahojnnes: Sorry, I meant to suggest this constructor: https://github.com/ceres-solver/ceres-solver/blob/master/include/ceres/jet.h#L229

src/colmap/sensor/models.h
ahojnnes: I think you already mentioned one reason: fewer recompilations upon code changes in the CamFromImg() methods. Otherwise, fewer inlined code thus faster compilation and possibly smaller overall binary sizes. Unless we have some profiling that shows us that we really want to inline the method on all the callsites, I suggest to move it to .cc. I can do this in a follow-up PR though, as it is not directly related to the changes here.

src/colmap/sensor/models.h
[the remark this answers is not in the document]
ahojnnes:
```
    x_jet[0] = ceres::Jet<double, 2>(x(0), 0);
    x_jet[1] = ceres::Jet<double, 2>(x(1), 1);
```

ahojnnes (approved):

src/colmap/sensor/models.h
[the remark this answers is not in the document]
B1ueber2y: Thanks. This is done.

src/colmap/sensor/models.h
B1ueber2y: I think ``CamFromImg`` will very rarely be touched upon in its current state. But I agree that it can benefit performance. I will leave it for the follow-up PR. Thanks : )

ahojnnes: Thanks, I'll follow up.

PR #2860, opened 2024-10-28
Improve the conditioning of covariance estimation

B1ueber2y:
This is a very crucial fix. (LD) can be very ill-conditioned on the diagonal, which results in nan at matrix inversion. So we should do inversion only on the L matrix.

### DIFF
--- a/src/colmap/estimators/covariance.cc
+++ b/src/colmap/estimators/covariance.cc
@@ -563,7 +563,7 @@ bool BundleAdjustmentCovarianceEstimator::FactorizeFull() {
   Eigen::SimplicialLDLT<Eigen::SparseMatrix<double>> ldltOfS(S_matrix_);
   int rank = 0;
   for (int i = 0; i < S_matrix_.rows(); ++i) {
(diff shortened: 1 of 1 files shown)


src/colmap/estimators/covariance.cc
```diff
@@ -669,16 +670,15 @@ bool BundleAdjustmentCovarianceEstimator::Factorize() {
...
-         ++k) {
-      L_poses.valuePtr()[k] *=
-          sqrt(std::max(ldltOfS_poses.vectorD().coeff(i), 0.));
-    }
-  }
   Eigen::MatrixXd L_poses_dense = L_poses;
   L_matrix_poses_inv_ = L_poses_dense.triangularView<Eigen::Lower>().solve(
       Eigen::MatrixXd::Identity(L_poses_dense.rows(), L_poses_dense.cols()));
+  for (int i = 0; i < S_poses.rows(); ++i) {
+    double sqrt_invd =
```
ahojnnes, line 677: Same here.

ahojnnes (approved):

PR #2862, opened 2024-10-29
Handle unsuccessful image point undistortion

ahojnnes:
When undistortion fails to converge (e.g., see https://github.com/colmap/colmap/issues/2802), we can detect it and then more robustly handle these failures on the callsite. For example, when undistorting images, we now consider such pixels as background pixels instead of arbitrarily warping them. The handling is not super clean in all instances, as implemented in this PR.

### DIFF
--- a/src/colmap/image/undistortion.cc
+++ b/src/colmap/image/undistortion.cc
@@ -851,17 +851,23 @@ Camera UndistortCamera(const UndistortCameraOptions& options,
 
     for (size_t y = roi_min_y; y < roi_max_y; ++y) {
       // Left border.
-      const Eigen::Vector2d point1_in_cam =
+      const std::optional<Eigen::Vector2d> point1_in_cam =
           camera.CamFromImg(Eigen::Vector2d(0.5, y + 0.5));
+      if (!point1_in_cam) {
+        continue;
+      }
       const Eigen::Vector2d undistorted_point1 =
-          undistorted_camera.ImgFromCam(point1_in_cam);
+          undistorted_camera.ImgFromCam(*point1_in_cam);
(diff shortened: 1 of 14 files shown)


src/colmap/image/undistortion.cc
```diff
@@ -851,17 +851,23 @@ Camera UndistortCamera(const UndistortCameraOptions& options,
 
     for (size_t y = roi_min_y; y < roi_max_y; ++y) {
       // Left border.
-      const Eigen::Vector2d point1_in_cam =
+      const std::optional<Eigen::Vector2d> point1_in_cam =
```
ahojnnes, line 854: This logic will likely fail for extreme fisheye lenses, where outer pixels fail to undistort. In this case, we may have to iterate over the whole image instead.

src/colmap/sensor/models.h
```diff
@@ -1091,14 +1097,19 @@ void OpenCVFisheyeCameraModel::ImgFromCam(
   ImgFromFisheye(params, uu + duu, vv + dvv, x, y);
 }
 
-void OpenCVFisheyeCameraModel::CamFromImg(
+bool OpenCVFisheyeCameraModel::CamFromImg(
     const double* params, double x, double y, double* u, double* v, double* w) {
-  // Lift points to normalized plane
   double uu, vv;
   FisheyeFromImg(params, x, y, &uu, &vv);
```
ahojnnes, line 1103: We could now also handle invalid pixels in the fisheye conversion, but then we should do the equivalent for ImgFromCam... At this point, the induced callsite changes become quite large. Not sure if worth it for a few exotic cameras.

PR #2864, opened 2024-10-29
Attempt ceres tiny solver for undistortion

ahojnnes:
Another attempt at fixing https://github.com/colmap/colmap/issues/2802 - Doesn't work. Sharing this PR just for reference.


ahojnnes:
Abandoning this PR, because tinysolver is both slower and less accurate/robust as compared to our handcrafted newton solver.

PR #2867, opened 2024-10-30
Various improvements for cost functors

ahojnnes:
* Avoids unnecessary copy of covariance matrix when computing sqrt information matrix.
* More consistent notation and naming for cost functors and its variables.
* RelativePosePriorCostFunctor takes prior pose and covariance in consistent coordinate frames.


src/colmap/estimators/cost_functions.h
```diff
@@ -349,182 +349,192 @@ inline void EigenQuaternionToAngleAxis(const T* eigen_quaternion,
...
+//    Rotation term: ΔR = log(j_R_w·i_R_w⁻¹·i_R_j)
+//    Translation term: Δt = j_t_w + j_R_w·i_R_w⁻¹·(i_t_j -i_t_w)
+struct RelativePosePriorCostFunctor {
  public:
-  Point3dAlignmentCostFunctor(const Eigen::Vector3d& ref_point,
-                              const Eigen::Matrix3d& covariance_point)
-      : ref_point_(ref_point),
-        sqrt_information_point_(SqrtInformation(covariance_point)) {}
+  RelativePosePriorCostFunctor(const Rigid3d& i_from_j_prior,
+                               const Eigen::Matrix6d& i_from_j_prior_cov)
```
ahojnnes, line 453: This was previously defined in an odd manner (if I read it correctly). If we supply a i_from_j_prior transformation, then one would typically obtain the covariance in the same direction. I changed this to now take both the transformation and the covariance in a consistent direction. @sarlinpe @B1ueber2y Any opinions?

B1ueber2y: Agreed from my side. This was for LaMAR (I never used this cost) so lets double check with @sarlinpe

src/colmap/estimators/cost_functions.h
```diff
@@ -349,182 +349,192 @@ inline void EigenQuaternionToAngleAxis(const T* eigen_quaternion,
...
+// in the reference frame of the camera. Its first and last three components
+// correspond to the rotation and translation errors, respectively.
+struct AbsolutePosePriorCostFunctor {
  public:
-  AbsolutePoseErrorCostFunctor(const Rigid3d& cam_from_world,
-                               const EigenMatrix6d& covariance_cam)
-      : world_from_cam_(Inverse(cam_from_world)),
-        sqrt_information_cam_(SqrtInformation(covariance_cam)) {}
+  AbsolutePosePriorCostFunctor(const Rigid3d& cam_from_world_prior,
+                               const Eigen::Matrix6d& cam_from_world_prior_cov)
```
sarlinpe, line 359: `cam_from_world_prior_cov` does not make it clear in which reference frame the covariance is defined, how about `cam_prior_cov` instead?

src/colmap/estimators/cost_functions.h
```diff
@@ -349,182 +349,192 @@ inline void EigenQuaternionToAngleAxis(const T* eigen_quaternion,
...
-//
-// Derivation:
-// j_T_w = ΔT_j·j_T_i·i_T_w
-// where ΔT_j = exp(η_j) is the residual in SE(3) and η_j in tangent space.
-// Thus η_j = log(j_T_w·i_T_w⁻¹·i_T_j)
-// Rotation term: ΔR = log(j_R_w·i_R_w⁻¹·i_R_j)
-// Translation term: Δt = j_t_w + j_R_w·i_R_w⁻¹·(i_t_j -i_t_w)
-struct MetricRelativePoseErrorCostFunctor {
+// 3-DoF error on the camera position in the world coordinate frame.
+struct AbsolutePosePositionPriorCostFunctor {
```
sarlinpe, line 402: The mixed-up diffs make it very difficult to check for changes, would you mind splitting this into 2 PRs for reordering the classes vs renaming all variables?

src/colmap/estimators/cost_functions.h
sarlinpe: Depending on how the covariance is obtained (left or right multiplication, which is for example different between Ceres and GTSAM), it might be defined in the source or target frame. I however agree that the previous definition (source frame) isn't consistent with the absolute cost functor (target frame), so I'm ok with this change.

src/colmap/estimators/cost_functions.h
```diff
@@ -349,182 +349,192 @@ inline void EigenQuaternionToAngleAxis(const T* eigen_quaternion,
...
+//    Rotation term: ΔR = log(j_R_w·i_R_w⁻¹·i_R_j)
+//    Translation term: Δt = j_t_w + j_R_w·i_R_w⁻¹·(i_t_j -i_t_w)
+struct RelativePosePriorCostFunctor {
  public:
-  Point3dAlignmentCostFunctor(const Eigen::Vector3d& ref_point,
-                              const Eigen::Matrix3d& covariance_point)
-      : ref_point_(ref_point),
-        sqrt_information_point_(SqrtInformation(covariance_point)) {}
+  RelativePosePriorCostFunctor(const Rigid3d& i_from_j_prior,
+                               const Eigen::Matrix6d& i_from_j_prior_cov)
```
sarlinpe, line 453: The name `i_from_j_prior_cov` does not make it clear in which reference frame the covariance is defined, how about `i_prior_cov` instead?

src/colmap/estimators/cost_functions.h
ahojnnes: How about cam_cov_from_world_prior ?

ahojnnes: Equivalently i_cov_from_j_prior in the other instance.

sarlinpe: Sure, this works too.

sarlinpe (approved):

PR #2869, opened 2024-10-31
Create requirements.txt for pycolmap examples to document ceres-pyceres version match

michael-hoss:
Hi, and thanks for your nice repo!

I just installed `pyceres==2.3` (on colmap 3.10), ran the `custom_bundle_adjustment` example, and got the issue that `BundleAdjustmentOptions.solver_options` could not be bound correctly. After aligning versions of ceres and pyceres to both 2.0.0, it worked fine. 

Maybe this is obvious to most people, but the additional documentation I am proposing might save some users some debugging time. If there is a nice way to programmatically enforce matching versions of ceres and pyceres without prescribing the actual version, that would be even better.

### DIFF
--- a/pycolmap/examples/requirements.txt
+++ b/pycolmap/examples/requirements.txt
@@ -1,0 +1,4 @@
+# Note that pyceres must have the same version as the libceres-dev apt package for the 
+# C++ build from source, otherwise there may be pybind issues for the used ceres types.
+pyceres
+enlighten
--- a/pycolmap/examples/requirements.txt
+++ b/pycolmap/examples/requirements.txt
@@ -1,4 +1,4 @@
-# Note that pyceres must have the same version as the libceres-dev apt package for the 
+# Note that pyceres must have the same version as the libceres-dev apt package of the 
 # C++ build from source, otherwise there may be pybind issues for the used ceres types.
 pyceres
 enlighten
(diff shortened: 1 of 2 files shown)


B1ueber2y:
Hey @michael-hoss, What pycolmap version are you using? Did you build from source or from pip? 

pyceres version does not necessary align with the ceres version. And the pyceres head should work with any ceres version that is >= 2.1.0. For 2.0.0 there might be some issues with backward compatibility for pyceres, as many interfaces have been changed.

B1ueber2y:
> Thank you for your contribution. I am wondering whether this is something that we would rather encode in the pyproject.toml by making specific pyceres versions a hard dependency of pycolmap? @sarlinpe @B1ueber2y ?

This wont work since the current head only works with installing pycolmap and pyceres both from source or both from pip wheels.  IIRC otherwise the cost functions and manifolds cannot linked correctly with ceres base class.

michael-hoss:
> Hey @michael-hoss, What pycolmap version are you using? Did you build from source or from pip?

Thanks for your quick reposes! And good to know that from ceres >= 2.1.0 onwards, the binding of types would work with other pyceres versions, too. I guess my PR is then obsolete, and the issue can be solved in a different way.

I used 
- pycolmap version 3.10, built from source
- pyceres 2.0.0, installed through pip
- ceres 2.0.0, installed through apt before

> This wont work since the current head only works with installing pycolmap and pyceres both from source or both from pip wheels. IIRC otherwise the cost functions and manifolds cannot linked correctly with ceres base class.

@B1ueber2y  Just to be sure, with my combined way of installing pyceres through pip and pycolmap from source, I should still be fine on colmap 3.10 right?

PR #2877, opened 2024-11-02
Add new and missing options to automatic reconstructor

ahojnnes:
* Adds missing single_camera_per_folder option.
* Allows optional feature extraction/matching.

### DIFF
--- a/src/colmap/controllers/automatic_reconstruction.cc
+++ b/src/colmap/controllers/automatic_reconstruction.cc
@@ -153,13 +153,17 @@ void AutomaticReconstructionController::Run() {
     return;
   }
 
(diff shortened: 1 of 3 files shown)


B1ueber2y (approved):
Thanks. Is there any use case for running automatic reconstruction but not requiring feature extraction or matching though?

ahojnnes:
Yes, my use case is the reconstruction benchmark for which I am using the automatic reconstructor, because it conveniently allows to automatically switch between parameter sets depending on the data type (individual, internet, sequential) and the desired quality level (low, medium, high).

sarlinpe (approved):

PR #2878, opened 2024-11-02
Shared auto diff cost function creation

ahojnnes:


B1ueber2y (approved):
Thanks. This looks really nice. The idea to remove inheritance is brilliant. LGTM from my side : )

sarlinpe (approved):
Great improvement, thank you!

PR #2880, opened 2024-11-03
Add covariance weighted cost functor

ahojnnes:


src/colmap/estimators/cost_functions.h
```diff
@@ -468,61 +457,67 @@ struct Point3DAlignmentCostFunctor
...
-      if (jacobians && *jacobians) {
-        **jacobians = s_;
-      }
-      return true;
-    }
+  // Covariance or sqrt information matrix type.
+  using CovMat = Eigen::Matrix<double, kNumResiduals, kNumResiduals>;
+
+  template <typename... Args>
+  explicit CovarianceWeightedCostFunctor(const CovMat& left_cov, Args&&... args)
```
B1ueber2y, line 490: I dont think ``left_cov`` is the right name. Rather it s ``left_sqrt_info`` if you want to force the "left" to be in the name.

B1ueber2y:
Thanks. Do we still want to support ``CovarianceType`` templating to retain the performance of reprojection error?

src/pycolmap/estimators/cost_functions.cc
```diff
@@ -132,28 +137,51 @@ void BindCostFunctions(py::module& m_parent) {
...
-                                                      const Eigen::Matrix3d&>,
+        &AbsolutePosePositionPriorCostFunctor::Create<const Eigen::Vector3d&>,
         "position_in_world_prior"_a,
-        "position_cov_in_world_prior"_a,
         "3-DoF error on the absolute camera pose's position.");
+  m.def(
+      "AbsolutePosePositionPriorCost",
+      &CovarianceWeightedCostFunctor<
+          AbsolutePosePositionPriorCostFunctor>::Create<const Eigen::Vector3d&>,
+      "position_cov_in_world_prior"_a,
```
B1ueber2y, line 158: The ordering changes in the following interfaces are not ideal (in terms of not only compatibility but also that we cannot support identity covariance as default), but I guess we dont have a good way to get around for now.

src/colmap/estimators/cost_functions.h
ahojnnes: Right, I fixed this. Thanks.

src/pycolmap/estimators/cost_functions.cc
ahojnnes: I think this is fine, because the standard cost function without the covariance argument is available. In this case, one should not use the CovarianceWeightedCostFunctor because it has some computational overhead.

ahojnnes:
> Thanks. Do we still want to support `CovarianceType` templating to retain the performance of reprojection error?

I think we discussed that the overhead is minimal? We can easily bring it back, if needed. The initial concern was that this would affect the performance of the unweighted cost functions but this is not the case anymore now.

B1ueber2y (approved):

src/colmap/estimators/cost_functions.h
```diff
@@ -468,61 +459,67 @@ struct Point3DAlignmentCostFunctor
...
-// A cost function that wraps another one and whiten its residuals with an
-// isotropic covariance, i.e. assuming that the variance is identical in and
-// independent between each dimension of the residual.
+template <typename... Args>
+auto LastValueParameterPack(Args&&... args) {
+  return std::get<sizeof...(Args) - 1>(std::forward_as_tuple(args...));
+}
+
+// A cost function that wraps another one and whitens its residuals with a given
+// left covariance. For example, to weight the reprojection error with a
```
B1ueber2y, line 475: left covariance -> covariance

ahojnnes: Good catch. Thanks.

src/colmap/estimators/cost_functions.h
```diff
@@ -113,8 +123,8 @@ class ReprojErrorConstantPoseCostFunctor
           3,
           CameraModel::num_params> {
  public:
-  ReprojErrorConstantPoseCostFunctor(const Rigid3d& cam_from_world,
-                                     const Eigen::Vector2d& point2D)
+  ReprojErrorConstantPoseCostFunctor(const Eigen::Vector2d& point2D,
+                                     const Rigid3d& cam_from_world)
```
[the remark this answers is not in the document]
ahojnnes: Without this change, the point2D_cov and point2D parameters are separated by the cam_from_world in the weighted covariance wrapper.

PR #2881, opened 2024-11-03
Skip all but latest Python version in PR builds

ahojnnes:


sarlinpe (approved):

.github/workflows/build-pycolmap.yml
```diff
@@ -41,11 +41,11 @@ jobs:
       - name: Select Python
         if: github.event_name == 'pull_request'
         run: |
-          echo 'CIBW_BUILD="cp312-{macosx,manylinux,win}*"' >> "$GITHUB_ENV"
+          echo 'CIBW_BUILD=cp312-{macosx,manylinux,win}*' >> "$GITHUB_ENV"
       - name: Select Python
         if: github.event_name != 'pull_request'
         run: |
-          echo 'CIBW_BUILD="cp3{8,9,10,11,12}-{macosx,manylinux,win}*"' >> "$GITHUB_ENV"
+          echo 'CIBW_BUILD=cp3{8,9,10,11,12}-{macosx,manylinux,win}*' >> "$GITHUB_ENV"
```
sarlinpe, line 48: This second CI step isn't necessary since the default is already given here:
https://github.com/colmap/colmap/blob/d6fe93871845cb7cd55335d4ed4fc79f325c648b/pycolmap/pyproject.toml#L40

ahojnnes: It doesn't seem to work by defining both the build in the env and to toml. The environment variable doesn't override the toml definition. You can see it in the previous successful CI run.

sarlinpe: Commit 7733ee2 did work for Linux and macOS but not for Windows - maybe because it should be powershell instead of bash, so `echo "CIBW_BUILD=..." >> "${env:GITHUB_ENV}"` instead?

ahojnnes: You are right. I only looked at the Windows build to check the runtime and saw that it didn't work and assumed the same for the other platforms.

sarlinpe: This seems to have worked!

ahojnnes: Finally, this was a hard nut to crack... :-)

PR #2882, opened 2024-11-04
Subproblem partitioning with BFS to support covariance estimation for extremely large-scale problem 

B1ueber2y:
- Abstract the logic of parameter block partitioning in ``ProblemPartitioner``
- Support covariance estimation with subproblem specified either by 1) a list of subset pose parameter blocks. 2) a list of subset image ids. If used only the subproblem is considered in covariance estimation, with all the other pose parameter blocks treated as fixed at the boundary.

This will make estimation of pose covariance for ultra large scale problem feasible with a reasonable selection of subset pose blocks.  One can do covariance estimation in sliding window manner spatially with this update.

### DIFF
--- a/src/colmap/estimators/CMakeLists.txt
+++ b/src/colmap/estimators/CMakeLists.txt
@@ -49,6 +49,7 @@ COLMAP_ADD_LIBRARY(
         generalized_relative_pose.h generalized_relative_pose.cc
         homography_matrix.h homography_matrix.cc
         pose.h pose.cc
+        problem_partitioner.h problem_partitioner.cc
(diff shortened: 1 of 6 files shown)


ahojnnes:
I have not had time to look into the details but the title sounds exciting. Independent of the algorithm itself, I would suggest to first converge on a simplified interface. I took a stab at this in this draft PR: https://github.com/colmap/colmap/pull/2788. Not sure if this is still the best approach. It was based on a discussion we had about only exposing the option to compute covariances for poses based on fixed points and vice versa.

PR #2884, opened 2024-11-04
Fix unused variable warnings under MSVC

ahojnnes:
Fixes the following warnings triggered on almost every file because of transitive includes:
```
colmap/scene/reconstruction.h(279): warning C4101: 'e': unreferenced local variable
```


sarlinpe (approved):
Nice, thanks! Shouldn't have this been caught by ClangTidy?

ahojnnes:
> Nice, thanks! Shouldn't have this been caught by ClangTidy?

There may be no check implemented for this or we have not activated it. In theory, I agree, that this is something that it should be able to detect.

sarlinpe:
It's apparently not caught by `bugprone-unused-local-non-trivial-variable`, Clang has a different warning but GCC doesn't have any: https://stackoverflow.com/questions/55977431/gcc-and-clang-not-warning-on-unused-exception-variable

PR #2886, opened 2024-11-04
[doc] Fix path to example in README.md

kielnino:
The path for the example in the pycolmap subfolder has changed. This MR incorporates this change into the corresponding README.


sarlinpe (approved):
Good catch, thank you!

Issue #1093, opened 2021-01-08
Why are there always 2 compile commands per target (most likely cmake related)


3ddiddE:
When exporting the compile_commands.json I see that there are always 2 compile commands per "target" or file.
For example:
```
{
  "directory": ".../colmap/build/src/exe",
  "command": "/usr/bin/c++ -DBOOST_ALL_NO_LIB -DCGAL_ENABLED -DCUDA_ENABLED -DEIGEN_INITIALIZE_MATRICES_BY_NAN -DGFLAGS_IS_A_DLL=0 -DOPENGL_ENABLED -DOPENMP_ENABLED -DQT_CORE_LIB -DQT_GUI_LIB -DQT_OPENGL_LIB -DQT_WIDGETS_LIB -I.../colmap/lib -I.../colmap/src -isystem /usr/include/eigen3 -isystem /usr/include/qt -isystem /usr/include/qt/QtCore -isystem /usr/lib/qt/mkspecs/linux-g++ -isystem /usr/include/qt/QtOpenGL -isystem /usr/include/qt/QtWidgets -isystem /usr/include/qt/QtGui  -Wno-maybe-uninitialized -fopenmp -Wall -std=c++14 -g -fPIE -fPIC -o CMakeFiles/colmap_exe.dir/colmap.cc.o -c .../colmap/src/exe/colmap.cc",
  "file": ".../colmap/src/exe/colmap.cc"
}
```
and
```
{
  "directory": ".../colmap/build",
  "command": "/usr/bin/c++ -DCGAL_ENABLED -DCUDA_ENABLED -DEIGEN_INITIALIZE_MATRICES_BY_NAN -DOPENGL_ENABLED -DOPENMP_ENABLED -I.../colmap/lib -I.../colmap/src -I/usr/include/eigen3 -I/usr/include/qt -I/usr/include/qt/QtCore -I/usr/lib/qt/mkspecs/linux-g++ -I/usr/include/qt/QtOpenGL -I/usr/include/qt/QtWidgets -I/usr/include/qt/QtGui  -Wno-maybe-uninitialized -fopenmp -g -fPIC -o CMakeFiles/colmap_sources.dir/src/exe/colmap.cc.o -c .../colmap/src/exe/colmap.cc",
  "file": ".../colmap/src/exe/colmap.cc"
}
```
The diff is: 
![image](https://user-images.githubusercontent.com/45269130/104001704-55bd0400-51a0-11eb-9382-1210997884c0.png)

[...]

PR #2889, opened 2024-11-04
Pr/remove all sources target

HernandoR:
1. ignore .cache path, which used by clangd for index caching.
2. comment out `COLMAP_SRC_ROOT_FOLDER` target since it's not compilable, missing linking target and would mess up static checks with vcpkg managed dependencies. relating to #1093


CMakeLists.txt
```diff
@@ -243,47 +243,47 @@ COLMAP_ADD_SOURCE_DIR(src/thirdparty/PoissonRecon THIRDPARTY_POISSON_RECON_SRCS
...
-    ${COLMAP_SRC_ROOT_FOLDER}
-    ${ALL_SRCS}
-)
-
-# Prevent the library from being compiled automatically.
-set_target_properties(
-    ${COLMAP_SRC_ROOT_FOLDER} PROPERTIES
-    EXCLUDE_FROM_ALL 1
-    EXCLUDE_FROM_DEFAULT_BUILD 1)
+# # Add all of the source files to a regular library target, as using a custom
```
ahojnnes, line 246: I think we want to retain the option to generate this target for Visual Studio / XCode users. How about introducing a new option that is off by default?

ahojnnes (approved):
Thank you, this looks good to me.

PR #2894, opened 2024-11-05
Gracefully handle missing image pose in viewer

ahojnnes:
Since we changed the behavior to throw an exception on missing image pose, this can cause a crash in the viewer. This case is now handled more gracefully.

### DIFF
--- a/src/colmap/ui/model_viewer_widget.cc
+++ b/src/colmap/ui/model_viewer_widget.cc
@@ -80,7 +80,7 @@ void BuildImageModel(const Image& image,
   // be in a partial, incorrect state. In rare circumstances, an image may be
   // registered but not yet have a pose. Instead of crashing the viewer, we
   // simply skip the visualization of these images and warn the user.
-  const std::optional<Rigid3d> cam_from_world = image.MaybeCamFromWorld();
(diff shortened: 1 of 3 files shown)


src/colmap/ui/model_viewer_widget.cc
```diff
@@ -75,6 +75,18 @@ void BuildImageModel(const Image& image,
                      const Eigen::Vector4f& frame_color,
                      std::vector<TrianglePainter::Data>* triangle_data,
                      std::vector<LinePainter::Data>* line_data) {
+  // Updating the reconstruction in the viewer (e.g., deleting an image or a
+  // point) is not thread-safe when the mapper is running, where some images may
+  // be in a partial, incorrect state. In rare circumstances, an image may be
+  // registered but not yet have a pose. Instead of crashing the viewer, we
+  // simply skip the visualization of these images and warn the user.
+  const std::optional<Rigid3d> cam_from_world = image.MaybeCamFromWorld();
```
sarlinpe, line 83: Should this be a const **ref**?

ahojnnes: Yes, thx.

sarlinpe (approved):

PR #2896, opened 2024-11-07
Modernized bundle adjustment interface

ahojnnes:
* Defines a single consistent abstract interface for bundle adjustment.
* Reuses functionality between different bundle adjusters using composition rather than inheritance.
* Extracts the pose prior alignment to the alignment module. Ideally needs tests but this remains a TODO.
* The problem setup is now performed when the bundle adjuster is created and, as such, disentangled from the solving of the problem. This renders the current awkward SetUpProblem unnecessary.
* The idea later is then to pass a bundle adjuster object to a simplified covariance estimation interface.

### DIFF
--- a/src/colmap/estimators/bundle_adjustment.cc
+++ b/src/colmap/estimators/bundle_adjustment.cc
@@ -44,36 +44,6 @@
 namespace colmap {
 
 ////////////////////////////////////////////////////////////////////////////////
-// BundleAdjustmentOptions
-////////////////////////////////////////////////////////////////////////////////
-
-ceres::LossFunction* BundleAdjustmentOptions::CreateLossFunction() const {
-  ceres::LossFunction* loss_function = nullptr;
-  switch (loss_function_type) {
-    case LossFunctionType::TRIVIAL:
-      loss_function = new ceres::TrivialLoss();
-      break;
-    case LossFunctionType::SOFT_L1:
-      loss_function = new ceres::SoftLOneLoss(loss_function_scale);
-      break;
-    case LossFunctionType::CAUCHY:
-      loss_function = new ceres::CauchyLoss(loss_function_scale);
-      break;
-  }
-  THROW_CHECK_NOTNULL(loss_function);
-  return loss_function;
-}
-
-bool BundleAdjustmentOptions::Check() const {
-  CHECK_OPTION_GE(loss_function_scale, 0);
-  CHECK_OPTION_LT(max_num_images_direct_dense_cpu_solver,
-                  max_num_images_direct_sparse_cpu_solver);
-  CHECK_OPTION_LT(max_num_images_direct_dense_gpu_solver,
-                  max_num_images_direct_sparse_gpu_solver);
-  return true;
-}
-
-////////////////////////////////////////////////////////////////////////////////
 // BundleAdjustmentConfig
 ////////////////////////////////////////////////////////////////////////////////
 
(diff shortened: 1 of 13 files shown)


B1ueber2y:
Thanks! The general idea is fantastic. Will try to find time later this week to look into details.

src/colmap/estimators/bundle_adjustment.cc
```diff
@@ -432,141 +377,33 @@ ceres::Solver::Options BundleAdjuster::SetUpSolverOptions(
...
-  const bool constant_camera = !options_.refine_focal_length &&
-                               !options_.refine_principal_point &&
-                               !options_.refine_extra_params;
-  for (const camera_t camera_id : camera_ids_) {
-    Camera& camera = reconstruction->Camera(camera_id);
-
-    if (constant_camera || config_.HasConstantCamIntrinsics(camera_id)) {
-      problem_->SetParameterBlockConstant(camera.params.data());
-      continue;
+void ParameterizeCameras(const BundleAdjustmentOptions& options,
```
B1ueber2y, line 380: This function is specific for bundle adjustment. If we make it no longer a class method, should we add another level of namespace to avoid potential collapsing in the future?

B1ueber2y: Same applies to other functions in this file.

B1ueber2y:
Thanks. Looks great! The pybind here needs updated: https://github.com/colmap/colmap/blob/main/src/pycolmap/estimators/bundle_adjustment.cc#L144-L164. Also the example runner here: https://github.com/colmap/colmap/blob/main/pycolmap/examples/custom_bundle_adjustment.py#L196-L208

src/colmap/estimators/bundle_adjustment.cc
ahojnnes: I probably don't fully understand but it is already in an anonymous namespace. Is this what you wanted to suggest?

B1ueber2y: oh i missed that. Thanks!

ahojnnes:
@B1ueber2y This should be ready for a final review. Thanks.

B1ueber2y:
> https://github.com/colmap/colmap/blob/main/pycolmap/examples/custom_bundle_adjustment.py#L196-L208

Thanks. Should we update the runner here as well: https://github.com/colmap/colmap/blob/main/pycolmap/examples/custom_bundle_adjustment.py#L196-L208? I expect this part of code to be broken with the new interface?

ahojnnes:
@B1ueber2y I fixed the custom_bundle_adjustment.py code with the new interface. The idea is that the bundle adjustment problem is configured purely through the config object instead of providing access to the internal methods. One still has full control over adding custom residuals to the existing problem. I verified that this works end-to-end together with the custom_incremental_mapper.py.

B1ueber2y:
The ``custom_bundle_adjustment.py`` script wont work if pyceres is not installed, since the solver options wont parse here: https://github.com/colmap/colmap/blob/user/jsch/ba-interface/pycolmap/examples/custom_bundle_adjustment.py#L40-L45. So I added pyceres back to avoid confusion when the user tries the code without pyceres installed.

B1ueber2y (approved):

B1ueber2y:
It might be helpful for stabilizing pycolmap interfaces if we added the example script (and potentially other python tests) to the CI pipeline.

ahojnnes:
> Also It might be helpful for stabilizing pycolmap interfaces if we added the example script (and potentially other python tests) to the CI pipeline.

Yes, agreed. On my TODO list. My idea would be to expose the colmap/scene/synthetic.h interface, so we can have equivalent end-to-end mapping tests as we have under colmap/controllers/incremental_pipeline_test.cc. Later, the synthetics also come in handy for implementing tests for other parts of the python interface.

PR #2915, opened 2024-11-17
Move geometry implementation of __repr__, __eq__ overloads to C++ side

ahojnnes:


src/colmap/geometry/gps.cc
```diff
@@ -197,4 +197,15 @@ std::vector<Eigen::Vector3d> GPSTransform::ENUToXYZ(
   return xyz;
 }
 
+std::ostream& operator<<(std::ostream& stream, const PosePrior& prior) {
+  const static Eigen::IOFormat kVecFmt(
+      Eigen::StreamPrecision, Eigen::DontAlignCols, ", ", ", ");
+  stream << "PosePrior(position=[" << prior.position.format(kVecFmt)
+         << "], position_covariance=["
+         << prior.position_covariance.format(kVecFmt)
+         << "], coordinate_system=" << static_cast<int>(prior.coordinate_system)
```
sarlinpe, line 206: The Python implementation prints the enum as string, is this not possible in C++?

ahojnnes: Needs a custom switch/case where we convert from enum to string.

B1ueber2y: I have a custom macro for enum to string in C++. Will open a PR

B1ueber2y (approved):

PR #2916, opened 2024-11-17
Initialize glog in custom gmock main function

ahojnnes:
Without the custom main function introduced in this PR, we get warnings about glog not being initialized, since the default gtest/gmock main functions do not initialize glog.

### DIFF
--- a/src/colmap/controllers/incremental_pipeline_test.cc
+++ b/src/colmap/controllers/incremental_pipeline_test.cc
@@ -203,7 +203,7 @@ TEST(IncrementalPipeline, PriorBasedSfMNoNoise) {
   Reconstruction gt_reconstruction;
   SyntheticDatasetOptions synthetic_dataset_options;
   synthetic_dataset_options.num_cameras = 2;
-  synthetic_dataset_options.num_images = 7;
+  synthetic_dataset_options.num_images = 10;
   synthetic_dataset_options.num_points3D = 100;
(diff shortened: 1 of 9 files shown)


sarlinpe:
Nice! Are the test failures related? Does it mean that tests have been silently failing on Windows?

ahojnnes:
> Nice! Are the test failures related? Does it mean that tests have been silently failing on Windows?

I am afraid that this might be the case. I don't have access to a Windows machine right now though :-)

ahojnnes:
@sarlinpe I fixed the Windows test failures. They were related in the sense that we didn't fail on test failures under Windows before the custom gtest main function.

src/colmap/controllers/incremental_pipeline_test.cc
```diff
@@ -203,7 +203,7 @@ TEST(IncrementalPipeline, PriorBasedSfMNoNoise) {
   Reconstruction gt_reconstruction;
   SyntheticDatasetOptions synthetic_dataset_options;
   synthetic_dataset_options.num_cameras = 2;
-  synthetic_dataset_options.num_images = 7;
+  synthetic_dataset_options.num_images = 10;
```
ahojnnes, line 206: We use unordered_map's in various places, which have an undefined iteration order, which differs between standard library implementations, in particular MSVC vs libc++/libstdc++. Using more images, makes the alignment in the prior mapper more stable.

sarlinpe (approved):

PR #2918, opened 2024-11-18
Custom macro for enum to string support

B1ueber2y:
I adapted only a few for now as examples. If we are happy with it we could apply on all enum classes.

### DIFF
--- a/src/colmap/util/enum_to_string.h
+++ b/src/colmap/util/enum_to_string.h
@@ -48,7 +48,7 @@ namespace colmap {
 // MAKE_ENUM_CLASS(MyEnum, 0, C1, C2, C3);
 
 #define ENUM_TO_STRING_PROCESS_ELEMENT(r, start_idx, idx, elem) \
-  case (idx + start_idx):                                       \
+  case ((idx) + (start_idx)):                                   \
     return BOOST_PP_STRINGIZE(elem);
 
 #define DEFINE_ENUM_TO_STRING(name, start_idx, ...)                   \
@@ -63,7 +63,8 @@ namespace colmap {
     }                                                                 \
   }
 
-#define ENUM_PROCESS_ELEMENT(r, start_idx, idx, elem) elem = idx + start_idx,
+#define ENUM_PROCESS_ELEMENT(r, start_idx, idx, elem) \
+  elem = (idx) + (start_idx),
 
 #define ENUM_VALUES(start_idx, ...) \
   BOOST_PP_SEQ_FOR_EACH_I(          \
--- a/src/colmap/util/enum_to_string.h
+++ b/src/colmap/util/enum_to_string.h
@@ -29,6 +29,8 @@
 
 #pragma once
 
+#include "colmap/util/logging.h"
+
 #include <iostream>
 #include <memory>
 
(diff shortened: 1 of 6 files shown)


B1ueber2y:
Also it currently only supports enum class. Need to support enum as well (should be straightforward).

B1ueber2y:
> Also it currently only supports enum class. Need to support enum as well (should be straightforward).

This is supported: https://github.com/colmap/colmap/pull/2918/commits/7e349790966525c7b0fa4eea766fb944a2aaee08

src/colmap/util/enum_to_string.h
```diff
@@ -0,0 +1,86 @@
...
+#define MAGIC_MAKE_ENUM(name, start_idx, ...)        \
+  enum name { ENUM_VALUES(start_idx, __VA_ARGS__) }; \
+  DEFINE_ENUM_TO_STRING(name, start_idx, __VA_ARGS__)
+
+#define MAGIC_MAKE_ENUM_CLASS(name, start_idx, ...)        \
+  enum class name { ENUM_VALUES(start_idx, __VA_ARGS__) }; \
+  DEFINE_ENUM_TO_STRING(name, start_idx, __VA_ARGS__)
+
+// This only works for non-nested enum classes.
+#define MAGIC_MAKE_ENUM_CLASS_OVERLOAD_STREAM(name, start_idx, ...) \
```
ahojnnes, line 80: nit: Shall we remove the "MAGIC" prefix?

src/colmap/util/enum_to_string.h
```diff
@@ -0,0 +1,86 @@
...
+// [Reference]: enum class MyEnum {C1, C2, C3};
+//
+// [New code]:
+// MAGIC_MAKE_ENUM_CLASS(MyEnum, 0, C1, C2, C3);
+
+#define ENUM_TO_STRING_PROCESS_ELEMENT(r, start_idx, idx, elem) \
+  BOOST_PP_COMMA_IF(idx) { idx + start_idx, BOOST_PP_STRINGIZE(elem) }
+
+#define DEFINE_ENUM_TO_STRING(name, start_idx, ...)                 \
+  const std::shared_ptr<std::map<int, std::string>> name##Strings = \
```
ahojnnes, line 53: This can be a static?

ahojnnes: Instead of a std::map, would it make sense to use a switch/case statement instead? It'll probably be translated into a lookup table by the compiled instead of branching.

src/colmap/util/enum_to_string.h
```diff
@@ -0,0 +1,86 @@
...
+
+#pragma once
+
+#include <iostream>
+#include <map>
+#include <memory>
+
+#include <boost/preprocessor.hpp>
+
+namespace colmap {
```
ahojnnes, line 38: It will be great to have a unit test for this new macro.

src/colmap/util/enum_to_string.h
B1ueber2y: Done.

ahojnnes (approved):
Looks great to me, thanks.

src/colmap/util/enum_to_string.h
```diff
@@ -0,0 +1,84 @@
...
+    return BOOST_PP_STRINGIZE(elem);
+
+#define DEFINE_ENUM_TO_STRING(name, start_idx, ...)                   \
+  constexpr const char* name##ToString(int value) {                   \
+    switch (value) {                                                  \
+      BOOST_PP_SEQ_FOR_EACH_I(ENUM_TO_STRING_PROCESS_ELEMENT,         \
+                              start_idx,                              \
+                              BOOST_PP_VARIADIC_TO_SEQ(__VA_ARGS__)); \
+      default:                                                        \
+        throw std::runtime_error("Invalid input value");              \
```
ahojnnes, line 59: nit: could log out the value here

sarlinpe: Use `LOG(FATAL_THROW) << message;` then

PR #2923, opened 2024-11-19
Consistent interface and various improvements for pycolmap/estimators

ahojnnes:
With a best effort of raising deprecation warnings for  renamed function names. For example:
```
>>> pycolmap.essential_matrix_estimation()
<stdin>:1: DeprecationWarning: essential_matrix_estimation() is deprecated, use estimate_essential_matrix() instead.
...
```
Other modules follow in separate PRs.

### DIFF
--- a/src/pycolmap/estimators/similarity_transform.cc
+++ b/src/pycolmap/estimators/similarity_transform.cc
@@ -16,7 +16,7 @@ using namespace pybind11::literals;
 namespace py = pybind11;
 
 void BindSimilarityTransformEstimator(py::module& m) {
-  auto est_options = m.attr("RANSACOptions")().cast<RANSACOptions>();
+  auto ransac_options = m.attr("RANSACOptions")().cast<RANSACOptions>();
 
   m.def(
       "estimate_sim3d",
@@ -33,9 +33,9 @@ void BindSimilarityTransformEstimator(py::module& m) {
           return py::none();
         }
       },
-      "points3D_src"_a,
-      "points3D_tgt"_a,
-      "Estimate the 3D similarity transform tgt_T_src.");
+      "src"_a,
+      "tgt"_a,
+      "Estimate the 3D similarity transform tgt_from_src.");
 
   m.def(
       "estimate_sim3d_robust",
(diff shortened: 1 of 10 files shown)


B1ueber2y (approved):
Looks great! Thanks for the DefDeprecation feature. Since this affects almost all the HLoc interfaces, I suggest to wait for Paul to give a feedback.

src/pycolmap/helpers.h
```diff
@@ -406,3 +407,28 @@ inline bool IsPyceresAvailable() {
...
+    std::optional<std::string> custom_warning = std::nullopt) {
+  parent.def(
+      old_name.c_str(),
+      [parent,
+       old_name = std::move(old_name),
+       new_name = std::move(new_name),
+       custom_warning = std::move(custom_warning)](const py::args& args,
+                                                   const py::kwargs& kwargs) {
+        if (custom_warning) {
+          PyErr_WarnEx(PyExc_DeprecationWarning, custom_warning->c_str(), 1);
```
sarlinpe, line 425: Does this print every single time it is called? if so, should we use `LOG_FIRST_N(WARNING, N)` instead?

src/pycolmap/estimators/similarity_transform.cc
```diff
@@ -54,8 +54,9 @@ void BindSimilarityTransformEstimator(py::module& m) {
                         "num_inliers"_a = report.support.num_inliers,
                         "inliers"_a = ToPythonMask(report.inlier_mask));
```
sarlinpe, line 55: Should this be `inlier_mask` to be consistent with the other changes?

src/pycolmap/helpers.h
ahojnnes: No, Python's deprecation warnings only get issued once upon first usage, so it is equivalent to LOG_FIRST_N() with the added benefit of working nicely together with Python's warning filtering system, etc.

src/pycolmap/estimators/similarity_transform.cc
ahojnnes: Good catch. Done.

PR #2925, opened 2024-11-20
Cleanup of remaining pycolmap interfaces

ahojnnes:
For Sift, I want to move to a consistent set of options, where we use the C++ ones as the default. We'll issue a deprecation warning:
```
>>> import pycolmap
>>> pycolmap.Sift(pycolmap.SiftExtractionOptions())
I20241120 09:56:59.974107 202499 sift.cc:727] Creating SIFT CPU feature extractor
<pycolmap._core.Sift object at 0x106511230>
>>> pycolmap.Sift()
<stdin>:1: DeprecationWarning: No SIFT extraction options specified. Setting them to peak_threshold=0.01, first_octave=0, max_image_size=7000 for backwards compatibility. If you want to keep the settings, explicitly pass them, because the defaults will change in the next major release.
I20241120 09:57:04.519587 202499 sift.cc:727] Creating SIFT CPU feature extractor
<pycolmap._core.Sift object at 0x106325630>
```

### DIFF
--- a/src/pycolmap/geometry/bindings.cc
+++ b/src/pycolmap/geometry/bindings.cc
@@ -1,4 +1,3 @@
-#include "colmap/geometry/essential_matrix.h"
 #include "colmap/geometry/gps.h"
 #include "colmap/geometry/pose.h"
 #include "colmap/geometry/rigid3.h"
@@ -19,11 +18,10 @@ using namespace colmap;
 using namespace pybind11::literals;
 namespace py = pybind11;
 
-void BindHomographyGeometry(py::module& m);
+void BindHomographyMatrixGeometry(py::module& m);
+void BindEssentialMatrixGeometry(py::module& m);
 
 void BindGeometry(py::module& m) {
-  BindHomographyGeometry(m);
-
   py::class_ext_<Eigen::Quaterniond> PyRotation3d(m, "Rotation3d");
   PyRotation3d.def(py::init([]() { return Eigen::Quaterniond::Identity(); }))
       .def(py::init<const Eigen::Vector4d&>(),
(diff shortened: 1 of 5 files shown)


src/pycolmap/geometry/bindings.cc
```diff
@@ -85,7 +83,6 @@ void BindGeometry(py::module& m) {
       .def_readwrite("translation", &Rigid3d::translation)
       .def("matrix", &Rigid3d::ToMatrix)
       .def("adjoint", &Rigid3d::Adjoint)
-      .def("essential_matrix", &EssentialMatrixFromPose)
```
sarlinpe, line 88: From a user standpoint, what is the rational for removing this method? It does not introduce annoying dependencies and it does not clutter the API. I suggest to keep this method in addition to the new binding in `geometry/essential_matrix.cc`.

ahojnnes: It does introduce a dependency from Rigid3d on essential matrix functionality and, semantically, not every Rigid3d is a meaningful essential matrix. If we are convinced about having it bound as a method to Rigid3d, we should likewise also introduce it on the C++ side? I understand this introduces a slight degradation in convenience when doing this conversion, but from a design perspective it is the cleaner and more consistent API. If you prefer, I can trigger an API deprecation warning instead?

sarlinpe: Keeping the C++ and Python API consistent indeed seems important. I don't have a strong opinion about this, we can remove the method without deprecation warning. 

Alternatively, we could move `EssentialMatrixFromPose`, `InterpolateCameraPoses` to rigid3.h and `TransformCameraWorld` to sim3.h to make the C++ & Python even more consistent.

ahojnnes: > Alternatively, we could move EssentialMatrixFromPose, InterpolateCameraPoses to rigid3.h and TransformCameraWorld to sim3.h to make the C++ & Python even more consistent.

I would keep them where they are right now. I currently like that rigid3d and sim3d are useful independent of colmap and projective geometry. One could copy it without changes into another project that needs euclidean/metric coordinate transformations.

sarlinpe (approved):

PR #2926, opened 2024-11-20
Move useful functions from scripts into pycolmap

B1ueber2y:
* Move database, IO and plyfile utilities into pycolmap.utils
* Adapt the import of other scripts with pycolmap (e.g., https://github.com/B1ueber2y/colmap/blob/features/move_scripts/scripts/python/migrate_database_pose_prior.py#L35, https://github.com/B1ueber2y/colmap/blob/features/move_scripts/scripts/python/merge_ply_files.py#L38)
* Fit linting issues for those moved scripts

``database.py`` and ``read_write_model.py`` are particularly useful scripts that are used a lot externally (e.g., HLoc, limap, etc. (436 search results) https://github.com/search?q=read_write_model.py&type=code). This could bring significant benefits and improve the exposure of pycolmap.


B1ueber2y:
> Thanks for this effort. I am actually thinking that we want to remove the legacy database.db, as all functionality and more is available through the pybind interfaces now? Similarly, for read_write_model.py?

Given the extensive usage of the two files, I would vote for keeping it in pycolmap such that the other projects could make an easy switch.

ahojnnes:
> > Thanks for this effort. I am actually thinking that we want to remove the legacy database.db, as all functionality and more is available through the pybind interfaces now? Similarly, for read_write_model.py?
> 
> Given the extensive usage of the two files, I would vote for keeping it in pycolmap such that the other projects could make an easy switch.

Fair enough. Would it make sense to trigger a deprecation warning upon import/usage of these modules?

B1ueber2y:
Closed as we aim to improve documentation rather than bring those non-maintained python interfaces back

PR #2931, opened 2024-11-22
fix LORANSAC::Estimate compilation error when build with visual studio C++ 17

yaozhou:
when build latest master branch of GLOMAP on windows with visual studio 2017, I found this compilation error, and found that colmap is using C++14 , but glomap is using C++17, which cause this problem.

### DIFF
--- a/src/colmap/optim/loransac.h
+++ b/src/colmap/optim/loransac.h
@@ -89,7 +89,7 @@ template <typename Estimator,
           typename LocalEstimator,
           typename SupportMeasurer,
           typename Sampler>
-typename LORANSAC<Estimator, LocalEstimator, SupportMeasurer, Sampler>::Report
+typename RANSAC<Estimator, SupportMeasurer, Sampler>::Report
 LORANSAC<Estimator, LocalEstimator, SupportMeasurer, Sampler>::Estimate(
     const std::vector<typename Estimator::X_t>& X,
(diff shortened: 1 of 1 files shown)


ahojnnes:
Thank for the fix. I am honestly surprised that colmap still compiles under VS 2017. We only test up to 2019 in CI. I am fine with this change, as long as it doesn't break any other compiler. Thanks.

yaozhou:
I thought it's not a problem of the version of the Visual Studio, it's a problem of C++ standard, the current code can compile with C++14, but would fail with C++17, no matter what version of Visual Studio you are using . 
Since the Glomap is using C++17 standard, so , cause this problem. I think C++17 is doing  more strict checking when using template.

ahojnnes:
colmap is also using C++17 (C++14 only for CUDA code).

yaozhou:
I also tried using VS2022, If I choose "ISO C++14 standard (/std:c++14)" in project options(C++/Language/C++ standard), It would be fine, but If I switch "ISO C++17 standard (/std:c++17)", It will cause the same compilation error,  But I haven't installed VS2019, not sure what would it be under vs2019.

ahojnnes:
Thank you. I am happy to merge the proposed change, but clang-tidy static analysis complains with the following error:
```
error: return type of out-of-line definition of 'colmap::LORANSAC::Estimate' differs from that in the declaration [clang-diagnostic-error]
```

PR #2942, opened 2024-11-25
Fix CI cache thrashing by inconsistent vcpkg binary caching

ahojnnes:
We are consistently overflowing the available actions cache (>30GB out of available 10GB). vcpkg natively supports reading/writing from the Github actions cache on a per package basis. We already used this mechanism in the native Windows build. The pycolmap builds still relied on the less efficient filesystem based caching.


sarlinpe (approved):
Nice, thanks!

PR #2946, opened 2024-11-26
Add test for Python mapping pipeline

ahojnnes:
Not yet integrated with CI (to be done later), but enables for quick verification of working pycolmap bindings and incremental mapping pipeline. Currently, the test_multi_reconstruction() test case fails and is skipped.


B1ueber2y (approved):
Thanks. This is very nice!

PR #2962, opened 2024-11-29
Use poselib's minimal solver for generalized relative pose

ahojnnes:
The poselib solver is much more stable and also only needs 6 vs. 8 points in the minimal case to work well in practice.

### DIFF
--- a/src/colmap/estimators/generalized_relative_pose.cc
+++ b/src/colmap/estimators/generalized_relative_pose.cc
@@ -37,8 +37,64 @@
 #include "colmap/util/logging.h"
 
 #include <Eigen/Dense>
+#include <PoseLib/solvers/gen_relpose_6pt.h>
 
 namespace colmap {
+
+void GR6PEstimator::Estimate(const std::vector<X_t>& points1,
+                             const std::vector<Y_t>& points2,
+                             std::vector<M_t>* rigs2_from_rigs1) {
+  THROW_CHECK_EQ(points1.size(), 6);
+  THROW_CHECK_EQ(points2.size(), 6);
+  THROW_CHECK_NOTNULL(rigs2_from_rigs1);
+
+  rigs2_from_rigs1->clear();
+
+  std::vector<Eigen::Vector3d> proj_centers1(6);
+  std::vector<Eigen::Vector3d> proj_centers2(6);
+  std::vector<Eigen::Vector3d> rays1(6);
+  std::vector<Eigen::Vector3d> rays2(6);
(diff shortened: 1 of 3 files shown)


B1ueber2y (approved):
Thanks. Looks good to me.

src/colmap/estimators/generalized_relative_pose.h
```diff
@@ -39,6 +39,37 @@
...
+  // The minimum number of samples needed to estimate a model. Note that in
+  // theory the minimum required number of samples is 6 but Laurent Kneip showed
+  // in his paper that using 8 samples is more stable.
+  static const int kMinNumSamples = 6;
+
+  // Estimate the most probable solution of the GR6P problem from a set of
+  // six 2D-2D point correspondences.
+  static void Estimate(const std::vector<X_t>& points1,
+                       const std::vector<Y_t>& points2,
+                       std::vector<M_t>* rigs2_from_rigs1);
```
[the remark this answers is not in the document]
ahojnnes: The naming is motivated by it being a vector. It is consistent with the naming in e.g. essential matrix estimation where we call the equivalent cams2_from_cams1. I'll keep it as is. Thanks.

src/colmap/estimators/generalized_relative_pose.cc
```diff
@@ -37,8 +37,64 @@
...
+        points2[i].cam_from_rig.rotation.inverse() * points2[i].ray_in_cam;
+  }
+
+  std::vector<poselib::CameraPose> poses;
+  poselib::gen_relpose_6pt(proj_centers1, rays1, proj_centers2, rays2, &poses);
+
+  rigs2_from_rigs1->reserve(poses.size());
+  for (const poselib::CameraPose& pose : poses) {
+    rigs2_from_rigs1->emplace_back(
+        Eigen::Quaterniond(pose.q(0), pose.q(1), pose.q(2), pose.q(3)), pose.t);
```
vlarsson, line 74: @ahojnnes, If I recall correctly the poselib quaternion and Eigen::Quaterniond is not in the same order? Maybe double check this

B1ueber2y: This reminds me though: we don't have any tests for the minimal solver being correct, and LO can oftentimes lead to correct results even with a wrongly configured minimal solver, if the convergence is good enough. Should we add some tests in the future for the Estimate() function. This can catch such inconsistent convention issues and also be potentially beneficial for PoseLib, as no such tests are available in PoseLib either.

PR #2971, opened 2024-12-02
Fix find_package logic for poselib in support of homebrew/vcpkg

chenrui333:
👋 add build patch to build against system poselib.

relates to https://github.com/Homebrew/homebrew-core/pull/199371/


src/colmap/estimators/CMakeLists.txt
```diff
@@ -27,6 +27,9 @@
 # ARISING IN ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE
 # POSSIBILITY OF SUCH DAMAGE.
 
+if(NOT FETCH_POSELIB)
+    find_package(PoseLib REQUIRED)
```
ahojnnes, line 31: This is already done in src/thirdparty/CMakeLists.txt. I guess the reason this doesn't work is because of the order in which we add sub-directories. The add_subdirectory() calls in the main CMakeLists.txt in L221 need to be swapped?

ahojnnes:
Thank for you updating the formulas in homebrew, this is great.

src/colmap/estimators/CMakeLists.txt
```diff
@@ -66,6 +69,7 @@ COLMAP_ADD_LIBRARY(
         colmap_optim
         Eigen3::Eigen
         Ceres::ceres
+        PoseLib::PoseLib
```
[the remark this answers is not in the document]
ahojnnes: OK, I got the answer myself and was confused with how other build systems handle this... CMake indeed only makes a dependency part of the link interface with the PUBLIC keyword. PRIVATE dependencies are only linked to but not added to the interface and thus not transitively propagated. See: https://cmake.org/cmake/help/latest/command/target_link_libraries.html#libraries-for-a-target-and-or-its-dependents

src/colmap/estimators/CMakeLists.txt
ahojnnes: I pushed some fixes to this branch. Please let me know if this doesn't work for you.

ahojnnes (approved):

chenrui333:
Thanks @ahojnnes for the help in here. 🙏

ahojnnes:
Thanks for updating the formula on homebrew! @chenrui333

PR #2975, opened 2024-12-02
ALIKED/LightGlue libtorch prototype

ahojnnes:
Seems to do the expected:
```
➜  build git:(main) ✗ src/colmap/util/colmap_util_torch_test
[==========] Running 1 test from 1 test suite.
[----------] Global test environment set-up.
[----------] 1 test from Torch
[ RUN      ] Torch.HelloTensor
 1  0  0
 0  1  0
 0  0  1
[ CPUFloatType{3,3} ]
[       OK ] Torch.HelloTensor (7 ms)
[----------] 1 test from Torch (7 ms total)

[----------] Global test environment tear-down
[==========] 1 test from 1 test suite ran. (7 ms total)
[  PASSED  ] 1 test.
```


src/thirdparty/LightGlue/matcher.cpp
```diff
@@ -0,0 +1,603 @@
...
+            output.insert("prune1", prune1);
+
+            return output;
+        }
+
+        // Remove padding and compute assignment
+        desc0 = desc0.index({torch::indexing::Slice(), torch::indexing::Slice(0, m), torch::indexing::Slice()});
+        desc1 = desc1.index({torch::indexing::Slice(), torch::indexing::Slice(0, n), torch::indexing::Slice()});
+
+        auto scores = log_assignment_[std::min(i, config_.n_layers - 1)]->forward(desc0, desc1);
```
ahojnnes, line 430: @MrNeRF I believe here is a bug. I had to add the std::min() here to prevent segfault. Will be great for you to confirm that the fix is correct. Thx.

src/thirdparty/ALIKED/deform_conv2d.cpp
```diff
@@ -0,0 +1,1279 @@
...
+ * \author Yuwen Xiong, Haozhi Qi, Jifeng Dai, Xizhou Zhu, Han Hu, Dazhi Cheng
+ */
+
+// modified from
+// https://github.com/chengdazhi/Deformable-Convolution-V2-PyTorch/blob/mmdetection/mmdet/ops/dcn/src/deform_conv_cuda_kernel.cu
+
+// modified from
+// https://github.com/open-mmlab/mmdetection/blob/master/mmdet/ops/dcn/src/deform_conv_cuda.cpp
+
+#include "deform_conv2d.h"
```
ahojnnes, line 69: @MrNeRF I had to copy over the cpu version of the get_patches implementation. Otherwise, there would be missing symbols.

src/thirdparty/ALIKED/get_patches.cpp
```diff
@@ -0,0 +1,150 @@
...
+      F::PadFuncOptions(
+          {pad_left_top, pad_right_bottom, pad_left_top, pad_right_bottom})
+          .mode(torch::kConstant);
+  auto map_pad = F::pad(map.unsqueeze(0), options)
+                     .squeeze(0);  // Cx(H+2*radius)x(W+2*radius)
+
+  // get patches
+  torch::Tensor patches =
+      torch::zeros({N, C, kernel_size, kernel_size}, map.options());
+  auto a_points = points.accessor<int, 2>();     // Nx2
```
ahojnnes, line 38: In a bunch of places, your original code here had `.accessor<long, 2>()` and I had to change it to `int` to avoid missing symbol errors.

src/thirdparty/LightGlue/matcher.cpp
MrNeRF: Indeed, great catch! That's a stupid bug. The i variable belongs into the loop and should not be defined and declared outside. This way you can define a variable that gets i assigned the correct value within the for-loop which will be always std::min(i, config_.n_layers - 1).

src/thirdparty/ALIKED/deform_conv2d.cpp
MrNeRF: Yeah, I see. I didn't test with cpu.

src/thirdparty/ALIKED/get_patches.cpp
MrNeRF: Interesting. I didn't check the code as it is taken from the original implementation https://github.com/Shiaoming/ALIKED/blob/main/custom_ops/get_patches.cpp#L33.

```
 torch::Tensor get_patches_forward(const torch::Tensor& map, torch::Tensor& points, int64_t kernel_size) {
        if (map.device() == torch::kCPU)
            return get_patches_forward_cpu(map, points, kernel_size);
        else
        {
            return get_patches_forward_cuda(map, points, kernel_size);
        }
    }
```
    
So, the points should have the same data type as in the cuda path. The data is accessed via points.data_ptr<int64_t>() in the cuda version. I guess accessing it as pointer to int would be a severe bug? 

I am not 100% sure why there are missing symbols? Does int64_t work?

MrNeRF: Ok, while thinking more about it, I am confused. Accessor is a safe way to access the tensor data, data_ptr is not. So if there are compile time issues with the datatype, something is really wrong. Maybe the cuda part is buggy but I didn't encounter with cuda any issues. Ok, I think I have too take a closer look into it.

src/thirdparty/ALIKED/deform_conv2d.cpp
MrNeRF: Btw, I used the ops from here: https://github.com/pytorch/vision/tree/main/torchvision/csrc/ops

src/thirdparty/ALIKED/get_patches.cpp
MrNeRF: I guess you are using the libtorch cpu verision. I only used exclusively the cuda 2.1 version which is why never ran into these issues.

ahojnnes:
@MrNeRF I am getting the following error when loading the models from your repo: `Warning: confidence_thresholds not found in model parameters or buffers`. Is this expected?

MrNeRF:
> @MrNeRF I am getting the following error when loading the models from your repo: `Warning: confidence_thresholds not found in model parameters or buffers`. Is this expected?

My understand is they are not learned parameters and are generated here: https://github.com/MrNeRF/Light_Glue_CPP/blob/master/src/matcher/lightglue/matcher.cpp#L181C1-L185C11
In other words, this can be safely ignored and the error msg can be deactivated.

Issue #2978, opened 2024-12-03
How to set pycolmap BA parameter/s


mattiasmar:
How does one set   max_num_iterations &   max_linear_solver_iterations when calling   pycolmap.bundle_adjustment?
```
options = pycolmap.BundleAdjustmentOptions()
pycolmap.bundle_adjustment(reconstruction, options)
```

As per the documentation the BundleAdjustmentOptions should hold these values, but not anymore in the latest commit.

**Context:**
```
#prior: reconstruction folder sfm_path with images.bin, cameras.bin, points3D.bin
os.makedirs(sfm_path,exist_ok=True)
rec.write(sfm_path)
  for _ in range(50):
      # Step 3: Perform triangulation using the existing camera poses
      triangulation.main(
          reference_model=sfm_path,
          sfm_dir = sfm_path,
          image_dir=Path(output_dir),
          pairs=sfm_pairs,
          features=features_path,
          matches=matches_path,
          verbose = False,
          skip_geometric_verification = True
      )
      # # Load the final reconstruction
      reconstruction = pycolmap.Reconstruction(sfm_path)

      # Define bundle adjustment options
      options = pycolmap.BundleAdjustmentOptions()
      options.refine_focal_length = False  # Do not refine focal length
      options.refine_principal_point = False  # Do not refine principal point
      options.refine_extra_params = False  # Do not refine extra parameters
      options.refine_extrinsics = True  # Refine extrinsics
      options.use_gpu = True
      pycolmap.bundle_adjustment(reconstruction, options)
      reconstruction.write(sfm_path)
```

ahojnnes:
You should be able to set these using options.solver_options.

sarlinpe:
Unfortunately the binding of Ceres::Solver::Options was removed in https://github.com/colmap/colmap/pull/2509 on prevent duplication with pyceres. Given how broken the pycolmap-pyceres compatibility is (at least for PyPI wheels on macOS), we maybe shouldn't have done this @B1ueber2y

mattiasmar:
Right... when I install pyceres I get an exception when importing pycolmap.
Any other idea on how to limit the BA? I want to do many repeated triangulation / BA, and thought that shorter iterations of BA would help me. I have initial poses that are quite OK, and I'm initializing the reconstruction from those and only finetuning the model with Colmap..
```
python3
Python 3.10.12 (main, Nov  6 2024, 20:22:13) [GCC 11.4.0] on linux
[...]

PR #2985, opened 2024-12-03
Add back some ceres bindings to use pycolmap bundle adjustment without pyceres

B1ueber2y:
Relevant discussion here: https://github.com/colmap/colmap/issues/2978

No need for ``import ceres`` in ``custom_bundle_adjustment.py`` anymore (relevant discussion: https://github.com/colmap/colmap/pull/2896)

### DIFF
--- a/pycolmap/examples/custom_bundle_adjustment.py
+++ b/pycolmap/examples/custom_bundle_adjustment.py
@@ -7,8 +7,6 @@ pyceres is needed as a dependency for this file.
 
 import copy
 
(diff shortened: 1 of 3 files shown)


B1ueber2y:
One would still need pyceres if the customized pycolmap cost functions and manifolds are needed, but this is a demand that only makes sense when the user is using pyceres to set up the problem.

ahojnnes (approved):

sarlinpe (approved):

Issue #2992, opened 2024-12-05
PosePriorBundleAdjusterOptions::ransac_max_error hasn't been used in PosePriorBundleAdjuster


StonerLing:
The maximum RANSAC error for Sim3 alignment from reconstruction to pose prior is calculated from average position covariance：
https://github.com/colmap/colmap/blob/290ec62f872260bcbbd44a31155a9f83b4be2fa1/src/colmap/estimators/bundle_adjustment.cc#L1023-L1029

But the member `ransac_max_error`  in `PosePriorBundleAdjusterOptions` isn't taken into consideration.
https://github.com/colmap/colmap/blob/290ec62f872260bcbbd44a31155a9f83b4be2fa1/src/colmap/estimators/bundle_adjustment.h#L200-L210

I think the interface options should have higher priority. It may be modified like this:
```cpp
if (num_covs > 0) { 
   ransac_options.max_error = prior_options_.ransac_max_error > 0. ? 
     prior_options_.ransac_max_error : (3. * (avg_cov / num_covs).cwiseSqrt()).norm(); 
 } 
```

ahojnnes:
Thanks for reporting this bug. Should be fixed in: https://github.com/colmap/colmap/pull/2993

PR #2993, opened 2024-12-05
Fix setting of RANSAC max error in pose prior BA alignment

ahojnnes:
Fixes: https://github.com/colmap/colmap/issues/2992

### DIFF
--- a/src/colmap/estimators/bundle_adjustment.cc
+++ b/src/colmap/estimators/bundle_adjustment.cc
@@ -1026,7 +1026,8 @@ class PosePriorBundleAdjuster : public BundleAdjuster {
         LOG(WARNING) << "No pose priors with valid covariance found.";
         return false;
       }
-      ransac_options.max_error = (3. * (avg_cov / num_covs).cwiseSqrt()).norm();
+      // Set max error at the 3 sigma confidence interval. Assumes no outliers.
+      ransac_options.max_error = 3 * std::sqrt((avg_cov / num_covs).sum());
     }
 
     Sim3d metric_from_orig;
--- a/src/colmap/estimators/bundle_adjustment.cc
+++ b/src/colmap/estimators/bundle_adjustment.cc
@@ -1018,7 +1018,7 @@ class PosePriorBundleAdjuster : public BundleAdjuster {
       size_t num_valid_covs = 0;
       for (const auto& [_, pose_prior] : pose_priors_) {
         if (pose_prior.IsCovarianceValid()) {
-          const Eigen::Vector3d max_stddev =
+          const double max_stddev =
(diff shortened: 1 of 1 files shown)


B1ueber2y (approved):

src/colmap/estimators/bundle_adjustment.cc
```diff
@@ -1011,16 +1011,21 @@ class PosePriorBundleAdjuster : public BundleAdjuster {
...
+          ++num_covs;
+        }
+      }
+      if (num_covs == 0) {
+        LOG(WARNING) << "No pose priors with valid covariance found.";
+        return false;
       }
-    }
-    if (num_covs > 0) {
       ransac_options.max_error = (3. * (avg_cov / num_covs).cwiseSqrt()).norm();
```
B1ueber2y, line 1029: Niche and not directly relevant to this PR: cWiseSqrt() + norm() should be numerically inferior to the square root of the sum?

ahojnnes: Not sure I understand. sqrt(sum(...)) would not produce the same result?

B1ueber2y: vec.cwiseSqrt().norm() and sqrt(vec.sum()) are mathematically equivalent?

ahojnnes: Actually, looking at this more now, I would have expected the multiplier 3 to be outside the norm? I guess it is meant as the confidence interval, where we set the max error as 3 * stddev?

ahojnnes: I am fully confused now. That's also the same. Excuse the brain fart :-)

sarlinpe (approved):

src/colmap/estimators/bundle_adjustment.cc
```diff
@@ -1011,17 +1011,28 @@ class PosePriorBundleAdjuster : public BundleAdjuster {
...
-        ++num_covs;
+    if (prior_options_.ransac_max_error > 0) {
+      ransac_options.max_error = prior_options_.ransac_max_error;
+    } else {
+      double max_stddev_sum = 0;
+      size_t num_valid_covs = 0;
+      for (const auto& [_, pose_prior] : pose_priors_) {
+        if (pose_prior.IsCovarianceValid()) {
+          const double max_stddev =
+              std::sqrt(Eigen::SelfAdjointEigenSolver<Eigen::Matrix3d>(
```
ahojnnes, line 1022: @B1ueber2y I think this makes more sense now. PTAL. Thanks.

PR #2997, opened 2024-12-05
Fix ccache installation in pycolmap windows CI

ahojnnes:
Installation current fails with the following error:
```
  Download CCache
  Installation failed with an error
  A parameter cannot be found that matches parameter name 'MaximumRetryCount'.
```
Likely due to powershell vs. powershell core feature differences. This PR switches to powershell core.


B1ueber2y (approved):

sarlinpe (approved):
Nice, thank you!

PR #2999, opened 2024-12-06
Use Azure blob storage as vcpkg binary cache

ahojnnes:
The Github actions cache only supporting 10GB makes our builds unusably slow. This PR attempts to configure the cache against an Azure blob storage container that I'll pay out of my pocket.


B1ueber2y (approved):
Thanks a lot!!

PR #3022, opened 2024-12-09
Add support for downloading files

ahojnnes:
Towards automatic download of resources like vocabulary trees or torch model weights. The idea is to support the automatic download to a `$HOME/.cache/colmap` folder.


src/colmap/util/file_test.cc
```diff
@@ -32,6 +32,7 @@
 #include "colmap/util/testing.h"
 
 #include <cstring>
+#include <thread>
```
sarlinpe, line 35: Why is this header needed?

CMakeLists.txt
```diff
@@ -72,6 +73,10 @@ else()
     message(STATUS "Disabling LSD support")
 endif()
 
+if(CURL_ENABLED)
+    list(APPEND VCPKG_MANIFEST_FEATURES "http")
```
sarlinpe, line 77: Do we want to keep the VCPKG feature called `http` or rename it to `curl` to be consistent with the CMake option?

sarlinpe (approved):

CMakeLists.txt
```diff
@@ -46,6 +46,7 @@ option(PROFILING_ENABLED "Whether to enable google-perftools linker flags" OFF)
 option(CCACHE_ENABLED "Whether to enable compiler caching, if available" ON)
 option(CGAL_ENABLED "Whether to enable the CGAL library" ON)
 option(LSD_ENABLED "Whether to enable the LSD library" ON)
+option(CURL_ENABLED "Whether to enable curl for auto-download of resources" ON)
```
sarlinpe, line 49: Should we call this option by its feature, e.g. `AUTODOWNLOAD_ENABLED`, instead of by the dependency, in case we later switch to a different downloading util than curl?

src/colmap/util/file_test.cc
ahojnnes: No, leftover from previous httplib implementation.

CMakeLists.txt
ahojnnes: Yes, this is better.

Issue #3017, opened 2024-12-09
Add Lat/Lon to UTM Conversion Feature


StonerLing:
Hi,

Now the transformatioin between ellipsoidal GPS coordinates to Cartesian GPS coordinate in [src/colmap/geometry/gps.h](https://github.com/colmap/colmap/blob/b2a93eb8ad329a20e8fafbad5f2f19f787a73878/src/colmap/geometry/gps.h) only supports LLA to ECEF or ENU.

I wonder if we could add a feature for converting LLA to UTM coordinates and vice versa. [UTM](https://en.wikipedia.org/wiki/Universal_Transverse_Mercator_coordinate_system) coordinates offer better precision in local areas than ENU and ECEF, minimizing distortion and maintaining a constant scale, which is essential for accurate mapping and alignment with other geographic data.

ahojnnes:
There is currently no plan to support UTM coordinates, but if somebody finds the time to add it, we would gladly accept a pull request. Thanks for the suggestion.

behnamasadi:
This appears to be implemented now: `GPSTransform::EllipsoidToUTM` / `GPSTransform::UTMToEllipsoid` exist in [`src/colmap/geometry/gps.h`](https://github.com/colmap/colmap/blob/main/src/colmap/geometry/gps.h) and are also exposed in pycolmap as `ellipsoid_to_utm` / `utm_to_ellipsoid`. Perhaps this issue can be closed?

PR #3030, opened 2024-12-11
Add conversion between LLA and UTM coords.

StonerLing:
- Add feature mentioned in #3017
- Add related unit tests, which have been passed on my machine.

### DIFF
--- a/src/colmap/geometry/gps.cc
+++ b/src/colmap/geometry/gps.cc
@@ -315,7 +315,7 @@ std::pair<std::vector<Eigen::Vector3d>, int> GPSTransform::EllToUTM(
     utm.emplace_back(E * 1000, N * 1000, lla[2]);  // converts back to meters
   }
 
-  return std::make_pair(utm, zone);
+  return std::make_pair(std::move(utm), zone);
 }
 
 std::vector<Eigen::Vector3d> GPSTransform::UTMToEll(
--- a/src/colmap/geometry/gps.cc
+++ b/src/colmap/geometry/gps.cc
@@ -32,34 +32,36 @@
 #include "colmap/math/math.h"
 
 namespace colmap {
-namespace utm {
-struct Params {
+namespace {
+struct UTMParams {
   // The order of the series expansion, determining the precision.
-  static constexpr int order = 4;
+  static constexpr int kOrder = 4;
+
+  // The following notations come from:
+  // https://en.wikipedia.org/wiki/Universal_Transverse_Mercator_coordinate_system
 
   // Powers of mathematical sign $n$, n[i] means $n^i$
-  std::array<double, order + 1> n;
+  std::array<double, kOrder + 1> n;
   // Alpha coefficients for the series expansion
-  std::array<double, order> alpha;
+  std::array<double, kOrder> alpha;
   // Beta coefficients for the series expansion
-  std::array<double, order> beta;
+  std::array<double, kOrder> beta;
   // Delta coefficients for the series expansion
-  std::array<double, order> delta;
+  std::array<double, kOrder> delta;
 
   // Multiplicative factor
   double A;
   // UTM scale factor at the central meridian, typically 0.9996
-  const double k0 = 0.9996;
+  static constexpr double k0 = 0.9996;
   // Easting of the origin, typically 500 km
-  const double E0 = 500;
+  static constexpr double E0 = 500;
   // Northing of the origin
-  double N0(double lat_or_hemi) const { return lat_or_hemi > 0 ? 0 : 1e4; }
+  static double N0(double lat_or_hemi) { return lat_or_hemi > 0 ? 0 : 1e4; }
 
-  Params() = delete;
-  Params(double a, double f) {
+  UTMParams(double a, double f) {
     n[0] = 1;
     n[1] = f / (2.0 - f);
-    for (size_t i = 2; i < order + 1; ++i) {
+    for (size_t i = 2; i < kOrder + 1; ++i) {
       n[i] = n[1] * n[i - 1];
     }
 
@@ -82,14 +84,14 @@ struct Params {
 
     A = a / (1.0 + n[1]) * (1.0 + n[2] / 4.0 + n[4] / 64.0);
   };
+
+  static double ZoneToCentralMeridian(int zone) { return 6 * zone - 183; }
+
+  static int MeridianToZone(double meridian) {
+    return static_cast<int>(std::floor((meridian + 180) / 6)) + 1;
+  }
 };
-
-double ZoneToCentralMeridian(int zone) { return 6 * zone - 183; }
-
-int MeridianToZone(double meridian) {
-  return static_cast<int>(floor((meridian + 180) / 6)) + 1;
-}
-}  // namespace utm
+}  // namespace
 
 GPSTransform::GPSTransform(const int ellipsoid) {
   switch (ellipsoid) {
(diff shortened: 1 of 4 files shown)


src/colmap/geometry/gps.cc
```diff
@@ -197,6 +255,115 @@ std::vector<Eigen::Vector3d> GPSTransform::ENUToXYZ(
...
+    const std::vector<Eigen::Vector3d>& ell, int* output_zone) const {
+  // https://en.wikipedia.org/wiki/Universal_Transverse_Mercator_coordinate_system
+
+  utm::Params params(a_ / 1000.0, f_);  // converts to kilometers
+
+  // For cases where points span different zones, we select the predominant zone
+  // as the reference frame.
+  std::vector<int> zone_counts(60, 0);
+  for (const Eigen::Vector3d& lla : ell) {
+    CHECK(lla[0] >= -90 && lla[0] <= 90)
```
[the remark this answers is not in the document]
ahojnnes: We use exceptions instead of hard checks in colmap.

src/colmap/geometry/gps.cc
```diff
@@ -197,6 +255,115 @@ std::vector<Eigen::Vector3d> GPSTransform::ENUToXYZ(
...
+    }
+
+    E = params.E0 + params.k0 * params.A * E;
+    N = params.N0(lla[0]) + params.k0 * params.A * N;
+
+    utm.push_back({E * 1000, N * 1000, lla[2]});  // converts back to meters
+  }
+
+  // returns zone number if needed
+  if (output_zone) {
```
ahojnnes, line 316:
```suggestion
  if (output_zone != nullptr) {
```

src/colmap/geometry/gps.cc
```diff
@@ -32,6 +32,64 @@
...
+    // clang-format on
+
+    A = a / (1.0 + n[1]) * (1.0 + n[2] / 4.0 + n[4] / 64.0);
+  };
+};
+
+double ZoneToCentralMeridian(int zone) { return 6 * zone - 183; }
+
+int MeridianToZone(double meridian) {
+  return static_cast<int>(floor((meridian + 180) / 6)) + 1;
```
ahojnnes, line 90:
```suggestion
  return static_cast<int>(std::floor((meridian + 180) / 6)) + 1;
```

src/colmap/geometry/gps.cc
```diff
@@ -32,6 +32,64 @@
...
+  // Multiplicative factor
+  double A;
+  // UTM scale factor at the central meridian, typically 0.9996
+  const double k0 = 0.9996;
+  // Easting of the origin, typically 500 km
+  const double E0 = 500;
+  // Northing of the origin
+  double N0(double lat_or_hemi) const { return lat_or_hemi > 0 ? 0 : 1e4; }
+
+  Params() = delete;
```
ahojnnes, line 58:
```suggestion
```
This should not be needed. The default constructor will be deleted by default already.

src/colmap/geometry/gps.cc
```diff
@@ -197,6 +255,115 @@ std::vector<Eigen::Vector3d> GPSTransform::ENUToXYZ(
...
+      double doubled_index = 2.0 * static_cast<double>(i + 1);
+      phi += params.delta[i] * std::sin(doubled_index * chi);
+    }
+
+    const double lat = RadToDeg(phi);
+    const double lon =
+        utm::ZoneToCentralMeridian(zone) +
+        RadToDeg(std::atan(std::sinh(eta_prime) / std::cos(xi_prime)));
+
+    ell.push_back({lat, lon, ena[2]});
```
ahojnnes, line 361:
```suggestion
    ell.emplace_back(lat, lon, ena[2]);
```

src/colmap/geometry/gps.cc
```diff
@@ -197,6 +255,115 @@ std::vector<Eigen::Vector3d> GPSTransform::ENUToXYZ(
...
+  }
+  return utm;
+}
+
+std::vector<Eigen::Vector3d> GPSTransform::UTMToEll(
+    const std::vector<Eigen::Vector3d>& utm, int zone, bool hemi) const {
+  CHECK(zone >= 1 && zone <= 60) << "Zone number must be between 1 and 60.";
+
+  // Setup params
+  utm::Params params(a_ / 1000.0, f_);  // converts to kilometers
```
ahojnnes, line 327:
```suggestion
  const utm::Params params(a_ / 1000.0, f_);  // converts to kilometers
```

src/colmap/geometry/gps.cc
```diff
@@ -197,6 +255,115 @@ std::vector<Eigen::Vector3d> GPSTransform::ENUToXYZ(
   return xyz;
 }
 
+std::vector<Eigen::Vector3d> GPSTransform::EllToUTM(
+    const std::vector<Eigen::Vector3d>& ell, int* output_zone) const {
+  // https://en.wikipedia.org/wiki/Universal_Transverse_Mercator_coordinate_system
+
+  utm::Params params(a_ / 1000.0, f_);  // converts to kilometers
```
ahojnnes, line 262:
```suggestion
  const utm::Params params(a_ / 1000.0, f_);  // converts to kilometers
```

src/colmap/geometry/gps.cc
```diff
@@ -32,6 +32,64 @@
...
+  std::array<double, order> delta;
+
+  // Multiplicative factor
+  double A;
+  // UTM scale factor at the central meridian, typically 0.9996
+  const double k0 = 0.9996;
+  // Easting of the origin, typically 500 km
+  const double E0 = 500;
+  // Northing of the origin
+  double N0(double lat_or_hemi) const { return lat_or_hemi > 0 ? 0 : 1e4; }
```
ahojnnes, line 56: Could be a static method.

src/colmap/geometry/gps.cc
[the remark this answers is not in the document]
ahojnnes: What would be even better here is to decompose the check into two lines and use THROW_CHECK_GE and THROW_CHECK_LE, such that violations of the check lead to a traceable error message with the specific values.

src/colmap/geometry/gps.h
```diff
@@ -74,6 +74,22 @@ class GPSTransform {
...
+  // The conversion utilizes a 4th-order expansion formula. Easting offset
+  // is 500 km, and northing offset is 10,000 km for the Southern Hemisphere.
+  std::vector<Eigen::Vector3d> EllToUTM(const std::vector<Eigen::Vector3d>& ell,
+                                        int* zone = nullptr) const;
+
+  // Converts UTM coords to GPS (lat / lon / alt).
+  // Requires the zone number and hemisphere (true for north, false for south).
+  std::vector<Eigen::Vector3d> UTMToEll(const std::vector<Eigen::Vector3d>& utm,
+                                        int zone,
+                                        bool hemi) const;
```
sarlinpe, line 91: Should call this `is_north` as it is less ambiguous?

src/colmap/geometry/gps.h
```diff
@@ -74,6 +74,22 @@ class GPSTransform {
...
                                         double alt0) const;
 
+  // Converts GPS (lat / lon / alt) to UTM coords.
+  // Returns the zone number for the converted coordinates if a pointer to an
+  // integer is provided. If the points span multiple zones, the zone with the
+  // most points is chosen as the reference frame.
+  //
+  // The conversion utilizes a 4th-order expansion formula. Easting offset
+  // is 500 km, and northing offset is 10,000 km for the Southern Hemisphere.
+  std::vector<Eigen::Vector3d> EllToUTM(const std::vector<Eigen::Vector3d>& ell,
```
sarlinpe, line 84: Is there really a use case where the zone is not needed? I suggest to always return the zone as `std::pair<std::vector<Eigen::Vector3d>, int>`, which we can conveniently unpack with structured bindings (C++17).

StonerLing: I did this to keep the interface consistent (though it may not matter). If the user only needs relative coordinates, the zone might not be necessary. However, your suggestion makes the function clearer and more directly transfers the responsibility of using the zone to the caller, so I'll update it accordingly.

ahojnnes (approved):
Thanks, looks good to me.

PR #3037, opened 2024-12-13
Integrate utm conversion and add prior position offset option

StonerLing:
- Add an option in the CLI to specify the Cartesian frame into which GPS coordinates are transformed, including the newly added UTM coordinates in #3030 
- Add an option in the CLI to determine whether to apply a mean shift (offset) to prior positions.


StonerLing:
This need to be disscuessed.

I added the `mean_shift_to_prior_positions` feature because I considered that after converting GPS coordinates to ENU/ECEF/UTM, the numerical values of the coordinates are quite large, reaching into the seventh or eighth digit. Specifically, it is intended to: 

- reduce the magnitude of the numerical values to enhance numerical stability;
- for related functions that require iteration, smaller numerical values are easier to set as initial values; 
-  the output model displays better in the colmap GUI.

To be honest, I thought of this because when I tried to use pose prior in `glomap`, the excessively large coordinate values did not converge well, and the range for glomap's global positioning, which is randomly generated in the code, is fixed at [-100,100].

ahojnnes:
Thank you, will have a look at this in the next days.

src/colmap/exe/model.cc
```diff
@@ -121,21 +121,27 @@ std::vector<Eigen::Vector3d> ConvertCameraLocations(
...
-    } else {
+    } else if (alignment_type == "enu") {
       LOG(INFO) << "Converting Alignment Coordinates from GPS (lat/lon/alt) "
                    "to ENU.";
       return gps_transform.EllToENU(
           ref_locations, ref_locations[0](0), ref_locations[0](1));
+    } else if (alignment_type == "utm") {
+      LOG(INFO) << "Converting Alignment Coordinates from GPS (lat/lon/alt) "
+                   "to UTM.";
+      const auto [utm, _] = gps_transform.EllToUTM(ref_locations);
```
B1ueber2y, line 136: Should we handle the possibility of images crossing utm zones. This can happen particularly when the video is shot at the zone boundaries.

src/colmap/geometry/gps.h
```diff
@@ -136,10 +137,29 @@ struct PosePrior {
 
   inline bool operator==(const PosePrior& other) const;
   inline bool operator!=(const PosePrior& other) const;
+
+  // Returns the offset applied to the original GPS position.
+  static Eigen::Vector3d Offset();
```
B1ueber2y, line 142: Should we use ``PositionOffset()`` to align with the naming?

B1ueber2y:
Looks good to me. Thanks a lot!

src/colmap/geometry/gps.h
```diff
@@ -45,6 +45,7 @@ namespace colmap {
 class GPSTransform {
  public:
   MAKE_ENUM(ELLPSOID, 0, GRS80, WGS84);
+  MAKE_ENUM(CartesianFrame, 0, ECEF, ENU, UTM);
```
ahojnnes, line 48:
```suggestion
  MAKE_ENUM(CARTESIAN_FRAME, 0, ECEF, ENU, UTM);
```

ahojnnes: Or maybe better actually rename the existing ELLIPSOID to Ellipsoid.

src/colmap/geometry/gps.h
```diff
@@ -136,10 +137,29 @@ struct PosePrior {
...
   inline bool operator!=(const PosePrior& other) const;
+
+  // Returns the offset applied to the original GPS position.
+  static Eigen::Vector3d Offset();
+  // Returns the "unoffset" position, i.e., the position after removing the
+  // offset.
+  inline Eigen::Vector3d UnoffsetPosition() const;
+
+ private:
+  friend class DatabaseCache;
```
ahojnnes, line 148: Is this really necessary? I don't see any usage of this declaration.

B1ueber2y: I think this is for setting the position_offset. I also got confused. Maybe a setter is better than friend class?

ahojnnes: OK, I see, it is for the position_offset... I am not so happy with the usage of static here, because it makes this a singleton state. It prevents using this class for two locations in the same process.

src/colmap/scene/database_cache.cc
```diff
@@ -260,21 +262,43 @@ bool DatabaseCache::SetupPosePriors() {
...
 
     auto xyz_prior_it = v_xyz_prior.begin();
     for (const auto& image_id : image_ids_with_prior) {
       struct PosePrior& pose_prior = PosePrior(image_id);
-      pose_prior.position = *xyz_prior_it;
+      pose_prior.position = *xyz_prior_it + offset;
       pose_prior.coordinate_system = PosePrior::CoordinateSystem::CARTESIAN;
       ++xyz_prior_it;
     }
+    PosePrior::position_offset = offset;
```
ahojnnes, line 301: I would keep the pose priors in the original coordinates and instead normalize the coordinates wherever we need the extra numerical precision. For example, this is currently only the case in the PosePriorBundleAdjuster.

StonerLing:
Sorry for any confusion caused by my use of `static` and `friend class` to control the GPS offset. It was a misguided idea on my part. And of course, GPS offset is currently unnecessary in the COLMAP pose prior mapper.

I plan to address these issues and possibly introduce 'utm-plane' and 'utm-plane-unscaled' options, similar to ENU in the model aligner, next week.

Though it may not be important, by the way, the primary reason for adding the offset was to facilitate pose prior usage in GLOMAP and to visualize the pose prior-based output model in the GUI. It's definitely more appropriate to handle this locally rather than setting the offset as a global configuration.

ahojnnes:
@StonerLing Thank you. If the primary concern is visualization in the GUI, then we can find other means to do so. For example, we could maintain such an offset just in the visualized reconstruction.

ahojnnes:
@StonerLing Please take a look at this PR and try whether it solves your main concern of visualizing models with large coordinates. https://github.com/colmap/colmap/pull/3044

src/colmap/geometry/gps.cc
```diff
@@ -266,6 +266,7 @@ std::pair<std::vector<Eigen::Vector3d>, int> GPSTransform::EllToUTM(
 
   // For cases where points span different zones, we select the predominant zone
   // as the reference frame.
+  // FUTURE: implement a more accurate method for merging UTM zones.
```
ahojnnes, line 269:
```suggestion
  // TODO: Implement a more accurate method for merging UTM zones.
```

src/colmap/scene/database_cache.h
```diff
@@ -98,7 +98,7 @@ class DatabaseCache {
   const class Image* FindImageWithName(const std::string& name) const;
 
   // Setup PosePriors for PosePriorBundleAdjustment
-  bool SetupPosePriors();
+  bool SetupPosePriors(GPSTransform::CartesianFrame cartesian_frame_from_lla);
```
ahojnnes, line 101: nit: using our current naming conventions `cartesian_frame_from_lla` implies a transformation. I suggest renaming this to just `cartesian_frame`.

ahojnnes (approved):

ahojnnes (approved):
Thank you for your contribution!

ahojnnes:
@StonerLing Not sure why the checks are not being run and reported.

StonerLing:
@ahojnnes That's strange... Should I make a new PR?

ahojnnes:
Either that or we wait for the next other PR to be merged so this branch needs an update. Hopefully then it triggers correctly.

src/colmap/controllers/incremental_pipeline.h
```diff
@@ -129,6 +129,12 @@ struct IncrementalPipelineOptions {
   // (chi2 for 3DOF at 95% = 7.815)
   double prior_position_loss_scale = 7.815;
 
+  // Specifies the Cartesian coordinate system used for converting ellipsoidal
+  // GPS coordinates when utilizing prior position.
+  GPSTransform::CartesianFrame cartesian_frame =
+      GPSTransform::CartesianFrame::ENU;
+
+  // bool shift_origin_to_avereage_gps = false;
```
sarlinpe, line 137: Orphan comment?

src/colmap/geometry/gps.h
```diff
@@ -44,7 +44,8 @@ namespace colmap {
 // representation and vice versa.
 class GPSTransform {
  public:
-  MAKE_ENUM(ELLPSOID, 0, GRS80, WGS84);
+  MAKE_ENUM(Ellpsoid, 0, GRS80, WGS84);
+  MAKE_ENUM(CartesianFrame, 0, ECEF, ENU, UTM);
```
sarlinpe, line 48: Would it make sense to merge both of these enums with `PosePrior::CoordinateSystem`? The pose priors could be equivalently in ECEF or GRS80 (but not UTM since it'd need to include the zone).

src/colmap/controllers/incremental_pipeline.cc
```diff
@@ -261,9 +261,9 @@ bool IncrementalPipeline::LoadDatabase() {
   }
 
   // If prior positions are to be used and setup from the database, convert
-  // geographic coords. to cartesian ones
+  // geographic coords to cartesian ones
```
sarlinpe, line 264:
```suggestion
  // geographic coordinates to cartesian ones.
```

src/colmap/exe/sfm.cc
```diff
@@ -392,6 +392,26 @@ int RunPosePriorMapper(int argc, char** argv) {
...
+  std::string cartesian_frame = "enu";
+  options.AddDefaultOption(
+      "cartesian_frame",
+      &cartesian_frame,
+      "Specifies the Cartesian coordinate frame for converting from latitude, "
+      "longitude, and altitude (LLA). Possible values are 'ENU', "
+      "'ECEF', and 'UTM'.");
+
+  StringToLower(&cartesian_frame);
+  if (cartesian_frame == "enu") {
```
sarlinpe, line 405: Use a switch statement instead.

src/colmap/geometry/gps.h
ahojnnes: The decomposition into Ellipsoid and CartesiamFrame seems to be a better one than the one in `PosePrior::CoordinateSystem`. One can use ECEF/ENU/UTM cartesian coordinates based on different ellipsoids. Is your proposal to replace the `PosePrior::CoordinateSystem` with the two enums here or flatten it into an enum with 6 values?

src/colmap/controllers/incremental_pipeline.h
StonerLing: Deleted.

src/colmap/exe/sfm.cc
StonerLing: The condition in switch statement can only be integral/enumeration/class(implicitly converted to integral/enumeration) for [C++17](https://en.cppreference.com/w/cpp/language/switch), and colmap seems don't have a method to transform string to enum(only have enum to string). So, to use switch statement here

- Should we consider to add function like `StringToEnum`?
- Or should we consider to transform string to some unique integral(e.g. hash value)?

src/colmap/geometry/gps.h
```diff
@@ -44,7 +44,8 @@ namespace colmap {
 // representation and vice versa.
 class GPSTransform {
  public:
-  MAKE_ENUM(ELLPSOID, 0, GRS80, WGS84);
+  MAKE_ENUM(Ellpsoid, 0, GRS80, WGS84);
```
ahojnnes, line 47:
```suggestion
  MAKE_ENUM(Ellipsoid, 0, GRS80, WGS84);
```

src/colmap/geometry/gps.h
ahojnnes: Looking at this again, I think the following serves all the practical use cases:
```
  MAKE_ENUM_CLASS(CoordinateSystem,
                  -1,
                  UNDEFINED,        // = -1
                  SPHERICAL_WGS84,  // = 0
                  CARTESIAN,        // = 1
                  SPHERICAL_GRS80,  // = 2
  );
```
With this, one can store lat/lon/alt either wrt to GRS80/WGS84 or directly in a suitable cartesian frame but we don't really care about which frame that is as long as it is cartesian. If we later care about whether the cartesian frame is metric, we could distinguish between `CARTESIAN_METRIC` and `CARTESIAN_NON_METRIC` later.

src/colmap/exe/sfm.cc
ahojnnes: I would keep it as is and address this separately in another PR, if we can make our macros produce a StringToEnum function. This can then also be applied to various other places with enums.

src/colmap/geometry/gps.h
StonerLing: @ahojnnes In my opinion, `CARTESIAN` is a high-level abstraction, while `SPHERICAL_WGS84` and `SPHERICAL_GRS80` are specific and accurate spherical frames. Should they be in the same enum class?

If we want to make one enum only, maybe this is better?
```cpp
MAKE_ENUM_CLASS(-1, 
                UNDEFINED, 
                SPHERICAL_WGS84, 
                SPHERICAL_GRS80,
                CARTESIAN_ECEF, 
                CARTESIAN_ENU,
                CARTESIAN_UTM);
```

Or more abstract, 
```cpp
MAKE_ENUM_CLASS(-1, 
                UNDEFINED, 
                SPHERICAL, 
                CARTESIAN);
```

src/colmap/exe/sfm.cc
sarlinpe: Don't we already have this for enums defined with `MAKE_ENUM`?
https://github.com/colmap/colmap/blob/590c5091590e3ea97c3a87229a688f56b643ce62/src/colmap/geometry/gps.cc#L376

https://github.com/colmap/colmap/blob/590c5091590e3ea97c3a87229a688f56b643ce62/src/colmap/util/enum_to_string.h#L54-L55

src/colmap/geometry/gps.h
ahojnnes: Currently, we cannot fully represent/serialize all information in CARTSIAN_UTM/CARTESIAN_ENU in the pose prior, because we lose the zone and origin. I am also unsure what we would do with this information.

Let's also be careful to preserve the existing enum values, otherwise existing database content will be broken.

sarlinpe: Since we are expanding the cartesian frames to include ECEF and UTM, I think that we should make the reference frame of each pose prior explicit (instead of only whether it is cartesian or not). This would allow:
- having databases with both WGS84 and cartesian priors (such that we can convert from the former to the latter)
- making a difference between pose priors in cartesian frames that are gravity-aligned (ENU, UTM) and those that are not (ECEF), which will be required when integrating IMU or image-based gravity priors.

Thus I indeed suggest to have an enum with 6 values, e.g. UNDEFINED, WGS84, GRS80, ECEF, ENU, UTM. The only problem is that UTM is not sufficient on its own since it needs a zone...

Alternatively, if we want to enforce that all pose priors are provided in the same reference frame, then we should make this an attribute of the database table rather than of each row.

StonerLing: I understand, so lat/lon/alt and ECEF(or say CARTESIAN) are worth being abstracted out separately, because they both represent sufficient information with just a three-dimensional vector. :D

`GPSTransform` is a helper class to convert gps coordinates, maybe the enum inside it could be more specific.

ahojnnes: In other words, I propose to go with (keeping it backwards compatible with old enum values):
```
MAKE_ENUM_CLASS(CoordinateSystem, -1, UNDEFINED, SPHERICAL_WGS84, CARTESIAN_ECEF , SPHERICAL_GRS80, CARTESIAN_ENU, CARTESIAN_UTM);
```

ahojnnes:
@StonerLing Unfortunately, we had to do a bit of a surgical operation on the main branch and overwrite git history for the last 5 commits (in order to undo enablement of git lfs). Not sure when you last merged with main, but you may have to rebase on top of main in this branch now.

xiemeilong:
Is there anything blocking this pull request?

StonerLing:
@xiemeilong, as we discussed earlier, to enhance the representation of pose position priors in COLMAP, a more general and complete approach is needed. This involves updating the underlying framework to better handle prior positions, such as fully storing the information in the database.

I have been working on adding a new class to represent prior positions, but it's still a work in progress(see the latest commit). I'm not sure if this meets your requirements......@ahojnnes.

ahojnnes:
@StonerLing Apologies, I will try to allocate some time to look at what is implemented now and provide feedback. Thank you.

StonerLing:
Thanks! You're right that serializing UTM context is one application — but `CoordinateSystemContext` is intended to be more general. e.g.

- **Local ENU**: GNSS on drones or robots often reports positions in ENU, defined relative to a specific geodetic origin
- **Sensor rigs**: Pose data from IMU or GPS/INS systems is usually in the sensor frame, requiring mounting transforms
- **Multi-agent systems**: Devices might operate in different UTM zones or local frames, so we need to track these separately
- **Survey/GCP data**: Sometimes provided in custom projections or local engineering frames — each with its own context

The idea is to support full transformation chains like:
`WGS84 -> ENU_1 -> sensor_1 -> camera_1 -> model`
`GRS80 -> ENU_2 -> sensor_1 -> camera_2 -> model`
`WGS84 -> UTM_1 ->  camera_1 -> model`

I understand the current implementation might not be ideal or elegent yet... the goal is to **make this flexible enough to support common real-world setups where pose priors are expressed in various coordinate systems.**

Let me know if there's a better way to approach this!

sarlinpe:
Do we really need this level of complexity? I'm afraid that this implementation is over-generalized - we will likely never want to do SfM across UTM timezones. Converting all pose priors to a local ENU would be just much simpler, no?

ahojnnes:
Thank you @StonerLing for your response. I have the same concerns as @sarlinpe about this being a bit too general for our use cases and thus also unnecessarily complex. I appreciate the amount of effort that already went into designing the CoordinateSystemContext and we can keep it in a branch as a backup in case we ever need the generality in the future. Otherwise, I suggest going with a simpler and more constrained solution.

xiemeilong:
I have been waiting for this feature for a long time. Will there be a new pull request?

StonerLing:
Thanks for the interest! Maybe I will try again in the future when I have time, but I’m not sure yet.

One reason I closed this PR is that it eventually tried to address too many things at once:

1. enabling SfM to directly operate in UTM coordinates (converted from GPS),

2. letting the database store full geodetic metadata (LLA ellipsoid, ENU origins, UTM zone/hemisphere),

3. fixing a potential issue in COLMAP’s LLA → ENU conversion where an arbitrary origin should be allowed.

Originally only (1) was my goal; (2) came from later discussions, and (3) was discovered during implementation.

Conceptually, COLMAP currently has only one well-defined path for registering georeferenced models: a similarity transform alignment. Even for ENU, the internal origin is implicitly chosen (the first image) and is not exposed to the user. Once we introduce an additional geospatial coordinate system such as UTM, this implicit origin—together with the gauge/scale information it provides—needs to be stored consistently in both the database and the exported models. That complicates the current design.

Moreover, for GPS data, regardless of converting to ENU/ECEF/UTM, there is meaningful scale/gauge information coming from the fixed origin and orientation. This interacts with COLMAP’s gauge fixing in BA. UTM in particular is mainly beneficial for large-scale reconstructions where geodetic priors need higher metric accuracy, or when interfacing with GIS, DEM, orthophotos, etc.

At this stage COLMAP is still more focused on the geometric/estimation aspect of reconstruction rather than full geospatial support. The recent refactoring of pose priors (for future GNSS/IMU integration) is a good step, but for small or medium-scale datasets ENU will likely remain the preferred coordinate frame.

ahojnnes:
Thanks again @StonerLing, I appreciate your effort on this and I still would love to get some of the pieces in this PR merged into the main branch. Especially the fixes in (3) are non-controversial and should make it into the main branch in a separate PR. What do you think?

StonerLing:
I’m glad to hear that, and I’ll try to extract (3) into a separate PR when I get a chance.

PR #3040, opened 2024-12-14
Abstract algorithm class IncrementalMapperImpl

B1ueber2y:
* Abstract struct ``RegistrationStatistics`` to simplify the IncrementalMapper header
* Abstract an algorithm class as discussed and move in some indepependent methods


src/colmap/sfm/incremental_mapper_impl.h
```diff
@@ -0,0 +1,100 @@
...
+  static std::vector<image_t> FindLocalBundle(
+      const IncrementalMapper::Options& options,
+      image_t image_id,
+      const std::shared_ptr<class Reconstruction>& reconstruction);
+
+  // Implement IncrementalMapper::EstimateInitialTwoViewGeometry
+  static bool EstimateInitialTwoViewGeometry(
+      const IncrementalMapper::Options& options,
+      const std::shared_ptr<const DatabaseCache>& database_cache,
+      TwoViewGeometry& two_view_geometry,
```
ahojnnes, line 95: Suggest making the output argument always the last. In this case, we could also change the interface to return an std::optional<TwoViewGeometry> instead?

ahojnnes:
LGTM, thanks.

src/colmap/sfm/incremental_mapper_impl.h
B1ueber2y: Thanks. I have switched the order. Making it optional will also lead to changes in the code here: https://github.com/colmap/colmap/blob/main/src/colmap/controllers/incremental_pipeline.cc#L280-L300, where the two_view_geometry object is shared across different function calls. I would leave it as it is for this PR.

ahojnnes (approved):

PR #3041, opened 2024-12-15
Perform linear interpolation in percentile computation

ahojnnes:
Makes Median and Percentile computation consistent. Enables to reuse Percentile for Median.


B1ueber2y (approved):
Thanks. The logic looks good to me. Would be good if we have some input examples in the nominal tests that are not initially sorted.

ahojnnes:
> Thanks. The logic looks good to me. Would be good if we have some input examples in the nominal tests that are not initially sorted.

Very good point, thank you. Done.

PR #3042, opened 2024-12-15
Templating the static methods in the algorithm class IncrementalMapperImpl

B1ueber2y:
In this way we achieve reusability without touching the IncrementalMapper class. I also added static_assert to ensure type safety between Reconstruction and ObservationManager, and DatabaseCache and ObservationManager. 

If we are happy with it I will move forward to update some core logics in the IncrementalTriangulator and ObservationManager to templated methods in an algorithm class in the following PRs as well.


B1ueber2y:
Made an update to change all the shared_ptr arguments to const reference in the algorithm class: https://github.com/colmap/colmap/pull/3042/commits/24f9776d531946fd1ec13c9382e97adf988dd9bb

B1ueber2y:
Actually we should remove dependencies across classes rather than doing static_assert. Will first improve the logic in a separate PR and then proceed with templating.

B1ueber2y:
> Actually we should remove dependencies across classes rather than doing static_assert. Will first improve the logic in a separate PR and then proceed with templating.

Submitted here: https://github.com/colmap/colmap/pull/3043

PR #3044, opened 2024-12-15
Support visualization of models with arbitrary origin and scale

ahojnnes:


StonerLing:
Thanks. It solves the issue of visualizing models with large coordinates in my test.

A small issue: the `View center grid` and `Coordinate axes` will move away from the `model origin` when I try to rotate, translate or zoom the model.

ahojnnes:
> Thanks. It solves the issue of visualizing models with large coordinates in my test.
> 
> A small issue: the `View center grid` and `Coordinate axes` will move away from the `model origin` when I try to rotate, translate or zoom the model.

This should be fixed now.

B1ueber2y (approved):

PR #3045, opened 2024-12-15
Templating the static methods in the algorithm class IncrementalMapperImpl

B1ueber2y:
In this way we achieve reusability without touching the IncrementalMapper class. If we are happy with it I will move forward to update some core logics in the IncrementalTriangulator and ObservationManager to templated methods in an algorithm class in the follow-up PRs as well. Reopening for same purpose as https://github.com/colmap/colmap/pull/3042


B1ueber2y:
Enhance type safety here: https://github.com/colmap/colmap/pull/3045/commits/f5f2cf168d02c4b10741c423bef3303b8b4326e6

PR #3047, opened 2024-12-16
Improvements for reconstruction normalization / bbox / centroid

ahojnnes:
* Use nth_element instead of sort for faster ordering.
* Normalize in double precision to better handle large coordinates (e.g., GPS).
* Use Eigen's AlignedBox instead of pair of Vector3d for bbox.
* Extract bbox and centroid computation into reusable geometry module function. Add some more tests.

### DIFF
--- a/src/colmap/exe/model.cc
+++ b/src/colmap/exe/model.cc
@@ -44,27 +44,26 @@
 namespace colmap {
 namespace {
 
-std::vector<std::pair<Eigen::Vector3d, Eigen::Vector3d>>
-ComputeEqualPartsBounds(const Reconstruction& reconstruction,
-                        const Eigen::Vector3i& split) {
-  std::vector<std::pair<Eigen::Vector3d, Eigen::Vector3d>> bounds;
-  const auto bbox = reconstruction.ComputeBoundingBox();
-  const Eigen::Vector3d extent = bbox.second - bbox.first;
+std::vector<Eigen::AlignedBox3d> ComputeEqualPartsBboxes(
+    const Reconstruction& reconstruction, const Eigen::Vector3i& split) {
+  std::vector<Eigen::AlignedBox3d> bboxes;
+  const Eigen::AlignedBox3d bbox = reconstruction.ComputeBoundingBox();
+  const Eigen::Vector3d extent = bbox.diagonal();
   const Eigen::Vector3d offset(
       extent(0) / split(0), extent(1) / split(1), extent(2) / split(2));
 
   for (int k = 0; k < split(2); ++k) {
     for (int j = 0; j < split(1); ++j) {
       for (int i = 0; i < split(0); ++i) {
-        Eigen::Vector3d min_bound(bbox.first(0) + i * offset(0),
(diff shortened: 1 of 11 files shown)


src/colmap/exe/model.cc
```diff
@@ -44,27 +44,26 @@
 namespace colmap {
 namespace {
 
-std::vector<std::pair<Eigen::Vector3d, Eigen::Vector3d>>
-ComputeEqualPartsBounds(const Reconstruction& reconstruction,
-                        const Eigen::Vector3i& split) {
-  std::vector<std::pair<Eigen::Vector3d, Eigen::Vector3d>> bounds;
-  const auto bbox = reconstruction.ComputeBoundingBox();
-  const Eigen::Vector3d extent = bbox.second - bbox.first;
+std::vector<Eigen::AlignedBox3d> ComputeEqualPartsBboxes(
```
ahojnnes, line 47: A lot of code in this file should be extracted into its own module with some tests. Left for future work.

src/colmap/scene/reconstruction.cc
```diff
@@ -301,101 +302,80 @@ Sim3d Reconstruction::Normalize(const bool fixed_scale,
...
-  Eigen::Vector3d mean_coord(0, 0, 0);
-  for (size_t i = P0; i <= P1; ++i) {
-    mean_coord(0) += coords_x[i];
-    mean_coord(1) += coords_y[i];
-    mean_coord(2) += coords_z[i];
-  }
-  mean_coord /= P1 - P0 + 1;
-
-  return std::make_tuple(bbox_min, bbox_max, mean_coord);
+  return ::colmap::ComputeBoundingBoxAndCentroid(min_percentile,
```
sarlinpe, line 374: Would it make sense to rename Reconstruction::ComputeBoundingBoxAndCentroid to avoid the name conflict?

src/pycolmap/geometry/bindings.cc
```diff
@@ -177,6 +177,42 @@ void BindGeometry(py::module& m) {
...
+                    [](Eigen::AlignedBox3d& self, const Eigen::Vector3d& max) {
+                      self.max() = max;
+                    })
+      .def("diagonal", &Eigen::AlignedBox3d::diagonal)
+      .def(
+          "contains_point",
+          [](const Eigen::AlignedBox3d& self, const Eigen::Vector3d& x) {
+            return self.contains(x);
+          },
+          "x"_a)
```
sarlinpe, line 201:
```suggestion
          "point"_a)
```

src/pycolmap/geometry/bindings.cc
```diff
@@ -177,6 +177,42 @@ void BindGeometry(py::module& m) {
...
+      .def(
+          "contains_bbox",
+          [](const Eigen::AlignedBox3d& self, const Eigen::AlignedBox3d& bbox) {
+            return self.contains(bbox);
+          },
+          "x"_a)
```
sarlinpe, line 207:
```suggestion
          "other"_a)
```

PR #3051, opened 2024-12-17
Consistent interface/tests for rigid3d/sim3d/affine2d, pycolmap bindings for rigid3d

ahojnnes:
* Add new functions EstimateRigid3d, EstimateRigid3dRobust, EstimateAffine2d, Estimate2dRobust (consistent with existing EstimateSim3d and Sim3dRobust).
* Improved tests for all of the transformations.
* Expose estimate_rigid3d and estimate_rigid3d_robust in pycolmap bindings.
* Add Rigid3d::FromMatrix (consistent with Sim3d::FromMatrix).

### DIFF
--- a/src/colmap/estimators/affine_transform.cc
+++ b/src/colmap/estimators/affine_transform.cc
@@ -102,6 +102,7 @@ bool EstimateAffine2d(const std::vector<Eigen::Vector2d>& src,
   if (models.empty()) {
     return false;
   }
+  THROW_CHECK_EQ(models.size(), 1);
   tgt_from_src = models[0];
   return true;
 }
--- a/src/colmap/estimators/affine_transform.cc
+++ b/src/colmap/estimators/affine_transform.cc
@@ -29,6 +29,7 @@
 
 #include "colmap/estimators/affine_transform.h"
 
+#include "colmap/optim/loransac.h"
 #include "colmap/util/eigen_alignment.h"
 #include "colmap/util/logging.h"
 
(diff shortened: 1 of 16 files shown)


src/colmap/estimators/affine_transform.cc
```diff
@@ -75,43 +76,47 @@ void AffineTransformEstimator::Estimate(const std::vector<X_t>& points1,
...
-                                         const M_t& A,
+void AffineTransformEstimator::Residuals(const std::vector<X_t>& src,
+                                         const std::vector<Y_t>& tgt,
+                                         const M_t& tgt_from_src,
                                          std::vector<double>* residuals) {
-  THROW_CHECK_EQ(points1.size(), points2.size());
-
-  residuals->resize(points1.size());
-
-  // Note that this code might not be as nice as Eigen expressions,
```
ahojnnes, line 91: This is not true anymore in recent profiling.

src/colmap/estimators/similarity_transform.h
```diff
@@ -129,44 +121,48 @@ EstimateSim3dRobust(const std::vector<Eigen::Vector3d>& src,
...
   models->clear();
 
-  Eigen::Matrix<double, kDim, Eigen::Dynamic> src_mat(kDim, src.size());
-  Eigen::Matrix<double, kDim, Eigen::Dynamic> dst_mat(kDim, dst.size());
-  for (size_t i = 0; i < src.size(); ++i) {
-    src_mat.col(i) = src[i];
-    dst_mat.col(i) = dst[i];
+  using MatrixType = Eigen::Matrix<double, kDim, Eigen::Dynamic>;
+  const Eigen::Map<const MatrixType, Eigen::Aligned16> src_mat(
+      reinterpret_cast<const double*>(src.data()), 2, src.size());
```
B1ueber2y, line 134: Maybe this is a dumb question. But I could not understand why here it is 2 rather than kDim?

ahojnnes: It's a bug :-) Thanks for spotting.

B1ueber2y (approved):

ahojnnes:
> Non-relevant to the update in this PR but relevant to the topic (spotted while reviewing): here the comment is wrong (actually from the initial commit of COLMAP in 2016 xD)? https://github.com/colmap/colmap/blob/main/src/colmap/estimators/similarity_transform.h#L64-L67
> 
> 
> 
> For both rigid and similarity transformation k is always minimal for k dimension and it is not overconstrained due to dependent constraints. 

I believe the comment is correct in that a 3D similarity transformation only has 7 degrees of freedom. While you need 3 points, you can drop 2 coordinates of the third point to obtain the true minimal system of equations.

B1ueber2y:
> > Non-relevant to the update in this PR but relevant to the topic (spotted while reviewing): here the comment is wrong (actually from the initial commit of COLMAP in 2016 xD)? https://github.com/colmap/colmap/blob/main/src/colmap/estimators/similarity_transform.h#L64-L67
> > For both rigid and similarity transformation k is always minimal for k dimension and it is not overconstrained due to dependent constraints.
> 
> I believe the comment is correct in that a 3D similarity transformation only has 7 degrees of freedom. While you need 3 points, you can drop 2 coordinates of the third point to obtain the true minimal system of equations.

But if one assumes this is a similarity transformation there are 2 dependent constraints. Its similar to the classic confusion: 3 points are minimal for Rigid3d although a Rigid3d object only has 6 DoF.

ahojnnes:
I don't think this is correct for 3D similarity transformation but I am also on my phone now... take it with a grain of salt. Ignoring degenerate configurations, you should be able to drop two of the coordinates of the third point.

ahojnnes:
Maybe to proof my point: I can easily find you two sets of three 3D points and you will not be able to fit a rigid3 or sim3 that always perfectly fits these points, ie the residuals of the transformed points will not all be perfectly zero. If it was indeed minimal, the residuals would by definition always be perfectly zero. Hence, it is an over constrained problem.

B1ueber2y:
> Maybe to proof my point: I can easily find you two sets of three 3D points and you will not be able to fit a rigid3 or sim3 that always perfectly fits these points, ie the residuals of the transformed points will not all be perfectly zero. If it was indeed minimal, the residuals would by definition always be perfectly zero. Hence, it is an over constrained problem.

With Rigid3d a solver will get perfect zero with two points, but it is always degenerate because the two point samples are not independent due to the distance preserving property. With three points satisfying the distance-preserving property there are always only 6 independent constraints so only one solution. I think one could not say that 2-point Rigid3d solver for Rigid3d is minimal and 3-point is not, but this can be inferred from your reasoning about residuals being perfect zero. So I am not very convinced. 

Sim3d is similar to Rigid3d with one additional degree of freedom but also one less dependent constraints. I agree that two points plus one dimension of the third point is minimal here, but this is because that all the 6 DoFs from the first two points can be used here with one of them for the scale.

I believe the fact that three perfect point correspondences gives minimal number of independent constraints is the rationale for its minimality, but in the end it comes to how the term "minimal solver" is defined.

B1ueber2y:
I think the confusion mostly comes from the fact that 2-point Rigid3d solver is minimal but always degenerate. I am happy to move on to leave the comments as it is. Thanks : )

PR #3061, opened 2024-12-18
Re-enable interface support for covariance estimation from a raw Ceres::Problem

B1ueber2y:
Sincere request to re-enable this feature. Exposing this interface will make dev much easier when problem is constructed with pyceres, for example, with ``pycolmap.cost_functions.ReprojErrorCost`` and non-identity keypoint covariance (e.g. with a scaled covariance due to resizing). Given that we expose the ``pycolmap.cost_functions`` module there is no downside to support this.

### DIFF
--- a/src/colmap/estimators/covariance.h
+++ b/src/colmap/estimators/covariance.h
@@ -115,6 +115,10 @@ std::optional<BACovariance> EstimateBACovariance(
     const BACovarianceOptions& options,
     const Reconstruction& reconstruction,
     BundleAdjuster& bundle_adjuster);
+std::optional<BACovariance> EstimateBACovariance(
+    const BACovarianceOptions& options,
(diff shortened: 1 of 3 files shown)


src/colmap/estimators/covariance.h
```diff
@@ -115,6 +115,10 @@ std::optional<BACovariance> EstimateBACovariance(
     const BACovarianceOptions& options,
     const Reconstruction& reconstruction,
     BundleAdjuster& bundle_adjuster);
+std::optional<BACovariance> EstimateBACovariance(
```
ahojnnes, line 118: Let's also name this EstimateBACovarianceFromProblem ?

B1ueber2y: Thanks. This is done.

ahojnnes (approved):

PR #3062, opened 2024-12-18
Fix python bindings of covariance estimation

B1ueber2y:
Bug fix for https://github.com/colmap/colmap/pull/2788


ahojnnes (approved):
Oh, bad bug. Thanks :-)

B1ueber2y:
Added another update to use ``py::array_t<double>`` for covariance estimation pybind interface instead of ``double*``, which could never be called before from Python side. Example:

```
In [1]: import pycolmap

In [2]: param = pycolmap.ExperimentalPoseParam()

In [3]: print(param.qvec)
None

In [4]: import numpy as np

In [5]: param.qvec = np.array([1., 0., 0., 0.])

In [6]: print(param.qvec)
[1. 0. 0. 0.]
```

ahojnnes (approved):

PR #3068, opened 2024-12-21
Add ALIKED/LightGlue feature based on libtorch

ahojnnes:


behnamasadi:
Does this mean no need for SIFT anymore?

src/colmap/controllers/feature_matching_utils.cc
```diff
@@ -72,12 +69,18 @@ void FeatureMatcherWorker::Run() {
...
   }
 
-  matching_options_.cpu_descriptor_index_cache =
-      &cache_->GetFeatureDescriptorIndexCache();
-  THROW_CHECK_NOTNULL(matching_options_.cpu_descriptor_index_cache);
+  if (matching_options_.type == FeatureMatcherType::SIFT) {
+    // TODO(jsch): This is a bit ugly, but currently cannot think of a better
+    // way to inject the shared descriptor index cache.
+    SiftMatchingOptions& sift_matching_options =
+        reinterpret_cast<SiftMatchingOptions&>(matching_options_);
```
yimingc, line 76: Should be reinterpret_cast<SiftMatchingOptions&>(matching_options_.sift)?

ahojnnes: Yes, good catch.

ahojnnes:
@MrNeRF We found two necessary fixes that you may want to also apply to your upstream version: https://github.com/colmap/colmap/pull/3068/commits/a6c360a33b6255682fef4778990311b6d0cb6c18 and https://github.com/colmap/colmap/pull/3068/commits/1bd573c55fb4a44112453e2ec99b804addce7d9d

MrNeRF:
Oh cool! Thanks for pinging me.

ahojnnes:
@MrNeRF Do you have an example for how you converted the original lightglue .pth files to your .pt files? I am trying to also convert the SIFT model but cannot get it to be in the same format. Thanks.

MrNeRF:
> > @MrNeRF Do you have an example for how you converted the original lightglue .pth files to your .pt files? I am trying to also convert the SIFT model but cannot get it to be in the same format. Thanks.
> 
> Try
> 
> ```python
> from collections import OrderedDict
> import torch
> 
> modelname = "sift_lightglue"
> filename = modelname + ".pth"
> checkpoint = torch.load(filename, map_location="cpu")
> 
> # Convert to a plain dict
> if isinstance(checkpoint, OrderedDict):
>     checkpoint = dict(checkpoint)
> 
> print(type(checkpoint))  # Now <class 'dict'>
> 
> # Save the result as a standard dict
> torch.save(checkpoint, modelname + ".pt")
> ```

Yeah! Beside heavily using the C++ debugger and asserts, I didn't do anything smart or fancy. Basically just this.

src/colmap/feature/aliked.h
```diff
@@ -0,0 +1,79 @@
...
+
+  // Maximum number of features to detect (in combination with score threshold).
+  int max_num_features = 4096;
+
+  // Minimum detection score threshold.
+  double score_threshold = 0.2;
+
+  // Detect fixed number of top k features independent of detection score.
+  // Ignored if negative.
+  int top_k = -1;
```
sarlinpe, line 52: Isn't this redundant with `max_num_features`? wouldn't `bool force_extract_max_num_features = false` be sufficient?

src/colmap/feature/aliked.cc
```diff
@@ -0,0 +1,257 @@
...
+      const double scale =
+          static_cast<double>(options_.aliked->max_image_size) /
+          static_cast<double>(image_size);
+      maybe_bitmap_copy->Rescale(scale * orig_width, scale * orig_height);
+    }
+
+    const int width = bitmap_ptr->Width();
+    const int height = bitmap_ptr->Height();
+    auto row_major_array = bitmap_ptr->ConvertToRowMajorArray();
+    // Clone to ensure ownership
```
sarlinpe, line 89: Nit: add to the line that calls `clone()`

src/colmap/feature/aliked.cc
```diff
@@ -0,0 +1,257 @@
...
+    const torch::Tensor torch_image =
+        torch::from_blob(
+            row_major_array.data(), {height, width, 3}, torch::kUInt8)
+            .clone()
```
sarlinpe, line 93: Is cloning needed given that casting to float creates a copy?

src/colmap/feature/lightglue.cc
```diff
@@ -0,0 +1,262 @@
...
+            .clone();
+    // Normalize to [-1, 1].
+    torch_keypoints.index({torch::indexing::Ellipsis, 0}) =
+        (2.0f * torch_keypoints.index({torch::indexing::Ellipsis, 0}) /
+         image.width) -
+        1.0f;
+    torch_keypoints.index({torch::indexing::Ellipsis, 1}) =
+        (2.0f * torch_keypoints.index({torch::indexing::Ellipsis, 1}) /
+         image.height) -
+        1.0f;
```
sarlinpe, line 193: Isn't this already done in `LightGlue::forward` by passing the image size as input?

sarlinpe (commented):
What a massive amount of work, thank you!

src/colmap/feature/aliked.cc
```diff
@@ -0,0 +1,257 @@
...
+    sim = torch::where(sim < options_.aliked->min_similarity, 0, sim);
+    const auto nn12 = torch::argmax(sim, /*axis=*/1).contiguous().cpu();
+    const int64_t* nn12_data = nn12.data_ptr<int64_t>();
+    const auto nn21 = torch::argmax(sim, /*axis=*/0).contiguous().cpu();
+    const int64_t* nn21_data = nn21.data_ptr<int64_t>();
+
+    matches->reserve(num_keypoints1);
+    for (int i1 = 0; i1 < num_keypoints1; ++i1) {
+      const int i2 = nn12_data[i1];
+      if (i1 == nn21_data[i2]) {
```
B1ueber2y, line 176: I believe we also need to check against the similarity here, to avoid the match of i1 == i2 == 0.

src/colmap/feature/lightglue.cc
```diff
@@ -0,0 +1,262 @@
...
+        FeatureMatch match;
+        match.point2D_idx1 = i;
+        match.point2D_idx2 = j;
+        THROW_CHECK_LT(match.point2D_idx2, num_keypoints2);
+        matches->push_back(match);
+      }
+    }
+  }
+
+  void MatchGuided(double max_error,
```
B1ueber2y, line 130: Could we add some comments in the virtual class for ``Match`` and ``MatchGuided``. From my understanding the current intention of ``MatchGuided`` is to make use of the two view geometry to do guided matching and store the matches in inlier matches? Why not output a ``FeatureMatches`` instead? 

On the other hand, since LightGlue does not benefit from guided matching, we should raise a warning on the call of this method. 

I also did not find any usage for this ``MatchGuided`` method throughout COLMAP.

B1ueber2y (commented):
Thank you for the amazing efforts!

xiemeilong:
It seems [LiftFeat](https://github.com/lyp-deeplearning/LiftFeat) is approximately twice as fast as ALIKED and also achieves higher accuracy. It's also designed to handle challenging scenarios such as drastic lighting changes, low-texture regions, and repetitive patterns. Maybe consider switching from ALIKED to LiftFeat or adding both?

xiemeilong:
Colmap ALIKED + Lightglue produces worse result than hloc.

HLOC ALIKED + Lightglue with default parameters
image

colmap ALIKED + Lightglue with default parameters（not all images registered）
image

colmap ALIKED + Lightglue with   --ALIKEDExtraction.max_image_size 1024  --ALIKEDMatching.min_similarity 0.1
image

ahojnnes:
@xiemeilong Thanks for the feedback. There are still some known issues related to torch under Windows that I cannot root cause. I am not sure if these hint at some larger issues also on other platforms. On which platform did you run this comparison? Unrelated to these issues, the current colmap implementation does not auto-rotate the images to be upright (e.g., based on EXIF image orientation information). ALIKED/LightGlue only work reliably up to ~45deg rotation change between images. Can it be that your images are inconsistently rotated?

xiemeilong:
@ahojnnes I tested this on Linux, and all images orientations are same. I found that LightGlue doesn't work well with aliked-n32 in my tests,  hloc uses aliked-n16 by default.

Issue #3070, opened 2024-12-22
pycolmap: Point2D.point3D_id cannot be assigned


ahojnnes:
```
>>> import pycolmap
>>> p = pycolmap.Point2D()
>>> p.point3D_id = -1
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
TypeError: (): incompatible function arguments. The following argument types are supported:
    1. (self: pycolmap._core.Point2D, arg0: int) -> None

Invoked with: Point2D(xy=[0, 0], point3D_id=-1), -1
```

B1ueber2y:
This was a long-standing issue not only for point2d but also for ``camera.camera_id``, ``image.image_id``, etc. The reason is that Pybind11 does not support auto casting from int to uint. Yes we should fix it.

B1ueber2y:
A temporary workaround could be:
```
import pycolmap
import ctypes
p = pycolmap.Point2D()
p.point3D_id = ctypes.c_uint32(-1).value
```

B1ueber2y:
> A temporary workaround could be:
> 
> ```
> import pycolmap
> import ctypes
> p = pycolmap.Point2D()
> p.point3D_id = ctypes.c_uint32(-1).value
> ```

Update: this is not correct, as point3D_id is uint64_t rather than uint32_t. However,  ``p.point3D_id = ctypes.c_uint64(-1).value`` does not seem to work.

B1ueber2y:
Another update (sorry): something messed up in my Python environment. After fixing it ``p.point3D_id = ctypes.c_uint64(-1).value`` does work for me.

PR #3071, opened 2024-12-22
Overload pybind11 type_caster from int to C++ uint types to support -1 as input in pycolmap

B1ueber2y:
Fixes: https://github.com/colmap/colmap/issues/3070

### DIFF
--- a/src/pycolmap/pybind11_extension.h
+++ b/src/pycolmap/pybind11_extension.h
@@ -13,9 +13,52 @@
 #include <pybind11/stl_bind.h>
 
 using namespace colmap;
+namespace py = pybind11;
 
 namespace PYBIND11_NAMESPACE {
 namespace detail {
+
+#include <limits>
+#include <stdexcept>
+
+#include <pybind11/pybind11.h>
+#include <pybind11/pytypes.h>
+
+// Custom type_caster from int to C++ uint types to support -1 as input.
+// Necessary for pycolmap id types that use -1 as the invalid value.
+#define TYPE_CASTER_UINT_T(DTYPE)                                           \
+  template <>                                                               \
+  struct type_caster<DTYPE> {                                               \
+   public:                                                                  \
+    PYBIND11_TYPE_CASTER(DTYPE, _(#DTYPE));                                 \
+    bool load(handle src, bool) {                                           \
+      if (!py::isinstance<py::int_>(src)) {                                 \
+        return false;                                                       \
+      }                                                                     \
+      int64_t val = py::cast<int64_t>(src);                                 \
+      if (val == -1) {                                                      \
+        value = std::numeric_limits<DTYPE>::max();                          \
+        return true;                                                        \
+      }                                                                     \
+      if (val < 0 || val > std::numeric_limits<DTYPE>::max()) {             \
+        LOG(WARNING) << "Value is out of range for uint types. Only -1 or " \
+                        "positive values are supported";                    \
+        return false;                                                       \
+      }                                                                     \
+      value = static_cast<DTYPE>(val);                                      \
+      return true;                                                          \
+    }                                                                       \
+    static handle cast(const DTYPE& src, return_value_policy, handle) {     \
+      return py::int_(src).release();                                       \
+    }                                                                       \
+  };
+
+TYPE_CASTER_UINT_T(uint8_t);
+TYPE_CASTER_UINT_T(uint16_t);
+TYPE_CASTER_UINT_T(uint32_t);
+TYPE_CASTER_UINT_T(uint64_t);
+
+#undef TYPE_CASTER_UINT_T
 
 // Bind COLMAP's backport implementation of std::span. This copies the content
 // into a list. We could instead create a view with an Eigen::Map but the cast
--- a/src/pycolmap/estimators/cost_functions.cc
+++ b/src/pycolmap/estimators/cost_functions.cc
@@ -3,6 +3,7 @@
 #include "colmap/geometry/rigid3.h"
 
 #include "pycolmap/helpers.h"
+#include "pycolmap/pybind11_extension.h"
 
 #include <pybind11/eigen.h>
 #include <pybind11/pybind11.h>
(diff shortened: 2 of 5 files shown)


B1ueber2y:
This update only supports -1 and positive values and add explicit logs. Example: 
```
In [1]: import pycolmap

In [2]: p = pycolmap.Point2D()

In [3]: p.point3D_id = -2
W20241222 14:41:54.504066 1408914 pybind11_extension.h:59] Value is out of range for uint types. Only -1 or positive values are supported
---------------------------------------------------------------------------
TypeError                                 Traceback (most recent call last)
Cell In[3], line 1
----> 1 p.point3D_id = -2

TypeError: (): incompatible function arguments. The following argument types are supported:
    1. (self: pycolmap._core.Point2D, arg0: uint64_t) -> None

Invoked with: Point2D(xy=[0, 0], point3D_id=-1), -2

In [4]: p.point3D_id = -1

In [5]:
```

B1ueber2y:
The stub seems to break after this update since it cannot parse uint32_t and uint64_t: https://github.com/colmap/colmap/actions/runs/12454820180/job/34766524324?pr=3071. No idea on this since I never used this feature. Help needed : )

ahojnnes (approved):
Great, thanks very much.

src/pycolmap/pybind11_extension.h
```diff
@@ -13,10 +13,53 @@
...
+  template <>                                                             \
+  struct type_caster<DTYPE> {                                             \
+   public:                                                                \
+    PYBIND11_TYPE_CASTER(DTYPE, _(#DTYPE));                               \
+    bool load(handle src, bool) {                                         \
+      if (!py::isinstance<py::int_>(src)) {                               \
+        return false;                                                     \
+      }                                                                   \
+      int64_t val = py::cast<int64_t>(src);                               \
+      if (val == -1) {                                                    \
```
sarlinpe, line 39: I don't like this approach because it will accept -1 as valid input for any function that accepts uint\*_t, while this is valid only for ID setters. Instead, how about wrapping `Set{Image,Camera,...}Id` with a templated function that accepts an int and casts appropriately?

sarlinpe (changes requested):

src/pycolmap/pybind11_extension.h
B1ueber2y: The use of pycolmap ids actually happen in a lot of places in the bindings (not only for the setters). For example, one can do ``grep -r '_id' ./src/pycolmap/`` and find a lot of relevant bindings. And the logic is also not consistent (e.g., here there is explicit casting supported for only one method: https://github.com/colmap/colmap/blob/main/src/pycolmap/sfm/incremental_mapper.cc#L358).

Actually I have a different proposal: how about we not supporting -1 as input and always recommend users to use ctypes to convert themselves from the Python side, as mentioned in the issue:
```
import pycolmap
import ctypes
p = pycolmap.Point2D()
p.point3D_id = ctypes.c_uint32(-1).value
```
We could update the logs here to inform the users who use -1 as input: https://github.com/B1ueber2y/colmap/blob/fix/pycolmap_uint/src/pycolmap/pybind11_extension.h#L44.

B1ueber2y: For example, something like: https://github.com/B1ueber2y/colmap/commit/1639ddb4b3b620386f822b1618f15bd3476d1a39

B1ueber2y: > The use of pycolmap ids actually happen in a lot of places in the bindings (not only for the setters). For example, one can do `grep -r '_id' ./src/pycolmap/` and find a lot of relevant bindings. And the logic is also not consistent (e.g., here there is explicit casting supported for only one method: https://github.com/colmap/colmap/blob/main/src/pycolmap/sfm/incremental_mapper.cc#L358).
> 
> Actually I have a different proposal: how about we not supporting -1 as input and always recommend users to use ctypes to convert from the Python side themselves, as the solution I suggested in the issue:
> 
> ```
> import pycolmap
> import ctypes
> p = pycolmap.Point2D()
> p.point3D_id = ctypes.c_uint32(-1).value
> ```
> 
> We could update the error logs here to inform the users who use -1 as input: https://github.com/B1ueber2y/colmap/blob/fix/pycolmap_uint/src/pycolmap/pybind11_extension.h#L44.

Update: this is not the right approach, since point3D_id is uint64_t, and ``p.point3D_id = ctypes.c_uint32(-1).value`` does not work with the following error: 
```
RuntimeError: Unable to cast Python instance of type <class 'int'> to C++ type '?' (#define PYBIND11_DETAILED_ERROR_MESSAGES or compile in debug mode for details)
```
@sarlinpe I agree that wrapping all functions with explicit appropriate casting is the best approach, though it will require going through all the relevant functions to fully fix the issue. Feel free to take this feature and go ahead : )

B1ueber2y: Another idea is to support type_caster for ``ctypes.c_ulong`` and ``c_types.c_uint`` and always suggest the users to do: ``p.point3D_id = ctypes.c_uint64(-1)``. This may be a cleaner alternative.

B1ueber2y:
The fix in this PR actually limits the range of uint64_t from 2^64 - 1 to 2^63 - 1 due to the int64_t in the middle. This will also be the case for explicit casting with int as input. So I suggest we not go ahead with the update and stick with ``p.point3D_id = ctypes.c_uint64(-1).value``. We might want to mention it somewhere in the documentation.

ahojnnes:
How about an explicit INVALID_POINT3D_ID Python/C++ type?

sarlinpe:
Using ctypes requires knowing the C++ type of the attribute (utin32 vs uint64), which is not reflected in the Python doc (int only) and cannot be asked users to know. Binding the invalid values seems like a better option, since we can then provide a nicer representation (e.g. `pycolmap.invalid_image_id`) to some default values, e.g. in `Image.__init__` (which for now is displayed as 4294967295, [see doc](https://colmap.github.io/pycolmap/pycolmap.html))
https://github.com/colmap/colmap/blob/cd97b6501178291f594274cf8741031ecf482775/src/colmap/util/types.h#L98-L103
https://github.com/colmap/colmap/blob/cd97b6501178291f594274cf8741031ecf482775/src/pycolmap/scene/image.cc#L46-L51

PR #3072, opened 2024-12-23
Add binding support for invalid values in pycolmap id types

B1ueber2y:
Fixes: https://github.com/colmap/colmap/issues/3070

Per discussion from https://github.com/colmap/colmap/pull/3071

### DIFF
--- a/src/colmap/feature/types.h
+++ b/src/colmap/feature/types.h
@@ -84,15 +84,15 @@ typedef Eigen::Matrix<float, Eigen::Dynamic, Eigen::Dynamic, Eigen::RowMajor>
 
 struct FeatureMatch {
   FeatureMatch()
-      : point2D_idx1(kInvalidPoint2DIdx), point2D_idx2(kInvalidPoint2DIdx) {}
+      : point2D_idx1(kInvalidPoint2DId), point2D_idx2(kInvalidPoint2DId) {}
   FeatureMatch(const point2D_t point2D_idx1, const point2D_t point2D_idx2)
       : point2D_idx1(point2D_idx1), point2D_idx2(point2D_idx2) {}
 
   // Feature index in first image.
-  point2D_t point2D_idx1 = kInvalidPoint2DIdx;
+  point2D_t point2D_idx1 = kInvalidPoint2DId;
 
   // Feature index in second image.
-  point2D_t point2D_idx2 = kInvalidPoint2DIdx;
+  point2D_t point2D_idx2 = kInvalidPoint2DId;
 };
 
 typedef std::vector<FeatureMatch> FeatureMatches;
--- a/src/colmap/feature/types.h
+++ b/src/colmap/feature/types.h
@@ -84,15 +84,15 @@ typedef Eigen::Matrix<float, Eigen::Dynamic, Eigen::Dynamic, Eigen::RowMajor>
 
 struct FeatureMatch {
   FeatureMatch()
-      : point2D_idx1(kInvalidPoint2DId), point2D_idx2(kInvalidPoint2DId) {}
+      : point2D_idx1(kInvalidPoint2DIdx), point2D_idx2(kInvalidPoint2DIdx) {}
   FeatureMatch(const point2D_t point2D_idx1, const point2D_t point2D_idx2)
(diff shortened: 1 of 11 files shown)


B1ueber2y:
Example:
```
import pycolmap
p = pycolmap.Point2D()
p.point3D_id = pycolmap.invalid_point3D_id
```

src/colmap/feature/types.h
```diff
@@ -84,15 +84,15 @@ typedef Eigen::Matrix<float, Eigen::Dynamic, Eigen::Dynamic, Eigen::RowMajor>
 
 struct FeatureMatch {
   FeatureMatch()
-      : point2D_idx1(kInvalidPoint2DIdx), point2D_idx2(kInvalidPoint2DIdx) {}
+      : point2D_idx1(kInvalidPoint2DId), point2D_idx2(kInvalidPoint2DId) {}
   FeatureMatch(const point2D_t point2D_idx1, const point2D_t point2D_idx2)
       : point2D_idx1(point2D_idx1), point2D_idx2(point2D_idx2) {}
 
   // Feature index in first image.
-  point2D_t point2D_idx1 = kInvalidPoint2DIdx;
```
sarlinpe, line 92: 1. I think that it should really be an index and not an ID since 2D points are stored in an ordered list while image/camera/points3D IDs are not guaranteed to be consecutive.
2. This would unnecessarily break many dependencies (incl. GLOMAP) so IMHO it's not worth it.
3. Even then, this would need to be changed in a separate PR.

B1ueber2y: Thanks. This makes sense. I have reverted it.

B1ueber2y:
On the other hand, I have made several attempts to improve the doc rendering of Image class (e.g., https://github.com/B1ueber2y/colmap/commit/e205e62e03190ef73894448e11e7793b476bcbdd). But no luck so far.

ahojnnes:
Thanks. Nit: For consistency across Python, I would suggest calling it `INVALID_POINT3D_ID`. It could also make sense to adjust the str `__repr__` to print out `kInvalidPoint3DId` or `INVALID_POINT3D_ID` instead of -1?

ahojnnes (approved):

B1ueber2y:
> Thanks. Nit: For consistency across Python, I would suggest calling it `INVALID_POINT3D_ID` etc. It could also make sense to adjust the str `__repr__` to print out `kInvalidPoint3DId` or `INVALID_POINT3D_ID` instead of -1?

Thanks. I have updated the names to upper cases. Now the ``__repr__`` is controlled in the C++ side after the recent update: https://github.com/colmap/colmap/blob/main/src/colmap/scene/image.cc#L107 (while the one for point2D is missing still), so the pybind part does not help.

sarlinpe:
> https://github.com/colmap/colmap/blob/main/src/colmap/scene/image.cc#L107

You need to replace
```c++
"camera_id"_a = kInvalidCameraId,
```
by
```c++
py::arg_v("camera_id", kInvalidCameraId, "INVALID_CAMERA_ID"),
```
and so on.

B1ueber2y:
> > On the other hand, I have made several attempts to improve the doc rendering of Image class (e.g., [B1ueber2y@e205e62](https://github.com/B1ueber2y/colmap/commit/e205e62e03190ef73894448e11e7793b476bcbdd)). But no luck so far.
> 
> You need to replace
> 
> ```c++
> "camera_id"_a = kInvalidCameraId,
> ```
> 
> by
> 
> ```c++
> py::arg_v("camera_id", kInvalidCameraId, "pycolmap.INVALID_CAMERA_ID"),
> ```
> 
> and so on.

Thanks. Updated.

sarlinpe (approved):
Thanks, can you also update this one? 
https://github.com/colmap/colmap/blob/cd97b6501178291f594274cf8741031ecf482775/src/pycolmap/scene/point2d.cc#L28

B1ueber2y:
> Thanks, can you also update this one?
> 
> https://github.com/colmap/colmap/blob/cd97b6501178291f594274cf8741031ecf482775/src/pycolmap/scene/point2d.cc#L28

I think this was already updated?

sarlinpe:
I missed it, sorry.

PR #3075, opened 2024-12-25
Only create OpenGL context in automatic reconstruction if necessary

ahojnnes:
Only creates OpenGL context if feature extraction and matching stages are executed on the GPU. Previously, we would always create a context, even if not necessary.

### DIFF
--- a/src/colmap/controllers/automatic_reconstruction.cc
+++ b/src/colmap/controllers/automatic_reconstruction.cc
@@ -109,35 +109,39 @@ AutomaticReconstructionController::AutomaticReconstructionController(
   option_manager_.mapper->ba_gpu_index = options_.gpu_index;
   option_manager_.bundle_adjustment->gpu_index = options_.gpu_index;
 
-  feature_extractor_ = CreateFeatureExtractorController(
-      reader_options, *option_manager_.sift_extraction);
-
-  exhaustive_matcher_ =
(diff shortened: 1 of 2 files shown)


src/colmap/controllers/automatic_reconstruction.cc
```diff
@@ -109,35 +109,39 @@ AutomaticReconstructionController::AutomaticReconstructionController(
...
-      CreateExhaustiveFeatureMatcher(*option_manager_.exhaustive_matching,
-                                     *option_manager_.sift_matching,
-                                     *option_manager_.two_view_geometry,
-                                     *option_manager_.database_path);
-
-  if (!options_.vocab_tree_path.empty()) {
-    option_manager_.sequential_matching->loop_detection = true;
-    option_manager_.sequential_matching->vocab_tree_path =
-        options_.vocab_tree_path;
+  if (options_.extraction) {
```
B1ueber2y, line 112: Should the setter of ``option_manager_.sift_extraction->use_gpu`` and ``gpu_index`` be moved inside the if branch? Also for the sift_matching

B1ueber2y (approved):

src/colmap/controllers/automatic_reconstruction.cc
ahojnnes: I would keep the options grouped as is above, so it becomes easier to keep it in sync and not forget the piping in future. The overhead of copying the parameter even when it is not used is probably negligible.

PR #3082, opened 2024-12-26
Add back detailed logs for covariance estimation

B1ueber2y:
In particular, the size of the problem in the log is very useful to estimate how long the estimation will take.


src/colmap/estimators/covariance.cc
```diff
@@ -95,7 +95,8 @@ bool ComputeSchurComplement(
     point_covs.reserve(points.size());
   }
 
-  VLOG(2) << "Schur elimination of point parameters";
+  VLOG(2) << StringPrintf("Schur elimination of point parameters (n = %d)",
```
ahojnnes, line 98: Nit: I suggest directly just using string streams to compose the strings instead of the extra hop through StringPrintf

ahojnnes (approved):

src/colmap/estimators/covariance.cc
B1ueber2y: Thanks. This is done.

PR #3084, opened 2024-12-26
Temporarily disable ccache in the pycolmap macOS CI

sarlinpe:


ahojnnes:
We could potentially pin the ccache version of supported by homebrew?

sarlinpe:
The [ccache formula](https://github.com/Homebrew/homebrew-core/blob/master/Formula/c/ccache.rb) hasn't been updated recently, so the issue is with one of the dependencies (most likely [fmt](https://github.com/Homebrew/homebrew-core/blob/master/Formula/f/fmt.rb)). I can reproduce this locally. I suggest turning off ccache until the issue is resolved upstream.

sarlinpe:
I however don't know why this does not affect the our regular non-pycolmap macOS CI...

sarlinpe:
The pycolmap CI pulls the latest fmt 11.1.0 (which likely leads to the segfault, maybe because the ccache binaries haven't been updated accordingly yet) while the regular colmap CI pulls fmt 11.0.2 (maybe because some other packages pin it).

ahojnnes (approved):

PR #3099, opened 2025-01-04
Fix download functionality under Windows

ahojnnes:
Fixes two issues:
* `curl[openssl]` does not pick up certificates from the system automatically. Now, we instead use Windows native schannel for this purpose.
* `OpenSSL::Crypto` leads to segfaults under Windows/ARM64. Here, we switch to CryptoPP as an alternative.


cmake/FindDependencies.cmake
```diff
@@ -88,14 +88,35 @@ if(CGAL_FOUND)
...
+    # The OpenSSL package in vcpkg seems broken under Windows and leads to
+    # missing certificate verification when connecting to SSL servers. We
+    # therefore use curl[schannel] (i.e., native Windows SSL/TLS) under Windows
+    # and curl[openssl] otherwise.
     find_package(CURL QUIET)
-    find_package(OpenSSL QUIET)
-    if(CURL_FOUND AND OpenSSL_FOUND)
+    set(CRYPTO_FOUND FALSE)
+    if(IS_MSVC AND IS_ARM64)
+        # OpenSSL crashes for ARM64 under Windows. We therefore fall back to
```
sarlinpe, line 98: Is this a temporary issue that we can expect to see fixed soon? If not, why using OpenSSL at all instead of Crypto++ on all platforms? Is it because Crypto++ is more difficult to install? It's unideal to have such complex dependency logic for such a small feature.

sarlinpe (approved):

cmake/FindDependencies.cmake
ahojnnes: I would expect this to be fixed eventually. The reason for not defaulting to CryptoPP is that I wanted to keep dependencies minimal on all other platforms. I expect essentially nobody to use colmap under Windows ARM except myself in a VM on my MacBook for debugging.

PR #3106, opened 2025-01-10
Fix reprojection error in camera rig cost function

binbin-xu:
Reorders arguments in BindCostFunctions to fix reprojection error in camera rig cost function. The parameters cam_from_rig and point2D were in incorrect order after a previous refactoring.


B1ueber2y (approved):
Thanks for the fix!

B1ueber2y:
Linking the PR where this comes from: https://github.com/colmap/colmap/pull/2880

PR #3125, opened 2025-01-25
Use poselib for generalized absolute pose minimal solver

ahojnnes:
The solver is faster and more stable. In preparation for upcoming rig support.


src/colmap/estimators/generalized_absolute_pose_test.cc
```diff
@@ -112,7 +111,7 @@ TEST(GeneralizedAbsolutePose, Estimate) {
       ransac.estimator.Residuals(
           points2D, points3D_faulty, report.model, &residuals);
       for (size_t i = 0; i < residuals.size(); ++i) {
-        EXPECT_GT(residuals[i], 1e-10);
```
B1ueber2y, line 115: This points3D_faulty seems not to be used in the whole estimation pipeline, and the input data to the ransac estimator in this test is noise-free. I am a bit confused here. Was it on purpose?

B1ueber2y (approved):

sarlinpe (approved):

src/colmap/estimators/generalized_absolute_pose_test.cc
ahojnnes: I added an explanation for why RANSAC is used but otherwise also improved the tests a bit.

PR #3126, opened 2025-01-26
Add code coverage reporting

ahojnnes:
image


sarlinpe (approved):
LGTM. As a follow up, one could post a summary of the coverage as PR comment using https://github.com/marocchino/sticky-pull-request-comment There are GH actions that handle the conversion from XML to Markdown, e.g. https://github.com/marketplace/actions/code-coverage-summary

ahojnnes:
@sarlinpe The idea about reporting to the PR is a nice one. However, it's not too useful as implemented by the github actions, because it doesn't perform a diff on the changed files and report the coverage specifically for those. WDYT?

sarlinpe:
Agree, diffs would be more informative but this requires knowing the coverage at head - where would we get this? fetching the artifact from the latest CI run on head? Sounds complicated.

Issue #3111, opened 2025-01-15
Center principal point in undistorter


Peamon:
**Is your feature request related to a problem? Please describe.**
To create gaussian splatting with a colmap result i need to center the principal points of the image but there is no option in the image_undistorter function.

**Describe the solution you'd like**
An boolean option to center the principal point of the image like it's possible in [opencv getOptimalNewCameraMatrix function](https://docs.opencv.org/4.x/d9/d0c/group__calib3d.html#ga7a6c4e032c97f03ba747966e6ad862b1).

**Describe alternatives you've considered**
The alternative is to do colmap mapper once to estimate intrinsics parameters, then undistord the image with opencv, then run colmap a second time. but it cost a lot of time to run colmap twice ...

ahojnnes:
This is a reasonable feature request. It could be better to address the issue on the gaussian splatting side though. In the meanwhile, there is an easier workaround: Run the colmap pipeline including undistortion and then post-process the images by cropping/padding the image by the delta between principal point and image center.

Peamon:
Thanks for the workaround tips.

StonerLing:
I met the same problem when work with odm, maybe this works: add a new camera model with (cx,cy) centered or simply modify the code in `UndistortCamera` like 
```cpp
Camera UndistortCamera(const UndistortCameraOptions& options,
                       const Camera& camera) {
  Camera undistorted_camera;
  undistorted_camera.SetModelId(PinholeCameraModel::model_id);
  undistorted_camera.SetWidth(camera.Width());
  undistorted_camera.SetHeight(camera.Height());

  // Copy focal length parameters.
  const std::vector<size_t>& focal_length_idxs = camera.FocalLengthIdxs();
  CHECK_LE(focal_length_idxs.size(), 2)
      << "Not more than two focal length parameters supported.";
  if (focal_length_idxs.size() == 1) {
    undistorted_camera.SetFocalLength(camera.FocalLength());
  } else if (focal_length_idxs.size() == 2) {
    undistorted_camera.SetFocalLengthX(camera.FocalLengthX());
    undistorted_camera.SetFocalLengthY(camera.FocalLengthY());
  }

  double target_principal_x = static_cast<double>(camera.Width()) / 2;
  double target_principal_y = static_cast<double>(camera.Height()) / 2;
  // Copy principal point parameters.
  undistorted_camera.SetPrincipalPointX(target_principal_x);
  undistorted_camera.SetPrincipalPointY(target_principal_y);
......
[...]

PR #3141, opened 2025-02-06
Add option to center principal point

ahojnnes:
Fixes: https://github.com/colmap/colmap/issues/3111

### DIFF
--- a/src/colmap/image/undistortion.cc
+++ b/src/colmap/image/undistortion.cc
@@ -802,8 +802,13 @@ Camera UndistortCamera(const UndistortCameraOptions& options,
   undistorted_camera.SetFocalLengthY(camera.FocalLengthY());
 
   // Copy principal point parameters.
-  undistorted_camera.SetPrincipalPointX(camera.PrincipalPointX());
-  undistorted_camera.SetPrincipalPointY(camera.PrincipalPointY());
+  if (options.center_principal_point) {
+    undistorted_camera.SetPrincipalPointX(0.5 * camera.width);
+    undistorted_camera.SetPrincipalPointY(0.5 * camera.height);
+  } else {
+    undistorted_camera.SetPrincipalPointX(camera.PrincipalPointX());
+    undistorted_camera.SetPrincipalPointY(camera.PrincipalPointY());
+  }
 
   // Modify undistorted camera parameters based on ROI if enabled
   size_t roi_min_x = 0;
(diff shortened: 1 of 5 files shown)


src/colmap/image/undistortion.cc
```diff
@@ -802,8 +802,13 @@ Camera UndistortCamera(const UndistortCameraOptions& options,
   undistorted_camera.SetFocalLengthY(camera.FocalLengthY());
 
   // Copy principal point parameters.
-  undistorted_camera.SetPrincipalPointX(camera.PrincipalPointX());
-  undistorted_camera.SetPrincipalPointY(camera.PrincipalPointY());
+  if (options.center_principal_point) {
+    undistorted_camera.SetPrincipalPointX(0.5 * camera.width);
```
B1ueber2y, line 806: This does not feel right, since the width and height will very likely be updated after the COLMAP undistortion?

ahojnnes: You are right. This cannot be combined with some of the other options. It's possible to make it happen but may be a bit complicated. I suggest that we throw an exception if this is combined with a rescaling of the image.

B1ueber2y: I believe that with radial distortion the image size will be changed by default, so this is not a corner case. Maybe we could fix the resolution when this option is enabled but this will result in a potential loss of FoV.

ahojnnes: PTAL

src/colmap/image/undistortion.cc
```diff
@@ -955,6 +950,11 @@ Camera UndistortCamera(const UndistortCameraOptions& options,
     }
   }
 
+  if (options.center_principal_point) {
+    undistorted_camera.SetPrincipalPointX(0.5 * undistorted_camera.width);
```
B1ueber2y, line 954: Thanks. But will this be erroneous when it comes to updating the ``point2D.xy`` at ``UndistortReconstruction``, as the principle points before and after the undistortion do not correspond to each other?

B1ueber2y: e.g. https://github.com/colmap/colmap/blob/main/src/colmap/image/undistortion.cc#L993

B1ueber2y: Yes I agree that this should work mathematically. Images will also have black region after warping if the principal point is far from the center, so this is not ideal as it means that we also will potentially lose information (pixels that map outside the image).

However, I feel that the alternative approach of cropping/increasing the size does not belong to the UndistortCamera function and should be done at the user side instead.

PR #3143, opened 2025-02-06
Add Rig serialization support to reconstruction+database

ahojnnes:


src/colmap/scene/reconstruction_io.cc
```diff
@@ -423,6 +493,46 @@ void ReadPoints3DBinary(Reconstruction& reconstruction,
...
+
+  // Ensure that we don't loose any precision by storing in text.
+  stream.precision(17);
+
+  stream << "# Rig calib list with one line of data per calib:" << std::endl;
+  stream << "#   RIG_ID, NUM_SENSORS, SENSORS[] as (SENSOR_TYPE, SENSOR_ID, QW, QX, QY, QZ, TX, TY, TZ)" << std::endl;
+  stream << "# Number of rig calibs: " << reconstruction.NumCameras() << std::endl;
+
+  for (const camera_t camera_id : ExtractSortedIds(reconstruction.Cameras())) {
+    const Camera& camera = reconstruction.Camera(camera_id);
```
B1ueber2y, line 508: This code block seems not to be updated? Maybe its because of the WIP : )

src/colmap/scene/synthetic.cc
```diff
@@ -183,6 +191,33 @@ void SynthesizeDataset(const SyntheticDatasetOptions& options,
     camera_ids[camera_idx] = camera_id;
     camera.camera_id = camera_id;
     reconstruction->AddCamera(std::move(camera));
+
+    sensor_t sensor_id(SensorType::CAMERA, camera_id);
+    const rig_t rig_id = camera_idx % options.num_rigs;
```
B1ueber2y, line 196: This does not seem to support options.num_rig == 0. This means that we wont support synthesizing datasets without rigs anymore. 

Also, this assumes that all the images belong to some rigs, and if num_rigs == 5, cameras with ids 1, 6, 11, 16, 21, 26, ...101 will belong the same rig. This is counterintuitive to real practice, where one would have cameras in the same rig stored adjacently.

src/colmap/scene/synthetic.cc
```diff
@@ -183,6 +191,33 @@ void SynthesizeDataset(const SyntheticDatasetOptions& options,
...
+    Rig& rig = reconstruction->Rig(rig_id);
+    if (rig.NumSensors() == 0) {
+      rig.AddRefSensor(sensor_id);
+    } else {
+      Rigid3d sensor_from_rig;
+      if (options.rig_from_sensor_translation_stddev > 0) {
```
B1ueber2y, line 202: This should be sensor_from_rig_translation_stddev instead?

src/colmap/scene/synthetic.h
```diff
@@ -37,10 +37,13 @@
 namespace colmap {
 
 struct SyntheticDatasetOptions {
+  int num_rigs = 1;
```
B1ueber2y, line 40: I would suggest to have num_rigs equal to zero as default (all independent images).

src/colmap/scene/synthetic_test.cc
```diff
@@ -50,6 +50,12 @@ TEST(SynthesizeDataset, Nominal) {
   CreateDirIfNotExists(sparse_path);
   reconstruction.Write(sparse_path);
 
+  EXPECT_EQ(database.NumRigs(), options.num_rigs);
+  EXPECT_EQ(reconstruction.NumRigs(), options.num_rigs);
+  for (const auto& rig : reconstruction.Rigs()) {
+    EXPECT_EQ(rig.second.NumSensors(), options.num_cameras);
```
B1ueber2y, line 56: This is only true when num_rigs == 1.

src/colmap/sensor/rig.cc
```diff
@@ -27,26 +27,40 @@
...
 }
 
+std::ostream& operator<<(std::ostream& stream, const Rig& rig) {
+  const std::string rig_id_str =
+      rig.RigId() != kInvalidRigId ? std::to_string(rig.RigId()) : "Invalid";
+  stream << "Rig(rig_id=" << rig_id_str << ", ref_sensor_id=("
+         << rig.RefSensorId().id << ", " << rig.RefSensorId().type
+         << "), sensors=[";
+  for (const auto& [sensor_id, _] : rig.Sensors()) {
+    stream << "(" << sensor_id.id << ", " << sensor_id.type << "), ";
```
B1ueber2y, line 57: Should we swap the order to (type, id) as type goes first and we can have both (CAMERA, 0) and (IMU, 0) present?

src/colmap/scene/reconstruction_io.h
```diff
@@ -39,6 +39,9 @@ namespace colmap {
 
 // Note that cameras must be read before images.
 
+void ReadRigsText(Reconstruction& reconstruction, std::istream& stream);
```
B1ueber2y, line 42: We should update the I/O methods in the Reconstruction class as well (e.g. ``Reconstruction::ReadBinary``), where we will write/read rigs when there are any.

B1ueber2y:
In the current code the rigs are not used as it is not associated through frames to images. Does it make sense to first add the frames into I/O and merge together? Also, after we fully support the frames I/O some code will be unnecessary (e.g., https://github.com/colmap/colmap/blob/user/jsch/rig-calib-reconstruction/src/colmap/scene/synthetic.cc#L243-L247)

src/colmap/scene/synthetic.cc
ahojnnes: I would have expected that, in the end, we will always have one rig per camera. If it's a single camera rig, it's a trivial frame.

src/colmap/scene/synthetic.h
ahojnnes: See other comment. I would have imagined that the end state of our work will result in a 1 to many relationship between rigs and cameras. Individual cameras will have one rig with trivial frames.

src/colmap/scene/synthetic.h
```diff
@@ -37,10 +37,13 @@
...
   int num_cameras = 2;
   int num_images = 10;
   int num_points3D = 100;
 
+  double rig_from_sensor_translation_stddev = 0.05;
```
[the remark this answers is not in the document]
ahojnnes: Yes, done.

src/colmap/scene/synthetic.cc
```diff
@@ -183,6 +191,33 @@ void SynthesizeDataset(const SyntheticDatasetOptions& options,
...
+      }
+      rig.AddSensor(sensor_id, sensor_from_rig);
+    }
+  }
+
+  if (database) {
+    for (int rig_idx = 0; rig_idx < options.num_rigs; ++rig_idx) {
+      LOG(INFO) << "Rig " << rig_idx << " has "
+                << reconstruction->Rig(rig_idx).NumSensors() << " sensors";
+      database->WriteRig(reconstruction->Rig(rig_idx), /*use_rig_id=*/true);
```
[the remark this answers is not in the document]
ahojnnes: Good point. Will fix.

src/colmap/scene/synthetic.h
ahojnnes: OK, I understand the issue here now. Will make the "zero" value a special token that will use independent cameras. Thanks.

src/colmap/scene/reconstruction_io.h
ahojnnes: I wanted to wait with doing the actual serialization until we reached feature completeness together with frames?

B1ueber2y (approved):

PR #3147, opened 2025-02-08
Support unregistered images in reconstruction I/O

B1ueber2y:
To make the reconstruction I/O utility more general for all reconstruction objects, although in the sfm pipeline we only save the reconstruction with all registered images.

### DIFF
--- a/src/colmap/scene/reconstruction_io.cc
+++ b/src/colmap/scene/reconstruction_io.cc
@@ -163,7 +163,8 @@ void ReadImagesText(Reconstruction& reconstruction, std::istream& stream) {
     std::getline(line_stream1, item, ' ');
     cam_from_world.translation.z() = std::stold(item);
 
-    image.SetCamFromWorld(cam_from_world);
+    if (cam_from_world.rotation.coeffs().norm() > 0)
+      image.SetCamFromWorld(cam_from_world);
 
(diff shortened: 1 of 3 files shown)


B1ueber2y:
For an unregistered image, all zeros will be written and we check if the norm of the quaternion is zero at reading to determine whether it is a valid pose. In this way we get full backwards compatibility. I also added a test for this.

ahojnnes:
Sorry, I am a bit confused. Why do we need this again? I think we had a misunderstanding in our chat :-) I thought we can just keep the current format and I would actually be careful changing this contract. It'll break many downstream applications parsing our files.

ahojnnes:
I guess the behavior is different for other people reading our reconstructions in that before they wouldn't find non registered images and now they do.

B1ueber2y:
> I guess the behavior is different for other people reading our reconstructions in that before they wouldn't find non registered images and now they do. I would prefer we keep the current contract and only store registered images.
> 
> To be fully backwards compatible, I would continue serializing the full cam from world transformation in the images.bin file. It would be redundant to sensor from rig from world stored in rigs.bin and frames.bin. Would this work?

Yes. This would mean that when the rig is available, we would use ``ComposeCamFromWorld`` when writing images.bin?

PR #3155, opened 2025-02-12
Fix covariance propagation of pose inverse

B1ueber2y:
The covariance propagation of pose inverse via adjoint introduced in https://github.com/colmap/colmap/pull/2598 followed the lamar convention, which was just found to be actually broken here: https://github.com/microsoft/lamar-benchmark/pull/84.

This PR fixes the propagation and added more cycle-consistent tests on the property of the adjoint matrix. To simplify the code for calculating adjoint inverse the function ``CrossProductMatrix`` is moved into ``rigid3.h``.

### DIFF
--- a/src/colmap/geometry/rigid3.cc
+++ b/src/colmap/geometry/rigid3.cc
@@ -33,6 +33,13 @@
 
 namespace colmap {
 
+Eigen::Matrix3d CrossProductMatrix(const Eigen::Vector3d& vector) {
+  Eigen::Matrix3d matrix;
+  matrix << 0, -vector(2), vector(1), vector(2), 0, -vector(0), -vector(1),
+      vector(0), 0;
+  return matrix;
+}
+
 std::ostream& operator<<(std::ostream& stream, const Rigid3d& tform) {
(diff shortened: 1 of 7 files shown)


src/colmap/geometry/rigid3.cc
```diff
@@ -33,6 +33,13 @@
 
 namespace colmap {
 
+Eigen::Matrix3d CrossProductMatrix(const Eigen::Vector3d& vector) {
```
ahojnnes, line 36: Do we have an existing test for this that needs to be moved? Otherwise, can be great to have one. Thx.

ahojnnes (approved):

src/colmap/geometry/rigid3.cc
B1ueber2y: We did one. I have moved it. Thanks for catching it!

B1ueber2y:
Edit: this PR proposal is still not correct. See update from https://github.com/microsoft/lamar-benchmark/pull/84. Will follow up another one as a fix.

PR #3159, opened 2025-02-15
[Spherical Camera Support] Change essential matrix estimation to use camera rays

ahojnnes:
This is the first in a series of PRs to support cameras with >180deg FOV (including spherical cameras). This PR should result in no behavioral change other than changes up to numerical precision.


B1ueber2y (approved):

sarlinpe (approved):

wgntn-cgs:
Does this mean that equirectangular images are now supported?

PR #3160, opened 2025-02-15
[Spherical Camera Support]  Change absolute pose estimation to use camera rays

ahojnnes:
This results in some behavioral change due to using angular errors in PNP RANSAC. To minimize behavioral change, I decided to convert the existing error thresholds in pixels to an angular error (at the center of the image). Benchmarking results on ETH3D show no significant change in performance (A: main, B: this PR) for scenes with stable initialization. The large difference in meadow is caused by random failure to initialize the scene and unrelated to this change.
```
I20250216 09:31:59.453352 990463 compare.py:main:58] Results A - B:

botanical_garden   0.01   0.00   0.00   0.00      0     0    0       0
boulders           0.02   0.01   0.00   0.00      0     0    0       0
bridge             0.01   0.00   0.00   0.00      0     0    0       0
courtyard          0.02   0.00   0.00   0.00      0     0    0       0
delivery_area      0.01   0.00   0.00   0.00      0     0    0       0
door              -0.02   0.01   0.00   0.00      0     0    0       0
electro            0.15   0.11   0.03   0.01      0     0    0       0
exhibition_hall    1.40   1.05   0.20   0.12      0     0    0       0
facade            -0.01  -0.00  -0.00  -0.00      0     0    0       0
kicker            -0.11  -0.06  -0.01  -0.01      0     0    0       0
lecture_room       0.08   0.05   0.01   0.01      0     0    0       0
living_room        0.20   0.13   0.03   0.02      0     0    0       0
lounge            -0.05  -0.02  -0.00  -0.00      0     0    0       0
meadow           -63.10 -70.82 -85.83 -88.37     -9     0   -1      -9
observatory        0.01   0.01   0.00   0.00      0     0    0       0
office            -0.15  -0.07  -0.08  -0.04      0     0    0       0
old_computer      -0.60  -0.27  -0.06  -0.03      0     0    0       0
pipes              1.25   0.81   0.16   0.08      0     0    0       0
playground        -0.05  -0.02  -0.01  -0.00      0     0    0       0
relief             0.03   0.02   0.00   0.00      0     0    0       0
relief_2          -0.13  -0.05  -0.01  -0.01      0     0    0       0
statue            -0.00  -0.00  -0.00  -0.00      0     0    0       0
terrace           -0.04  -0.02  -0.00  -0.00      0     0    0       0
terrace_2          0.15   0.07   0.01   0.01      0     0    0       0
terrains           0.12   0.07   0.02   0.01      0     0    0       0
overall           -0.16  -0.22  -0.38  -0.40     -9     0   -1      -9
average           -2.43  -2.76  -3.42  -3.53      0     0    0       0
```

### DIFF
--- a/src/colmap/estimators/pose.cc
+++ b/src/colmap/estimators/pose.cc
@@ -84,7 +84,7 @@ bool EstimateAbsolutePose(const AbsolutePoseEstimationOptions& options,
   } else {
     // Convert normalized error to angular error.
     custom_ransac_options.max_error =
-      std::atan(custom_ransac_options.max_error);
+        std::atan(custom_ransac_options.max_error);
     std::vector<Eigen::Vector3d> rays(points2D.size());
     for (size_t i = 0; i < points2D.size(); ++i) {
       rays[i] = camera->CamFromImg(points2D[i]).homogeneous().normalized();
--- a/src/colmap/estimators/pose.cc
+++ b/src/colmap/estimators/pose.cc
@@ -55,16 +55,16 @@ bool EstimateAbsolutePose(const AbsolutePoseEstimationOptions& options,
   *num_inliers = 0;
   inlier_mask->clear();
 
-  std::vector<Eigen::Vector2d> points2D_normalized(points2D.size());
-  for (size_t i = 0; i < points2D.size(); ++i) {
-    points2D_normalized[i] = camera->CamFromImg(points2D[i]);
-  }
-
   auto custom_ransac_options = options.ransac_options;
   custom_ransac_options.max_error =
       camera->CamFromImgThreshold(options.ransac_options.max_error);
 
   if (options.estimate_focal_length) {
+    // Focal length estimator only works for non-spherical cameras.
+    std::vector<Eigen::Vector2d> points2D_normalized(points2D.size());
+    for (size_t i = 0; i < points2D.size(); ++i) {
+      points2D_normalized[i] = camera->CamFromImg(points2D[i]);
+    }
     // TODO(jsch): Implement non-minimal solver for LORANSAC refinement.
     // Experiments showed marginal difference between RANSAC/LORANSAC for PNPF
     // after refining the estimates of this function using RefineAbsolutePose.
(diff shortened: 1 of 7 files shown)


src/colmap/estimators/pose.cc
```diff
@@ -82,8 +82,15 @@ bool EstimateAbsolutePose(const AbsolutePoseEstimationOptions& options,
       return true;
     }
   } else {
+    // Convert normalized error to angular error.
+    custom_ransac_options.max_error =
+        std::atan(custom_ransac_options.max_error);
```
B1ueber2y, line 87: This will result in large threshold in pixels for rays that are far away from the camera center, and is therefore not ideal for perspective images, though this should be the way to go for spherical images. This also makes the localization workaround difficult to debug via 2D visualization in the future. 

I feel that ideally we should explicitly separate the two models when it comes to measurements for perspective and spherical images, and should create a new pose estimator via virtualization and inheritance for spherical images. Although this has the drawback that it wont work when it comes to multi-camera with both models present.

B1ueber2y: I dont have a better idea for now. Ideally we would love to benchmark such change on single-image visual localization (e.g., on LaMAR), as this will change all the behaviors on HLoc as well if the users updates pycolmap.

ahojnnes: I am not sure I fully agree with the concern. One could similarly argue that the current approach is unnecessarily constrained at the image boundaries for wide FOV cameras. The normalized image plane is imo not always better than angular errors. They both come with their pros and cons.

This brings me to another idea, which is maybe better: If we inject the cameras into the RANSAC and the P3P estimator, then we could directly measure errors in pixels, which would then also account for lens distortion. This is closer to the actual cost function we want to optimize (and optimize in BA already). However, it's not clear whether that is always better either. Let me know what you think.

ahojnnes: Tagging @vlarsson @tsattler in case you have insights or experiments that would suggest one or the other way. Thanks.

B1ueber2y: > I am not sure I fully agree with the concern. One could similarly argue that the current approach is unnecessarily constrained at the image boundaries for wide FOV cameras. The normalized image plane is imo not always better than angular errors. They both come with their pros and cons.
> 
> This brings me to another idea, which is maybe better: If we inject the cameras into the RANSAC and the P3P estimator, then we could directly measure errors in pixels, which would then also account for lens distortion. This is closer to the actual cost function we want to optimize (and optimize in BA already). However, it's not clear whether that is always better either. Let me know what you think.

I like the idea of injecting the cameras into the P3P estimator. We could, for example, template the camera model for the estimator.

ahojnnes: I have this implemented here now: https://github.com/colmap/colmap/pull/3166

PR #3161, opened 2025-02-16
Improve incremental mapper initialization logic

ahojnnes:
Currently, if we manage to initialize from one pair but get stuck after a few images and none of the other pairs satisfy the initialization thresholds, then we never try to initialize from a relaxed set of thresholds anymore for the remaining images, even though there is a high chance to succeed. In addition, we unnecessarily tried the same pairs redundantly for the same initialization constraints. This should result in more complete reconstruction results and faster initialization.

### DIFF
--- a/src/colmap/controllers/incremental_pipeline.cc
+++ b/src/colmap/controllers/incremental_pipeline.cc
@@ -284,93 +284,6 @@ bool IncrementalPipeline::LoadDatabase() {
   return true;
 }
 
-void IncrementalPipeline::Reconstruct(
-    IncrementalMapper& mapper,
-    const IncrementalMapper::Options& mapper_options,
-    bool continue_reconstruction) {
-  for (int num_trials = 0; num_trials < options_->init_num_trials;
-       ++num_trials) {
-    if (CheckIfStopped()) {
-      break;
-    }
-    size_t reconstruction_idx;
-    if (!continue_reconstruction || num_trials > 0) {
-      reconstruction_idx = reconstruction_manager_->Add();
-    } else {
-      reconstruction_idx = 0;
-    }
-    std::shared_ptr<Reconstruction> reconstruction =
-        reconstruction_manager_->Get(reconstruction_idx);
-
-    const Status status =
-        ReconstructSubModel(mapper, mapper_options, reconstruction);
-    switch (status) {
-      case Status::INTERRUPTED: {
-        LOG(INFO) << "Keeping reconstruction due to interrupt";
-        mapper.EndReconstruction(/*discard=*/true);
-        return;
-      }
-
-      case Status::NO_INITIAL_PAIR: {
-        LOG(INFO) << "Disacarding reconstruction due to no initial pair";
-        mapper.EndReconstruction(/*discard=*/true);
-        reconstruction_manager_->Delete(reconstruction_idx);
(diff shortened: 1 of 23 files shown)


src/colmap/controllers/incremental_pipeline.cc
```diff
@@ -270,6 +284,93 @@ bool IncrementalPipeline::LoadDatabase() {
   return true;
 }
 
+void IncrementalPipeline::Reconstruct(
+    IncrementalMapper& mapper,
+    const IncrementalMapper::Options& mapper_options,
+    bool continue_reconstruction) {
+  for (int num_trials = 0; num_trials < options_->init_num_trials;
```
sarlinpe, line 291: How about splitting this function into 2 levels with one called `ReconstructWithTrials` to reduce the nesting? This will make it more readable and modular.

src/colmap/controllers/incremental_pipeline.cc
```diff
@@ -270,6 +284,93 @@ bool IncrementalPipeline::LoadDatabase() {
   return true;
 }
 
+void IncrementalPipeline::Reconstruct(
```
sarlinpe, line 287: Unfortunately it's hard to carefully diff this function because it was moved - consider reverting the move to make it easier to review and catch unintended changes.

src/colmap/controllers/incremental_pipeline.cc
```diff
@@ -326,20 +427,6 @@ IncrementalPipeline::Status IncrementalPipeline::InitializeReconstruction(
   return Status::SUCCESS;
 }
 
-bool IncrementalPipeline::CheckRunGlobalRefinement(
```
sarlinpe, line 329: Same as above, moving it is likely not necessary.

sarlinpe (approved):
Nice improvements, thank you! LGTM, but let's wait that @B1ueber2y reviews too.

src/colmap/controllers/incremental_pipeline.cc
ahojnnes: Tried to undo. Hope it's more readable now.

src/colmap/controllers/incremental_pipeline.cc
ahojnnes: Agreed, but this will again make it hard to review. Let's do this reshuffling of code in another PR and focus here on the logic changes.

src/colmap/controllers/incremental_pipeline.cc
```diff
@@ -502,25 +508,33 @@ void IncrementalPipeline::Reconstruct(
         ReconstructSubModel(mapper, mapper_options, reconstruction);
     switch (status) {
       case Status::INTERRUPTED: {
-        mapper.EndReconstruction(/*discard=*/false);
+        LOG(INFO) << "Keeping reconstruction due to interrupt";
+        mapper.EndReconstruction(/*discard=*/true);
```
sarlinpe, line 512: Is this change intended? Don't GUI users expect the reconstruction to not be discarded if they interrupt the process?

ahojnnes: Good catch. Unintended. The log message is already saying the right thing. :-)

B1ueber2y (approved):
Thanks. LGTM

PR #3164, opened 2025-02-17
Fix mask path

kstoneriv3:
I think the current code creates double extensions like "mask_path/image0001.png.png" instead of "mask_path/image_0001.png".


kstoneriv3:
Ah, sorry! Looks like I didn't read the document properly. 

Thank you for the clarification!

PR #3168, opened 2025-02-17
Add support for propagating relative pose covariance.

B1ueber2y:


src/colmap/geometry/rigid3.h
```diff
@@ -112,6 +112,17 @@ inline Eigen::Matrix6d GetCovarianceForRigid3dInverse(
   return adjoint_inv * covar * adjoint_inv.transpose();
 }
 
+// Calculate covariance (6 x 6) for relative transformation T_1^{-1}T_2.
+inline Eigen::Matrix6d GetCovarianceForRelativeRigid3d(
+    const Rigid3d& pose1,
+    const Rigid3d& pose2,
+    const Eigen::Matrix<double, 12, 12>& covar) {
```
ahojnnes, line 119: It will be good to provide some more documentation for this and specify what the covariance matrix contains. I also suggest using the a_from_b notation. For example:
```suggestion
// Returns covariance for relative pose a_from_c. Expects covariance as ...
inline Eigen::Matrix6d GetCovarianceForRelativeRigid3d(
    const Rigid3d& b_from_a,
    const Rigid3d& b_from_c,
    const Eigen::Matrix<double, 12, 12>& cov) {
```

B1ueber2y: This should not matter. I believe the math works for both cam_from_world and world_from_cam. So its really the composition of rigid3d as long as the covariance is in the same convention of how the rigid3d is defined.

B1ueber2y: Also I just realized that the second pose does not contribute to the covariance propagation and is thus not required. I have made a simplification and made some improvements to the comments: https://github.com/colmap/colmap/pull/3168/commits/1f8ed3c17ade78e3a7fe5cbbe9595133c0c30fc2

ahojnnes: What I wanted to imply is that if the left hand side of the two transformations is not the same, then the resulting covariance is not meaningful. It then also matters how the covariance is specified. In the case where the input poses are b_from_an and b_from_c, it's the covariance blocks b_from_a_cov and b_from_c_cov.

For our use case, I am also thinking that it may be more useful to design the interface in the other direction where the relative pose in your notation is composed as T_1 * T_2^{-1}. I assume we'll use this for computing relative pose covariances and we will always have cam_from_world poses as input?

B1ueber2y: You are absolutely right. I updated the notation to T_2 * T_1^{-1} and did the derivation again. The tests are also updated. I also changed the convention to a_from_c and b_from_c and the relative pose will be b_from_a.

B1ueber2y: PTAL

B1ueber2y: One thing to mention: with the updated convention the second pose does contribute to the covariance propagation, so I added it back to the interface.

src/colmap/geometry/rigid3.h
```diff
@@ -98,20 +98,32 @@ inline Rigid3d Inverse(const Rigid3d& b_from_a) {
...
     const Rigid3d& rigid3, const Eigen::Matrix6d& covar) {
   const Eigen::Matrix6d adjoint_inv = rigid3.AdjointInverse();
   return adjoint_inv * covar * adjoint_inv.transpose();
 }
 
+// Given a (12 x 12) covariance on top of two rigid3d objects, (T_a, T_b), this
+// function calculates the (6 x 6) covariance of the relative transformation
+// T_a^{-1}T_b. T_b does not contribute to the covariance propagation and is
+// thus not required.
+inline Eigen::Matrix6d GetCovarianceForRelativeRigid3d(
```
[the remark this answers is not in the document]
B1ueber2y: I dont think the way you proposed can handle the correlation between the two poses, which is crucial in the covariance propagation for relative pose. Please refer to the two tests I added, where in both cases a 12x12 Jacobian is applied to the covariance when the pose inversion happens.

B1ueber2y:
Added another interface for the pose composition, and added a consistency test.

src/colmap/geometry/rigid3.h
```diff
@@ -98,20 +98,45 @@ inline Rigid3d Inverse(const Rigid3d& b_from_a) {
...
   const Eigen::Matrix6d adjoint_inv = rigid3.AdjointInverse();
   return adjoint_inv * covar * adjoint_inv.transpose();
 }
 
+// Given a (12 x 12) covariance on top of two rigid3d objects, a_from_b,
+// b_from_c, this function calculates the (6 x 6) covariance of the composed
+// transformation a_from_c (a_T_b * b_T_c). b_T_c does not contribute to the
+// covariance propagation and is thus not required.
+inline Eigen::Matrix6d GetCovarianceForComposedRigid3d(
+    const Rigid3d& a_from_b, const Eigen::Matrix<double, 12, 12>& covar) {
```
ahojnnes, line 120: nit: for consistency of notation with what we did in `https://github.com/colmap/colmap/blob/main/src/pycolmap/estimators/cost_functions.cc` ?

```suggestion
    const Rigid3d& a_from_b, const Eigen::Matrix<double, 12, 12>& a_cov_from_b) {
```

ahojnnes: Same for other "covar" variable names.

B1ueber2y: But this is the joint covariance (12 x 12) on a_T_b and b_T_c. Do you think this is not explained in the comments? : )

B1ueber2y: On the other hand, I have fixed the convention in the tests.

ahojnnes: Right, didn't look carefully enough. Maybe we can be more explicit in the comments.
```
[a_cov_from_b, b_cov_from_a]
[a_cov_from_c, b_cov_from_c]
```
Hope I got it right (on my phone).

src/colmap/geometry/rigid3.h
```diff
@@ -98,20 +98,45 @@ inline Rigid3d Inverse(const Rigid3d& b_from_a) {
...
+  return J * covar * J.transpose();
+}
+
+// Given a (12 x 12) covariance on top of two rigid3d objects, a_from_c,
+// b_from_c, this function calculates the (6 x 6) covariance of the relative
+// transformation b_from_a (b_T_c * a_T_c^{-1}).
+inline Eigen::Matrix6d GetCovarianceForRelativeRigid3d(
+    const Rigid3d& a_from_c,
+    const Rigid3d& b_from_c,
+    const Eigen::Matrix<double, 12, 12>& covar_x_from_c) {
```
ahojnnes, line 133: nit
```suggestion
    const Eigen::Matrix<double, 12, 12>& cov_x_from_c) {
```

ahojnnes (approved):

src/colmap/geometry/rigid3.h
B1ueber2y: The understanding is correct, but in your proposal its a 6 x 12 matrix? You have overlooked the correlation part, which is a bit hard to describe in the comments.

B1ueber2y: I made a minor improvement on the comments. Hope it looks better.

src/colmap/geometry/rigid3.h
B1ueber2y: I changed it back to ``covar`` rather than ``covar_x_from_c`` as it is a joint covariance (12 x 12) on two transformation and ``cov_x_from_c`` can be confusing.

src/colmap/geometry/rigid3.h
B1ueber2y: Wait. But this is not correct then? b_cov_from_a does not equal to the correlation of a_from_b and b_from_c.

ahojnnes: Still on the phone. Given how tricky this is, we should be very explicit about which covariance is expected here. We shouldn't expect people to go to the paper to figure this out.

src/colmap/geometry/rigid3_test.cc
```diff
@@ -172,5 +172,98 @@ TEST(Rigid3d, CovariancePropagation) {
...
+  // Compose T_1T_2^{-1} in right
+  // [Reference] Joan Solà, Jeremie Deray, Dinesh Atchuthan, A micro Lie theory
+  // for state estimation in robotics, 2018.
+  // Eqs. (177) and (178)
+  Eigen::Matrix<double, 6, 12> J_in_right;
+  J_in_right.block<6, 6>(0, 0) = b_from_world.Adjoint();
+  J_in_right.block<6, 6>(0, 6) = Eigen::Matrix6d::Identity();
+  const Eigen::Matrix6d a_cov_from_b_right =
+      J_in_right * covar_in_right * J_in_right.transpose();
+  EXPECT_LT((b_cov_from_a - a_cov_from_b_right).norm(), 1e-6);
```
ahojnnes, line 239: `EXPECT_THAT(b_cov_from_a, EigenMatrixNear(a_cov_from_b_right, 1e-6))`

ahojnnes: Same for other places.

src/colmap/geometry/rigid3.h
ahojnnes: I am not saying that what I wrote above is correct. It's probably wrong but I want us to be explicit.

B1ueber2y: Its not in the paper. In the relative pose for example, the covariance is defined on (a_from_world, and b_from_world), and the off-diagonal would be just the result of the function we already have here: https://github.com/colmap/colmap/blob/main/src/colmap/estimators/covariance.h#L66. 

However, I see the issue now. The naming of the function is wrong, as it is not Cam2FromCam1Cov. Instead, it is the correlation of cam1_from_world and cam2_from_world, which is a different thing.

B1ueber2y: > Its not in the paper. In the relative pose for example, the covariance is defined on (a_from_world, and b_from_world), and the off-diagonal would be just the result of the function we already have here: https://github.com/colmap/colmap/blob/main/src/colmap/estimators/covariance.h#L66.
> 
> However, I see the issue now. The naming of the function is wrong, as it is not Cam2FromCam1Cov. Instead, it is the correlation of cam1_from_world and cam2_from_world, which is a different thing.

We should fix that naming in a separate PR, once we have a good idea how to name it.

src/colmap/geometry/rigid3_test.cc
B1ueber2y: Done.

src/pycolmap/geometry/bindings.cc
```diff
@@ -102,6 +102,19 @@ void BindGeometry(py::module& m) {
                                            const Eigen::Matrix6d&)>(
                &GetCovarianceForRigid3dInverse),
            py::arg("covar"))
+      .def("get_covariance_for_composed_rigid3d",
+           static_cast<Eigen::Matrix6d (*)(
+               const Rigid3d&, const Eigen::Matrix<double, 12, 12>&)>(
+               &GetCovarianceForComposedRigid3d),
+           py::arg("covar"))
```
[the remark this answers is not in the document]
B1ueber2y: No. This is defined on the Rigid3d object, so the first argument is the object itself.

src/colmap/geometry/rigid3.h
B1ueber2y: PR created here: https://github.com/colmap/colmap/pull/3170

src/pycolmap/geometry/bindings.cc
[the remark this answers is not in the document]
ahojnnes: I see. I would suggest keeping it as a separate function just like in C++.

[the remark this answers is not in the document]
B1ueber2y: Sure. Updated.

PR #3169, opened 2025-02-18
Add docker instruction link to docs

j3soon:
I found that some of my collaborators aren't aware of the pre-built Docker image. I think it would be helpful to mention it somewhere on the documentation's installation page.


ahojnnes (approved):
Thank for your contribution. This will surely be helpful for folks.

PR #3170, opened 2025-02-18
Fix naming of cross covariance and add relative pose covariance interface

B1ueber2y:
Now the name is too long. Creating this PR to initiate the thread.

### DIFF
--- a/src/colmap/estimators/covariance.cc
+++ b/src/colmap/estimators/covariance.cc
@@ -270,15 +270,15 @@ std::optional<Eigen::MatrixXd> BACovariance::GetCam2CovFromCam1(
   auto cov_11 = GetCamCovFromWorld(image_id1);
   if (!cov_11.has_value()) return std::nullopt;
   if (cov_11->rows() != 6) {
-    LOG(WARNING) << "cam1_from_world not fully in the problem (likely to due "
-                    "to being set constant / partially constant). ";
+    LOG(WARNING) << "cam1_from_world is not fully in the problem. This is "
+                    "likely due to the pose being set (partially) constant. ";
     return std::nullopt;
   }
   auto cov_22 = GetCamCovFromWorld(image_id2);
   if (!cov_22.has_value()) return std::nullopt;
   if (cov_22->rows() != 6) {
-    LOG(WARNING) << "cam2_from_world not fully in the problem (likely to due "
-                    "to being set constant / partially constant). ";
+    LOG(WARNING) << "cam2_from_world is not fully in the problem. This is "
+                    "likely due to the pose being set (partially) constant. ";
     return std::nullopt;
   }
   auto cov_12 = GetCamCrossCovFromWorld(image_id1, image_id2);
--- a/src/colmap/estimators/covariance.cc
+++ b/src/colmap/estimators/covariance.cc
@@ -250,8 +250,7 @@ std::optional<Eigen::MatrixXd> BACovariance::GetCamFromWorldCov(
   return ExtractCovFromLInverse(L_inv_, start, start, size, size);
 }
 
-std::optional<Eigen::MatrixXd>
-BACovariance::GetCorrelation_Cam1FromWorld_Cam2FromWorld(
+std::optional<Eigen::MatrixXd> BACovariance::GetCamFromWorldCorr(
     image_t image_id1, image_t image_id2) const {
   const auto it1 = pose_L_start_size_.find(image_id1);
   const auto it2 = pose_L_start_size_.find(image_id2);
--- a/src/colmap/estimators/covariance.cc
+++ b/src/colmap/estimators/covariance.cc
@@ -290,6 +290,24 @@ std::optional<Eigen::MatrixXd> BACovariance::GetCamFromWorldCov(
   return cov;
 }
 
+std::optional<Eigen::MatrixXd> BACovariance::GetCam2CovFromCam1(
+    image_t image_id1,
+    const Rigid3d& cam1_from_world,
+    image_t image_id2,
+    const Rigid3d& cam2_from_world) const {
+  std::vector<image_t> image_ids = {image_id1, image_id2};
+  auto cov = GetCamFromWorldCov(image_ids);
+  if (!cov.has_value()) return std::nullopt;
+  if (cov->rows() != 12) {
+    LOG(WARNING) << "Either cam1_from_world or cam2_from_world are not fully "
+                    "in the problem. This is likely due to one of the two "
+                    "being set constant / partially constant.";
+    return std::nullopt;
+  }
+  return GetCovarianceForRelativeRigid3d(
+      cam1_from_world, cam2_from_world, *cov);
+}
+
 std::optional<Eigen::MatrixXd> BACovariance::GetOtherParamsCov(
     const double* params) const {
   const auto it = other_L_start_size_.find(params);
--- a/src/colmap/estimators/covariance.cc
+++ b/src/colmap/estimators/covariance.cc
@@ -240,7 +240,7 @@ std::optional<Eigen::MatrixXd> BACovariance::GetPointCov(
   return it->second;
 }
 
-std::optional<Eigen::MatrixXd> BACovariance::GetCamFromWorldCov(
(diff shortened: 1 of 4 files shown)


src/colmap/estimators/covariance.h
```diff
@@ -63,8 +63,8 @@ struct BACovariance {
   // dimensions are kept constant, the respective rows/columns are omitted.
   // Returns null if image not a variable in the problem.
   std::optional<Eigen::MatrixXd> GetCamFromWorldCov(image_t image_id) const;
-  std::optional<Eigen::MatrixXd> GetCam1FromCam2Cov(image_t image_id1,
-                                                    image_t image_id2) const;
+  std::optional<Eigen::MatrixXd> GetCorrelation_Cam1FromWorld_Cam2FromWorld(
```
ahojnnes, line 66: Another option could be GetCamFromWorldCorr ?

B1ueber2y: Sounds good. Updated.

ahojnnes (approved):

ahojnnes:
Actually, this is not really a correlation but still a covariance. How about just also GetCamFromWorldCov but with two instead of one image argument?

B1ueber2y:
> Actually, this is not really a correlation but still a covariance. How about just also GetCamFromWorldCov but with two instead of one image argument?

Indeed, true. How about we say ``GetCamFromWorldCrossCov``?

ahojnnes:
SGTM

B1ueber2y:
I added another interface to directly get covariance for multiple images.

src/colmap/estimators/covariance.h
```diff
@@ -61,14 +61,17 @@ struct BACovariance {
...
   // dimensions are kept constant, the respective rows/columns are omitted.
-  // Returns null if image not a variable in the problem.
+  // Returns null if image is not a variable in the problem.
   std::optional<Eigen::MatrixXd> GetCamFromWorldCov(image_t image_id) const;
-  std::optional<Eigen::MatrixXd> GetCam1FromCam2Cov(image_t image_id1,
-                                                    image_t image_id2) const;
+  std::optional<Eigen::MatrixXd> GetCamFromWorldCrossCov(
+      image_t image_id1, image_t image_id2) const;
+  std::optional<Eigen::MatrixXd> GetCamFromWorldCov(
+      const std::vector<image_t>& image_ids) const;
```
ahojnnes, line 69: nit: could be colmap::span.

src/colmap/estimators/covariance.cc
```diff
@@ -262,6 +262,33 @@ std::optional<Eigen::MatrixXd> BACovariance::GetCam1FromCam2Cov(
   return ExtractCovFromLInverse(L_inv_, start1, start2, size1, size2);
 }
 
+std::optional<Eigen::MatrixXd> BACovariance::GetCamFromWorldCov(
+    const std::vector<image_t>& image_ids) const {
+  int n_dims = 0;
+  std::vector<std::pair<int, int>> start_sizes;
```
ahojnnes, line 268:
```suggestion
  std::vector<std::pair<int, int>> start_sizes;
  start_sizes.reserve(image_ids.size());
```

src/colmap/estimators/covariance.h
```diff
@@ -61,14 +61,17 @@ struct BACovariance {
...
   // Tangent space covariance in the order [rotation, translation]. If some
   // dimensions are kept constant, the respective rows/columns are omitted.
-  // Returns null if image not a variable in the problem.
+  // Returns null if image is not a variable in the problem.
   std::optional<Eigen::MatrixXd> GetCamFromWorldCov(image_t image_id) const;
-  std::optional<Eigen::MatrixXd> GetCam1FromCam2Cov(image_t image_id1,
-                                                    image_t image_id2) const;
+  std::optional<Eigen::MatrixXd> GetCamFromWorldCrossCov(
+      image_t image_id1, image_t image_id2) const;
+  std::optional<Eigen::MatrixXd> GetCamFromWorldCov(
```
ahojnnes, line 68: So, why do we call one cov and the other cross cov? Let's just call all of them cov?

I am also curious about what we need the new method for in practice?

B1ueber2y: This function estimates the 6 x 6 cross covariance between cam1_from_world and cam2_from_world. I was worried that if we use covariance it can confuse with the 12 x 12 covariance of [cam1_from_world, cam2_from_world]. 

To use the relative pose covariance interface, we could directly use the new method rather than concatenating four 6 x 6 matrices. However, in the future we could also add a method to directly do all the logic regarding relative pose in this CovarianceEstimatorClass.

src/colmap/estimators/covariance.h
ahojnnes: Yes, see pybind_extension.h:25

src/colmap/estimators/covariance.h
B1ueber2y: I have added such method as discussed. The ``BACovariance`` class does not store the poses so they need to be fed as arguments.

The function returns null when one of the poses are (partially) kept constant. This will require the users to fill in zero block themselves in the initial covariance before composing relative pose. Thus, We could still keep the joint covariance interface for use case. I added comments to explain this.

src/colmap/estimators/covariance.h
```diff
@@ -61,14 +61,27 @@ struct BACovariance {
 
   // Tangent space covariance in the order [rotation, translation]. If some
   // dimensions are kept constant, the respective rows/columns are omitted.
-  // Returns null if image not a variable in the problem.
+  // Returns null if image is not a variable in the problem.
   std::optional<Eigen::MatrixXd> GetCamFromWorldCov(image_t image_id) const;
```
ahojnnes, line 65: GetCamCovFromWorld?

src/colmap/estimators/covariance.cc
```diff
@@ -250,7 +250,7 @@ std::optional<Eigen::MatrixXd> BACovariance::GetCamFromWorldCov(
   return ExtractCovFromLInverse(L_inv_, start, start, size, size);
 }
 
-std::optional<Eigen::MatrixXd> BACovariance::GetCam1FromCam2Cov(
+std::optional<Eigen::MatrixXd> BACovariance::GetCamFromWorldCrossCov(
```
ahojnnes, line 253: GetCamCrossCovFromWorld ?

src/colmap/estimators/covariance.cc
```diff
@@ -262,6 +262,52 @@ std::optional<Eigen::MatrixXd> BACovariance::GetCam1FromCam2Cov(
   return ExtractCovFromLInverse(L_inv_, start1, start2, size1, size2);
 }
 
+std::optional<Eigen::MatrixXd> BACovariance::GetCamFromWorldCov(
```
ahojnnes, line 265: I would also call this GetCamCrossCovFromWorld

B1ueber2y: This is a general covariance rather than the cross covariance? In my mind cross covariance refers to those off-diagonal blocks.

src/colmap/estimators/covariance.h
B1ueber2y: Thanks. Updated.

src/colmap/estimators/covariance.cc
ahojnnes: But it also contains the off-diagonal blocks?

B1ueber2y: Yeah but i think the covariance of [A, B] includes four blocks, the covariance of A, the cross covariance of A and B /  B and A, and the covariance of B. So this function returns a covariance rather than cross covariance.

ahojnnes: Do we use this method for anything? Let's just get rid of it and resolve the discussion this way?

On another note, we could also change the interface of the other method and not take a vector. Are we every going to use this for more than 2 images?

B1ueber2y: Okay I have removed this method and updated the relative pose covariance impl: https://github.com/colmap/colmap/pull/3170/commits/b8a77101974224dcd9b58d4617680f2740746df3.

PR #3172, opened 2025-02-19
Camera models perform valid projection test

ahojnnes:
Towards supporting spherical / large FOV cameras for which a different logic will have to be implemented than for perspective projection models. The logic is now specialized and localized within the camera model and the user just knows about valid or invalid projections. This can also be used to more robustly deal with numerical issues in distortion computation (as we have for some of the fisheye models).

The next PRs will do the equivalent for CamFromImg to return camera rays rather than image points as well as remove redundant "HasPointPositiveDepth/Cheirality" checks.

### DIFF
--- a/src/colmap/estimators/bundle_adjustment_test.cc
+++ b/src/colmap/estimators/bundle_adjustment_test.cc
@@ -164,14 +164,11 @@ void GenerateReconstruction(const size_t num_images,
             RandomUniformReal(-1.0, 1.0), RandomUniformReal(-1.0, 1.0), 10)));
     reconstruction->AddImage(image);
 
-    const Eigen::Matrix3x4d cam_from_world_matrix =
-        image.CamFromWorld().ToMatrix();
-
     std::vector<Eigen::Vector2d> points2D;
     for (const auto& point3D : reconstruction->Points3D()) {
       // Get exact projection of 3D point.
-      std::optional<Eigen::Vector2d> point2D = camera.ImgFromCam(
-          (image.CamFromWorld() * point3D.second.xyz).hnormalized());
+      std::optional<Eigen::Vector2d> point2D =
+          camera.ImgFromCam(image.CamFromWorld() * point3D.second.xyz);
       ASSERT_TRUE(point2D.has_value());
       // Add some uniform noise.
       *point2D += Eigen::Vector2d(RandomUniformReal(-2.0, 2.0),
--- a/src/colmap/estimators/bundle_adjustment_test.cc
+++ b/src/colmap/estimators/bundle_adjustment_test.cc
@@ -162,21 +162,21 @@ void GenerateReconstruction(const size_t num_images,
         Eigen::Quaterniond::Identity(),
         Eigen::Vector3d(
             RandomUniformReal(-1.0, 1.0), RandomUniformReal(-1.0, 1.0), 10)));
-    reconstruction->AddImage(image);
 
     std::vector<Eigen::Vector2d> points2D;
     for (const auto& point3D : reconstruction->Points3D()) {
       // Get exact projection of 3D point.
       std::optional<Eigen::Vector2d> point2D =
           camera.ImgFromCam(image.CamFromWorld() * point3D.second.xyz);
-      ASSERT_TRUE(point2D.has_value());
+      CHECK(point2D.has_value());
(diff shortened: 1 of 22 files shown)


src/colmap/estimators/bundle_adjustment_test.cc
```diff
@@ -713,7 +711,9 @@ TEST(RigBundleAdjuster, FourView) {
   std::vector<CameraRig> camera_rigs;
   camera_rigs.emplace_back();
   camera_rigs[0].AddCamera(0, Rigid3d());
-  camera_rigs[0].AddCamera(1, Rigid3d());
+  camera_rigs[0].AddCamera(1,
```
ahojnnes, line 714: This was a bug in the test setup before leading to points behind the camera, which now lead to a solver failure. This is fixed now.

B1ueber2y:
Thanks! This is a great improvement for spherical cameras and bundle adjustment. My only concern is the updated interface of ``ImgFromCam``where 3D ray rather than 2D point becomes the argument. There are mainly two issues:

1) ``CamFromImg`` and ``ImgFromCam`` is not consistent anymore. So this should be updated together in a same PR. 
2) The updated ``ImgFromCam`` could potentially break pycolmap users and downstream codebase that relies on COLMAP.

One easy way out for both issues is support backwards compatibility with an overloaded ``ImgFromCam`` with original Vector2d argument.

ahojnnes:
> CamFromImg and ImgFromCam is not consistent anymore. So this should be updated together in a same PR.

As I mentioned in the description of this PR, this change follows in a subsequent PR to keep reviewing scope manageable.

> The updated ImgFromCam could potentially break pycolmap users and downstream codebase that relies on COLMAP.

Yes, it's a breaking change. I don't see a way for us to make this change without breaking the API. There will be many other breaking changes related to rig support. Not sure if we need to be extra concerned now about this PR?

src/colmap/scene/camera.h
```diff
@@ -131,7 +131,8 @@ struct Camera {
   inline double CamFromImgThreshold(double threshold) const;
 
   // Project point from camera frame to image plane.
-  inline Eigen::Vector2d ImgFromCam(const Eigen::Vector2d& cam_point) const;
+  inline std::optional<Eigen::Vector2d> ImgFromCam(
+      const Eigen::Vector3d& cam_point) const;
```
B1ueber2y, line 135: We could keep the original function with overloading to avoid breaking backward incompatibility?

``inline std::optional<Eigen::Vector2d> ImgFromCam(const Eigen::Vector2d& cam_point_2d) {return ImgFromCam(cam_point.homogeneous());}``

sarlinpe: Agree, let's avoid breaking user code for widely used functions, consider also providing an overload for pycolmap.Camera.img_from_cam.

ahojnnes: It's a nice idea but I would suggest not doing it:
* I actually think, we want this to be a breaking change, so people correctly update their code and don't continue using this going forward. It's an easy fix to make. We could flag it with `[[deprecated]]` but there is another issue.
* Overloading the method, causes compiler issues like this due to Eigen:
```
colmap/src/colmap/image/warp.cc:86:25: error: call to member function 'ImgFromCam' is ambiguous
   86 |           source_camera.ImgFromCam(cam_point.homogeneous());
...
colmap/estimators/pose.cc:84:9: error: no matching function for call to 'bind'
   84 |         std::bind(&Camera::ImgFromCam, camera, std::placeholders::_1);
...
```

B1ueber2y: I see. Flagging it as deprecated in pycolmap sounds good to me then (if we find the way to do that). Thanks!

sarlinpe: Ok thanks for the clarification, I agree with Johannes here. I suggest providing an overload only for `pycolmap.Camera.img_from_cam`, with a deprecation warning. This already supports `Eigen::MatrixX2d` and `Eigen::MatrixX3d`, so we should have both `Vector2d` and `Vector3d` - and we can drop support for the 2D interfaces at the next version.

sarlinpe (approved):
Thank you!

B1ueber2y (approved):
LGTM. Thanks a lot!

PR #3176, opened 2025-02-21
Various improvements and extensions for pycolmap

ahojnnes:


src/pycolmap/scene/point2d.cc
```diff
@@ -28,9 +28,12 @@ void BindPoint2D(py::module& m) {
...
                "point3D_id", kInvalidPoint3DId, "pycolmap.INVALID_POINT3D_ID"))
       .def_readwrite("xy", &Point2D::xy)
+      .def("x", [](const Point2D& self) -> double { return self.xy[0]; })
+      .def("y", [](const Point2D& self) -> double { return self.xy[1]; })
       .def_readwrite("point3D_id", &Point2D::point3D_id)
       .def("has_point3D", &Point2D::HasPoint3D);
   MakeDataclass(PyPoint2D);
 
-  py::bind_vector<Point2DVector>(m, "ListPoint2D");
+  py::bind_vector<Point2DVector>(m, "Points2D");
```
B1ueber2y, line 37: Points2D is not a type in colmap, and could be potentially ambiguous, as Points3D in colmap is a map. I would suggest to keep the name ``ListPoint2D`` to be clear.

ahojnnes: Where is Points3D defined as a map?

B1ueber2y: ``reconstruction.points3D`` is a map: https://github.com/colmap/colmap/blob/main/src/pycolmap/scene/point3d.cc#L30

ahojnnes: OK. The naming is a bit unfortunate, because it's not easily discoverable due to the List prefix. I change it to Point2DList().

B1ueber2y:
The moved line was in ``scene/bindings.cc`` for a reason that we did not understand well: https://github.com/colmap/colmap/pull/2558

ahojnnes:
Good callout. We should add a comment then, so we don't attempt to make the same move in a few weeks from now :-)

src/pycolmap/scene/bindings.cc
```diff
@@ -33,5 +33,6 @@ void BindScene(py::module& m) {
   BindDatabaseCache(m);
   BindSynthetic(m);
 
-  py::implicitly_convertible<py::iterable, Point2DVector>();
+  // Must be defined here, see: https://github.com/colmap/colmap/pull/2558
+  py::bind_vector<Point2DVector>(m, "Point2DList");
```
sarlinpe, line 37: I think that the `implicitly_convertible` should be here and the `bind_vector` should be in `point3d.cc`.

ahojnnes: Good catch. Thanks.

B1ueber2y (approved):

src/pycolmap/scene/database_cache.cc
```diff
@@ -19,6 +19,8 @@ void BindDatabaseCache(py::module& m) {
                   "min_num_matches"_a,
                   "ignore_watermarks"_a,
                   "image_names"_a)
+      .def("add_camera", &DatabaseCache::AddCamera)
+      .def("add_image", &DatabaseCache::AddImage)
```
sarlinpe, line 23: AFAIK `DatabaseCache` does not have such methods?

ahojnnes: Right. Not done here. I am now expanding the scope of this PR.

src/pycolmap/sensor/bitmap.cc
```diff
@@ -0,0 +1,126 @@
...
+                     : (output_into.shape[1] * output_into.shape[2]);
+             for (ssize_t y = 0; y < output_into.shape[0]; ++y) {
+               if (is_rgb) {
+                 for (ssize_t x = 0; x < output_into.shape[1]; ++x) {
+                   BitmapColor<uint8_t> color;
+                   THROW_CHECK(self.GetPixel(x, y, &color));
+                   output_row_ptr[3 * x] = color.r;
+                   output_row_ptr[3 * x + 1] = color.g;
+                   output_row_ptr[3 * x + 2] = color.b;
+                 }
```
sarlinpe, line 46: It would be nice to have a comment that explains why we cannot copy the row in this case. Is it about the ordering of the axes?

src/pycolmap/geometry/bindings.cc
```diff
@@ -1,224 +1,26 @@
...
+void BindEigenGeometry(py::module& m);
+void BindRigid3(py::module& m);
+void BindSim3(py::module& m);
+void BindPosePrior(py::module& m);
 void BindHomographyMatrixGeometry(py::module& m);
 void BindEssentialMatrixGeometry(py::module& m);
+void BindTriangulation(py::module& m);
 
 void BindGeometry(py::module& m) {
-  py::class_ext_<Eigen::Quaterniond> PyRotation3d(m, "Rotation3d");
```
ahojnnes, line 25: Simple copy&paste into their own files.

sarlinpe (approved):
Thank you!

B1ueber2y (approved):

PR #3189, opened 2025-02-27
Define bindings in the correct order

sarlinpe:
Following an issue introduced in https://github.com/colmap/colmap/pull/3185, since the type annotation of `QueryOptions.add` depends on `FeatureKeypoint`.


ahojnnes (approved):

ahojnnes:
I can bet I tried exactly this without success. Thanks.

PR #3199, opened 2025-03-03
Fix maybe-uninitialized warnings

papjuli:
Fix some code that result in maybe-uninitialized compiler warnings:
 - best_model in ransac and loransac can be uninitialized -> use optional,
 - initialize Eigen matrices to 0 in EstimateRigid3d and other similar functions.

Also, don't turn the warning type off CMakeLists.txt.

Unfortunately, there are 3 other such warnings that originate outside, namely in boost graph.


ahojnnes (approved):
Thank you. LGTM.

PR #3202, opened 2025-03-04
Add support for frame serialization

ahojnnes:


src/colmap/scene/frame.h
```diff
@@ -114,10 +91,12 @@ class Frame {
   // case, where rig modeling is no longer needed.
   std::optional<Rigid3d> frame_from_world_;
 
-  // [Optional] Rig calibration
-  std::shared_ptr<class Rig> rig_ = nullptr;
+  // [Optional] Rig calibration.
+  const class Rig* rig_ = nullptr;
```
ahojnnes, line 95: I made this change for consistency with the Image::CameraPtr. In a follow-up, I will homogeneously use shared_ptr's for the CameraPtr/FramePtr/RigPtr, which requires us to store them as shared_ptr's in the reconstruction object.

src/colmap/scene/image.cc
```diff
@@ -53,10 +53,7 @@ Image::Image(const Image& other)
     camera_ptr_ = other.CameraPtr();
   }
   if (other.HasTrivialFrame()) {
-    frame_ = std::make_shared<class Frame>();
-    frame_->SetFrameFromWorld(other.MaybeCamFromWorld());
-  } else {
-    frame_ = other.Frame();
+    frame_ptr_ = other.FramePtr();
```
B1ueber2y, line 56: This changes behaviors from deepcopy to shallow copy, which can be dangerous.

src/colmap/scene/image.cc
```diff
@@ -72,10 +69,7 @@ Image& Image::operator=(const Image& other) {
     num_points3D_ = other.NumPoints3D();
     points2D_ = other.Points2D();
     if (other.HasTrivialFrame()) {
-      frame_ = std::make_shared<class Frame>();
-      frame_->SetFrameFromWorld(other.MaybeCamFromWorld());
-    } else {
-      frame_ = other.Frame();
+      frame_ptr_ = other.FramePtr();
```
B1ueber2y, line 72: Same here

src/colmap/util/types.h
```diff
@@ -77,38 +78,91 @@ namespace colmap {
...
+  }
+  inline bool operator!=(const sensor_t& other) const {
+    return !(*this == other);
+  }
+};
+
+constexpr sensor_t kInvalidSensorId =
+    sensor_t(SensorType::INVALID, std::numeric_limits<uint32_t>::max());
+
+struct data_t {
```
B1ueber2y, line 141: Should also move the corresponding hash function from frame.h?

src/colmap/scene/frame.h
```diff
@@ -102,9 +78,15 @@ class Frame {
   inline bool HasPose() const;
   inline void ResetPose();
 
-  // Get the sensor from world transformation
+  // Get the sensor from world transformation.
   inline Rigid3d SensorFromWorld(sensor_t sensor_id) const;
 
+  // Set the world to frame from the given camera to world transformation.
+  void ApplyCamFromWorld(camera_t camera_id, const Rigid3d& cam_from_world);
```
B1ueber2y, line 85: The naming is confusing. From the first look it feels to be to apply some transformation on top of the pose, and only after looking at the implementation I understood that it is a setter.

src/colmap/scene/image.h
```diff
@@ -93,26 +93,22 @@ class Image {
...
-
-  // The following methods only works for non-composed pose. Will throw an
-  // error if the image has a non-trivial frame (rig) attached to it.
-  inline const Rigid3d& CamFromWorld() const;
-  inline Rigid3d& CamFromWorld();
-  inline const std::optional<Rigid3d>& MaybeCamFromWorld() const;
-  inline std::optional<Rigid3d>& MaybeCamFromWorld();
-  inline void SetCamFromWorld(const Rigid3d& cam_from_world);
-  inline void SetCamFromWorld(const std::optional<Rigid3d>& cam_from_world);
+  // Composition of sensor_from_rig and frame_from_world transformations.
```
B1ueber2y, line 110: Should we mention about the trivial frame in the comment here, to be more clear?

src/colmap/scene/frame.h
ahojnnes: ApplyCamToFrameFromWorld? SetCamToFrameFromWorld? Without making the method name unreasonably long, I cannot really come up with something good.

src/colmap/scene/image.h
ahojnnes: Done.

src/colmap/scene/reconstruction_io.h
```diff
@@ -45,6 +45,9 @@ void ReadRigsText(Reconstruction& reconstruction, const std::string& path);
 void ReadCamerasText(Reconstruction& reconstruction, std::istream& stream);
 void ReadCamerasText(Reconstruction& reconstruction, const std::string& path);
 
+void ReadFramesText(Reconstruction& reconstruction, std::istream& stream);
```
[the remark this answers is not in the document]
ahojnnes: I would wait with this a bit more until some of the remaining TODOs are resolved related to supporting non-trivial frames.

src/colmap/estimators/bundle_adjustment_test.cc
```diff
@@ -154,20 +154,31 @@ void GenerateReconstruction(const size_t num_images,
...
-        Eigen::Quaterniond::Identity(),
-        Eigen::Vector3d(
-            RandomUniformReal(-1.0, 1.0), RandomUniformReal(-1.0, 1.0), 10)));
 
     std::vector<Eigen::Vector2d> points2D;
     for (const auto& point3D : reconstruction->Points3D()) {
       // Get exact projection of 3D point.
       std::optional<Eigen::Vector2d> point2D =
-          camera.ImgFromCam(image.CamFromWorld() * point3D.second.xyz);
+          camera.ImgFromCam(frame.FrameFromWorld() * point3D.second.xyz);
```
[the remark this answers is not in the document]
ahojnnes: In theory, yes. In practice, the FramePtr() is not set yet. The follow-up PR to this one actually gets rid of this manual reconstruction generation and replaces it with the synthetic dataset function. I'll leave it as is.

src/colmap/scene/reconstruction.h
```diff
@@ -116,6 +121,9 @@ class Reconstruction {
   // might be taken by the same camera.
   void AddCamera(struct Camera camera);
```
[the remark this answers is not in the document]
ahojnnes: That feels a bit magic to me.

B1ueber2y (approved):

src/colmap/estimators/alignment.cc
```diff
@@ -456,15 +455,28 @@ bool MergeReconstructions(const double max_reproj_error,
...
+          src_reconstruction.Frame(tgt_image.FrameId()));
     }
-    tgt_reconstruction.AddImage(src_image);
+    tgt_image.SetFramePtr(&tgt_reconstruction.Frame(tgt_image.FrameId()));
+    const Rigid3d cam_from_tgt_world =
+        TransformCameraWorld(tgt_from_src, src_image.CamFromWorld());
+    tgt_reconstruction.Frame(tgt_image.FrameId())
+        .SetCamFromWorld(tgt_image.CameraId(), cam_from_tgt_world);
+    tgt_reconstruction.AddImage(tgt_image);
     tgt_reconstruction.RegisterImage(image_id);
```
sarlinpe, line 480: Consider splitting this code block into functions `MergeReconstructions(src_reconstruction, tgt_reconstruction, tgt_from_src)` and `CopyImageFromReconstruction(src_reconstruction, tgt_reconstruction, tgt_from_src, image_id)` that can be more easily tested later on - can be done in a later PR to keep the diffs clean.

src/colmap/scene/frame.h
```diff
@@ -34,65 +34,41 @@
...
-  }
-  inline bool operator!=(const data_t& other) const {
-    return !(*this == other);
-  }
-};
-
-constexpr data_t kInvalidDataId =
-    data_t(kInvalidSensorId, std::numeric_limits<uint32_t>::max());
-
 class Frame {
```
sarlinpe, line 43: Since the code is getting more complex, it would be nice to add class comments documenting the objects, e.g. what is a Frame and how does it differ from a Rig.

src/colmap/scene/frame.h
```diff
@@ -114,10 +96,13 @@ class Frame {
   // case, where rig modeling is no longer needed.
   std::optional<Rigid3d> frame_from_world_;
 
-  // [Optional] Rig calibration
-  std::shared_ptr<class Rig> rig_ = nullptr;
+  // Rig calibration.
+  rig_t rig_id_ = kInvalidRigId;
```
sarlinpe, line 100: Does the frame need to store the rig id given that it is already stored in the rig? If so, consider documenting why. I understand that it is convenient to only populate the rig id before calling `Reconstruction::AddFrame`, but is this really necessary?

src/colmap/scene/image.h
```diff
@@ -161,14 +158,15 @@ class Image {
   camera_t camera_id_;
   struct Camera* camera_ptr_;
 
+  // The corresponding frame of the image. Note that multiple images might
+  // share the same frame. If not specified `kInvalidFrameId`.
+  frame_t frame_id_;
```
sarlinpe, line 163: Similarly as above, is storing the frame id necessary? This carries logic to keep it in sync with the frame ptr.

src/colmap/scene/reconstruction_io.cc
```diff
@@ -481,10 +684,23 @@ void ReadImagesBinary(Reconstruction& reconstruction, std::istream& stream) {
...
+      image.SetFrameId(image.ImageId());
+      image.SetFramePtr(&reconstruction.Frame(image.ImageId()));
+    } else {
+      Frame* frame = image_to_frame.at(image.ImageId());
+      image.SetFrameId(frame->FrameId());
+      image.SetFramePtr(frame);
+    }
+
+    if (image.HasTrivialFrame()) {
+      image.FramePtr()->SetFrameFromWorld(cam_from_world);
```
sarlinpe, line 701: Do I understand correctly that each image pose is now written both to the images and to the frames files? And that it is again read twice? Is this better than incrementing the file version (is there one?) and storing the pose in the frames file only?

src/colmap/scene/reconstruction_io.cc
```diff
@@ -843,6 +1115,49 @@ void WriteCamerasBinary(const Reconstruction& reconstruction,
...
+  THROW_CHECK(stream.good());
+
+  WriteBinaryLittleEndian<uint64_t>(&stream, reconstruction.NumFrames());
+
+  for (const frame_t frame_id : ExtractSortedIds(reconstruction.Frames())) {
+    const Frame& frame = reconstruction.Frame(frame_id);
+    WriteBinaryLittleEndian<frame_t>(&stream, frame_id);
+    WriteBinaryLittleEndian<rig_t>(&stream, frame.RigId());
+    WriteBinaryLittleEndian<uint8_t>(&stream, frame.HasPose() ? 1 : 0);
+    if (frame.HasPose()) {
```
sarlinpe, line 1129: `WriteImages` requires the image to be registered (otherwise CamFromWorld throws), but this doesn't - isn't this inconsistent?

sarlinpe (commented):
Thank you for this monumental effort!

src/colmap/scene/frame.h
ahojnnes: This is necessary for the same reason we need to store the camera_id separately from the camera_ptr in the image: When loading the data from the database (or from disk), we do not yet have the object available to assign the right object pointers.

src/colmap/scene/reconstruction_io.cc
ahojnnes: We don't have a file version unfortunately. I did this to be fully forwards and backwards compatible. Anybody that deserializes the image poses (e.g., for gaussian splatting), can read the files with the exact same code and get valid image poses. Eventually, we should move to a format versioning and/or protobuf. Since the pose data is so small compared to the keypoints, tracks, etc. I am not worried about the duplication.

src/colmap/scene/reconstruction_io.cc
ahojnnes: Good callout. I changed it.

PR #3214, opened 2025-03-12
Handle non-trivial frames in bundle adjustment

ahojnnes:


src/colmap/estimators/bundle_adjustment.cc
```diff
@@ -557,6 +566,96 @@ class DefaultBundleAdjuster : public BundleAdjuster {
...
+        continue;
+      }
+
+      num_observations += 1;
+      point3D_num_observations_[point2D.point3D_id] += 1;
+
+      Point3D& point3D = reconstruction.Point3D(point2D.point3D_id);
+      assert(point3D.track.Length() > 1);
+
+      // Notice that the constant_cam_pose && !constant_rig_pose is rare/exotic
```
B1ueber2y, line 605: Should we raise a not implemented error then?

src/colmap/estimators/bundle_adjustment.cc
```diff
@@ -557,6 +566,96 @@ class DefaultBundleAdjuster : public BundleAdjuster {
...
+        if (config_.HasConstantCamPositions(image.ImageId())) {
+          const std::vector<int>& constant_position_idxs =
+              config_.ConstantCamPositions(image.ImageId());
+          SetSubsetManifold(3,
+                            constant_position_idxs,
+                            problem_.get(),
+                            rig_from_world_translation);
+        }
+      }
+      if (!constant_rig_pose) {
```
B1ueber2y, line 653: But where does the constant rig pose get set? The whole parameterization of rig pose seems to be missing. Also, the current implementation set the quaternion manifold of ``cam_from_rig_rotation`` for each image, which will lead to multiple calls of ``SetQuaternionManifold`` on the rig extrinsics.

ahojnnes: I think redundantly setting the manifold is okay even though there may be a more efficient way to do it. One would have to benchmark whether it's cheaper to do the bookkeeping vs. doing it redundantly. Without the numbers, I tend to do the simpler thing (i.e., no bookeeping).

src/colmap/estimators/bundle_adjustment.cc
```diff
@@ -540,23 +639,110 @@ class DefaultBundleAdjuster : public BundleAdjuster {
...
+    double* frame_from_world_translation = frame_from_world.translation.data();
+    double* camera_params = camera.params.data();
+
+    const bool constant_sensor_from_rig =
+        !options_.refine_sensor_from_rig ||
+        config_.HasConstantSensorFromRigPose(image.CameraPtr()->SensorId());
+    const bool constant_frame_from_world =
+        !options_.refine_frame_from_world ||
+        config_.HasConstantFrameFromWorldPose(image.FrameId());
+    const bool constant_pose_cost_functor =
```
B1ueber2y, line 678: The naming of these two flags is a bit confusing to me. We could at least add some comments here?

src/colmap/estimators/bundle_adjustment.cc
```diff
@@ -540,23 +639,110 @@ class DefaultBundleAdjuster : public BundleAdjuster {
...
+        continue;
+      }
+
+      num_observations += 1;
+      point3D_num_observations_[point2D.point3D_id] += 1;
+
+      Point3D& point3D = reconstruction.Point3D(point2D.point3D_id);
+      THROW_CHECK_GT(point3D.track.Length(), 1);
+
+      // Notice that the constant_frame_from_world && !constant_sensor_from_rig
```
B1ueber2y, line 696: Again, should we throw an error when the rare/exotic case happens?

src/colmap/estimators/bundle_adjustment.cc
```diff
@@ -540,23 +639,110 @@ class DefaultBundleAdjuster : public BundleAdjuster {
...
+            camera_params);
+      }
+    }
+
+    if (num_observations > 0) {
+      parameterized_camera_ids_.insert(image.CameraId());
+      parameterized_image_ids_.insert(image.ImageId());
+
+      // Set pose parameterization.
+      if (constant_frame_from_world && !constant_pose_cost_functor) {
```
B1ueber2y, line 734: Why ``!constant_pose_cost_functor``? We would still need to fix ``frame_from_world`` for that?

src/colmap/estimators/bundle_adjustment.cc
```diff
@@ -540,23 +639,110 @@ class DefaultBundleAdjuster : public BundleAdjuster {
...
+        problem_->SetParameterBlockConstant(frame_from_world_rotation);
+        problem_->SetParameterBlockConstant(frame_from_world_translation);
+      } else {
+        SetQuaternionManifold(problem_.get(), frame_from_world_rotation);
+      }
+      if (!constant_sensor_from_rig) {
+        SetQuaternionManifold(problem_.get(), cam_from_rig_rotation);
+      }  // else: special cased cost function above.
```
B1ueber2y, line 742: Do we fix ``cam_from_rig`` somewhere when it s constant? Maybe I overlooked something but I did not find it.

src/colmap/estimators/bundle_adjustment.cc
```diff
@@ -419,22 +422,104 @@ void ParameterizeCameras(const BundleAdjustmentOptions& options,
...
+    } else if (image1 != nullptr && image1->FrameId() != image.FrameId() &&
+               image.FramePtr()->RigPtr()->IsRefSensor(
+                   image.CameraPtr()->SensorId())) {
+      image2 = &image;
+      break;
+    }
+  }
+
+  // TODO(jsch): Once we support IMUs or other sensors, we have to
+  // fix the Gauge differently, as we are not guaranteed to find two
```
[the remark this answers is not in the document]
ahojnnes: I assume, this comment doesn't suggest to change anything but is a confirmation of what is already written.

src/colmap/estimators/bundle_adjustment.cc
```diff
@@ -419,22 +422,104 @@ void ParameterizeCameras(const BundleAdjustmentOptions& options,
...
+  // images/cameras that are reference sensors in different frames.
+  THROW_CHECK(image1 != nullptr && image2 != nullptr);
+
+  Rigid3d& cam1_from_world = image1->FramePtr()->FrameFromWorld();
```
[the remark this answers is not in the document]
ahojnnes: Done

src/colmap/estimators/bundle_adjustment.cc
```diff
@@ -419,22 +422,104 @@ void ParameterizeCameras(const BundleAdjustmentOptions& options,
   }
 }
 
+void ParameterizeImages(const BundleAdjustmentOptions& options,
```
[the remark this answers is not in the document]
ahojnnes: > Given that we now have this function, maybe we can move all the QuaternionManifold parameterization and SetParameterBlockConstant to here as well? It does not hurt to parameterize even if num_observations == 0.

Ceres has checks that prevent setting a manifold when the parameters are not part of the problem. I can attempt to move the parameterization of image poses to this function, even though it'll lead to some code duplication.

> I suggest to make this parameterization function a protected method in the BundleAdjuster, as we did for ParameterizeCameras and ParamerterizePoints. And after moving in the manifold parameterization and set constant, we will only add residuals without doing any parameterization and set constant in the AddImages, which would be much cleaner and readable IMO.

Not sure I understand the proposal. ParameterizeCameras and ParamerterizePoints are free-standing functions in anonymous namespace equivalent to the new ParameterizeImages function.

src/colmap/sfm/incremental_mapper.cc
```diff
@@ -464,6 +464,7 @@ IncrementalMapper::AdjustLocalBundle(
   // Do the bundle adjustment only if there is any connected images.
   if (local_bundle.size() > 0) {
     BundleAdjustmentConfig ba_config;
+    ba_config.FixGauge(BundleAdjustmentGauge::THREE_POINTS);
```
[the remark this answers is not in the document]
ahojnnes: It'll behave differently but I think it should be overall superior to the previous behavior. As you correctly point out, the previous behavior actually lead to an over-constraining of the Gauge, because typically there would already be a lot of constant points at the "boundary" of the local bundle.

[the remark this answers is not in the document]
ahojnnes: I'll evaluate the impact on the ETH3D benchmark.

src/colmap/estimators/bundle_adjustment.cc
ahojnnes: The naming matches the cost functions, which are a bit confusing, agreed.

src/colmap/estimators/bundle_adjustment.cc
ahojnnes: I don't think it makes sense to throw an exception. It all works as expected - it's just a little less efficient as with a customized cost function for the special case.

src/colmap/estimators/bundle_adjustment.cc
ahojnnes: It's handled correctly as far as I can see. It is now refactored and handled in ParameterizeImages() which is hopefully more readable.

src/colmap/estimators/bundle_adjustment.cc
```diff
@@ -419,21 +422,214 @@ void ParameterizeCameras(const BundleAdjustmentOptions& options,
...
+                        const std::set<image_t>& image_ids,
+                        Reconstruction& reconstruction,
+                        ceres::Problem& problem) {
+  std::unordered_set<sensor_t> parameterized_sensor_ids;
+  std::unordered_set<frame_t> parameterized_frame_ids;
+  for (const image_t image_id : image_ids) {
+    Image& image = reconstruction.Image(image_id);
+    if (image.HasTrivialFrame()) {
+      // Parameterize cam_from_world (==frame_from_world).
+      if (parameterized_frame_ids.insert(image.FrameId()).second) {
```
B1ueber2y, line 498: Duplicate code. This block is always needed and does not depend on whether it is a trivial frame or not. Is it for readability purpose?

src/colmap/estimators/bundle_adjustment.cc
```diff
@@ -419,21 +422,214 @@ void ParameterizeCameras(const BundleAdjustmentOptions& options,
...
+      // Parameterize sensor_from_rig.
+      const sensor_t sensor_id = image.CameraPtr()->SensorId();
+      if (parameterized_sensor_ids.insert(sensor_id).second) {
+        Rigid3d& sensor_from_rig =
+            image.FramePtr()->RigPtr()->SensorFromRig(sensor_id);
+        // CostFunction assumes unit quaternions.
+        sensor_from_rig.rotation.normalize();
+        if (problem.HasParameterBlock(
+                sensor_from_rig.rotation.coeffs().data())) {
+          SetQuaternionManifold(&problem,
```
B1ueber2y, line 525: nit: Do you think it is makes sense to do the ``HasParameterBlock`` check in the ``SetManifold`` functions the colmap? We can also create a ``SetParameterBlockConstant`` function similarly.

B1ueber2y (approved):
Thanks a lot! LGTM.

src/colmap/estimators/bundle_adjustment.cc
ahojnnes: Done

src/colmap/estimators/bundle_adjustment.cc
ahojnnes: I'd like to avoid that so we don't silently swallow errors. In this specific use case, we only want to set the manifold when it was added to the problem. However, in other cases, this behavior is not desirable (imo).

B1ueber2y: Okay. This makes sense to me. We could still add function like SetParameterBlockConstantIfPresent though xD.

src/colmap/sfm/incremental_mapper.cc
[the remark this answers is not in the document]
ahojnnes: @B1ueber2y A/B comparison against main branch. Seems to be in the order of random noise.

```
I20250317 21:43:15.484681 807426 compare.py:main:58] Results A - B:

botanical_garden   0.05   0.02   0.00   0.00      0     0    0       0
boulders           0.07   0.03   0.01   0.00      0     0    0       0
bridge             0.02   0.01   0.00   0.00      0     0    0       0
courtyard          1.00   0.62   0.14   0.07      0     0    0       0
delivery_area      0.01   0.01   0.00   0.00      0     0    0       0
door               0.04   0.06   0.01   0.01      0     0    0       0
electro            0.02   0.02   0.00   0.00      0     0    0       0
exhibition_hall   -5.61 -11.14 -16.90 -17.78      0     0   -1       0
facade             0.00   0.00   0.00   0.00      0     0    0       0
kicker             1.25   0.63   0.13   0.06      0     0    0       0
lecture_room      -0.11  -0.07  -0.02  -0.01      0     0    0       0
living_room       -0.11  -0.08  -0.02  -0.01      0     0    0       0
lounge             0.04   0.02   0.00   0.00      0     0    0       0
meadow            -1.57  -1.06  -0.21  -0.10      0     0    0       0
observatory        0.32   0.18   0.04   0.02      0     0    0       0
office            -1.12  -2.23  -4.67  -5.13     -1     0    0      -1
old_computer      -0.42  -0.19  -0.04  -0.02      0     0    0       0
pipes              0.71   0.37   0.07   0.04      0     0    0       0
playground        -0.19  -0.10  -0.02  -0.01      0     0    0       0
relief            -0.04  -0.02  -0.00  -0.00      0     0    0       0
relief_2          -0.14  -0.05  -0.01  -0.01      0     0    0       0
statue            -0.04  -0.02  -0.00  -0.00      0     0    0       0
terrace           -2.54  -1.40  -0.28  -0.14      0     0    0       0
terrace_2          0.13   0.07   0.01   0.01      0     0    0       0
terrains          -0.20  -0.11  -0.02  -0.01      0     0    0       0
overall           -0.60  -1.16  -1.77  -1.87     -1     0   -1      -1
average           -0.34  -0.58  -0.87  -0.92      0     0    0       0
```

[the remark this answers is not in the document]
B1ueber2y: Thinking again on it, do we really need to fix the Gauge for the local bundle adjustment? Now we fix three points which can be potentially already constrained by the boundary points in the local BA. This will lead to noise at point fixing. 

Also, fixing three points is more likely to introduce noise than fixing two cameras (6 + 1) as it fixed more degrees of freedom (3x3 = 9) than needed, where the three points are very likely starting at a quite noisy initialization.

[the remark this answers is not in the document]
ahojnnes: If you look at the implementation, it only fixes three points, if not already three other points have been fixed (that have rank 3).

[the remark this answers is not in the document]
B1ueber2y: I see. I missed that. Great then!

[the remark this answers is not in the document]
ahojnnes: If we wanted to be extra safe, we could fix two cameras when the local bundle contains all images in the beginning? WDYT?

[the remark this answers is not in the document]
B1ueber2y: Yes but I guess it probably does not matter practically though. 

IMO it might be already safe to just drop the Gauge in local bundle. I believe it may not affect much the performance there. In my experience even without Gauge fixing in global bundle it generally produces marginal difference (as long as we dont hit numerical issues).

PR #3156, opened 2025-02-13
fix pip install error

taomiao:
fix #2986
fixed install pycolmap error

PR #3217, opened 2025-03-13
Change the root of the Python package

sarlinpe:
Copy https://github.com/colmap/colmap/pull/3156, as there seems to be a problem with CI checks.


ahojnnes (approved):
Great, thanks.

PR #3220, opened 2025-03-14
Export optimized camera rig configuration when running rigba

binbin-xu:
The current rigba can refine the relative poses between the camera and the rig. As a result, the input configuration may change after this optimization. Exporting the optimized rig calibration file will facilitate downstream tasks


ahojnnes:
I appreciate your contribution, however, we are in the process of enabling proper rig support very soon. I'd like to avoid adding more functionality to the legacy rig support. Thank you.

PR #3238, opened 2025-03-27
WIP: Support rigs/frames in incremental mapper

ahojnnes:
First proof of concept leveraging rig constraints in incremental mapping. Known TODOs:
- Use generalized absolute/relative pose solvers in RegisterInitialImagePair / RegisterNextImage
- Fix scene normalization to not estimate scale when rig calibration exists.
- Fix gauge freedom for rigs, especially when rigs are under-constrained.
- FindLocalBundle needs to consider rigs.
-

### DIFF
--- a/.github/workflows/build-ubuntu.yml
+++ b/.github/workflows/build-ubuntu.yml
@@ -71,26 +71,6 @@ jobs:
             checkCodeFormat: false,
             coverageEnabled: false,
           },
-          {
-            os: ubuntu-20.04,
-            cmakeBuildType: Release,
-            asanEnabled: false,
-            guiEnabled: true,
-            cudaEnabled: false,
-            e2eTests: false,
-            checkCodeFormat: false,
-            coverageEnabled: false,
-          },
-          {
-            os: ubuntu-20.04,
-            cmakeBuildType: Release,
-            asanEnabled: false,
-            guiEnabled: false,
-            cudaEnabled: true,
-            e2eTests: false,
-            checkCodeFormat: false,
-            coverageEnabled: false,
-          },
         ]
 
     env:
--- a/.github/workflows/build-ubuntu.yml
+++ b/.github/workflows/build-ubuntu.yml
@@ -108,7 +108,6 @@ jobs:
           key: v${{ env.COMPILER_CACHE_VERSION }}-${{ matrix.config.os }}-${{ matrix.config.cmakeBuildType }}-${{ matrix.config.asanEnabled }}--${{ matrix.config.cudaEnabled }}-${{ github.run_id }}-${{ github.run_number }}
           restore-keys: v${{ env.COMPILER_CACHE_VERSION }}-${{ matrix.config.os }}-${{ matrix.config.cmakeBuildType }}-${{ matrix.config.asanEnabled }}--${{ matrix.config.cudaEnabled }}
           path: ${{ env.COMPILER_CACHE_DIR }}
-      
       - name: Install compiler cache
         run: |
           mkdir -p "$CCACHE_DIR" "$CTCACHE_DIR"
@@ -248,6 +247,8 @@ jobs:
           Xvfb :99 &
           sleep 3
           cd build
+          export GLOG_v=1
+          export GLOG_logtostderr=1
           ctest -E "$ctestExclusions" --output-on-failure
 
       - name: Run E2E tests
@@ -295,7 +296,6 @@ jobs:
           indicators: true
           output: both
           thresholds: '75 90'
-      
       # TODO: Add code coverage comment to PR. Currently, this action reports
       # coverage for the entire repository, not just the changed files in the PR.
       # We could manually filter the coverage report to only include the changed
--- a/src/colmap/estimators/gauge.cc
+++ b/src/colmap/estimators/gauge.cc
@@ -1,0 +1,126 @@
+// Fixes the gauge by setting the rotation and translation of one camera as
+// constant and one translational dimension of another camera. Notice that there
+// are some degenerate cases where the current logic fails to properly fix the
+// gauge (e.g., when the observations in the image are on a line, etc.).
+void FixGaugeWithTwoCamsFromWorld(const BundleAdjustmentOptions& options,
+                                  const BundleAdjustmentConfig& config,
+                                  const std::set<image_t>& image_ids,
+                                  Reconstruction& reconstruction,
+                                  ceres::Problem& problem) {
+  // Check if two camera poses are already constant. Then, we don't need to
+  const size_t num_constant_images = std::count_if(
+      image_ids.begin(),
+      image_ids.end(),
+      [&config, &reconstruction](image_t image_id) {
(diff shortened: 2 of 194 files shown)


src/colmap/estimators/gauge.cc
```diff
@@ -0,0 +1,126 @@
+// Fixes the gauge by setting the rotation and translation of one camera as
```
ahojnnes, line 1: Temp commit. Remove.

.github/workflows/build-ubuntu.yml
```diff
@@ -71,26 +71,6 @@ jobs:
             checkCodeFormat: false,
             coverageEnabled: false,
           },
-          {
```
ahojnnes, line 74: Restore.

src/colmap/exe/database.cc
```diff
@@ -110,4 +115,262 @@ int RunDatabaseMerger(int argc, char** argv) {
...
+//           {
+//               "image_prefix": "images_rig_cam6_undistorted/"
+//           },
+//           {
+//               "image_prefix": "images_rig_cam7_undistorted/"
+//           }
+//       ]
+//   }
+// ]
+int RunRigConfigurator(int argc, char** argv) {
```
ahojnnes, line 229: Extract into helper function and add test.

pyproject.toml
```diff
@@ -45,7 +45,7 @@ test-requires = "mypy==1.12.1"
 test-command = "python -c \"import pycolmap; print(pycolmap.__version__)\" &&  python -m mypy --package pycolmap --implicit-optional"
 
 [tool.cibuildwheel.environment]
-VCPKG_COMMIT_ID = "3b57fb2e1ff55613db14d2aaf0a30529289c7050"
+VCPKG_COMMIT_ID = "460551b0ec06be1ba6b918448bf3b0f44add813d"
```
B1ueber2y, line 48: Should we make this change in a separate commit? The current CI failure seems to be relevant to this change?

ahojnnes: Makes sense. Done.

src/colmap/estimators/gauge.cc
```diff
@@ -0,0 +1,126 @@
...
+// constant and one translational dimension of another camera. Notice that there
+// are some degenerate cases where the current logic fails to properly fix the
+// gauge (e.g., when the observations in the image are on a line, etc.).
+void FixGaugeWithTwoCamsFromWorld(const BundleAdjustmentOptions& options,
+                                  const BundleAdjustmentConfig& config,
+                                  const std::set<image_t>& image_ids,
+                                  Reconstruction& reconstruction,
+                                  ceres::Problem& problem) {
+  // Check if two camera poses are already constant. Then, we don't need to
+  const size_t num_constant_images = std::count_if(
```
B1ueber2y, line 11: nit: this can benefit from the ``has_constant_cam_from_world`` below.

src/colmap/scene/reconstruction.cc
```diff
@@ -551,10 +609,10 @@ double Reconstruction::ComputeMeanTrackLength() const {
 }
 
 double Reconstruction::ComputeMeanObservationsPerRegImage() const {
-  if (NumRegImages() == 0) {
+  if (NumRegFrames() == 0) {
     return 0.0;
   } else {
-    return ComputeNumObservations() / static_cast<double>(NumRegImages());
+    return ComputeNumObservations() / static_cast<double>(NumRegFrames());
```
B1ueber2y, line 615: This update should correspond to a newly created method ``ComputeMeanObservationsPerRegFrame``.

src/colmap/scene/reconstruction_test.cc
```diff
@@ -158,8 +159,8 @@ TEST(Reconstruction, Print) {
   std::ostringstream stream;
   stream << reconstruction;
   EXPECT_EQ(stream.str(),
-            "Reconstruction(num_rigs=1, num_cameras=1, num_images=2, "
-            "num_reg_images=2, num_points3D=3)");
+            "Reconstruction(num_rigs=1, num_cameras=1, num_frames=2, "
```
B1ueber2y, line 162: Should we add more logics here such that we dont print rigs and frames for the cases where all images are trivial frames. This can still be very common in the practical use cases since most users are using COLMAP without rigs for now.

B1ueber2y (approved):
LGTM. Exciting efforts! Left some minor comments.

For the whole development,  we might also benefit from a regression test to make sure that there is negligible change when we run the new rig incremental mapper in the non-rig setup, where all images are trivial frames.

src/colmap/scene/reconstruction_test.cc
ahojnnes: I don't think there is much harm in printing out a bit more information in case all frames are trivial. I'd keep it as is.

src/colmap/scene/reconstruction.h
```diff
@@ -158,14 +162,11 @@ class Reconstruction {
...
-  // Register an existing image.
-  void RegisterImage(image_t image_id);
+  // Register an existing frame.
+  void RegisterFrame(frame_t frame_id);
 
-  // De-register an existing image, and all its references.
-  void DeRegisterImage(image_t image_id);
-
-  // Check if image is registered.
-  inline bool IsImageRegistered(image_t image_id) const;
```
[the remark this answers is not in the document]
ahojnnes: Will extract these in a separate PR against the main branch.

src/colmap/estimators/gauge.cc
ahojnnes: This file was now removed. Not meant to be checked in. Thanks.

src/pycolmap/sfm/incremental_mapper.cc
```diff
@@ -376,8 +379,8 @@ void BindIncrementalMapperImpl(py::module& m) {
           "find_initial_image_pair",
           [](IncrementalMapper& self,
              const IncrementalMapper::Options& options,
-             int image_id1,
-             int image_id2)
+             image_t image_id1,
```
B1ueber2y, line 382: This broke the python pipeline. There is the explicit handling in the code block from int to uint32_t. This is because ``options.init_image_id1`` in Python is auto cast to int type.

src/colmap/scene/database_cache_test.cc
```diff
@@ -122,47 +125,86 @@ TEST(DatabaseCache, ConstructFromDatabase) {
...
+
+  const std::vector<Image> images = database.ReadAllImages();
+  auto cache = DatabaseCache::Create(database,
+                                     /*min_num_matches=*/0,
+                                     /*ignore_watermarks=*/false,
+                                     /*image_names=*/{images[0].Name()});
+
+  EXPECT_EQ(cache->NumRigs(), 1);
+  EXPECT_EQ(cache->NumCameras(), 2);
+  EXPECT_EQ(cache->NumFrames(), 2);
```
B1ueber2y, line 193: This should be 1 instead of 2?

PR #3248, opened 2025-04-06
Fix chained match synthesis

ahojnnes:
The chained match synthesis was meant to work differently than currently implemented, which would create a separate chain per 3D point track. This PR creates a consistent chain across all tracks leading to sufficient matches for downstream tests. Noticed the issue while observing non-deterministic behavior in our e2e pipeline tests depending on order of tests.


sarlinpe (approved):

B1ueber2y:
Thanks. What if the image ids are not consecutive, where there are some image ids that are not present in the database?

ahojnnes:
> Thanks. What if the image ids are not consecutive, where there are some image ids that are not present in the database?

This cannot really happen in the current synthesis (unless the number of outlier reprojections is tuned up significantly). In this case, it would be expected to then also have no matches in the chain.

PR #3249, opened 2025-04-06
Fix missing PRNG initialization

ahojnnes:
In a few places, we directly dereference the PRNG pointer for use in std::shuffle. This leads to null pointer dereference if the PRNG was not initialized before, which requires an explicit call to SetPRGNSeed. In most conditions this would be done by a previous call to the random number generator helpers but not under some other conditions.


ahojnnes:
This PR will not pass because it fails the two-view estimation tests. I subsumed the changes into the other PR.

PR #3250, opened 2025-04-07
Fix two-view geometry pose estimation for homography

ahojnnes:
Fixing the unrelated issue of correctly initializing the PRNG in our synthetic scene generator revealed issues with homography pose estimation. Out of the 4 decomposed solutions, the cheirality check can only eliminate 2 and we would randomly pick one of them. I added a reprojection check to resolve the ambiguity. This PR addresses both issues jointly. Previously, on synthetically generated image pairs, half of the pairs would fail to correctly estimate the relative pose from the homography. Now, we achieve 100% success rate. I also added more test coverage.

### DIFF
--- a/src/colmap/estimators/two_view_geometry_test.cc
+++ b/src/colmap/estimators/two_view_geometry_test.cc
@@ -144,6 +144,7 @@ TwoViewGeometryTestData CreateTwoViewGeometryTestData(
 TEST(EstimateTwoViewGeometryPose, Calibrated) {
   constexpr int kNumTests = 100;
   for (int seed = 0; seed < kNumTests; ++seed) {
+    SetPRNGSeed(seed);
     const TwoViewGeometryTestData test_data = CreateTwoViewGeometryTestData(
         TwoViewGeometry::ConfigurationType::CALIBRATED);
 
@@ -169,6 +170,7 @@ TEST(EstimateTwoViewGeometryPose, Calibrated) {
 }
 
 TEST(EstimateTwoViewGeometryPose, FailureDueToInsufficientMatches) {
+  SetPRNGSeed(0);
   for (const auto config : {TwoViewGeometry::ConfigurationType::CALIBRATED,
                             TwoViewGeometry::ConfigurationType::UNCALIBRATED,
                             TwoViewGeometry::ConfigurationType::PLANAR,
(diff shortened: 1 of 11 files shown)


src/colmap/estimators/two_view_geometry_test.cc
```diff
@@ -154,97 +142,134 @@ TwoViewGeometryTestData CreateTwoViewGeometryTestData(
...
+        geometry.cam2_from_cam1.translation,
+        EigenMatrixNear(
+            test_data.geometry.cam2_from_cam1.translation.normalized(), 1e-6));
+  }
 }
 
 TEST(EstimateTwoViewGeometryPose, Planar) {
-  const TwoViewGeometryTestData test_data =
-      CreateTwoViewGeometryTestData(TwoViewGeometry::ConfigurationType::PLANAR);
+  constexpr int kNumTests = 100;
```
ahojnnes, line 220: This test would fail in ~50% of the times.

src/colmap/geometry/homography_matrix.cc
```diff
@@ -202,9 +247,17 @@ void PoseFromHomographyMatrix(const Eigen::Matrix3d& H,
 
   points3D->clear();
   std::vector<Eigen::Vector3d> tentative_points3D;
+  double best_reproj_error_sum = std::numeric_limits<double>::max();
   for (size_t i = 0; i < cams2_from_cams1.size(); ++i) {
-    CheckCheirality(cams2_from_cams1[i], points1, points2, &tentative_points3D);
-    if (tentative_points3D.size() >= points3D->size()) {
+    // Note that we can typically eliminate 2 of the 4 solutions using the
```
ahojnnes, line 252: This was the main underlying issue. Before we would randomly pick 1 of the 2 remaining solutions but it was wrong half of the times.

src/colmap/math/math.h
```diff
@@ -166,7 +167,7 @@ bool NextCombination(Iterator first1,
 
 template <typename T>
 int SignOfNumber(const T val) {
-  return (T(0) < val) - (val < T(0));
+  return val >= 0 ? 1 : -1;
```
ahojnnes, line 170: Only used in the homography decomposition. There are some rare edge cases, where the previous implementation would return 0 (~2-3% of cases) and then the decomposition falsely returns identical rotations for all 4 solutions.

B1ueber2y (approved):
Thanks. LGTM

sarlinpe (approved):
Thank you very much!

PR #3257, opened 2025-04-10
Include Boost headers on build

jonahjnewton:
When building, I got an 'include not found' error for the `boost/preprocessor.hpp` header-only library, as it seems the Boost headers are not included when building COLMAP. This simple PR includes the Boost headers and fixes this issue.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -238,6 +238,7 @@ endif()
 include(GenerateVersionDefinitions)
 
 include_directories(src)
+include_directories(${COLMAP_INCLUDE_DIRS})
 link_directories(${COLMAP_LINK_DIRS})
 
 add_subdirectory(src/colmap)
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -238,7 +238,6 @@ endif()
 include(GenerateVersionDefinitions)
 
 include_directories(src)
-include_directories(${COLMAP_INCLUDE_DIRS})
 link_directories(${COLMAP_LINK_DIRS})
 
 add_subdirectory(src/colmap)
--- a/cmake/FindDependencies.cmake
+++ b/cmake/FindDependencies.cmake
@@ -126,6 +126,7 @@ if(NOT FETCH_POSELIB)
     find_package(PoseLib ${COLMAP_FIND_TYPE})
 endif()
 
+set(COLMAP_INCLUDE_DIRS ${Boost_INCLUDE_DIRS})
 set(COLMAP_LINK_DIRS ${Boost_LIBRARY_DIRS})
 
 set(CUDA_MIN_VERSION "7.0")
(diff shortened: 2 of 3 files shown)


ahojnnes:
Thank you. Could you please share under which OS and CMake/Boost versions this issue occurs?

ahojnnes:
I assume this happens because of a bug in the specific boost/cmake version under RHEL9. What's not great is that we now add the boost include directory to all colmap targets even if the targets don't depend on boost. In the cmake/FindDependencies.cmake, we can probably add the include directory to the imported boost target manually?  Can you try whether the following works:
```
find_package(Boost ${COLMAP_FIND_TYPE} COMPONENTS
             graph
             program_options
             system)
if(TARGET Boost::boost)
    # Fix missing include directory definition for specific CMake/Boost versions.
    target_link_libraries(Boost::boost INTERFACE ${Boost_INCLUDE_DIRS})
    target_link_libraries(Boost::graph INTERFACE ${Boost_INCLUDE_DIRS})
    target_link_libraries(Boost::program_options INTERFACE ${Boost_INCLUDE_DIRS})
endif()
```

jonahjnewton:
Yep, moving the Boost::boost to being a PUBLIC link for the colmap_util target fixes the issue as well!

jonahjnewton:
Actually, upon further digging and looking at other files that link against `colmap_util` (such as `src/colmap/math/CMakeLists.txt`), it just seems to be happenstance that `colmap_sensor` (`src/colmap/sensor/CMakeLists.txt`) is the only target that does not also independently link against a Boost library, and thus the Boost headers (causing this error). 

Only the **headers** are required publicly when linking against `colmap_util` for the header-only libs, so a more minimal approach would be to keep the `Boost::boost` include `PRIVATE` for `colmap_util`, and include `Boost::headers` publicly for targets that include `colmap_util`.

I've updated the PR.

```
COLMAP_ADD_LIBRARY(
    NAME colmap_util
    SRCS
        base_controller.h base_controller.cc
        cache.h
        controller_thread.h
        eigen_alignment.h
        endian.h endian.cc
        enum_utils.h
        file.h file.cc
        logging.h logging.cc
        misc.h misc.cc
        opengl_utils.h opengl_utils.cc
        ply.h ply.cc
        sqlite3_utils.h
        string.h string.cc
        threading.h threading.cc
        timer.h timer.cc
        types.h
        version.h version.cc
    PUBLIC_LINK_LIBS
        Eigen3::Eigen
        glog::glog
        SQLite::SQLite3
        Boost::headers
    PRIVATE_LINK_LIBS
        Boost::boost
        $<$<AND:$<CXX_COMPILER_ID:GNU>,$<VERSION_LESS:$<CXX_COMPILER_VERSION>,9.0>>:-lstdc++fs>
)
```

jonahjnewton:
My mistake. Boost::headers is just an alias for Boost::boost. Rolling that back.

ahojnnes (approved):

PR #3265, opened 2025-04-16
Use generalized absolute pose estimation for non-trivial frames

ahojnnes:


src/colmap/sfm/incremental_mapper.cc
```diff
@@ -175,26 +176,47 @@ void IncrementalMapper::RegisterInitialImagePair(
...
+           reg_stats_.num_reg_images_per_camera[frame_image.CameraId()] == 0) ||
+          frame_image.CameraPtr()->HasBogusParams(
+              options.min_focal_length_ratio,
+              options.max_focal_length_ratio,
+              options.max_extra_param)) {
+        all_cameras_have_good_focal_length = false;
+        break;
+      }
+    }
+    if (all_cameras_have_good_focal_length) {
```
B1ueber2y, line 206: Should we also check whether the sensor_from_rig transformation for the corresponding rig has been initialized to a reasonable value?

src/colmap/sfm/incremental_mapper.cc
```diff
@@ -175,26 +176,47 @@ void IncrementalMapper::RegisterInitialImagePair(
   RegisterFrameEvent(image2.FrameId());
 }
 
-// TODO(jsch): Better handle non-trivial frames by performing generalized
-// absolute pose estimation.
 bool IncrementalMapper::RegisterNextImage(const Options& options,
```
B1ueber2y, line 179: In the long term, does it make sense to operate fully on the frame level, where we have ``FindNextFrames`` and ``RegisterNextFrame``?

src/colmap/sfm/incremental_mapper.cc
ahojnnes: At the moment, the pipeline won't work at all if the sensor_from_rig poses are unknown. I am not sure it makes sense to support this scenario. If somebody has unknown sensor_from_rig poses, I would suggest first running colmap with trivial frames and then using the rig_configurator to derive the average rig poses from the images. Does that make sense?

B1ueber2y: Why not use single-image registration when the sensor_from_rig is not set, and initialize the rig extrinsics when we accumulate a frame with each of the data registered?

I thought we plan to support unknown sensor_from_rig poses since this can be useful for non-experienced user, and was in the comments here as well: https://github.com/colmap/colmap/blob/feature/rig/src/colmap/exe/database.cc#L217. The only thing we need to be careful about in such unknown rig extrinsic case is the scale of the sensor_from_rig transformation, which is ambiguous due to Gauge but globally consistant across different rigs.

ahojnnes: What is wrong with the approach described above? Run colmap twice. Once with no rig constraints and then with rig constraints and poses derived from the first run?

src/colmap/sfm/incremental_mapper.cc
```diff
@@ -257,6 +286,8 @@ bool IncrementalMapper::RegisterNextImage(const Options& options,
   // hence we skip some of the 2D-3D correspondences.
   if (tri_points2D.size() <
       static_cast<size_t>(options.abs_pose_min_num_inliers)) {
+    VLOG(2) << "Insufficient number of " << tri_points2D.size()
```
B1ueber2y, line 289: nit: "Insufficient number of xx 2D-3D correspondences" -> "Insufficient number of 2D-3D correspondences (xx < xx) for registration"

src/colmap/sfm/incremental_mapper.cc
```diff
@@ -321,10 +364,13 @@ bool IncrementalMapper::RegisterNextImage(const Options& options,
                             &camera,
                             &num_inliers,
                             &inlier_mask)) {
+    VLOG(2) << "Absolute pose estimation failed";
     return false;
   }
 
   if (num_inliers < static_cast<size_t>(options.abs_pose_min_num_inliers)) {
+    VLOG(2) << "Absolute pose estimation failed due to insufficient "
+            << num_inliers << " inliers";
```
B1ueber2y, line 373: nit: -> due to insufficient inliers (xx < xx).

B1ueber2y (approved):
Thanks. LGTM

PR #3269, opened 2025-04-18
Add changes needed for MP-SfM

Zador-Pataki:
### Summary
This PR introduces the core functionality required for MP-SfM. It adds new Ceres cost functions and integrates them into COLMAP’s pipeline via ExtendBundleAdjusterWithDepth, with full Python bindings. Additionally, it includes several utility functions to improve the robustness and efficiency of the MP-SfM pipeline, particularly in challenging reconstruction scenarios.

### Main Changes

- Added two new Ceres cost functions: ScaledDepthErrorCostFunction and TruncatedLogScaledDepthErrorCostFunction
- ExtendBundleAdjusterWithDepth: adds per-3D point depth constraints efficiently to an existing BA problem 
- Utility pybindings:
  - get_xy and get_point3D_ids methods in Image avoid iterating in python 
  - FindPoints3DWithGoodTriangulationAngle: follows "FilterPoints3DWithSmallTriangulationAngle" but instead returns the points. Needed in MP-SfM and is inefficient in python. However, it is unclear if this should be in observation manager. 
- Extended Retringulate() with optional exclusion of specific image IDs. This is needed because for low-overlap pairs, introducing high error 3D points can be harmful

### DIFF
--- a/python/pyproject.toml
+++ b/python/pyproject.toml
@@ -1,0 +1,57 @@
+[build-system]
+requires = [
+    "scikit-build-core>=0.3.3",
+    "pybind11==2.13.6",
+    "pybind11_stubgen @ git+https://github.com/sarlinpe/pybind11-stubgen@sarlinpe/fix-2024-11",
+    "numpy",
+    "ruff==0.6.7",
+    "clang-format==19.1.0",
+]
+build-backend = "scikit_build_core.build"
+
+[project]
+name = "pycolmap"
+# WARNING: This version must follow the MAJOR.MINOR.PATCH format. If only
+# MAJOR.MINOR is used, cibuildwheel will add a .dev0 patch version, which
+# results in releasing a pre-release version on PyPI.
+version = "3.12.0.dev0"
+description = "COLMAP bindings"
+readme = "README.md"
+authors = [
+  { name = "Johannes Schönberger", email = "jsch@demuc.de" },
+  { name = "Mihai Dusmanu", email = "mihai.dusmanu@gmail.com" },
+  { name = "Paul-Edouard Sarlin", email = "psarlin@ethz.ch" },
+  { name = "Shaohui Liu", email = "b1ueber2y@gmail.com" },
+  { name = "Philipp Lindenberger", email = "plindenbe@ethz.ch" },
+]
+license = {text = "BSD-3-Clause"}
+urls = {Repository = "https://github.com/colmap/colmap"}
+requires-python = ">=3.8"
+dependencies = ["numpy"]
+classifiers = [
+  "License :: OSI Approved :: BSD License",
+  "Programming Language :: Python :: 3 :: Only",
+]
+
+[tool.scikit-build]
+wheel.expand-macos-universal-tags = true
+
+[tool.cibuildwheel]
+build = "cp3{8,9,10,11,12}-{macosx,manylinux,win}*"
+archs = ["auto64"]
+test-requires = "mypy==1.12.1"
+test-command = "python -c \"import pycolmap; print(pycolmap.__version__)\" &&  python -m mypy --package pycolmap --implicit-optional"
+
+[tool.cibuildwheel.environment]
+VCPKG_COMMIT_ID = "3b57fb2e1ff55613db14d2aaf0a30529289c7050"
+
+[tool.cibuildwheel.linux]
+before-all = "{package}/ci/install-colmap-almalinux.sh"
+
+[tool.cibuildwheel.macos]
+before-all = "{package}/ci/install-colmap-macos.sh"
+
+[tool.cibuildwheel.windows]
+before-all = "pwsh -File {package}/ci/install-colmap-windows.ps1"
+before-build = "pip install delvewheel"
+test-command = "pwsh -File {package}/ci/test-colmap-windows.ps1"
--- a/src/colmap/estimators/bundle_adjustment.cc
+++ b/src/colmap/estimators/bundle_adjustment.cc
@@ -1175,4 +1175,70 @@ void PrintSolverSummary(const ceres::Solver::Summary& summary,
   LOG(INFO) << log.str();
 }
 
+ceres::LossFunction* CreateLossFunctionPerPoint(const std::string& loss_name,
+                                                  double loss_param,
+                                                  double loss_magnitude) {
+  ceres::LossFunction* lf = nullptr;
+  if (loss_name == "trivial") {
+    lf = new ceres::TrivialLoss();
+  } else if (loss_name == "soft_l1") {
+    lf = new ceres::SoftLOneLoss(loss_param);
+  } else if (loss_name == "cauchy") {
+    lf = new ceres::CauchyLoss(loss_param);
+  } else {
+    throw std::runtime_error("Unknown loss name: " + loss_name);
+  }
+  if (loss_magnitude != 1.0) {
+    lf = new ceres::ScaledLoss(lf, loss_magnitude, ceres::TAKE_OWNERSHIP);
+  }
+  return lf;
+}
+void ExtendBundleAdjusterWithDepth(ceres::Problem* problem,
+    image_t image_id,
+    const std::vector<point3D_t>& point3D_ids,
+    const std::vector<double>& depths,
+    const std::vector<double>& loss_magnitudes,
+    const std::vector<double>& loss_params,
+    const std::string& loss_name,
+    double* shift_scale_ptr,  
+    Reconstruction& reconstruction,
+    bool logloss,
+    bool fix_shift,
+    bool fix_scale) {
+  if (point3D_ids.size() != depths.size() ||
+      point3D_ids.size() != loss_magnitudes.size() ||
+      point3D_ids.size() != loss_params.size()) {
+    throw std::runtime_error("Input vectors must have the same size");
+  }
+
+  Image& image = reconstruction.Image(image_id);
+
+  double* cam_rot = image.CamFromWorld().rotation.coeffs().data();
+  double* cam_trans = image.CamFromWorld().translation.data();
+
+  for (size_t i = 0; i < point3D_ids.size(); ++i) {
+    point3D_t pt_id = point3D_ids[i];
+    double depth = depths[i];
+    double loss_magnitude = loss_magnitudes[i];
+    double loss_param = loss_params[i];
+
+    Point3D& point3D = reconstruction.Point3D(pt_id);
+
+    ceres::CostFunction* cost_function = nullptr;
+    if (logloss)
+      cost_function = TruncatedLogScaledDepthErrorCostFunction::Create(depth);
+    else
+      cost_function = ScaledDepthErrorCostFunction::Create(depth);
+
+    ceres::LossFunction* loss_function = CreateLossFunctionPerPoint(loss_name, loss_param, loss_magnitude);
+
+    problem->AddResidualBlock(cost_function,
+        loss_function,
+        cam_rot,
+        cam_trans,
+        point3D.xyz.data(),
+        shift_scale_ptr);
+    }
+}
+
 }  // namespace colmap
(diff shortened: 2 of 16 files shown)


python/pyproject.toml
sarlinpe, line 1: Unnecessary, pycolmap should now be installed with `python -m pip install .` from the top-level directory.

src/pycolmap/scene/image.cc
```diff
@@ -105,6 +105,58 @@ void BindSceneImage(py::module& m) {
            py::overload_cast<camera_t>(&Image::Point2D),
            "point2D_idx"_a,
            "Direct accessor for a point2D.")
+      .def("get_xy",
+           [](const Image& self, std::optional<std::vector<point2D_t>> point_ids = std::nullopt) -> py::array {
+               std::vector<double> xy_coords;
```
sarlinpe, line 110: Better create a `Eigen::MatrixX2d` and return it as is, pybind will handle the conversion to a numpy array.

src/pycolmap/scene/image.cc
```diff
@@ -105,6 +105,58 @@ void BindSceneImage(py::module& m) {
            py::overload_cast<camera_t>(&Image::Point2D),
            "point2D_idx"_a,
            "Direct accessor for a point2D.")
+      .def("get_xy",
```
sarlinpe, line 108: Move to the C++ Image implementation, add a test, rename to `KeypointCoordinates` or so?

src/pycolmap/scene/image.cc
```diff
@@ -105,6 +105,58 @@ void BindSceneImage(py::module& m) {
...
+                    py::format_descriptor<double>::format(),
+                    2,
+                    shape,
+                    strides
+               ));
+           },
+           "point_ids"_a = std::nullopt,
+           "Get an Nx2 numpy array of xy coordinates for the specified 2D point IDs. "
+           "If no IDs are provided, return all.")
+      .def("get_point3D_ids",
```
sarlinpe, line 140: Move to the C++ Image implementation, add a test, rename to `Point3DIds` or so?

src/pycolmap/scene/reconstruction.cc
```diff
@@ -252,5 +252,84 @@ void BindReconstruction(py::module& m) {
            << "\n\tmean_reprojection_error = "
            << self.ComputeMeanReprojectionError();
         return ss.str();
-      });
-}
+     })
+      .def("find_local_bundle_ids", [](const Reconstruction& self,
```
sarlinpe, line 256: Isn't this already implemented in the incremental mapper? If so, can't we reuse this code, making it a standalone function? I think that this shouldn't be part of the reconstruction object.

src/pycolmap/scene/reconstruction.cc
```diff
@@ -252,5 +252,84 @@ void BindReconstruction(py::module& m) {
...
+          local_bundle_image_ids.reserve(num_eff_images);
+
+          for (size_t i = 0; i < num_eff_images; ++i) {
+               local_bundle_image_ids.push_back(overlapping_images[i].first);
+          }
+
+          return local_bundle_image_ids;
+          }, py::arg("image_id"), py::arg("local_ba_num_images"),
+          "Find a local bundle of images based on shared observations with the specified image.")
+      .def("get_point3d_coordinates",
```
sarlinpe, line 307: Move to the C++ implementation file of Reconstruction, make it a `Eigen::MatrixX2d` instead.

src/colmap/estimators/cost_functions.h
```diff
@@ -530,6 +530,104 @@ class CovarianceWeightedCostFunctor {
...
+                               shift_scale,
+                               residuals);
+   }
+  private:
+   const Rigid3d& cam_from_world_;
+ };
+
+ // Cost function for constraining the depth of a point in the camera frame
+ // with depth priors, with a fixed camera pose, in log space.
+ struct TruncatedLogScaledDepthErrorCostFunction {
```
sarlinpe, line 598: Why `Truncated`? Shouldn't this be called `LogScaledDepthErrorCostFunction` instead?

src/colmap/estimators/bundle_adjustment.cc
```diff
@@ -1175,4 +1175,70 @@ void PrintSolverSummary(const ceres::Solver::Summary& summary,
   LOG(INFO) << log.str();
 }
 
+ceres::LossFunction* CreateLossFunctionPerPoint(const std::string& loss_name,
```
sarlinpe, line 1178: Seems like a duplicate of `BundleAdjustmentOptions::CreateLossFunction`? Maybe take this out of `BundleAdjustmentOptions` to reuse?

src/colmap/estimators/bundle_adjustment.cc
```diff
@@ -1175,4 +1175,70 @@ void PrintSolverSummary(const ceres::Solver::Summary& summary,
...
+
+    Point3D& point3D = reconstruction.Point3D(pt_id);
+
+    ceres::CostFunction* cost_function = nullptr;
+    if (logloss)
+      cost_function = TruncatedLogScaledDepthErrorCostFunction::Create(depth);
+    else
+      cost_function = ScaledDepthErrorCostFunction::Create(depth);
+
+    ceres::LossFunction* loss_function = CreateLossFunctionPerPoint(loss_name, loss_param, loss_magnitude);
```
sarlinpe, line 1233: Can be called only once instead of at each iteration, since the parameters are identical.

src/colmap/estimators/bundle_adjustment.h
```diff
@@ -245,6 +248,20 @@ std::unique_ptr<BundleAdjuster> CreatePosePriorBundleAdjuster(
     std::unordered_map<image_t, PosePrior> pose_priors,
     Reconstruction& reconstruction);
 
+// Adds a depth prior to the bundle adjustment problem.
+void ExtendBundleAdjusterWithDepth(ceres::Problem* problem,
```
sarlinpe, line 252: Should this be a `DepthPriorBundleAdjuster` instead, similarly to `PosePriorBundleAdjuster`? The downside is that we then cannot combine Depth and Pose priors, these interface would need to be later refactored.

src/colmap/estimators/covariance_test.cc
```diff
@@ -299,6 +299,45 @@ TEST_P(ParameterizedBACovarianceTests, CompareWithCeres) {
   }
 }
 
+TEST(ScaledDepthError, Nominal) {
```
sarlinpe, line 302: Wrong file for this test?

src/colmap/estimators/bundle_adjustment.h
```diff
@@ -117,6 +117,9 @@ struct BundleAdjustmentOptions {
 
   // Scaling factor determines residual at which robustification takes place.
   double loss_function_scale = 1.0;
+  
+  // Magnitude determines the scaling of the residual
+  double loss_function_magnitude = 1.0;
```
B1ueber2y, line 122: This added option seems useful but was not actually used in the code.

src/colmap/estimators/covariance.cc
```diff
@@ -416,6 +416,9 @@ std::vector<PoseParam> GetPoseParams(const Reconstruction& reconstruction,
   std::vector<PoseParam> params;
   params.reserve(reconstruction.NumImages());
   for (const auto& [image_id, image] : reconstruction.Images()) {
+    if (!image.HasPose()) {
```
B1ueber2y, line 419: Nice fix. Consider cherry-picking this out to a separate commit.

src/colmap/sfm/incremental_triangulator.h
```diff
@@ -134,7 +134,8 @@ class IncrementalTriangulator {
   // Image pairs are under-reconstructed if less than `Options::tri_re_min_ratio
   // > tri_ratio`, where `tri_ratio` is the number of triangulated matches over
   // inlier matches between the image pair.
-  size_t Retriangulate(const Options& options);
+  size_t Retriangulate(const Options& options,
+    const std::unordered_set<image_t>& ignore_image_ids = {});
```
B1ueber2y, line 138: This does not seem to be used anywhere?

src/pycolmap/scene/reconstruction.cc
B1ueber2y: Agreed.

src/colmap/sfm/observation_manager.cc
```diff
@@ -402,6 +402,65 @@ size_t ObservationManager::FilterPoints3DWithSmallTriangulationAngle(
   return num_filtered;
 }
 
+std::vector<bool> ObservationManager::FindPoints3DWithGoodTriangulationAngle(
```
B1ueber2y, line 405: Can this be implemented outside COLMAP, given that it is not used anywhere in COLMAP?

B1ueber2y: Now I see that this does the same thing as ``FilterPoints3DWithSmallTriangulationAngle``. Does it make sense to refactor and share some code between the two methods?

src/pycolmap/scene/image.cc
Zador-Pataki: sounds good

src/pycolmap/scene/reconstruction.cc
Zador-Pataki: Yes exactly. It's in the incremental mapper which MPSFM does not use. I will implement what is proposed.

src/colmap/estimators/cost_functions.h
Zador-Pataki: truncated because residuals for negative depths are set to 0. But yes LogScaledDepthErrorCostFunction is better

src/colmap/estimators/bundle_adjustment.h
Zador-Pataki: name "Extend.." is used because this function only adds the residuals from a single depth map. In MP-SfM, I loop over the depth maps in python. But I like DepthPriorBundleAdjuster. Let's go with this?

src/colmap/sfm/incremental_triangulator.h
Zador-Pataki: Indeed. This is used only in MP-SfM. Problem?

src/colmap/estimators/bundle_adjustment.cc
Zador-Pataki: unclear which line. loss_param, loss_magnitude and depth are all per 3D point in this implementation. the if statement could be done once but not sure how to implement this in cpp

src/pycolmap/sfm/incremental_mapper.cc
```diff
@@ -360,6 +361,14 @@ void BindIncrementalMapperImpl(py::module& m) {
       .def_readwrite("num_adjusted_observations",
                      &LocalBAReport::num_adjusted_observations);
   MakeDataclass(PyLocalBAReport);
+  
+  py::class_<IncrementalMapperImpl>(m, "IncrementalMapperImpl")
+    .def(py::init<>())
+    .def_static("find_local_bundle",
+        &IncrementalMapperImpl::FindLocalBundle,
```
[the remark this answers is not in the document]
sarlinpe: Fine by me.

src/pycolmap/sfm/observation_manager.cc
```diff
@@ -134,11 +134,29 @@ void BindObservationManager(py::module& m) {
...
-          "min_tri_angle"_a,
-          "point3D_ids"_a,
-          "Find 3D points with sufficient triangulation angle. "
-          "Returns a boolean mask.")
+      .def("find_points3D_with_small_triangulation_angle",
+           &ObservationManager::FindPoints3DWithSmallTriangulationAngle,
+           "min_tri_angle"_a,
+           "point3D_ids"_a,
+           "Return point3D IDs with triangulation angle below the threshold.")
+      .def("find_small_angle_points_mask",
```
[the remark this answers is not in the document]
sarlinpe: Comparison between two sets should be relatively fast in Python. We can't checkin find_small_angle_points_mask, this is too specific - you could simply maintain a dict of point IDs to indices? Alternatively np.searchsorted might the handy if the IDs are always sorted.

reynoldscem-oculo:
@Zador-Pataki is the intent still to get this stuff merged? I'm quite interested in using some functionality that exists in both MP-SfM and which has since been added to colmap. Wondering what the best way to do so would be.

PR #3275, opened 2025-04-23
Improved code/docs and tests for rig configuration

ahojnnes:


src/colmap/scene/rig.h
```diff
@@ -0,0 +1,131 @@
...
+// [
+//   {
+//     "cameras": [
+//         {
+//             "image_prefix": "0/",
+//             "ref_sensor": true
+//         },
+//         {
+//             "image_prefix": "1/",
+//             "cam_from_rig_rotation": [
```
B1ueber2y, line 82: Would be good to document in the header about the wxyz convention.

src/colmap/scene/rig.h
```diff
@@ -0,0 +1,131 @@
...
+//             ],
+//             "cam_from_rig_translation": [
+//                 0,
+//                 0,
+//                 0
+//             ]
+//         },
+//         {
+//             "image_prefix": "2/",
+//             "cam_from_rig_rotation": [
```
B1ueber2y, line 96: This cam_from_rig transformation seems to be identity. Should the rotation here be (0, 0, 1, 0)?

ahojnnes: Fixed. Good catch :-)

src/colmap/scene/rig.h
ahojnnes: Done

B1ueber2y (approved):

PR #3281, opened 2025-04-24
Rig bundle adjuster uses default bundle adjustment routine

ahojnnes:
Uses new rig/frame functionality to implement support for the existing rig bundle adjuster functionality. This should be fully backwards compatible. In a follow-up PR, I will delete all code related to legacy rig support (CameraRig struct, RigBundleAdjuster, etc.).


B1ueber2y (approved):
Thanks! We are finally getting there soon : )

PR #3284, opened 2025-04-25
Allow png mask without double extension

MotivaCG:
By default, COLMAP searches for mask files by appending ".png" to the image filename. However, this can lead to unexpected behavior when the original image already has a ".png" extension. For example, an image named "image0001.png" would require a mask named "image0001.png.png", which is unintuitive and may cause compatibility issues with other tools that rely on masks, potentially requiring duplicate files.

This patch retains the current behavior as the primary lookup method but introduces a fallback: if the image already has a .png extension and the initial mask file is not found, it also checks for a mask with the exact same name. Since COLMAP allows specifying separate directories for images and masks, having identical filenames is not problematic.

Thank you for considering this patch.


src/colmap/controllers/image_reader.cc
```diff
@@ -156,11 +156,22 @@ ImageReader::Status ImageReader::Next(Rig* rig,
...
-    if (!ExistsFile(mask_path)) {
-      LOG(ERROR) << "Mask at " << mask_path << " does not exist.";
-      return Status::MASK_ERROR;
+    if (!ExistsFile(mask_path)) {      
+      bool exists_mask = false;
+      if (HasFileExtension(image->Name(), ".png")){
+        std::string alt_mask_path =
+            JoinPaths(options_.mask_path, image->Name());
+        if (ExistsFile(alt_mask_path)) {
+          mask_path = alt_mask_path;
```
ahojnnes, line 167:
```suggestion
          mask_path = std::move(alt_mask_path);
```

src/colmap/mvs/fusion.cc
```diff
@@ -339,9 +339,13 @@ void StereoFusion::InitFusedPixelMask(int image_idx,
                                       size_t height) {
   Bitmap mask;
   Mat<char>& fused_pixel_mask = fused_pixel_masks_.at(image_idx);
-  const std::string mask_path =
-      JoinPaths(options_.mask_path,
-                workspace_->GetModel().GetImageName(image_idx) + ".png");
+  const std::string mask_image_name =
+      workspace_->GetModel().GetImageName(image_idx);
+  std::string mask_path =
```
ahojnnes, line 344:
```suggestion
  const std::string mask_path =
```

ahojnnes (approved):
Thank you. LGTM.

MotivaCG:
Updated from master (workflows failed!)

MotivaCG:
@ahojnnes I've removed const qualifier from mask_path to enable its modification during the fallback check for mask files without a .png extension. That was the reason why the workflows were not compiling.

MotivaCG:
I'm seeing a code format check failure specifically on the COLMAP (Ubuntu) / ubuntu-22.04 Release (pull_request) CI job, while other checks are passing. I cannot imagine why this specific job is failing on the format check. Any insights?

ahojnnes:
The code format check is only implemented in this one CI job.

MotivaCG:
That's weird, I've run Clang locally and it triggers no conflicts.

ahojnnes:
Did you run `./scripts/format/c++.sh`? Different versions of clang-format will produce different results. This script ensures you use a consistent command and clang-format version.

MotivaCG:
Done, fingers crossed.

PR #3288, opened 2025-04-26
Fix the logic bug of PosePriorBA

whuaegeanse:
The Normalize function of Reconstruction should be called before creating DefaultBundleAdjuster.


ahojnnes:
Thank you. This should be fixed by the recent change: https://github.com/colmap/colmap/commit/3e55b506d7e2ce163a0131da7af1472849b545c8

PR #3289, opened 2025-04-27
[Bugfix] Center 2D points by principal point for absolute pose estimation with unknown focal length

xjiangan:


ahojnnes:
Thank you, this looks good to me. We should create some unit tests for the EstimateAbsolutePose function as a follow-up. Similar to what we have in the estimators/absolute_pose_test. This should have ideally been caught by a test failure.

ahojnnes (approved):

PR #3290, opened 2025-04-28
Documentation for rig support

ahojnnes:


doc/tutorial.rst
```diff
@@ -298,7 +300,9 @@ matching modes, that are intended for different input scenarios:
   images are explicitly ordered according to their file names. Note that loop
   detection requires a pre-trained vocabulary tree. A default tree will be
   automatically downloaded and cached. More trees are available and can be
-  downloaded from https://demuc.de/colmap/.
+  downloaded from https://demuc.de/colmap/. In case rigs and frames are
+  configured appropriately in the database, sequential matching will
+  automatically match all images in consecutive frames against reach other.
```
B1ueber2y, line 305: Typo: "against each other"

doc/rigs.rst
```diff
@@ -101,22 +113,21 @@ Unknown rig sensor poses
...
-Note that this can be a partial reconstruction from a small subset of the full
-set of input images. The only requirement is that each camera modeled in a rig
-must have at least one registered image. If the reconstruction was successful
-and the relative poses between registered images look roughly correct, we can
-proceed with the next step. Otherwise, you can try to adjust COLMAP's options
-and proceed when happy with the results.
+Next, reconstruct the scene without rig constraints by modeling each camera as
+its own rig (the default behavior of COLMAP without further configuration). Note
+that this can be a partial reconstruction from a subset of the full set of input
+images. The only requirement is that each camera must have at least one
```
B1ueber2y, line 123: This only requirement seems not to be accurate. We need a fully registered frame rather than at least one registered image for each camera, since they also need to be synchronized to calculate the rig extrinsics.

doc/rigs.rst
```diff
@@ -187,6 +198,10 @@ with the `rig_config.json`::
         }
     ]
 
+Notice that we do not specify the sensor poses, because we used the groundtruth
```
B1ueber2y, line 201: "the groundtruth reconstruction" -> an existing reconstruction (either groundtruth or from the first run in the two step approach)?

doc/tutorial.rst
```diff
@@ -36,7 +36,9 @@ reconstruction tool, the folder would look similar to this::
     │   +── ...
     +── sparse
     │   +── 0
+    │   │   +── rigs.bin
```
B1ueber2y, line 39: Do we say something about the backward compatibility here?

doc/rigs.rst
ahojnnes: I guess it's a bit more complicated. We need each image at least registered once together with an image of the reference camera in the same frame. In any case, I'll fix it, good catch.

doc/tutorial.rst
ahojnnes: Done

PR #3304, opened 2025-05-01
Add an example script for SfM with 360 spherical images

sarlinpe:
Works like a charm on the datasets given in https://github.com/json87/SphereSfM.

What can be improved:
- make `render_perspective_images` faster (currently 1-2s per pano) by moving the logic to C++ (though cv2.remap is already multi-threaded).
  - We could create a new CLI steps that outputs a rig_config.json to be consumed after feature extraction by the `rig_configurator` (unfortunately this can't be done in a single step).
- add an option to make feature matching ignore pairs between images of the same frame (since we know here that they are not covisible)
- geometric verification with rig constraints (I have a quick prototype based on PoseLib, for another PR)

### DIFF
--- a/python/examples/panorama_sfm.py
+++ b/python/examples/panorama_sfm.py
@@ -111,7 +111,8 @@ def render_perspective_images(
         pano_path = pano_image_dir / pano_name
         pano_image = pycolmap.Bitmap.read(pano_path, as_rgb=True)
         if pano_image is None:
-            raise RuntimeError(f"Cannot read {pano_path}")
+            logging.info(f"Skipping file {pano_path} as it cannot be read.")
+            continue
         pano_image = pano_image.to_array()
         pano_height, pano_width, *_ = pano_image.shape
         if pano_width != pano_height * 2:
--- a/python/examples/panorama_sfm.py
+++ b/python/examples/panorama_sfm.py
@@ -1,0 +1,222 @@
+"""
+An example for running incremental SfM on 360 spherical panorama images.
+"""
+
+import argparse
+from collections.abc import Sequence
+from pathlib import Path
+
+import cv2
+import numpy as np
+from scipy.spatial.transform import Rotation
+from tqdm import tqdm
+
+import pycolmap
+from pycolmap import logging
+
+
+def create_virtual_camera(
+    pano_height: int, fov_deg: float = 90
+) -> pycolmap.Camera:
+    """Create a virtual perspective camera."""
+    image_size = int(pano_height * fov_deg / 180)
+    focal = image_size / (2 * np.tan(np.deg2rad(fov_deg) / 2))
+    return pycolmap.Camera.create(0, "PINHOLE", focal, image_size, image_size)
+
+
+def get_virtual_camera_rays(camera: pycolmap.Camera) -> np.ndarray:
+    size = (camera.width, camera.height)
+    y, x = np.indices(size).astype(np.float32)
+    xy = np.column_stack([x.ravel(), y.ravel()])
+    # The center of the upper left most pixel has coordinate (0.5, 0.5)
+    xy += 0.5
+    xy_norm = camera.cam_from_img(xy)
+    rays = np.concatenate([xy_norm, np.ones_like(xy_norm[:, :1])], -1)
+    rays /= np.linalg.norm(rays, axis=-1, keepdims=True)
+    return rays
+
+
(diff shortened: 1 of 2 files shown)


python/examples/panorama_sfm.py
```diff
@@ -0,0 +1,222 @@
...
+        camera_mode=pycolmap.CameraMode.PER_FOLDER,
+    )
+
+    db = pycolmap.Database(database_path)
+    pycolmap.apply_rig_config([rig_config], db)
+    db.close()
+
+    pycolmap.match_vocabtree(database_path)
+
+    opts = pycolmap.IncrementalPipelineOptions(ba_refine_sensor_from_rig=False)
```
B1ueber2y, line 210: We can also fix the intrinsics here?

B1ueber2y (approved):
Thanks! LGTM. Will give it a try later today.

ahojnnes:
Great, thanks very much. Shall we also add reference this example in `docs/rigs.rst`?

python/examples/panorama_sfm.py
```diff
@@ -0,0 +1,222 @@
...
+        database_path.unlink()
+
+    rec_path = args.output_path / "sparse"
+    rec_path.mkdir(exist_ok=True, parents=True)
+
+    # Search for input images.
+    pano_image_dir = args.input_image_path
+    pano_image_names = sorted(
+        p.relative_to(pano_image_dir).as_posix()
+        for p in pano_image_dir.rglob("*")
```
ahojnnes, line 188: nit: On Mac, I am getting some annoying .DS_Store files every time I look at the image folder. It breaks the pipeline:
```
I20250502 09:20:26.679360 17029462 process_rigs.py:run:191] Found 38 images in /Users/jsch/data/spheresfm/campus_parterre/images.
  0%|                                                                                                                                                                                                                                                                          | 0/38 [00:00<?, ?it/s]
Traceback (most recent call last):
  File "/Users/jsch/dev/colmap/process_rigs.py", line 222, in <module>
    run(parser.parse_args())
  File "/Users/jsch/dev/colmap/process_rigs.py", line 193, in run
    rig_config = render_perspective_images(
                 ^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/jsch/dev/colmap/process_rigs.py", line 114, in render_perspective_images
    raise RuntimeError(f"Cannot read {pano_path}")
RuntimeError: Cannot read /Users/jsch/data/spheresfm/campus_parterre/images/.DS_Store
```

python/examples/panorama_sfm.py
```diff
@@ -0,0 +1,222 @@
...
+    )
+    pycolmap.set_random_seed(0)
+    pycolmap.extract_features(
+        database_path,
+        image_dir,
+        reader_options={"mask_path": mask_dir},
+        camera_mode=pycolmap.CameraMode.PER_FOLDER,
+    )
+
+    db = pycolmap.Database(database_path)
```
ahojnnes, line 204: nit: We can now use the database with a context manager as:
```
with pycolmap.Database(database_path) as db:
    ...
```

python/examples/panorama_sfm.py
sarlinpe: How does COLMAP deal with this without hardwiring image extensions? glob everything and skip if reading the file fails?
https://github.com/colmap/colmap/blob/4d99352aff28a16be4df8e53d2ac640a11515166/src/colmap/controllers/image_reader.cc#L150-L152

python/examples/panorama_sfm.py
sarlinpe: Good catch, done.

python/examples/panorama_sfm.py
sarlinpe: I forgot about this, thank you.

python/examples/panorama_sfm.py
ahojnnes: Yes, COLMAP simply skips if it cannot read the file.

sarlinpe:
I added a short doc. I will try to record a small dataset (with GPS pose priors) and add it to the example.

ahojnnes (approved):

python/examples/panorama_sfm.py
sarlinpe: Updated the script to have the same behavior.

B1ueber2y:
Just gave it a try. This works smoothly in my end. Thanks for the great example script!
campus_parterre

PR #3305, opened 2025-05-02
Recognize URIs for vocab_tree_path in GUI feature matching

ahojnnes:
The defaults were broken and lead to an "invalid vocab tree path" error when running matching in the GUI.

### DIFF
--- a/doc/rigs.rst
+++ b/doc/rigs.rst
@@ -220,3 +220,17 @@ sensor rig poses and camera parameters fixed::
         --Mapper.ba_refine_focal_length 0 \
         --Mapper.ba_refine_extra_params 0 \
         --output_path terrains/sparse
+
+
+Reconstruction from 360° spherical images
+-----------------------------------------
+
+COLMAP can handle collections of 360° panoramas by rendering virtual pinhole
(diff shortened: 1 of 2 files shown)


sarlinpe (approved):

src/colmap/util/file.h
```diff
@@ -133,6 +133,10 @@ void WriteBinaryBlob(const std::string& path, const span<const char>& data);
 // ignored and leading/trailing whitespace is removed.
 std::vector<std::string> ReadTextFileLines(const std::string& path);
 
+// Detect if given string is a URI
+// (i.e., starts with http://, https://, file://).
```
B1ueber2y, line 137: "ftp://" can also be relevant here?

ahojnnes: I am not sure if ftp works out of the box at the moment. Probably, if curl was compiled with ftp support. I'd keep it as is for now and add ftp if the use case arises and we have explicitly verified and tested its support.

B1ueber2y: Okay. Sounds good.

B1ueber2y (approved):

PR #3311, opened 2025-05-04
Bind GPSTransform and make GPSTransform::Ellipsoid an enum class

sarlinpe:


src/pycolmap/geometry/gps.cc
```diff
@@ -0,0 +1,57 @@
...
+      m, "GPSTransfromEllipsoid");
+  PyGPSTransfromEllipsoid.value("GRS80", GPSTransform::Ellipsoid::GRS80)
+      .value("WGS84", GPSTransform::Ellipsoid::WGS84);
+  AddStringToEnumConstructor(PyGPSTransfromEllipsoid);
+
+  py::class_ext_<GPSTransform> PyGPSTransform(m, "GPSTransform");
+  PyGPSTransform
+      .def(py::init<const GPSTransform::Ellipsoid>(),
+           "ellipsoid"_a = GPSTransform::Ellipsoid::GRS80)
+      .def("ellipsoid_to_ecef", &GPSTransform::EllToXYZ, "lat_lon_alt"_a)
```
sarlinpe, line 27: I find these names clearer than the C++ names - I can update them accordingly if you agree.

B1ueber2y: Agreed. This update makes it clearer.

B1ueber2y (approved):

ahojnnes (approved):
Thanks and nice!

src/pycolmap/geometry/gps.cc
ahojnnes: How about we make the equivalent changes on the C++ side?

ahojnnes (approved):

PR #3325, opened 2025-05-05
Cosmetic improvement on some geometry python bindings

B1ueber2y:


src/pycolmap/geometry/triangulation.cc
```diff
@@ -15,7 +15,7 @@ namespace py = pybind11;
 
 void BindTriangulation(py::module& m) {
   m.def(
-      "TriangulatePoint",
+      "triangulate_point",
```
B1ueber2y, line 18: to be consistent with our pycolmap conventions.

ahojnnes (approved):

PR #3328, opened 2025-05-06
Add gmock matchers for rigid3 and sim3

ahojnnes:
More convenience for defining test expectations for rigid3 and sim3.


B1ueber2y (approved):
Thanks. LGTM. It would also make sense to adapt the new changes to the other PR on added absolute pose tests : )

PR #3331, opened 2025-05-07
Colmap utility functions and editing colmap functions for MP-SfM

Zador-Pataki:
Splitting my last PR into 2 Part 1. 

These functions are needed for MP-SfM. They are also generic extensions of COLMAP that could be used elsewhere. 

See original PR https://github.com/colmap/colmap/pull/3269

### DIFF
--- a/src/colmap/estimators/bundle_adjustment.h
+++ b/src/colmap/estimators/bundle_adjustment.h
@@ -117,6 +117,9 @@ struct BundleAdjustmentOptions {
 
   // Scaling factor determines residual at which robustification takes place.
   double loss_function_scale = 1.0;
+  
+  // Magnitude determines the scaling of the residual
+  double loss_function_magnitude = 1.0;
 
   // Whether to refine the focal length parameter group.
   bool refine_focal_length = true;
--- a/src/colmap/scene/image.cc
+++ b/src/colmap/scene/image.cc
@@ -118,6 +118,31 @@ void Image::ResetPoint3DForPoint2D(const point2D_t point2D_idx) {
   }
 }
 
+std::vector<point3D_t> Image::Point3DIds(const std::optional<std::vector<point2D_t>>& point2D_idxs) const {
+  std::vector<point3D_t> point3D_ids;
+  if (!point2D_idxs.has_value()) {
+    point3D_ids.resize(NumPoints2D());
+    for (point2D_t idx = 0; idx < NumPoints2D(); ++idx) {
+      point3D_ids[idx] = Point2D(idx).point3D_id;
+    }
+  } else {
+    point3D_ids.resize(point2D_idxs->size());
+    for (size_t i = 0; i < point2D_idxs->size(); ++i) {
+      point3D_ids[i] = Point2D((*point2D_idxs)[i]).point3D_id;
+    }
+  }
+
+  return point3D_ids;
+}
+
+Eigen::MatrixXd Image::KeypointCoordinates(const std::vector<point2D_t>& point2D_idxs) const {
+  Eigen::MatrixXd coords(point2D_idxs.size(), 2);
+  for (size_t i = 0; i < point2D_idxs.size(); ++i) {
+    coords.row(i) = Point2D(point2D_idxs[i]).xy;
+  }
+  return coords;
+}
+
 bool Image::HasPoint3D(const point3D_t point3D_id) const {
   return std::find_if(points2D_.begin(),
                       points2D_.end(),
--- a/src/colmap/scene/image.h
+++ b/src/colmap/scene/image.h
@@ -125,6 +125,14 @@ class Image {
   void SetPoints2D(const std::vector<Eigen::Vector2d>& points);
   void SetPoints2D(const std::vector<struct Point2D>& points);
 
+  std::vector<point3D_t> Point3DIds(
+    const std::optional<std::vector<point2D_t>>& point2D_idxs = std::nullopt
+  ) const;
+
+  Eigen::MatrixXd KeypointCoordinates(
+    const std::vector<point2D_t>& point2D_idxs
+  ) const;
+
   // Set the point as triangulated, i.e. it is part of a 3D point track.
   void SetPoint3DForPoint2D(point2D_t point2D_idx, point3D_t point3D_id);
 
--- a/src/colmap/scene/reconstruction.cc
+++ b/src/colmap/scene/reconstruction.cc
@@ -271,6 +271,17 @@ void Reconstruction::DeleteAllPoints2DAndPoints3D() {
   }
 }
 
+Eigen::MatrixXd Reconstruction::Point3DCoordinates(
+    const std::vector<point3D_t>& point3D_ids) const {
+  Eigen::MatrixXd coords(point3D_ids.size(), 3);
+  for (size_t i = 0; i < point3D_ids.size(); ++i) {
+    auto it = points3D_.find(point3D_ids[i]);
+    THROW_CHECK(it != points3D_.end());
+    coords.row(i) = it->second.xyz;
+  }
+  return coords;
+}
+
 void Reconstruction::RegisterImage(const image_t image_id) {
   THROW_CHECK(Image(image_id).HasPose());
   reg_image_ids_.insert(image_id);
--- a/src/colmap/scene/reconstruction.h
+++ b/src/colmap/scene/reconstruction.h
@@ -139,6 +139,9 @@ class Reconstruction {
   // Delete all 2D points of all images and all 3D points.
   void DeleteAllPoints2DAndPoints3D();
 
+  // Return the 3D coordinates of the given point IDs.
+  Eigen::MatrixXd Point3DCoordinates(const std::vector<point3D_t>& point3D_ids) const;
+
   // Register an existing image.
   void RegisterImage(image_t image_id);
 
--- a/src/colmap/sfm/incremental_triangulator.cc
+++ b/src/colmap/sfm/incremental_triangulator.cc
@@ -302,7 +302,9 @@ size_t IncrementalTriangulator::MergeAllTracks(const Options& options) {
   return num_merged;
 }
 
-size_t IncrementalTriangulator::Retriangulate(const Options& options) {
+size_t IncrementalTriangulator::Retriangulate(
+  const Options& options,
+  const std::unordered_set<image_t>& ignore_image_ids) {
   THROW_CHECK(options.Check());
 
   size_t num_tris = 0;
@@ -328,6 +330,11 @@ size_t IncrementalTriangulator::Retriangulate(const Options& options) {
     std::tie(image_id1, image_id2) =
         Database::PairIdToImagePair(image_pair.first);
 
+    if (!ignore_image_ids.empty() &&
+        (ignore_image_ids.count(image_id1) > 0 || ignore_image_ids.count(image_id2) > 0)) {
+      continue;
+    }
+    
     const Image& image1 = reconstruction_.Image(image_id1);
     if (!image1.HasPose()) {
       continue;
--- a/src/colmap/sfm/observation_manager.cc
+++ b/src/colmap/sfm/observation_manager.cc
@@ -343,9 +343,19 @@ size_t ObservationManager::FilterObservationsWithNegativeDepth() {
 size_t ObservationManager::FilterPoints3DWithSmallTriangulationAngle(
     const double min_tri_angle,
     const std::unordered_set<point3D_t>& point3D_ids) {
-  // Number of filtered points.
-  size_t num_filtered = 0;
+  const std::unordered_set<point3D_t> to_remove = FindPoints3DWithSmallTriangulationAngle(min_tri_angle, point3D_ids);
+  for (const point3D_t id : to_remove) {
+    DeletePoint3D(id);
+  }
+  return to_remove.size();
+}
 
+std::unordered_set<point3D_t> ObservationManager::FindPoints3DWithSmallTriangulationAngle(
+    const double min_tri_angle,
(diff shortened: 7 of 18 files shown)


src/colmap/estimators/bundle_adjustment.h
```diff
@@ -117,6 +117,9 @@ struct BundleAdjustmentOptions {
 
   // Scaling factor determines residual at which robustification takes place.
   double loss_function_scale = 1.0;
+  
+  // Magnitude determines the scaling of the residual
+  double loss_function_magnitude = 1.0;
```
ahojnnes, line 122:
```suggestion
  double loss_function_scale = 1.0;
```

src/colmap/scene/image.cc
```diff
@@ -118,6 +118,31 @@ void Image::ResetPoint3DForPoint2D(const point2D_t point2D_idx) {
...
+    point3D_ids.resize(point2D_idxs->size());
+    for (size_t i = 0; i < point2D_idxs->size(); ++i) {
+      point3D_ids[i] = Point2D((*point2D_idxs)[i]).point3D_id;
+    }
+  }
+
+  return point3D_ids;
+}
+
+Eigen::MatrixXd Image::KeypointCoordinates(const std::vector<point2D_t>& point2D_idxs) const {
```
ahojnnes, line 138:
```suggestion
Eigen::MatrixXd Image::Point2DCoords(const std::vector<point2D_t>& point2D_idxs) const {
```

src/colmap/scene/image.h
```diff
@@ -125,6 +125,14 @@ class Image {
   void SetPoints2D(const std::vector<Eigen::Vector2d>& points);
   void SetPoints2D(const std::vector<struct Point2D>& points);
 
+  std::vector<point3D_t> Point3DIds(
+    const std::optional<std::vector<point2D_t>>& point2D_idxs = std::nullopt
```
ahojnnes, line 129: Otherwise, it's easy to create an unnecessary copy of the input vector.
```suggestion
    const std::vector<point2D_t>* point2D_idxs = nullptr
```

src/colmap/scene/image.h
```diff
@@ -125,6 +125,14 @@ class Image {
...
+  ) const;
+
+  Eigen::MatrixXd KeypointCoordinates(
+    const std::vector<point2D_t>& point2D_idxs
+  ) const;
```
ahojnnes, line 134: Would it be possible to just have `Point3DIds()` and `Point2DCoords()` methods without the point2D_idxs input and always return all data? Then, on the Python side, you can do a simple `point3D_ids[point2D_idxs]` filtering. This should be fast enough?

src/colmap/scene/reconstruction.cc
```diff
@@ -271,6 +271,17 @@ void Reconstruction::DeleteAllPoints2DAndPoints3D() {
   }
 }
 
+Eigen::MatrixXd Reconstruction::Point3DCoordinates(
```
ahojnnes, line 274: nit
```suggestion
Eigen::MatrixXd Reconstruction::Point3DCoords(
```

src/colmap/sfm/incremental_triangulator.cc
```diff
@@ -302,7 +302,9 @@ size_t IncrementalTriangulator::MergeAllTracks(const Options& options) {
   return num_merged;
 }
 
-size_t IncrementalTriangulator::Retriangulate(const Options& options) {
+size_t IncrementalTriangulator::Retriangulate(
+  const Options& options,
+  const std::unordered_set<image_t>& ignore_image_ids) {
```
ahojnnes, line 307: Let's use a pointer and default to a nullptr instead of checking for ".empty()".

```suggestion
  const std::unordered_set<image_t>* ignore_image_ids) {
```

src/pycolmap/sfm/incremental_mapper.cc
```diff
@@ -360,6 +361,14 @@ void BindIncrementalMapperImpl(py::module& m) {
       .def_readwrite("num_adjusted_observations",
                      &LocalBAReport::num_adjusted_observations);
   MakeDataclass(PyLocalBAReport);
+  
+  py::class_<IncrementalMapperImpl>(m, "IncrementalMapperImpl")
```
ahojnnes, line 365: We shouldn't expose the impl class here. Instead, if we really need the find_local_bundle, let's expose it in the main incremental mapper and then bind the method there.

src/colmap/sfm/observation_manager.h
```diff
@@ -102,6 +102,11 @@ class ObservationManager {
 
   size_t FilterPoints3DWithSmallTriangulationAngle(
       double min_tri_angle, const std::unordered_set<point3D_t>& point3D_ids);
+
+  std::unordered_set<point3D_t> FindPoints3DWithSmallTriangulationAngle(
```
ahojnnes, line 106:
```suggestion
  std::vector<point3D_t> FindPoints3DWithSmallTriangulationAngle(
```

ahojnnes (changes requested):
Thanks, looks good to me for the most part. Left a few small comments.

src/colmap/scene/image.h
Zador-Pataki: @ahojnnes I am okay with this. In the case of Point3DIds, the 2D points that don't have a 3D point will return `kInvalidPoint3DId`. Would you also do something like that for Point2DCoords, i.e., assign something like `std::numeric_limits<...>::max()` to each row of Point2DCoords that doesn't have a 3D point? Alternatively we could follow `get_observation_point2D_idxs` and `get_observation_points2D`, and create the function `get_observation_points2D_coords` instead. This would return only the coordinates of 2D points with 3D points.

src/pycolmap/sfm/incremental_mapper.cc
```diff
@@ -495,7 +490,7 @@ void BindIncrementalMapperImpl(py::module& m) {
       .def("clear_modified_points3D",
            &IncrementalMapper::ClearModifiedPoints3D);
 
-      m.def("find_local_bundle_impl",
+      m.def("find_rec_local_bundle",
```
Zador-Pataki, line 493: @ahojnnes directly exposing static function from incremental_mapper_impl works. Do you like this solution?

src/colmap/estimators/bundle_adjustment.h
ahojnnes: Right, my bad. How about loss_function_weight ?

src/colmap/scene/image.h
ahojnnes: I suggest Point2DCoords just returns valid coordinates for all keypoints independent of whether they have a corresponding 3D point or not. If you want a list of only 2D point observations, you can simply do the masking on the Python side as `observations = point2D_coords[point3D_ids ~= INVALID_POINT3D_ID]`.

src/colmap/sfm/observation_manager.cc
```diff
@@ -343,18 +343,19 @@ size_t ObservationManager::FilterObservationsWithNegativeDepth() {
 size_t ObservationManager::FilterPoints3DWithSmallTriangulationAngle(
     const double min_tri_angle,
     const std::unordered_set<point3D_t>& point3D_ids) {
-  const std::unordered_set<point3D_t> to_remove = FindPoints3DWithSmallTriangulationAngle(min_tri_angle, point3D_ids);
+  const std::vector<point3D_t> ids_vec(point3D_ids.begin(), point3D_ids.end());
```
Zador-Pataki, line 346: @ahojnnes  updated `FindPoints3DWithSmallTriangulationAngle` inputs/outputs following https://github.com/colmap/colmap/pull/3331#discussion_r2079130567. However, I did not convert the `FilterPoints3DWithSmallTriangulationAngle` input to `std::unordered_set<point3D_t>`. 

As a result, I need to convert ids from set to vec. Changing the input to vec as well would also mean changing the inputs to this function in the COLMAP pipeline

src/pycolmap/sfm/incremental_mapper.cc
```diff
@@ -488,13 +488,12 @@ void BindIncrementalMapperImpl(py::module& m) {
...
       .def("get_modified_points3D", &IncrementalMapper::GetModifiedPoints3D)
       .def("clear_modified_points3D",
-           &IncrementalMapper::ClearModifiedPoints3D);
-
-      m.def("find_rec_local_bundle",
-            &IncrementalMapperImpl::FindLocalBundle,
-            "options"_a,
-            "image_id"_a,
-            "reconstruction"_a);
+           &IncrementalMapper::ClearModifiedPoints3D)
```
Zador-Pataki, line 491: @ahojnnes done following https://github.com/colmap/colmap/pull/3331#discussion_r2084929178

Note that `find_local_bundle` is taken (uses the reconstruction object in self), hence the new name. I need to find a local bundle for an input reconstruction. I've made it static because I don't use `IncrementalMapper` in MP-SfM. This why I can do `pycolmap.IncrementalMapper.find_rec_local_bundle(options, refimid, self.rec)`.

src/colmap/scene/reconstruction.h
```diff
@@ -139,6 +139,9 @@ class Reconstruction {
   // Delete all 2D points of all images and all 3D points.
   void DeleteAllPoints2DAndPoints3D();
 
+  // Return the 3D coordinates of the given point IDs.
+  Eigen::MatrixXd Point3DCoords(const std::vector<point3D_t>& point3D_ids) const;
```
B1ueber2y, line 143: ``reconstruction.Point3DCoords`` and ``image.Point2DCoords`` should ideally have identical interfaces.

src/colmap/estimators/bundle_adjustment.h
```diff
@@ -117,6 +117,9 @@ struct BundleAdjustmentOptions {
 
   // Scaling factor determines residual at which robustification takes place.
   double loss_function_scale = 1.0;
+  
+  // Magnitude determines the scaling of the residual
```
B1ueber2y, line 121: nit: end mark missing.

src/pycolmap/sfm/incremental_triangulator.cc
```diff
@@ -100,7 +100,9 @@ void BindIncrementalTriangulator(py::module& m) {
            &IncrementalTriangulator::MergeAllTracks,
            "options"_a)
       .def(
-          "retriangulate", &IncrementalTriangulator::Retriangulate, "options"_a)
+          "retriangulate", &IncrementalTriangulator::Retriangulate,
+          "options"_a,
+          "ignore_image_ids"_a = std::unordered_set<image_t>())
```
B1ueber2y, line 105: Does this work? You used a pointer as an input to the method.

src/pycolmap/sfm/incremental_mapper.cc
B1ueber2y: Maybe ``find_local_bundle_from_reconstruction``? Your current naming does not give intuitive meaning.

ahojnnes: OK, I guess I didn't look close enough in that this takes a custom 3D reconstruction and not the one managed by the mapper? Sorry for the confusion, but in this case a free-standing function could indeed be more suitable.

src/pycolmap/sfm/incremental_triangulator.cc
Zador-Pataki: nice catch thanks

src/colmap/scene/reconstruction.h
Zador-Pataki: @ahojnnes @B1ueber2y I agree, however, I don't think there is an elegant solution for this one. The biggest differences is that `image.Point2DCoords` returns all point2D coordinates, while `reconstruction.Point3DCoords` takes list of 3D points. Since COLMAP filters 3D points, not all ids between 1 and the maximum point3D id has a 3D point. Thus, if we omit the input ids, we either 
1. return as many coordinates as there are 3D points, such that they each correspond to a sorted list of ids, or 
2. we return as many coordinates as the max point3D id, with some filler value where there is a missing 3D point. This way, in python, we could access the point using a 1D array of point3D ids (like for  `image.Point2DCoords`)

src/pycolmap/sfm/incremental_mapper.cc
```diff
@@ -488,15 +488,16 @@ void BindIncrementalMapperImpl(py::module& m) {
...
-           "image_id"_a,
-           "reconstruction"_a);
+           &IncrementalMapper::ClearModifiedPoints3D);
 }
 
 void BindIncrementalMapper(py::module& m) {
   BindIncrementalMapperImpl(m);
   BindIncrementalPipeline(m);
+
+  m.def("find_local_bundle_from_reconstruction",
```
Zador-Pataki, line 498: @ahojnnes following your suggestion https://github.com/colmap/colmap/pull/3331#discussion_r2091701970, I have made this into a free-standing function. What do you think of this placement?

Zador-Pataki:
Seems like only https://github.com/colmap/colmap/pull/3331#discussion_r2085590228, https://github.com/colmap/colmap/pull/3331#discussion_r2086168442 and  https://github.com/colmap/colmap/pull/3331#discussion_r2094589775 remain

src/colmap/scene/reconstruction.h
B1ueber2y: I think we should go with your option 1, and make the argument point3D_ids optional if you would like to keep it.

B1ueber2y:
The new utilities ``Image::Point2DCoords``, ``Image::Point3DIds``, and ``Reconstruction::Point3DCoords`` are indeed quite useful to speed up the operations in Python (I also found myself in need of these interfaces). I suggest we move forward with this PR. 

I am happy with its current state except for a minor change requested for ``Point3DCoords``.

PR #3334, opened 2025-05-09
Add generalized relative pose estimation and pose binding cleanups

ahojnnes:


B1ueber2y (approved):

src/pycolmap/estimators/pose.cc
```diff
@@ -181,4 +230,20 @@ void BindAbsolutePoseEstimator(py::module& m) {
         "followed by non-linear refinement.");
   DefDeprecation(
       m, "absolute_pose_estimation", "estimate_and_refine_absolute_pose");
+
+  m.def("estimate_relative_pose",
+        &PyEstimateRelativePose,
+        "points1"_a,
+        "points2"_a,
```
sarlinpe, line 237: Should we make it clear that the points are expected to be in normalized coordinates? e.g. `points1_in_cam` or so? Similarly for the C++ definitions and for the refinement function.

src/colmap/estimators/generalized_pose_test.cc
```diff
@@ -166,5 +168,116 @@ TEST(RefineGeneralizedAbsolutePose, Nominal) {
...
+      problem.points2D2.push_back(image2_ptr->Point2D(point2D_idx2).xy);
+      problem.camera_idxs1.push_back(maybe_add_and_get_camera(image1));
+      problem.camera_idxs2.push_back(maybe_add_and_get_camera(*image2_ptr));
+    }
+  }
+
+  return problem;
+}
+
+TEST(EstimateGeneralizedRelativePose, Nominal) {
```
sarlinpe, line 253: Just curious - does this test pass if all camera centers coincide? left for future PR?

ahojnnes: Comes as a follow-up to this PR and I am planning to test it at the minimal solver level rather than here.

src/pycolmap/estimators/pose.cc
ahojnnes: Done.

PR #3340, opened 2025-05-14
Covariance-aware Umeyama for pose prior mapper

clementinboittiaux:
### Summary
Pose prior covariance matrices are used during pose prior BA but not for the Sim3 alignment. In the case where pose priors have significantly lower/higher uncertainties for different axes, this leads to a very bad alignment, loosing all the benefits of the pose prior BA.

### Practical application

For underwater surveys, we often only have access to depth/altitude measurements. In addition, calibration is tougher due to refraction. Also, we mainly observe flat surfaces from a top-down view and for long distances. Because of this, a small difference in distortion parameters can lead to significant curve-like drift (see first figure below). Incorporating depth/altitude measurements into the BA is a very good way to regularize the camera parameters and enable a flat reconstruction.

We ran COLMAP pose prior mapper with and without the covariance-weighted Umeyama on a sample of our data. Since we only have access to depth measurements, the X and Y axes of our pose priors are uniform random values between -10 and 10.

Both following models were obtained using this command:
```
colmap pose_prior_mapper --database_path database.db --image_path images/ --output_path weighted/ --Mapper.ba_refine_principal_point 1 --overwrite_priors_covariance 1 --prior_position_std_x 10000.0 --prior_position_std_y 10000.0 --prior_position_std_z 0.01 --Mapper.ba_global_max_num_iterations 20 --Mapper.ba_global_max_refinements 3 --Mapper.ba_global_function_tolerance 0.00000001
```

**Without weighted Umeyama:**
classic

**With weighted Umeyama:**
weighted

You can find the images, database and COLMAP models here: https://drive.google.com/file/d/1dDQOfejPgSZYCfdlnQcSFlefAb819ZW4/view?usp=sharing

### Pull request
The code is dirty, this should not be merged. Please let me know if you feel that this is a useful feature that should be integrated in the project. In this case I can do some cleaning.

**Side note:** it seems that the pose prior mapper is performing Sim3 alignment with RANSAC by default - this feels quite hard to setup RANSAC with covariance-aware Umeyama - maybe a parameter could control whether or not to use RANSAC for this step (or maybe I am missing something).

### DIFF
--- a/src/colmap/estimators/alignment.cc
+++ b/src/colmap/estimators/alignment.cc
@@ -238,6 +238,84 @@ bool AlignReconstructionToLocations(
   return true;
 }
 
+void PrintAxisErrors(const std::vector<Eigen::Vector3d>& src_points,
+                    const std::vector<Eigen::Vector3d>& tgt_points,
+                    const Sim3d& transform,
+                    const std::string& stage) {
+  if (src_points.empty() || src_points.size() != tgt_points.size()) {
+    return;
+  }
+  
+  // Initialize error accumulators for each axis
+  double x_abs_error_sum = 0.0;
+  double y_abs_error_sum = 0.0;
+  double z_abs_error_sum = 0.0;
+  double total_abs_error_sum = 0.0;
+  
+  std::vector<double> x_errors, y_errors, z_errors, total_errors;
+  
+  // For each point pair
+  for (size_t i = 0; i < src_points.size(); ++i) {
+    // Transform the source point if we're in the "after" stage
+    Eigen::Vector3d transformed_src;
+    if (stage == "After") {
+      transformed_src = transform * src_points[i];
+    } else {
+      transformed_src = src_points[i];
+    }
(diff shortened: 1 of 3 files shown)


ahojnnes:
Thanks for your (first) contribution. This sounds like a useful feature. I'll try to review your implementation soon. Regarding your questions: (a) Yes, we could certainly control the use of RANSAC vs. no RANSAC with an option, (b) if we have covariance for points in the target coordinate frame, we could use the whitened error and consider points with < X times the standard deviation as inliers? In other words, compute `X = sqrt(r^T C^-1 r)` where `r` is the residual between transformed source point in target and `C` is the covariance matrix.

clementinboittiaux:
Got it, thanks! I'll look into it and try to clean the code in the next weeks.
Also the commit title is misleading, I am not using the full covariance matrices but just the diagonal per-axis variance. I'll try to incorporate the full covariance matrices.

PR #3373, opened 2025-06-02
Fix deadlock during feature matching

ahojnnes:
The switch to faiss revealed a rare deadlock that can occur when simultaneously reading from the feature matching cache while acquiring the database transaction. This was reliably reproducible with a debugger. Removing the transaction leads to minor performance degradation when writing to the database.


B1ueber2y (approved):

jhacsonmeza:
Hi @ahojnnes. I wanted to report that I'm noticing an important runtime impact when using `matches_importer` after this change. After adding the lines that were removed, the performance was the same than previous versions.

PR #3375, opened 2025-06-02
expose loading database into database cache from DatabaseCache::Create.

B1ueber2y:
This can avoid code duplication when COLMAP is used as a third party.

### DIFF
--- a/src/colmap/scene/database_cache.h
+++ b/src/colmap/scene/database_cache.h
@@ -65,11 +65,10 @@ class DatabaseCache {
   //                              frame is included, all other images in the
   //                              same frame will also be included. All images
   //                              are used if empty.
-  void LoadDatabase(
-      const Database& database,
-      size_t min_num_matches,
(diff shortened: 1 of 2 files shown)


src/colmap/scene/database_cache.h
```diff
@@ -65,6 +65,11 @@ class DatabaseCache {
   //                              frame is included, all other images in the
   //                              same frame will also be included. All images
   //                              are used if empty.
+  void LoadDatabase(const Database& database,
```
ahojnnes, line 68: nit: would probably just name this "Load".

ahojnnes (approved):

src/colmap/scene/database_cache.h
B1ueber2y: Okay I shouldnt have enabled auto-merge. Fixed here: https://github.com/colmap/colmap/pull/3376. Will be careful to use it in the future : )

PR #3383, opened 2025-06-06
Unit tests for image reader, remove redundant definition of database

ahojnnes:


src/colmap/controllers/image_reader.cc
```diff
@@ -187,7 +187,16 @@ ImageReader::Status ImageReader::Next(Rig* rig,
     }
 
     prev_camera_ = std::move(current_camera);
-    prev_rig_ = database_->ReadRigWithSensor(prev_camera_.SensorId()).value();
+    if (std::optional<Rig> rig =
+            database_->ReadRigWithSensor(prev_camera_.SensorId());
+        rig.has_value()) {
+      prev_rig_ = std::move(rig.value());
+    } else {
+      // For backwards compatibility with old databases, we create a rig.
```
B1ueber2y, line 195: Does this logic need to be employed here as well? https://github.com/colmap/colmap/blob/main/src/colmap/controllers/image_reader.cc#L219. If so, we should also make sure that the new rig is not written into the database twice: https://github.com/colmap/colmap/blob/main/src/colmap/controllers/image_reader.cc#L262. This logic appears to be a bit convoluted to me.

B1ueber2y (approved):

PR #3386, opened 2025-06-08
Add ReadAllFramesData method to Database class for efficient frame data retrieval

qchenevier:
Fixes https://github.com/colmap/colmap/issues/3385

This new method retrieves all frame data in a single query, improving performance by reducing the number of database calls. The ReadAllFrames method has been updated to utilize this new functionality, matching data to frames more efficiently.

I'm mainly write python or javascript and I'm not very used to C++ so this PR might not be compliant with C++ best practices. Feel free to reject or ask for modifications.

I have tested it on my own datasets with the `colmap hierarchical_mapper`.

PR #3387, opened 2025-06-10
Speedup database reads of rigs/frames with single SQL outer join query

ahojnnes:
When writing the code, I got lazy and thought it would be fast enough to use separate queries per frame/rigs. This can be done with a single query by joining the rigs/rig_sensors and frames/frame_data tables.Supersedes https://github.com/colmap/colmap/pull/3386.

### DIFF
--- a/src/colmap/scene/database.cc
+++ b/src/colmap/scene/database.cc
@@ -218,7 +218,7 @@ std::optional<std::stringstream> BlobColumnToStringStream(
 }
 
 void ReadRigRows(sqlite3_stmt* sql_stmt,
-                 const std::function<void(Rig)> new_rig_callback) {
+                 const std::function<void(Rig)>& new_rig_callback) {
   Rig rig;
   while (SQLITE3_CALL(sqlite3_step(sql_stmt)) == SQLITE_ROW) {
     const rig_t rig_id = static_cast<rig_t>(sqlite3_column_int64(sql_stmt, 0));
@@ -302,7 +302,7 @@ Camera ReadCameraRow(sqlite3_stmt* sql_stmt) {
 }
 
 void ReadFrameRows(sqlite3_stmt* sql_stmt,
-                   const std::function<void(Frame)> new_frame_callback) {
+                   const std::function<void(Frame)>& new_frame_callback) {
   Frame frame;
   while (SQLITE3_CALL(sqlite3_step(sql_stmt)) == SQLITE_ROW) {
     const frame_t frame_id =
--- a/src/colmap/scene/database.cc
+++ b/src/colmap/scene/database.cc
@@ -217,50 +217,66 @@ std::optional<std::stringstream> BlobColumnToStringStream(
   return stream;
 }
 
-Rig ReadRigRow(sqlite3_stmt* sql_stmt) {
+void ReadRigRows(sqlite3_stmt* sql_stmt,
+                 const std::function<void(Rig)> new_rig_callback) {
   Rig rig;
+  while (SQLITE3_CALL(sqlite3_step(sql_stmt)) == SQLITE_ROW) {
+    const rig_t rig_id = static_cast<rig_t>(sqlite3_column_int64(sql_stmt, 0));
+    if (rig_id != rig.RigId()) {
+      if (rig.RigId() != kInvalidRigId) {
+        new_rig_callback(std::move(rig));
+      }
+      rig = Rig();
+      rig.SetRigId(rig_id);
+      sensor_t ref_sensor_id;
+      ref_sensor_id.id =
+          static_cast<uint32_t>(sqlite3_column_int64(sql_stmt, 1));
+      ref_sensor_id.type =
+          static_cast<SensorType>(sqlite3_column_int64(sql_stmt, 2));
+      rig.AddRefSensor(ref_sensor_id);
+    }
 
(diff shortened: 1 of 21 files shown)


src/colmap/scene/database.cc
```diff
@@ -217,50 +217,66 @@ std::optional<std::stringstream> BlobColumnToStringStream(
   return stream;
 }
 
-Rig ReadRigRow(sqlite3_stmt* sql_stmt) {
+void ReadRigRows(sqlite3_stmt* sql_stmt,
+                 const std::function<void(Rig)>& new_rig_callback) {
   Rig rig;
+  while (SQLITE3_CALL(sqlite3_step(sql_stmt)) == SQLITE_ROW) {
+    const rig_t rig_id = static_cast<rig_t>(sqlite3_column_int64(sql_stmt, 0));
+    if (rig_id != rig.RigId()) {
```
paulinus, line 248: Here we are assuming that all rows for a particular rig will be consecutive. I'm not sure that an SQL JOIN guaranties this. So it is likely that an `ORDER BY rigs.rig_id` is needed in the query below.

src/colmap/scene/database.cc
```diff
@@ -217,50 +217,66 @@ std::optional<std::stringstream> BlobColumnToStringStream(
...
+      sensor_from_rig->rotation.y() =
+          ReadBinaryLittleEndian<double>(sensor_from_rig_stream_ptr);
+      sensor_from_rig->rotation.z() =
+          ReadBinaryLittleEndian<double>(sensor_from_rig_stream_ptr);
+      sensor_from_rig->translation.x() =
+          ReadBinaryLittleEndian<double>(sensor_from_rig_stream_ptr);
+      sensor_from_rig->translation.y() =
+          ReadBinaryLittleEndian<double>(sensor_from_rig_stream_ptr);
+      sensor_from_rig->translation.z() =
+          ReadBinaryLittleEndian<double>(sensor_from_rig_stream_ptr);
```
paulinus: aside: is this the only place a Rigid3D is read? i wonder if it would make sense to have a dedicated ReadRigid3DFromStream function.

src/colmap/scene/database.cc
```diff
@@ -217,50 +217,66 @@ std::optional<std::stringstream> BlobColumnToStringStream(
...
 
-Rig ReadRigRow(sqlite3_stmt* sql_stmt) {
+void ReadRigRows(sqlite3_stmt* sql_stmt,
+                 const std::function<void(Rig)>& new_rig_callback) {
   Rig rig;
+  while (SQLITE3_CALL(sqlite3_step(sql_stmt)) == SQLITE_ROW) {
+    const rig_t rig_id = static_cast<rig_t>(sqlite3_column_int64(sql_stmt, 0));
+    if (rig_id != rig.RigId()) {
+      if (rig.RigId() != kInvalidRigId) {
+        new_rig_callback(std::move(rig));
```
paulinus, line 250: Something unclear to me. Here we move the rig. That in theory, invalidates the rig object. But on the next iteration of the loop, we do `if (rig_id != rig.RigId())`. That is accessing the rig that we just invalidated. Should we instead either pass a copy here or use a `new_rig_needed` boolean variable to track if a new rig is needed at the beginning of the loop?

src/colmap/scene/database.cc
```diff
@@ -285,24 +301,39 @@ Camera ReadCameraRow(sqlite3_stmt* sql_stmt) {
...
-
-void ReadFrameData(sqlite3_stmt* sql_stmt, Frame* frame) {
-  SQLITE3_CALL(sqlite3_bind_int64(sql_stmt, 1, frame->FrameId()));
   while (SQLITE3_CALL(sqlite3_step(sql_stmt)) == SQLITE_ROW) {
+    const frame_t frame_id =
+        static_cast<rig_t>(sqlite3_column_int64(sql_stmt, 0));
+    if (frame_id != frame.FrameId()) {
+      if (frame.FrameId() != kInvalidFrameId) {
+        new_frame_callback(std::move(frame));
+      }
```
paulinus, line 321: same thoughts as in the rigs case here. Consider ORDERing the rows first and not accessing the frame variable after moving it.

paulinus (changes requested):
LGTM, just two minor potential issues that could break if SQL stops sorting rows or if the receiver of the moved objects modifies them.

src/colmap/scene/database.cc
ahojnnes: Very good catch. It probably depends on the order in which the data is added, which is guaranteed by the C++ interface and why the tests passed. However, there is no guarantee for this ordering to be correct for manually inserted data by users directly modifying the database. Fixed.

src/colmap/scene/database.cc
ahojnnes: Thanks, same response. Fixed.

src/colmap/scene/database.cc
ahojnnes: It's the only place where this is currently being used, but it's a good improvement to extract this into a separate helper function for readability. Done, same for the counter-part of writing a rigid3d to a stream.

src/colmap/scene/database.cc
ahojnnes: Two lines further down we assign a new rig object. Same for the frame.

PR #3389, opened 2025-06-10
Add missing VocabTreeMatching.num_threads in option manager

ahojnnes:


ahojnnes:
> Any way that could have found automatically?

Good question, but I don't think this can be tested (with reasonable effort).

PR #3393, opened 2025-06-10
Use MKL as BLAS vendor for faiss

ahojnnes:


doc/install.rst
```diff
@@ -138,6 +135,10 @@ CUDA package and GCC, and you must compile against GCC 10::
     export CUDAHOSTCXX=/usr/bin/g++-10
     # ... and then run CMake against COLMAP's sources.
 
+Notice that compiling against OpenBLAS instead of MKL requires the OpenMP
```
B1ueber2y: Sorry that I did not get the meaning of this sentence. Do you mean minimum OpenMP version? In the linked faiss documentation it seems to suggest that libopenblas-openmp-dev can work?

doc/install.rst
```diff
@@ -115,7 +112,7 @@ Configure and compile COLMAP::
     cd colmap
     mkdir build
     cd build
-    cmake .. -GNinja
+    cmake .. -GNinja -DBLA_VENDOR=Intel10_64lp
```
sarlinpe, line 115: Should we comment on this obscure argument? when is it needed and when is it not?

doc/install.rst
ahojnnes: There are different implementations of BLAS/LAPACK. Intel's MKL is one, OpenBLAS is another, and there are more. It means that if you decide to link against the OpenBLAS implementation under Ubuntu, you need to use the OpenMP and not the pthread version (which is the default).

doc/install.rst
ahojnnes: Did my best :-)

PR #3396, opened 2025-06-11
Propagate macros to top-level CMakeLists.txt files

jhacsonmeza:
When COLMAP is used as a third-party library with `add_subdirectory` or with FetchContent (like GLOMAP does) all the macro definitions are not properly propagated to the top-level `CMakeLists.txt` project. I noticed this after getting a very long compilation error related to Glog:
```
FAILED: CMakeFiles/utils.dir/src/utils.cpp.o 
/dir/include/colmap/util/logging.h:163:38: error: 'CheckOpString' in namespace 'google' does not name a type
  163 |                        const google::CheckOpString& result)
      |                                      ^~~~~~~~~~~~~
```
This was because `GLOG_VERSION_MAJOR` and `GLOG_VERSION_MINOR` macros were not visible in the top-level project. Building GLOMAP with `FETCH_COLMAP` also causes the same compilation error.

The problem with `add_compile_definitions` is that the macros have a scope limited to the current directory and all the subdirectories, but they don't propagate upwards to parent directories. I was able to quickly ignore this error with a `find_package(colmap REQUIRED)` call which makes all the macros visible in the main CMakeLists.txt file, even if I don't use the target there.

I'm proposing a solution by replacing all the `add_compile_definitions` with `target_compile_definitions` which allows to better control the scope and propagate macros upwards with `PUBLIC`. These are the macro definitions I changed:
* From [`cmake/FindDependencies.cmake`](https://github.com/colmap/colmap/blob/main/cmake/FindDependencies.cmake): `GLOG_VERSION_MAJOR`, `GLOG_VERSION_MINOR`, `COLMAP_CGAL_ENABLED`, `COLMAP_DOWNLOAD_ENABLED`, `COLMAP_CUDA_ENABLED`, `COLMAP_GUI_ENABLED`, `COLMAP_GPU_ENABLED`.
* From the main [`CMakeLists.txt`](https://github.com/colmap/colmap/blob/main/CMakeLists.txt) file: `COLMAP_LSD_ENABLED`.

[`COLMAP_OPENGL_ENABLED`](https://github.com/colmap/colmap/blob/e175fb1a02412d25fe81e2f5348e914e89fa3f9c/cmake/FindDependencies.cmake#L230) is defined in `cmake/FindDependencies.cmake` but it's never used in the code. That's why I didn't add it anywhere. I also removed it from the `FindDependencies.cmake` file. However, I'd like to know if this is correct.

I used `PUBLIC` for the cases where the macros are used in header files, and `PRIVATE` otherwise. This solved the compilation error for me.

I kept the `add_compile_definitions` call for the macros `WIN32_LEAN_AND_MEAN`, `GLOG_USE_GLOG_EXPORT`, `GLOG_NO_ABBREVIATED_SEVERITIES`, `GL_GLEXT_PROTOTYPES`, `NOMINMAX` and `EIGEN_INITIALIZE_MATRICES_BY_NAN`. I think these compile definitions are more for internal use and don't need to be propagated to parent directories. However, I would like to know if I'm wrong.

I also decided to explicitly add the compile definitions to all the static libraries even if they can be inherited from other libraries. For example, `COLMAP_CGAL_ENABLED` needs to be public for `colmap_mvs` target and `colmap_controllers` library links against `colmap_mvs`, so adding `target_compile_definitions(colmap_controllers PRIVATE COLMAP_CGAL_ENABLED)` is unnecessary but I decided to do it. This can be improved if you consider so. 

An alternative solution to this problem could be to keep the `add_compile_definitions` calls and keep track of all the macros that needs to be visible in parent directories and add them to the [COLMAP interface library](https://github.com/colmap/colmap/blob/e175fb1a02412d25fe81e2f5348e914e89fa3f9c/CMakeLists.txt#L396) that holds all the `colmap_<name>` targets.

Also as a side note I wanted to ask, the [`colmap_exe`](https://github.com/colmap/colmap/blob/e175fb1a02412d25fe81e2f5348e914e89fa3f9c/src/colmap/exe/CMakeLists.txt#L51) static lib is created in `src/colmap/exe/CMakeLists.txt` but I see that it's not used at all. It's not even linked against `colmap_main`. Does it serve any purpose?


jhacsonmeza:
Hi. I think I didn't get it. You mean to modify `COLMAP_ADD_LIBRARY` and add the option to provide compile definitions when we create the static libraries?

ahojnnes:
Apologies, I wasn't clear enough. Here is my proposal:

FindDependencies.cmake
```
...
set(COLMAP_COMPILE_DEFINITIONS)
...
if(CGAL_FOUND)
    list(APPEND COLMAP_COMPILE_DEFINITIONS COLMAP_CGAL_ENABLED)
...
```

CMakeHelper.cmake
```
...
macro(COLMAP_ADD_LIBRARY)
    ...
    target_compile_definitions(${COLMAP_ADD_LIBRARY_NAME} PUBLIC ${COLMAP_COMPILE_DEFINITIONS})
endmacro(COLMAP_ADD_LIBRARY)
...
```

jhacsonmeza:
I did the changes. I think it's better in this way. I moved the `COLMAP_LSD_ENABLED` macro definition from the main CMakeLists.txt file to `cmake/FindDependencies.cmake`. Do you also think that the other macros like `WIN32_LEAN_AND_MEAN`, `GLOG_USE_GLOG_EXPORT`, `GLOG_NO_ABBREVIATED_SEVERITIES`, etc. defined in the main CMakeLists.txt file must also be added to `COLMAP_COMPILE_DEFINITIONS` variable? I think they don't have a public scope.

cmake/CMakeHelper.cmake
```diff
@@ -92,6 +92,7 @@ macro(COLMAP_ADD_LIBRARY)
     target_link_libraries(${COLMAP_ADD_LIBRARY_NAME}
         PRIVATE ${COLMAP_ADD_LIBRARY_PRIVATE_LINK_LIBS}
         PUBLIC ${COLMAP_ADD_LIBRARY_PUBLIC_LINK_LIBS})
+    target_compile_definitions(${COLMAP_ADD_LIBRARY_NAME} PUBLIC ${COLMAP_COMPILE_DEFINITIONS})
```
ahojnnes, line 95: We'll also need to add this to the COLMAP_ADD_EXECUTABLE and COLMAP_ADD_TEST macros.

ahojnnes:
> I moved the COLMAP_LSD_ENABLED macro definition from the main CMakeLists.txt file to cmake/FindDependencies.cmake.

Thanks, makes sense. This is a relic from when some special LSD logic needed to be applied for vcpkg.

> Do you also think that the other macros like WIN32_LEAN_AND_MEAN, GLOG_USE_GLOG_EXPORT, GLOG_NO_ABBREVIATED_SEVERITIES, etc. defined in the main CMakeLists.txt file must also be added to COLMAP_COMPILE_DEFINITIONS variable?

No, these should stay private.

One other minor comment, otherwise this looks good to me. Thanks.

cmake/CMakeHelper.cmake
jhacsonmeza: Regarding that, I actually have a question. When the executables or tests link against any of the libraries, they'll inherit all the compile definitions from the library targets. So, shouldn't we rely on that and add explicitly the compile definitions?

jhacsonmeza:
I also have an additional question, should we keep `colmap_exe` lib? It seems it's not used at all.

cmake/CMakeHelper.cmake
ahojnnes: There could be executables that don't depend on libraries (and I guess in theory also tests, but that practically doesn't make sense). It's safer to just add them in both cases.

ahojnnes:
> I also have an additional question, should we keep `colmap_exe` lib? It seems it's not used at all.

Some external libraries depend on it.

cmake/CMakeHelper.cmake
jhacsonmeza: Thanks for clarifying! I just wanted to make sure. I will add the changes.

jhacsonmeza:
I added the compile definitions to `COLMAP_ADD_EXECUTABLE` and `COLMAP_ADD_TEST` with private scope because I think it doesn't make sense to use public for executables.

cmake/CMakeHelper.cmake
```diff
@@ -141,4 +143,5 @@ macro(COLMAP_ADD_TEST)
             install(TARGETS ${COLMAP_ADD_TEST_NAME} DESTINATION ${CMAKE_INSTALL_BINDIR})
         endif()
     endif()
+    target_compile_definitions(${COLMAP_ADD_TEST_NAME} PRIVATE ${COLMAP_COMPILE_DEFINITIONS})
```
ahojnnes: Needs to move inside the `if(TESTS_ENABLED)` condition.

ahojnnes:
```suggestion
        target_compile_definitions(${COLMAP_ADD_TEST_NAME} PRIVATE ${COLMAP_COMPILE_DEFINITIONS})
    endif()
```

PR #3401, opened 2025-06-12
Compute residuals with threadpool for model_merger

clementinboittiaux:
Hey @ahojnnes ,

We've recently had a computation bottleneck when trying to merge big reconstructions (2,500 to 35,000 common images). The heaviest step is computing the residuals for alignment. This PR simply computes the residuals with a threadpool.
On two models with 2,706 common images the alignment step drops from 813s to 209s with 64 threads.

*N.B.:* `THROW_CHECK_EQ` exceptions in threads are propagated in the main thread through `future.get()`.

### DIFF
--- a/src/colmap/estimators/alignment.cc
+++ b/src/colmap/estimators/alignment.cc
@@ -34,6 +34,7 @@
 #include "colmap/optim/loransac.h"
 #include "colmap/scene/projection.h"
 #include "colmap/util/logging.h"
+#include "colmap/util/threading.h"
 
 #include <unordered_map>
 
@@ -49,10 +50,12 @@ struct ReconstructionAlignmentEstimator {
 
   ReconstructionAlignmentEstimator(double max_reproj_error,
                                    const Reconstruction* src_reconstruction,
-                                   const Reconstruction* tgt_reconstruction)
+                                   const Reconstruction* tgt_reconstruction,
+                                   ThreadPool* thread_pool = nullptr)
       : max_squared_reproj_error_(max_reproj_error * max_reproj_error),
         src_reconstruction_(src_reconstruction),
-        tgt_reconstruction_(tgt_reconstruction) {
+        tgt_reconstruction_(tgt_reconstruction),
+        thread_pool_(thread_pool) {
     THROW_CHECK_GE(max_reproj_error, 0);
     THROW_CHECK_NOTNULL(src_reconstruction_);
     THROW_CHECK_NOTNULL(tgt_reconstruction_);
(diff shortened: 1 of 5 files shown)


src/colmap/estimators/alignment.cc
```diff
@@ -103,81 +106,112 @@ struct ReconstructionAlignmentEstimator {
 
     residuals->resize(src_images.size());
 
-    for (size_t i = 0; i < src_images.size(); ++i) {
-      const Image& src_image = *src_images[i];
-      const Image& tgt_image = *tgt_images[i];
+    if (thread_pool_ && src_images.size() > 10) {
```
ahojnnes, line 109:
```suggestion
    if (thread_pool_ && src_images.size() > thread_pool_->NumThreads()) {
```

src/colmap/estimators/alignment.cc
```diff
@@ -103,81 +106,112 @@ struct ReconstructionAlignmentEstimator {
...
+      std::vector<std::future<void>> futures;
```
ahojnnes: No need to accumulate the futures. You can just do thread_pool_->Wait(); to finish all tasks in the pool.

src/colmap/sfm/observation_manager.h
```diff
@@ -33,13 +33,15 @@
 #include "colmap/scene/image.h"
 #include "colmap/scene/reconstruction.h"
 #include "colmap/scene/track.h"
+#include "colmap/util/threading.h"
```
ahojnnes: Let's forward-decl threadpool here as well instead of including this header.

clementinboittiaux: Fixed

src/colmap/estimators/alignment.cc
clementinboittiaux: Fixed but does not propagate `THROW_CHECK_EQ` exceptions. How could this be solved?

clementinboittiaux: I removed `THROW_CHECK_EQ` from `ComputeImageResidual`, and only perform the checks once before computing the residuals.

clementinboittiaux:
Hey @ahojnnes ,
Do you think this would be okay to merge ?

src/colmap/estimators/alignment.cc
```diff
@@ -106,78 +109,108 @@ struct ReconstructionAlignmentEstimator {
...
+                               residuals]() {
+          ComputeImageResidual(
+              i, src_images, tgt_images, tgt_from_src, src_from_tgt, residuals);
+        });
+      }
+      thread_pool_->Wait();
+    } else {
+      for (size_t i = 0; i < src_images.size(); ++i) {
+        ComputeImageResidual(
+            i, src_images, tgt_images, tgt_from_src, src_from_tgt, residuals);
```
sarlinpe: How about instead passing `src_images[i]`, `tgt_images[i]`, and `residuals[i]`?

clementinboittiaux: Done!

clementinboittiaux:
Hey @sarlinpe , is this okay to merge?

PR #3402, opened 2025-06-12
Fix the seed to get rid of randomness/non-deterministic behaviour

patodichayan:


src/colmap/estimators/two_view_geometry.cc
```diff
@@ -100,6 +100,11 @@ TwoViewGeometry EstimateCalibratedHomography(
     const std::vector<Eigen::Vector2d>& points2,
     const FeatureMatches& matches,
     const TwoViewGeometryOptions& options) {
+  // Set a fixed random seed to ensure deterministic results
+  // This is necessary because in multi-threaded mode, different threads
+  // will process different image pairs in arbitrary order
+  SetPRNGSeed(42);
```
ahojnnes, line 106: Why is this not using the newly introduced `kRandomPRNGSeed` ?

ahojnnes:
Thanks, unless you set num_threads=1, the results will still not be perfectly deterministic, because of small numerical differences in various optimization routines, where accumulating sums/products is non-deterministic depending on the order of execution.

src/colmap/estimators/two_view_geometry.cc
patodichayan: True, will make that change, and push.

patodichayan:
> Thanks, unless you set num_threads=1, the results will still not be perfectly deterministic, because of small numerical differences in various optimization routines, where accumulating sums/products is non-deterministic depending on the order of execution.

Agreed, but after building on these changes, the results were more consistent compared to what I was getting before for the same set of images. 

If there are tests you want me to run? or do some sort of benchmarking, let me know, and I'd be happy to do that.

ahojnnes:
@StonerLing I believe your recent changes should have already addressed this as well?

StonerLing:
Yes. The RANSAC-based estimation is now fully reproducible, while the mapper’s results are nearly reproducible.

PR #3414, opened 2025-06-23
Add legacy docs from 3.8 to 3.11.

B1ueber2y:
Dependent on https://github.com/colmap/colmap.github.io/pull/14

### DIFF
--- a/doc/index.rst
+++ b/doc/index.rst
@@ -119,3 +119,4 @@ support (special credits to `Torsten Sattler <https://tsattler.github.io>`_).
    contribution
    license
    bibliography
+   legacy
--- a/doc/legacy.rst
+++ b/doc/legacy.rst
@@ -1,0 +1,10 @@
+Legacy Documentations
+=====================
+
+.. toctree::
+   :maxdepth: 1
+
+   v3.11 (2024-11-28) <https://colmap.github.io/legacy/3.11/index.html>
+   v3.10 (2024-07-23) <https://colmap.github.io/legacy/3.10/index.html>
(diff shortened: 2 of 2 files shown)


sarlinpe:
My concern with this approach is that switching between versions is more difficult, and the file tree in the left panel will look messier. Would it be more difficult to add a drop-down menu? sphinx_rtd_theme seems to support it: https://sphinx-rtd-theme.readthedocs.io/en/stable/configuring.html#confval-version_selector (or maybe this requires the doc to be hosted on readthedocs.com, then [that could be an alternative](https://www.codingwiththomas.com/blog/my-sphinx-best-practice-for-a-multiversion-documentation-in-different-languages))

B1ueber2y:
> My concern with this approach is that switching between versions is more difficult, and the file tree in the left panel will look messier. Would it be more difficult to add a drop-down menu? sphinx_rtd_theme seems to support it: https://sphinx-rtd-theme.readthedocs.io/en/stable/configuring.html#confval-version_selector (or maybe this requires the doc to be hosted on readthedocs.com, then [that could be an alternative](https://www.codingwiththomas.com/blog/my-sphinx-best-practice-for-a-multiversion-documentation-in-different-languages))

The thing about switching between versions is that, if we want to do switching back and forth, we would need to add this feature back in the previous versions as well, to update the sidebars, which is not maintainable without recompiling everything at each release. 

The versions are by default folded in the sidebars under "legacy documentations" so I dont think this will make it much messier : )

sarlinpe (approved):
Ok fair. If it's just about the version dropdown, patching old docs could work but is maybe not worth the effort.

ahojnnes (approved):

PR #3418, opened 2025-06-26
Return bad initial pair when number of triangulation is less than abs_pose_min_num_inliers

B1ueber2y:
Otherwise ``FindNextImages`` will return empty and the process will save a reconstruction with only two frames: https://github.com/colmap/colmap/blob/main/src/colmap/controllers/incremental_pipeline.cc#L575

### DIFF
--- a/src/colmap/controllers/incremental_pipeline.cc
+++ b/src/colmap/controllers/incremental_pipeline.cc
@@ -346,7 +346,8 @@ IncrementalPipeline::Status IncrementalPipeline::InitializeReconstruction(
   }
 
   // Number of triangulated points not enough for registering future images.
-  if (int(reconstruction.NumPoints3D()) < mapper_options.abs_pose_min_num_inliers) {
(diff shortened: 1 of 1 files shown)


B1ueber2y:
Not sure in practice we should even increase this threshold, since otherwise at the boundary condition it would require all points to be inliers at registration, which could end up trying to register forever.

src/colmap/controllers/incremental_pipeline.cc
```diff
@@ -345,6 +345,12 @@ IncrementalPipeline::Status IncrementalPipeline::InitializeReconstruction(
     return Status::BAD_INITIAL_PAIR;
   }
 
+  // Number of triangulated points not enough for registering future images.
+  if (int(reconstruction.NumPoints3D()) <
```
ahojnnes:
```suggestion
  if (static_cast<int>(reconstruction.NumPoints3D()) <
```

ahojnnes (approved):

PR #3419, opened 2025-06-27
Initialize from non-trivial frame pairs using generalized relative pose

ahojnnes:


src/colmap/sfm/incremental_mapper_impl.cc
```diff
@@ -539,18 +536,142 @@ std::vector<image_t> IncrementalMapperImpl::FindLocalBundle(
...
-
   const Image& image2 = database_cache.Image(image_id2);
+  const Frame& frame1 = database_cache.Frame(image1.FrameId());
+  const Frame& frame2 = database_cache.Frame(image2.FrameId());
+  const Rig& rig1 = database_cache.Rig(frame1.RigId());
+  const Rig& rig2 = database_cache.Rig(frame2.RigId());
+  const Camera& camera1 = database_cache.Camera(image1.CameraId());
   const Camera& camera2 = database_cache.Camera(image2.CameraId());
 
+  if (!rig1.IsRefSensor(camera1.SensorId()) ||
```
B1ueber2y: I think it should not matter here whether they are reference sensors or not?

ahojnnes: This is the mechanism by which we can check whether the image has a trivial frame or not. Note that in the database cache, the frame/rig ptrs are not set yet, so we cannot use image.HasTrivialFrame().

B1ueber2y: But what if they are non-trivial frames but both cameras are reference sensors?

ahojnnes: Great catch, should be fixed now.

sarlinpe (approved):
Thank you!

PR #3420, opened 2025-06-27
Add option to build with thread sanitizer flags

ahojnnes:


lpanaf:
Should this flag be tested in the pipeline as well? According to our tests, adding this flag can sometimes introduce very strange results

B1ueber2y:
> Should this flag be tested in the pipeline as well? According to our tests, adding this flag can sometimes introduce very strange results

Adding this into the tested pipeline would slower the CI process for quite a bit, which is not favorable for the project iterations. So I would vote for not having it.

ahojnnes:
Agreed with Shaohui. The intent was to introduce this option for convenience of local debugging.

PR #3422, opened 2025-06-27
Add test to ensure we can recover generalized relative pose scale for case of one panoramic rig

ahojnnes:


B1ueber2y (approved):

ahojnnes:
Merged as part of https://github.com/colmap/colmap/pull/3419

PR #3423, opened 2025-06-27
Estimate essential matrix using camera rays instead of points

ahojnnes:
Enables us to robustly initialize from panoramic camera rigs.


src/colmap/estimators/essential_matrix_test.cc
```diff
@@ -85,117 +84,45 @@ class EssentialMatrixFivePointEstimatorTests
...
-    ExpectAtLeastOneValidModel(estimator, points1, points2, expected_E, models);
+    ExpectAtLeastOneValidModel(estimator, rays1, rays2, expected_E, models);
   }
 }
 
 INSTANTIATE_TEST_SUITE_P(EssentialMatrixFivePointEstimator,
                          EssentialMatrixFivePointEstimatorTests,
                          ::testing::Values(5, 20, 1000));
 
-TEST(EssentialMatrixEightPointEstimator, Reference) {
```
ahojnnes, line 109: This test was a bit unnecessary with the improved tests. Just removed it.

src/colmap/geometry/homography_matrix.cc
```diff
@@ -192,41 +192,27 @@ namespace {
...
+    const Eigen::Vector3d point3D_in_cam2 = cam2_from_cam1 * point3D_in_cam1;
     const double error1 =
-        (points1[i] - point3D_in_cam1.hnormalized()).squaredNorm();
+        1 -
+        std::clamp(cam_rays1[i].dot(point3D_in_cam1.normalized()), -1.0, 1.0);
     const double error2 =
-        (points2[i] - point3D_in_cam2.hnormalized()).squaredNorm();
+        1 -
+        std::clamp(cam_rays2[i].dot(point3D_in_cam2.normalized()), -1.0, 1.0);
     reproj_residual_sum += error1 + error2;
```
sarlinpe, line 214: Should we update the names of this variable and of the function? This isn't a reprojection error anymore but a cosine similarity.

src/colmap/geometry/homography_matrix.cc
```diff
@@ -259,7 +245,7 @@ void PoseFromHomographyMatrix(const Eigen::Matrix3d& H,
     // non-squared reprojection errors other than avoid sqrt for efficiency and
     // consistency with the RANSAC cost function.
     const double reproj_residual_sum = CheckCheiralityAndReprojErrorSum(
-        cams2_from_cams1[i], points1, points2, &tentative_points3D);
+        cams2_from_cams1[i], cam_rays1, cam_rays2, &tentative_points3D);
     if (tentative_points3D.size() > points3D->size() ||
         (tentative_points3D.size() == points3D->size() &&
          reproj_residual_sum < best_reproj_residual_sum)) {
```
sarlinpe, line 248: For scoring based on the cosine similarity, the best model has a *higher* value. This might be the cause of the failing test.

ahojnnes: ... hm, I am computing 1 - cos_dist, so this should still be correct? (on my phone right now though).

ahojnnes: This whole decomposition logic is a bit fragile. The original paper only said that the solution can be narrowed down to two using cheirality testing and I added the reprojection-based scoring, which seemed to have worked in practice with some limited testing. Maybe it actually doesn't work (or there is a bug in the change I made).

sarlinpe: > ... hm, I am computing 1 - cos_dist, so this should still be correct?

You're right, I misread -_-

ahojnnes: The paper this is based on is: "Deeper understanding of the homography decomposition
for vision-based control".

src/colmap/geometry/triangulation_test.cc
```diff
@@ -78,14 +78,54 @@ TEST(TriangulatePoint, Nominal) {
...
       Rigid3d().ToMatrix(),
       Rigid3d(Eigen::Quaterniond::Identity(), Eigen::Vector3d(1, 0, 0))
           .ToMatrix(),
       Eigen::Vector2d(0, 0),
       Eigen::Vector2d(0, 0),
-      &xyz));
+      &point3D));
+}
+
+TEST(TriangulateMidPoint, Nominal) {
```
B1ueber2y, line 91: Maybe add a test on mid point when the rays do not perfectly intersect?

src/colmap/geometry/triangulation.h
```diff
@@ -45,29 +46,42 @@ namespace colmap {
...
 
+// Triangulate 3D mid-point from corresponding camera ray observations.
+//
+// @param cam2_from_cam1    Relative pose between camera pair.
+// @param cam_ray1          Corresponding 2D ray in first camera.
+// @param cam_ray2          Corresponding 2D ray in second camera.
+// @param point3D           Triangulated 3D point in first camera.
+//
+// @return                  Whether triangulation was successful.
+bool TriangulateMidPoint(const Rigid3d& cam2_from_cam1,
```
B1ueber2y, line 70: The three types of triangulation functions in this file should have consistent interfaces for whether to use two cam_from_worlds and triangulate in the world space, or use a relative pose and triangulate in the first camera?

src/colmap/geometry/homography_matrix.cc
B1ueber2y: Scoring based on cosine similarity is not ideal in terms of uncertainty modeling, but it should be fine if we only use it in the decomposition.

src/colmap/geometry/triangulation.h
ahojnnes: This was intentional for efficiency reasons. I think it's okay, if these do not have consistent interfaces.

B1ueber2y: Why was the other case inefficient? We could unify the interface here and input an identity rigid3d for cam1 when this is called?

B1ueber2y: Or, alternatively, overload the functions : )

src/colmap/geometry/triangulation_test.cc
ahojnnes: Done

src/colmap/geometry/triangulation.h
ahojnnes: The only place where this is currently used, it is used with a relative transformation. If we take absolute camera poses, then we'll have to recompose the relative pose again.

B1ueber2y (approved):

PR #3424, opened 2025-06-28
Fix ObservationManager

whuaegeanse:
1.Fix the bug in ObservationManager::FilterPoints3DWithSmallTriangulationAngle where the count of observations filtered by triangulation angle was incorrect.
2. Renamed num_filtered to num_filtered_observations to better express the meaning of the variable.

### DIFF
--- a/src/colmap/sfm/observation_manager.cc
+++ b/src/colmap/sfm/observation_manager.cc
@@ -280,12 +280,12 @@ size_t ObservationManager::FilterPoints3D(
     const double max_reproj_error,
     const double min_tri_angle,
     const std::unordered_set<point3D_t>& point3D_ids) {
-  size_t num_filtered = 0;
-  num_filtered +=
+  size_t num_filtered_observations = 0;
+  num_filtered_observations +=
       FilterPoints3DWithLargeReprojectionError(max_reproj_error, point3D_ids);
-  num_filtered +=
+  num_filtered_observations +=
       FilterPoints3DWithSmallTriangulationAngle(min_tri_angle, point3D_ids);
-  return num_filtered;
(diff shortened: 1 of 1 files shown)


ahojnnes:
Thank for catching this bug, the fix look valid to me. Ideally, we would adjust the unit tests to catch this issue by checking the return values of the filtering function calls. Is this something you could add as part of this PR? Thank you.

whuaegeanse:
The unit tests of filtering functions have been adjusted.

src/colmap/sfm/observation_manager_test.cc
```diff
@@ -85,39 +85,47 @@ TEST(ObservationManager, FilterPoints3D) {
...
   EXPECT_EQ(reconstruction.NumPoints3D(), 1);
-  obs_manager.FilterPoints3D(
-      0.0, 0.0, std::unordered_set<point3D_t>{point3D_id1 + 1});
+  EXPECT_EQ(obs_manager.FilterPoints3D(
+                0.0, 0.0, std::unordered_set<point3D_t>{point3D_id1 + 1}),
+            0);
   EXPECT_EQ(reconstruction.NumPoints3D(), 1);
-  obs_manager.FilterPoints3D(
-      0.0, 0.0, std::unordered_set<point3D_t>{point3D_id1});
+  EXPECT_EQ(obs_manager.FilterPoints3D(
```
ahojnnes, line 97: This is odd. Wouldn't we expect this to be non-zero?

ahojnnes: Ah, no, because the point has an empty track. I suggest to add observations, so we can make the expectation more meaningful? Same for other instances below?

PR #3434, opened 2025-06-30
Add missing include in mat.h

sarlinpe:
Used by `std::fill`


ahojnnes:
Let's get this one in first: https://github.com/colmap/colmap/pull/3440, so we have new changes associated with the new dev version.

PR #3438, opened 2025-06-30
Fix ReadPositionPriorData to return valid and numerically more stable Position prior data

whuaegeanse:
1. Fix ReadPositionPriorData to return valid and numerically more stable Position prior data.
2. Adjust the unit test of SpatialPairGenerator to test that the matrix returned by ReadPositionPriorData has the correct number of rows.


src/colmap/feature/pairing.cc
```diff
@@ -639,51 +639,105 @@ Eigen::RowMajorMatrixXf SpatialPairGenerator::ReadPositionPriorData(
...
+            static_cast<float>(position_prior(0) - position_offset(0));
+        position_matrix(position_idx, 1) =
+            static_cast<float>(position_prior(1) - position_offset(1));
+        position_matrix(position_idx, 2) = static_cast<float>(
+            options_.ignore_z ? 0 : position_prior(2) - position_offset(2));
     }
-
-    num_positions += 1;
   }
   return position_matrix;
```
ahojnnes: Wouldn't it be easier to `return position_matrix.topRows(num_positions);` instead of the proposed fix of iterating twice over the data?

src/colmap/feature/pairing_test.cc
```diff
@@ -239,6 +239,176 @@ TEST(SequentialPairGenerator, Quadratic) {
 }
 
 TEST(SpatialPairGenerator, Nominal) {
+  {
+    constexpr int kNumImages = 3;
```
ahojnnes, line 365: I suggest to decouple this into different `TEST` cases with descriptive names for what we test.

src/colmap/feature/pairing_test.cc
```diff
@@ -239,6 +239,176 @@ TEST(SequentialPairGenerator, Quadratic) {
...
+
+    auto cache = std::make_shared<FeatureMatcherCache>(
+        options.CacheSize(), THROW_CHECK_NOTNULL(database));
+    SpatialPairGenerator generator(options, cache);
+
+    Eigen::RowMajorMatrixXf position_matrix =
+        generator.ReadPositionPriorData(*cache);
+    EXPECT_EQ(position_matrix.rows(), 3);
+
+    return;
```
ahojnnes: I assume this return statement is not intentional?

src/colmap/feature/pairing.cc
whuaegeanse: Yes

src/colmap/feature/pairing_test.cc
whuaegeanse: Sorry, this is the test code I wrote and forgot to delete it

whuaegeanse:
> Thank you for the fix. The implementation indeed had a bug. Left some comments.

The bug in this PR has been fixed.

ahojnnes:
@whuaegeanse Thank you, I pushed another commit with some improvements and an explicit test for large coordinates.

PR #3443, opened 2025-07-01
Redefine glog's branch prediction macros

ahojnnes:
Older glog versions do not have these macros defined, which leads to a compilation error.


B1ueber2y (approved):
Thanks. Does it make sense to put these macros into a separate file and include it in util/logging.h?

sarlinpe:
Yes I would also move them to `logging_macros.h` or so, such that we can use a different license header with clearer marking of which code it applies to.

ahojnnes:
Thanks, done.

sarlinpe (approved):
Thank you!

Issue #3450, opened 2025-07-03
Regression in spatial matching pair generation between 3.11.0 and 3.13.0


dlichy:
**Describe the bug**
There appears to be a regression in the spatial matching logic between COLMAP versions `3.11.0.dev0` and `3.13.0.dev0`, specifically in how matching pairs are generated using pose priors read from image metadata.

---

In version `3.13.0.dev0` (Commit `f2137129`), significantly more pairs are generated during spatial matching, but far fewer of them contain usable matches. This is in contrast to version `3.11.0.dev0` (Commit `189478b6`), which produces fewer total pairs but far more that contain high numbers of feature matches.

---

**Setup:**
- Dataset: 790 images with pose priors embedded in image metadata (EXIF/GPS).
- Feature extraction: Identical between both runs — same features and descriptors stored in the database.
- Matching mode: `spatial_matcher`

---

**Comparison:**

| Version   | Total `matches` rows | Pairs with >100 matches |
|-----------|----------------------|--------------------------|
| 3.11.0    | ~19,100              | ~7,200                   |
| 3.13.0    | ~39,000              | ~730                     |

Despite the number of pairs doubling in 3.13.0, the number of meaningful matches dropped significantly.

Notably, for the `pair_id`s that exist in both versions' `matches` tables, the number of matches is consistent, suggesting the issue is with **pair generation**, likely in the spatial matching logic.

---

**Suspected cause:**
- Regression in the logic determining which image pairs to match when pose priors are used.
- Possibly related to changes in ENU coordinate conversion or spatial filtering thresholds.
- The large number of low-match pairs in 3.13.0 indicates many non-overlapping or irrelevant pairs are being selected.

---

**To Reproduce**
I cannot share this dataset, but I will try to replicate this on an open source one and update.

---

**Expected behavior**
Database should be the same or similar.

---

**Environment:**
 - OS: Ubuntu 22.04.5 LTS
 - COLMAP Version 3.13.0.dev0` (Commit `f2137129`)

ahojnnes:
Thank you for reporting. @whuaegeanse Can this be related to your recent bugfix?

ahojnnes:
@dlichy To narrow down the problem, can you confirm whether the database contains identical GPS pose priors?

dlichy:
[...]

PR #3451, opened 2025-07-03
Fix bug in spatial matching (issue #3450)

dlichy:
Correctly subtract mean of position priors after conversion to ECEF coordinates.


ahojnnes:
Thank you!

PR #3453, opened 2025-07-03
Fix nvidia/docker container runtime

MasahiroOgawa:
# Objective
Fix colmap docker container issues.

# Problem and solutions
## Problem
When I tested colmap autoreconstructor inside docker container using original repoo's South Building dataset in [here](https://demuc.de/colmap/datasets/), it failed with below error message.
>colmap: error while loading shared libraries: libmkl_intel_thread.so: cannot open shared object file: No such file or directory

## Cause 
In docker build (build.sh), installation of libmkl_intel_thread silently failed because of dependency issues. Necessary dependencies wasn't installed because it uses --no-install-recommends --no-install-suggests options in apt install, but it is necessary  to reduce the final image size. 

## Solution
Add necessary dependencies manually one by one including other necessary libraries (e.g. libopengl0, libmetis5,libomp5). And originally, run.sh didn't use the docker container built by build.sh, so I also fixed to use it.
(I noticed that adding libraries are also done in main today, so I merged main repo's changes.)

# Test
Using South Building dataset, I ran below full colmap reconstruction.
```
cd docker/
./build.sh
./run.sh ../dataset
colmap automatic_reconstructor --image_path ./images --workspace_path .
```
And I successfully got below result.
image


ahojnnes:
Thanks very much, greatly appreciated.

PR #3454, opened 2025-07-04
Autoselect compatible nvidia for docker

MasahiroOgawa:
# Objective
Select compatible nvidia libraries automatically.
Select automatically in this order; Nvidia driver -> Nvidia container toolkit -> Cuda -> nvidia container

# Original Problem and solution
## Problem
When I ran 
```
 ./run.sh ../dataset
```
I encountered below error; 
> docker: Error response from daemon: failed to create task for container: failed to create shim task: OCI runtime create failed: runc create failed: unable to start container process: error during container init: error running prestart hook #0: exit status 1, stdout: , stderr: Auto-detected mode as 'legacy'
nvidia-container-cli: requirement error: unsatisfied condition: cuda>=12.9

## Cause
This is because I didn't install the proper Nvidia driver.

## Solution
 But it is preferable that setup-ubuntu.sh automatically select compatible drivers and Cuda version.

# Test
I confirmed all below procedure finished successfully using [South building dataset](https://demuc.de/colmap/datasets/)
```
cd docker
./setup-ubuntu.sh
./build.sh
./run.sh ../dataset
colmap automatic_reconstructor --image_path ./images --workspace_path .
```

### DIFF
--- a/docker/README.md
+++ b/docker/README.md
@@ -42,8 +42,8 @@
 
 ## Build from Scratch
 
-After completing steps 1,2, you can alternatively build the docker image from
-scratch based on the **Dockerfile** (e.g., with your own modifications) using:
+After completing steps 1-2, you can build the Docker image from scratch using the **Dockerfile**.
+First, update the CUDA and Ubuntu versions in Dockerfile lines 1-2 to match your system, then:
 
 ```
 ./build.sh


ahojnnes:
Thanks, looks good to me.

PR #3459, opened 2025-07-05
Replace FreeImage with OpenImageIO

ahojnnes:
* Simple image reader/writer tests with 1000 random JPG images shows 2.5x speedup vs FreeImage.
* Image resampling / bilinear interpolation operations should be faster due to direct/inlined access to underlying image buffers.
* pycolmap package size increases by ~20% due to additional OpenImageIO dependencies.

### DIFF
--- a/src/pycolmap/sensor/bitmap.cc
+++ b/src/pycolmap/sensor/bitmap.cc
@@ -87,55 +87,53 @@ void BindBitmap(pybind11::module& m) {
            "width"_a,
            "height"_a,
            "as_rgb"_a,
-           "linear_colorspace"_a = false),
+           "linear_colorspace"_a = false)
       .def("to_array", &ArrayFromBitmap)
-          .def_static("from_array",
-                      &BitmapFromArray,
-                      "array"_a,
-                      "linear_colorspace"_a = false,
-                      "Create bitmap as a copy of array. Returns RGB bitmap, "
-                      "if array has shape (H, W, 3), or grayscale bitmap, if "
-                      "array has shape (H, W[, 1]).")
-          .def("write",
-               &Bitmap::Write,
-               "path"_a,
-               "delinearize_colorspace"_a = true,
-               "Write bitmap to file at given path. Defaults to converting to "
-               "sRGB colorspace for file storage.")
-          .def_static(
-              "read",
-              [](const std::string& path,
-                 bool as_rgb,
-                 bool linearize_colorspace) -> py::typing::Optional<Bitmap> {
-                Bitmap bitmap;
-                if (!bitmap.Read(
-                        path,
-                        /*as_rgb=*/as_rgb,
-                        /*linearize_colorspace=*/linearize_colorspace)) {
-                  return py::none();
-                }
-                return py::cast(bitmap);
-              },
-              "path"_a,
-              "as_rgb"_a,
-              "linearize_colorspace"_a = false,
-              "Read bitmap at given path and convert to grey- or colorscale. "
-              "Defaults to keeping the original colorspace (potentially "
-              "non-linear) for image processing.")
-          .def("rescale",
-               &Bitmap::Rescale,
-               "new_width"_a,
-               "new_height"_a,
-               "filter"_a = BitmapRescaleFilter::kBilinear,
-               "Rescale image to the new dimensions.")
-          .def_property_readonly("width", &Bitmap::Width, "Width of the image.")
-          .def_property_readonly(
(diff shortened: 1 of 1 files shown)


src/colmap/controllers/feature_extraction.cc
```diff
@@ -205,7 +205,7 @@ class SiftFeatureExtractorThread : public Thread {
                             &image_data.keypoints,
                             &image_data.descriptors);
             }
-            if (image_data.mask.Data()) {
+            if (image_data.mask.NumBytes() > 0) {
```
sarlinpe: How about adding `Bitmap::IsEmpty()`?

src/colmap/sensor/bitmap.cc
```diff
@@ -35,267 +35,179 @@
...
+struct OIIOInitializer {
+  OIIOInitializer() {
+    OIIO::attribute("threads", 1);
+    OIIO::attribute("exr_threads", 1);
+  }
 
-struct FreeImageInitializer {
-  FreeImageInitializer() { FreeImage_Initialise(); }
-  ~FreeImageInitializer() { FreeImage_DeInitialise(); }
+  ~OIIOInitializer() { OIIO::shutdown(); }
```
sarlinpe: This seems to be available only since v2.5.3.0 https://github.com/AcademySoftwareFoundation/OpenImageIO/blob/v2.5.3.0-beta1/src/include/OpenImageIO/imageio.h#L2788-L2793

sarlinpe (approved):

sarlinpe:
Really nice update, thank you!

ahojnnes:
Thanks for the review. This needs a bit more eval. It seems to be quite a bit slower than FreeImage for reading images and there are some discrepancies in terms of extracted features, which are probably related to different filters when resizing images, etc.

src/colmap/sensor/bitmap.cc
ahojnnes: Thank you!

src/colmap/sensor/bitmap.h
```diff
@@ -274,6 +274,8 @@ bool Bitmap::GetPixel(const int x,
 
   if (IsGrey()) {
     color->r = data_[y * width_ + x];
+    color->g = color->r;
```
B1ueber2y, line 279: Need to update the tests as well.

B1ueber2y (approved):
Thank you! This is a very nice and necessary migration, especially that FreeImage is not actively maintained anymore.

src/colmap/sensor/bitmap.h
ahojnnes: Done

.github/workflows/build-pycolmap.yml
```diff
@@ -174,7 +174,6 @@ jobs:
         env:
           CIBW_ARCHS_MACOS: ${{ matrix.config.arch }}
           CIBW_TEST_REQUIRES: 'pytest enlighten==1.13.0'
-          CIBW_TEST_COMMAND: 'pytest {project}/python/examples/custom_incremental_pipeline_test.py'
```
B1ueber2y: Why removing this test? I think the CI failure is not relevant? Rather it was failing here: https://github.com/colmap/colmap/blob/main/pyproject.toml#L45. 

I have a suspicion that ``CIBW_TEST_REQUIRES`` somehow overrides the variable from pyproject.toml. Adding mypy==1.12.1 in the previous line might fix it.

B1ueber2y: Strange that it was never a problem before though.

B1ueber2y: Actually, we should move this pytest into pyproject.toml as well.

ahojnnes: I am debugging windows which gets stuck in this test. No idea why and I fail to build the package locally on my windows machine.

chpatrick:
Hi, how this is going? Anything I can do to help?

ahojnnes:
The main blocker is that pycolmap seems to freeze for an unknown reason. Couldn't get to the bottom of the issue though. If you have time to debug, it would be greatly appreciated. Thanks.

B1ueber2y:
> Thanks for the review. This needs a bit more eval. It seems to be quite a bit slower than FreeImage for reading images and there are some discrepancies in terms of extracted features, which are probably related to different filters when resizing images, etc.

I cannot clearly recall what happened here. Was this performance issue resolved?

ahojnnes:
I also don't perfectly remember but I think I fixed the perf issues.

PR #3460, opened 2025-07-05
Add hierarchical_pose_prior_mapper to CLI

dlichy:
### Summary

This adds a new CLI utility, `hierarchical_pose_prior_mapper`, which enables running the hierarchical mapping pipeline with pose priors.

### Implementation Notes

- The implementation closely mirrors the existing `hierarchical_mapper` and `pose_prior_mapper` logic.
- Currently, there is significant code duplication between `mapper` and `pose_prior_mapper`, and likewise between `hierarchical_mapper` and `hierarchical_pose_prior_mapper`.
- As a potential improvement, we could consolidate `mapper` and `pose_prior_mapper` into a single tool, with a flag to enable pose prior support — and do the same for their hierarchical counterparts. I'm happy to implement this refactor if it's aligned with project direction.


sarlinpe (approved):
Thank you! It would be great if, as you suggest, you could consolidate the regular and pose_prior versions of the mappers in a follow-up PR, as there is currently a lot of duplicate code.

PR #3462, opened 2025-07-06
Decouple image import and feature extraction

sarlinpe:
This PR moves the image import step (creating images, cameras, rigs, pose priors in the database) outside of the feature extraction. Benefits:
- reduces code complexity
- makes it possible to first import images and then run feature extraction outside of COLMAP based on the imported images -- this was possible only in pycolmap and incurred code duplication
- guarantees that images are always sorted by name since image import is not multi-threaded

One downside of this is that we now need to read image pixels twice (to read the EXIF for import and later for extraction) but this is likely negligible, and we can later make it possible to read EXIF without the pixels.

Some points should be discussed:
1. Are we OK with such breaking change of the CLI?
2. If not, should we have `feature_extractor` call `image_importer` if the database is empty? We will still break the argument names of the `mask_path` and `camera_mask_path` options, but these are not widely used.
3. Are we OK adding an extra step in the GUI?
4. Do we want to make `mask_path` and `camera_mask_path` part of the `SiftExtractionOptions`?

TODO:

- [ ] Add image importer widget -- once the above points are resolved.
- [ ] Merge `ImageReader` and `ImageImporter` -- in a follow-up PR.
- [ ] Add tests for `FeatureExtractorController::ReadImageData` -- in a follow-up PR.
- [ ] Update the documentation.

### DIFF
--- a/cmake/FindMetis.cmake
+++ b/cmake/FindMetis.cmake
@@ -85,6 +85,11 @@ else()
         PATHS
         ${METIS_CHECK_LIBRARY_DIRS})
 
+    if(GK_LIBRARIES)
+        set(METIS_LIBRARIES ${METIS_LIBRARIES} ${GK_LIBRARIES})
+        message(STATUS "Found GKlib")
+    endif()
+
     if(METIS_INCLUDE_DIRS AND METIS_LIBRARIES)
         set(METIS_FOUND TRUE)
         message(STATUS "Found Metis")
--- a/docker/run-gui.sh
+++ b/docker/run-gui.sh
@@ -1,11 +1,89 @@
-docker pull colmap/colmap:latest
-docker run \
-    -e QT_XCB_GL_INTEGRATION=xcb_egl \
-    -e DISPLAY=:0 \
-    -w /working \
-    -v $1:/working \
-    -v /tmp/.X11-unix:/tmp/.X11-unix \
-    --gpus all \
-    --privileged \
-    -it colmap/colmap:latest \
-    colmap gui
+#!/bin/bash
+#
+# A robust script to run the COLMAP GUI inside a Docker container,
+# with automatic GPU detection and flexible argument passing.
+
+# --- Initial Checks ---
+# Exit immediately if a command exits with a non-zero status.
+set -e
+
+# --- Cleanup Function and Trap ---
+# This function will be called automatically when the script exits
+# to ensure that X server permissions are always restored.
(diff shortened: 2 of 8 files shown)


src/colmap/controllers/feature_extraction.cc
```diff
@@ -423,10 +417,59 @@ class FeatureExtractorController : public Thread {
...
+          if (ExistsFile(alt_mask_path)) {
+            mask_path = std::move(alt_mask_path);
+            exists_mask = true;
+          }
+        }
+      }
+      if (exists_mask) {
+        if (!image_data.mask.Read(mask_path, false)) {
+          LOG(ERROR) << image.Name() << " Failed to read the mask file";
+          image_data.mask.Deallocate();  // Skip the mask but not the image!
```
sarlinpe, line 465: Previously, no features would be extracted for an image when its mask cannot be found or read. This seems wrong to me.

src/colmap/controllers/feature_extraction.cc
```diff
@@ -84,13 +84,33 @@ void MaskKeypoints(const Bitmap& mask,
...
+    case FeatureExtractionStatus::FAILURE:
+      return "FAILURE: Failed to process the image.";
+    case FeatureExtractionStatus::IMAGE_EXISTS:
+      return "IMAGE_EXISTS: Features for image were already extracted.";
+    case FeatureExtractionStatus::BITMAP_ERROR:
+      return "BITMAP_ERROR: Failed to read the image file format.";
+    default:
+      return "Unknown";
+  }
+}
```
sarlinpe, line 107: All of this is unnecessary if we decide to not propagate the image data through the queues when it is unsuccessful. The only downside of this is that error messages won't appear in the order of the `FeatureWriterThread`.

ahojnnes:
Thanks, this is a good initiative and I had the same in mind in the past. The main motivation for keeping them coupled was the one you already mentioned in that I wanted to avoid reading images twice. Another one is that an image in the database has valid features. I don't think we can start assuming images to be ordered by name, in particular, as we want to support the use case, where one can run multiple rounds of image import / feature extraction.

ahojnnes:
Another note is that this PR will have a lot of conflicts with the changes here: https://github.com/colmap/colmap/pull/3068

I should probably extract the non-ALIKED/LightGlue related refactoring into a separate PR until we land the torch changes (which still need some more testing due to cross-platform issues).

PR #3463, opened 2025-07-07
Add min_num_neighbors constraint for spatial matching

StonerLing:
This PR adds a `min_num_neighbors` option to `spatial_matcher`, ensuring that each image is matched with at least the specified number of neighbors. This makes matching more adaptive to different image distributions and removes the need to manually tune `max_distance` in some cases.

- Adds `min_num_neighbors` to `SpatialMatchingOptions`. When set to a positive value, the matcher will always include at least that number of neighbors, even if some exceed max_distance
- Adds `CMakeUserPresets.json` and `.clangd` to `.gitignore`

### DIFF
--- a/src/colmap/feature/pairing.cc
+++ b/src/colmap/feature/pairing.cc
@@ -134,6 +134,8 @@ VocabTreeMatchingOptions SequentialMatchingOptions::VocabTreeOptions() const {
 bool SpatialMatchingOptions::Check() const {
   CHECK_OPTION_GT(max_num_neighbors, 0);
   CHECK_OPTION_GT(max_distance, 0.0);
+  CHECK_OPTION_GE(min_num_neighbors, 0);
+  CHECK_OPTION_LE(min_num_neighbors, max_num_neighbors);
   return true;
 }
 
@@ -557,6 +559,8 @@ SpatialPairGenerator::SpatialPairGenerator(
   Eigen::RowMajorMatrixXf position_matrix = ReadPositionPriorData(*cache);
   const size_t num_positions = position_idxs_.size();
 
+  THROW_CHECK_LT(options_.min_num_neighbors, num_positions);
+
   LOG(INFO) << StringPrintf(" in %.3fs", timer.ElapsedSeconds());
   if (num_positions == 0) {
     LOG(INFO) << "=> No images with location data.";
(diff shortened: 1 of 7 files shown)


src/colmap/feature/pairing.cc
```diff
@@ -613,16 +645,14 @@ std::vector<std::pair<image_t, image_t>> SpatialPairGenerator::Next() {
...
-      static_cast<float>(options_.max_distance * options_.max_distance);
   for (int j = 0; j < knn_; ++j) {
     // Check if query equals result.
     if (index_matrix_(current_idx_, j) == static_cast<int>(current_idx_)) {
       continue;
     }
 
     // Since the nearest neighbors are sorted by distance, we can break.
-    if (distance_matrix_(current_idx_, j) > max_distance) {
+    if (distance_matrix_(current_idx_, j) > max_distance_squared_) {
```
ahojnnes: Wouldn't it be easier to change this single line to:
```suggestion
    if (distance_matrix_(current_idx_, j) > max_distance_squared_ && j + 1 >= min_num_neighbors) {
```

StonerLing: True, much cleaner. I'll update the PR accordingly.

src/colmap/feature/pairing.cc
```diff
@@ -621,8 +626,10 @@ std::vector<std::pair<image_t, image_t>> SpatialPairGenerator::Next() {
       continue;
     }
 
-    // Since the nearest neighbors are sorted by distance, we can break.
-    if (distance_matrix_(current_idx_, j) > max_distance) {
+    // Since the nearest neighbors are sorted by distance, we can break
+    // once the distance is too large and enough neighbors are collected.
+    if (distance_matrix_(current_idx_, j) > max_distance &&
```
ahojnnes: I liked your previous change of calling it max_distance_squared. Maybe we should also call it distance_squared_matrix_ ?

src/colmap/feature/pairing.cc
```diff
@@ -557,6 +560,8 @@ SpatialPairGenerator::SpatialPairGenerator(
   Eigen::RowMajorMatrixXf position_matrix = ReadPositionPriorData(*cache);
   const size_t num_positions = position_idxs_.size();
 
+  THROW_CHECK_LT(options_.min_num_neighbors, num_positions);
```
ahojnnes: Do we really need this check and crash the program? I would maybe turn it into a LOG(WARNING) instead.

ahojnnes (approved):

src/colmap/feature/pairing.cc
StonerLing: Done, makes sense.

src/colmap/feature/pairing.cc
StonerLing: Done.

ahojnnes:
Thank you!

PR #3470, opened 2025-07-10
Update FindMetis.cmake to also link GK_LIBRARIES

yeicor:
I might be missing something, but I was encountering missing symbols from GKlib during the linking stage. It seems like an oversight: FindMetis.cmake correctly locates the GKlib library, but it doesn't actually link it after finding it. This pull request adds GKlib to the list of linked libraries to resolve that issue.


yeicor:
It appears that Docker, Mac, and Ubuntu were using a statically compiled GKlib in the metis library. I changed the file again so that both ways of linking should work.

ahojnnes (approved):

ahojnnes:
Thank you, LGTM.

PR #3480, opened 2025-07-12
Fix potential deadlock in job queue

huluoboge:
Fix a bug where the Stop function's state issue may cause the Pop() function to enter an infinite loop.
template <typename T>
void JobQueue<T>::Stop() {
    stop_ = true; // problem code
  push_condition_.notify_all();
  pop_condition_.notify_all();
}

typename JobQueue<T>::Job JobQueue<T>::Pop() {
  std::unique_lock<std::mutex> lock(mutex_);
  while (jobs_.empty() && !stop_) {
    push_condition_.wait(lock);
  ....

​​Bug Description:​​
There is a race condition between Stop() and Pop() that can lead to deadlock. The sequence is:

In Pop(), the thread evaluates while (jobs_.empty() && !stop_) and sees stop_ = false
Before entering push_condition_.wait(lock), the Stop() function gets executed:
Sets stop_ = true
Calls push_condition_.notify_all()
The Pop() thread then proceeds to push_condition_.wait(lock)
Because the notification was sent before the wait began, it's missed
Now the thread is stuck waiting indefinitely with stop_ = true, causing a deadlock

### DIFF
--- a/src/colmap/controllers/image_reader.cc
+++ b/src/colmap/controllers/image_reader.cc
@@ -50,14 +50,6 @@ ImageReader::ImageReader(const ImageReaderOptions& options, Database* database)
     : options_(options), database_(database), image_index_(0) {
   THROW_CHECK(options_.Check());
 
-  // Ensure trailing slash, so that we can build the correct image name.
-  options_.image_path =
-      EnsureTrailingSlash(StringReplace(options_.image_path, "\\", "/"));
-  if (!options_.mask_path.empty()) {
-    options_.mask_path =
-        EnsureTrailingSlash(StringReplace(options_.mask_path, "\\", "/"));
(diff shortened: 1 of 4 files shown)


src/colmap/util/threading.h
```diff
@@ -409,7 +409,10 @@ void JobQueue<T>::Wait() {
 
 template <typename T>
 void JobQueue<T>::Stop() {
-  stop_ = true;
+  {
+    std::unique_lock<std::mutex> lock(mutex_);
+    stop_ = true;
```
ahojnnes, line 413: Thanks, we should be able to make the variable non atomic now, since it is always protected by a mutex?

ahojnnes (approved):

PR #3481, opened 2025-07-13
Update GetPathBaseName to rely on std::filesystem

sarlinpe:


B1ueber2y (approved):
Thanks. Nit: does it make sense to use the std::filesystem support ``fs_path.has_filename()`` instead of ``basename.empty()``? Though it should be equivalent.

sarlinpe:
Indeed, good catch, thanks.

PR #3482, opened 2025-07-14
Ensure UTF-8 encoding for database paths passed to SQLite open

StonerLing:
`sqlite3_open_v2` only accepts UTF-8 encoded paths. This PR adds a function to convert ANSI (platform) strings to UTF-8 on Windows to correctly handle paths containing non-ASCII characters.

### DIFF
--- a/src/colmap/scene/database.cc
+++ b/src/colmap/scene/database.cc
@@ -432,7 +432,7 @@ void Database::Open(const std::string& path) {
   // Modifications to the database will still be serialized, but multiple
   // connections can read concurrently.
   SQLITE3_CALL(sqlite3_open_v2(
-      path.c_str(),
+      PlatformToUTF8(path).c_str(),
       &database_,
       SQLITE_OPEN_READWRITE | SQLITE_OPEN_CREATE | SQLITE_OPEN_NOMUTEX,
       nullptr));
--- a/src/colmap/util/string.h
+++ b/src/colmap/util/string.h
@@ -34,6 +34,22 @@
 
 namespace colmap {
 
+namespace internal {
+#ifdef _WIN32
+// Convert a string from the specified Windows code page to UTF-8.
+//
+// This function is used internally for unit testing to verify roundtrip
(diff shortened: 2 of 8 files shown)


src/colmap/util/string.h
```diff
@@ -34,6 +34,22 @@
 
 namespace colmap {
 
+namespace internal {
```
ahojnnes: Nit: suggest to move the internal namespace to the end of the header.

src/colmap/util/string_test.cc
```diff
@@ -225,5 +228,86 @@ TEST(StringContains, Nominal) {
   EXPECT_FALSE(StringContains("ab", "c"));
 }
 
+#ifdef _WIN32
+TEST(ConversionBetweenPlatformAndUTF8, NonASCIIStringRoundtripWin) {
+  const std::array<std::pair<std::string_view, unsigned int>, 12>
```
ahojnnes: nit: We could deduplicate the definition of the test strings as a static variable outside of the test cases, as they seem identical on both platforms?

src/colmap/scene/database.cc
```diff
@@ -432,7 +432,7 @@ void Database::Open(const std::string& path) {
   // Modifications to the database will still be serialized, but multiple
   // connections can read concurrently.
   SQLITE3_CALL(sqlite3_open_v2(
-      path.c_str(),
+      PlatformToUTF8(path).c_str(),
```
ahojnnes, line 435: We could add a minimal test case, where we open a database file with UTF-8 characters in a local test directory?

src/colmap/util/string.h
StonerLing: Done

src/colmap/util/string_test.cc
StonerLing: Done

src/colmap/scene/database.cc
StonerLing: Done

ahojnnes (approved):

PR #3492, opened 2025-07-17
Add random_seed option to RANSACOptions for reproducible TwoViewGeometry estimation

StonerLing:
This PR adds a `random_seed` parameter to `RANSACOptions` for deterministic results in multithreaded environments.
- `random_seed = -1` (default) preserves the original nondeterministic behavior.
- Setting it to a fixed value makes estimation deterministic and reproducible.
- Currently applied and tested only in `TwoViewGeometry` estimation.


src/colmap/optim/ransac_test.cc
```diff
@@ -136,9 +136,63 @@ TEST(RANSAC, SimilarityTransform) {
...
+
+  ASSERT_TRUE(report1.success);
+  ASSERT_TRUE(report2.success);
+
+  // Results should be exactly the same.
+  EXPECT_EQ(report1.support.num_inliers, report2.support.num_inliers);
+  EXPECT_EQ(report1.inlier_mask, report2.inlier_mask);
+  EXPECT_THAT(
+      report1.model,
+      EigenMatrixNear(report2.model, std::numeric_limits<double>::epsilon()));
```
ahojnnes: Should we test for exact equality without approximation?

src/colmap/estimators/two_view_geometry_test.cc
```diff
@@ -311,5 +339,156 @@ TEST(EstimateTwoViewGeometryPose, PlanarOrPanoramic) {
...
+      camera1, points1, camera2, points2, matches, two_view_geometry_options);
+
+  two_view_geometry_options.ransac_options.random_seed = 42;
+  const TwoViewGeometry geometry2 = EstimateTwoViewGeometry(
+      camera1, points1, camera2, points2, matches, two_view_geometry_options);
+
+  // Using the same random seed should produce identical results.
+  EXPECT_THAT(
+      geometry1.H,
+      EigenMatrixNear(geometry2.H, std::numeric_limits<double>::epsilon()));
```
ahojnnes: Should we test for identical results here and equivalent cases above?

src/colmap/optim/ransac.h
```diff
@@ -199,6 +205,11 @@ RANSAC<Estimator, SupportMeasurer, Sampler>::Estimate(
     const std::vector<typename Estimator::Y_t>& Y) {
   THROW_CHECK_EQ(X.size(), Y.size());
 
+  if constexpr (is_randomized_sampler<Sampler>::value) {
+    options_.random_seed == -1 ? SetPRNGSeed(std::random_device{}())
```
ahojnnes: Do we need to overwrite the prng in the case of -1 or just keep it as is?

ahojnnes (commented):
Thanks, LGTM apart from a few minor comments.

src/colmap/optim/ransac_test.cc
StonerLing: EXPECT_EQ passed on my machine. I used EigenMatrixNear to handle possible floating-point precision issues, but it can switch to exact comparison if preferred.

src/colmap/optim/ransac.h
StonerLing: I think keeping the existing seed is preferable.

src/colmap/optim/ransac_test.cc
ahojnnes: Thanks, great.

ew-pallon:
Thanks for this contribution, I was just looking for this!

Do you already know when this will be available as part of an official release?

PR #3497, opened 2025-07-18
Configure trivial rigs for unconfigured images

ahojnnes:
This PR adds support for configuring rigs for only a subset of images. The images without configuration are assigned to trivial rigs/frames.

### DIFF
--- a/src/colmap/scene/rig.cc
+++ b/src/colmap/scene/rig.cc
@@ -214,6 +214,7 @@ void ApplyRigConfig(const std::vector<RigConfig>& configs,
   database.ClearRigs();
 
   const std::vector<Image> images = database.ReadAllImages();
+  std::set<image_t> configured_image_ids;
 
   for (const RigConfig& config : configs) {
     Rig rig;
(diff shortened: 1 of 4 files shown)


src/colmap/scene/rig.cc
```diff
@@ -290,33 +292,59 @@ void ApplyRigConfig(const std::vector<RigConfig>& configs,
...
+        // input reconstruction, e.g., when self-calibrating the rigs and
+        // cameras from a subset of images.
+        const Image* image = reconstruction->FindImageWithName(
+            database.ReadImage(data_id.id).Name());
+        if (image == nullptr || !image->HasPose()) {
+          continue;
         }
+        const sensor_t sensor_id = image->CameraPtr()->SensorId();
+        const Rig& rig = database.ReadRig(frame.RigId());
+        if (rig.IsRefSensor(sensor_id)) {
```
B1ueber2y, line 337: The following block can use ``frame.SetCamFromWorld`` instead.

B1ueber2y (approved):

src/colmap/scene/rig.cc
ahojnnes: It looks as if, but it can unfortunately not, because the frame read from the database doesn't have a valid rig pointer yet.

PR #3498, opened 2025-07-18
Add option to perform geometric verification with rig constraints

ahojnnes:


src/colmap/estimators/two_view_geometry.cc
```diff
@@ -325,6 +326,150 @@ TwoViewGeometry EstimateTwoViewGeometry(
...
+  auto maybe_add_camera = [&cameras_vec, &cams_from_rig, &camera_id_to_idx](
+                              const Rig& rig, const Camera& camera) {
+    const auto [it, inserted] =
+        camera_id_to_idx.emplace(camera.camera_id, cameras_vec.size());
+    if (inserted) {
+      cameras_vec.push_back(camera);
+      if (rig.IsRefSensor(camera.SensorId())) {
+        cams_from_rig.push_back(Rigid3d());
+      } else {
+        if (!rig.Sensors().count(camera.SensorId())) {
```
B1ueber2y: THROW_CHECK(rig.HasSensor(camera.SensorId()))?

src/colmap/estimators/two_view_geometry.cc
```diff
@@ -325,6 +326,150 @@ TwoViewGeometry EstimateTwoViewGeometry(
...
+          LOG(ERROR) << "Camera " << camera.camera_id
+                     << " not found in rig: " << rig;
+        }
+        cams_from_rig.push_back(rig.SensorFromRig(camera.SensorId()));
```
B1ueber2y: Actually, we never have a mechanism to enforce that each camera belongs to only one rig. The logic here can be broken if this is not the case?

Does it make sense to check here, either rig1 and rig2 are equivalent, or they have exclusive camera ids? And furthermore, does it make sense to add this exclusive camera id check also for the reconstruction during AddRig?

B1ueber2y: Or, since now we enforce that each camera must belong to a rig, does it make sense to add rig_id_ in the Camera class?

ahojnnes: The database enforces that each camera is only part of one rig but we don't have the same constraint in the reconstruction.

src/colmap/estimators/two_view_geometry.cc
ahojnnes: Oh, this was just for debugging. Removed. The next line will throw anyway.

src/colmap/estimators/two_view_geometry.cc
ahojnnes: If we want to do the same in the reconstruction, we could add a rig_id_ in the camera struct or maybe just internal to the reconstruction.

B1ueber2y: > Actually, we never have a mechanism to enforce that each camera belongs to only one rig. The logic here can be broken if this is not the case?
> 
> Does it make sense to check here, either rig1 and rig2 are equivalent, or they have exclusive camera ids? And furthermore, does it make sense to add this exclusive camera id check also for the reconstruction during AddRig?

Shall we first add a check in this PR for rig1 and rig2 either equivalent or having exclusive camera ids, and add the constraints the reconstruction or camera class in a separate PR?

ahojnnes: I think this PR works without a problem if this constraint is violated. The only thing that will happen is that the same camera will be added twice in the cameras vector. However, as we keep the cameras constant, this causes no issue other than a bit of extra memory usage?

B1ueber2y: > I think this PR works without a problem if this constraint is violated. The only thing that will happen is that the same camera will be added twice in the cameras vector. However, as we keep the cameras constant, this causes no issue other than a bit of extra memory usage?

I dont think so. camera_id_to_idx is an ``unordered_map``, so the second camera will be skipped and therefore the corresponding cam_from_rig will be only for the first rig, if this constraint is violated.

ahojnnes: Thanks, you are right. Fixed.

ahojnnes:
@B1ueber2y Added a test.

src/colmap/estimators/two_view_geometry.h
```diff
@@ -134,6 +139,16 @@ TwoViewGeometry EstimateTwoViewGeometry(
     FeatureMatches matches,
     const TwoViewGeometryOptions& options);
 
+std::vector<std::pair<std::pair<image_t, image_t>, TwoViewGeometry>>
```
B1ueber2y, line 153: Should we add some comments to align with the other functions in this file?

ahojnnes: Done

PR #3515, opened 2025-07-30
Upgrade ruff and automatically apply fixes

ahojnnes:


scripts/format/python.sh
```diff
@@ -23,4 +23,4 @@ echo "Formatting ${num_files} files"
 
 # shellcheck disable=SC2086
 ruff format --config ${root_folder}/ruff.toml ${all_files}
-ruff check --config ${root_folder}/ruff.toml ${all_files}
+ruff check --config ${root_folder}/ruff.toml ${all_files} --fix --unsafe-fixes
```
B1ueber2y: I would not include unsafe-fixes here.

B1ueber2y (approved):

PR #3519, opened 2025-08-01
Replace force_H_use with homography_usage in TwoViewGeometryOptions

StonerLing:
Replace `force_H_use ` with `homography_usage`  to control whether and how homography is estimated during two-view geometry:

- `AUTO` (default): Estimate H, E, F, and choose the best model based on inlier ratios.
- `DISABLE`: Skip H.
- `FORCE`: Only use H.

Helps avoid instability in H estimation in cases of poor match quality (set `homography_usage` to `DISABLE`).

### DIFF
--- a/src/colmap/controllers/option_manager.cc
+++ b/src/colmap/controllers/option_manager.cc
@@ -327,9 +327,10 @@ void OptionManager::AddMatchingOptions() {
 
   AddAndRegisterDefaultOption("TwoViewGeometry.min_num_inliers",
                               &two_view_geometry->min_num_inliers);
-  AddAndRegisterDefaultOption("TwoViewGeometry.homography_usage",
-                              &two_view_geometry->homography_usage,
-                              "{0 = AUTO, 1 = FORCE, 2 = DISABLE}");
+  AddAndRegisterDefaultOption(
+      "TwoViewGeometry.homography_usage",
+      reinterpret_cast<int*>(&two_view_geometry->homography_usage),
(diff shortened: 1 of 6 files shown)


sarlinpe:
Thank you! What exactly is the instability in H estimation that you mention?

StonerLing:
> Thank you! What exactly is the instability in H estimation that you mention?

I add a unit test to show the instability in H estimation. It may not be very rigorous, but the issue showed up clearly in my natural scene datasets. Exposing `min_E_F_inlier_ratio` and `max_H_inlier_ratio` might be another option.

ahojnnes:
I am not sure I understand the test. The points are neither on a plane nor do the cameras undergo pure rotation, so the two view geometry cannot be modeled by a homography. I wouldn't expect the resulting homographies to be consistent in this degenerate setup?

StonerLing:
> I am not sure I understand the test. The points are neither on a plane nor do the cameras undergo pure rotation, so the two view geometry cannot be modeled by a homography. I wouldn't expect the resulting homographies to be consistent in this degenerate setup?

I’ve removed the inaccurate test. It originated from partially planar-like scenes with low texture in my degenerate dataset, where COLMAP repeatedly picks H with large inconsistency.

Back to this PR — it just adds an option to control H usage, which I believe is a reasonable addition?

ahojnnes:
Would you be able to share some sample imagery with the problem? We may be able to address the issue in a more principled manner by detecting these degeneracies explicitly.

PR #3521, opened 2025-08-02
Add option to filter stationary points in two-view geometry estimation

ahojnnes:


src/colmap/estimators/two_view_geometry.cc
```diff
@@ -93,6 +93,92 @@ inline bool IsImagePointInBoundingBox(const Eigen::Vector2d& point,
          point.y() <= maxy;
 }
 
+bool DetectWatermarkMatches(const Camera& camera1,
```
ahojnnes: Simply moved to anonymous namespace.

src/colmap/estimators/two_view_geometry.cc
```diff
@@ -93,6 +93,92 @@ inline bool IsImagePointInBoundingBox(const Eigen::Vector2d& point,
...
+  const double matches_in_border_ratio =
+      static_cast<double>(num_matches_in_border) / num_inliers;
+
+  if (matches_in_border_ratio < options.watermark_min_inlier_ratio) {
+    return false;
+  }
+
+  // Check if matches follow a translational model.
+
+  RANSACOptions ransac_options = options.ransac_options;
```
B1ueber2y: Looking again at the code here, I think it would be better to set the max error here in the ransac options, as the default is for two view geometry estimation and 4px is too large to describe a detection noise in stationary matches. Can we use the stationary_matches_max_error option here?

src/colmap/estimators/two_view_geometry.cc
```diff
@@ -310,8 +397,12 @@ TwoViewGeometry EstimateTwoViewGeometry(
     const std::vector<Eigen::Vector2d>& points1,
     const Camera& camera2,
     const std::vector<Eigen::Vector2d>& points2,
-    const FeatureMatches& matches,
+    FeatureMatches matches,
     const TwoViewGeometryOptions& options) {
+  if (options.ignore_stationary_matches) {
```
B1ueber2y: If ignore_stationary_matches is set true, we will never use the configuration type WATERMARK anymore, which means that multiple_ignore_watermark option will be broken when used together.

ahojnnes: I am not sure I understand. Why would we ignore the watermark configuration?

src/colmap/estimators/two_view_geometry.cc
ahojnnes: I also don't understand. Do you suggest to introduce a new "watermark_detection_max_error" option field instead of using the ransac error?

src/colmap/estimators/two_view_geometry.cc
B1ueber2y: As far as I understand, if the stationary matches are filtered out at the start, detect watermark will never return true as the inliers in ransac have already been mostly filtered?

src/colmap/estimators/two_view_geometry.cc
B1ueber2y: Yes.

src/colmap/estimators/two_view_geometry.cc
ahojnnes: Watermark detection will still trigger if the watermark is in different regions of the image. Notice that watermark detection estimates a translational model.

B1ueber2y: Yes agreed. But this would result in bad watermark detection for static watermarks.

B1ueber2y (approved):

src/colmap/estimators/two_view_geometry.cc
ahojnnes: Yes, I guess the watermark detection is not great. It would be nicer if it was "ignore_watermark" and we would equivalently filter out the matches before the estimation. It has the advantage that we are still able to estimate a valid model from the remaining matches, which is currently not possible if we are unlucky and latch onto the watermark matches as the dominant model. WDYT?

PR #3523, opened 2025-08-03
Upgrade pybind11 to 3.0.0

ahojnnes:


sarlinpe:
This needs https://github.com/sizmailov/pybind11-stubgen/pull/259 as well. Which make me think that I still haven't pushed my patches upstream, will do so.

ahojnnes:
> This needs https://github.com/sizmailov/pybind11-stubgen/pull/259 as well. Which make me think that I still haven't pushed my patches upstream, will do so.

That would be great. Thank you!

sarlinpe (approved):

PR #3525, opened 2025-08-04
Fix comment for FilterStationaryMatches

sarlinpe:


src/colmap/estimators/two_view_geometry.h
```diff
@@ -176,8 +176,8 @@ bool DetectWatermarkMatches(const Camera& camera1,
                             const std::vector<char>& inlier_mask,
                             const TwoViewGeometryOptions& options);
 
-// Detect if inlier matches are caused by a watermark.
-// A watermark causes a pure translation in the boundaries of the images.
+// Remove matches that are caused by a static watermark, which has the same
```
ahojnnes: Thanks, this comment was wrongly copied from the previous function... The main purpose for implementing this feature are robotic / data capture platforms with rigidly mounted cameras, where the platform itself is visible in the images and the data is crowd-sourced (i.e., think of mapillary), so we cannot easily generate a mask to ignore the features.

sarlinpe: Got it. It would be nice to unify this with the watermark detector, e.g. make the detection stationary or translation-only and discard matches in both cases.

Issue #3479, opened 2025-07-11
Performance slowdown in Mapper in 3.13 compared to 3.11


dlichy:
I’ve noticed a pretty significant performance drop in the mapper command when upgrading from COLMAP 3.11.0.dev0 to 3.13.0.dev0.

On my machine:

- **3.11.0.dev0** (commit \`189478b6\`) runs the [Person-Hall dataset](https://demuc.de/colmap/datasets/) in about **7.5 minutes**  
- **3.13.0.dev0** (commit \`c3ebe1ff\`) takes about **11 minutes**

Same database (from link above), same system, using default settings.

### Some extra info:

- **CPU:** 13th Gen Intel(R) Core(TM) i9-13950HX (24 cores, 32 threads)  
- **System:** Ubuntu-based, running COLMAP inside Docker  
- **Build:** Used the official Dockerfile from the COLMAP repo to build both versions  
- **Runs:** Used the same command-line settings and flags

On larger datasets (especially when using Hierarchical Mapping), the slowdown can be even worse, sometimes more than **2× slower** in 3.13.

Is this a known regression? Any idea on how to debug this?
Happy to give you more info if needed.

ahojnnes:
I suggest to enable profiling and use the profiling helper script to identify the bottleneck. We'd be very interested to learn about the root cause and fix it.

dlichy:
Hi @ahojnnes , I ran the profiling script on both COLMAP 3.11 and 3.13 using the same setup. I'm not entirely sure how to interpret the results or pinpoint what might be causing the slowdown, but I’ve attached the output of the following commands below:

```bash
./profile_binary.sh ...
perf report -i .perf.data --stdio --percent-limit=0.1 -g graph,0.5,caller > perf_pl_0.1_colmap3.11.txt
```
and similarly for colmap 3.13.

If you have any insights based on this, I’d really appreciate it. Let me know if there’s anything else I can try or provide to help debug this further.

[perf_pl_0.1_colmap3.11.txt](https://github.com/user-attachments/files/21554170/perf_pl_0.1_colmap3.11.txt)

[perf_pl_0.1_colmap3.13.txt](https://github.com/user-attachments/files/21554171/perf_pl_0.1_colmap3.13.txt)

Thanks again!

ahojnnes:
Thank you again for reporting the issue. I believe, I managed to reproduce and root-cause the issue. Here is a potential fix: https://github.com/colmap/colmap/pull/3527

[...]

PR #3527, opened 2025-08-04
Fix bundle adjustment performance regression due to changed Gauge

ahojnnes:
Based on reported issue https://github.com/colmap/colmap/issues/3479, I root-caused a performance regression in bundle adjustment to different gauge fixing behavior. In particular, the change from fixing the gauge with two cameras vs. fixing it with three points causes a ~40% slowdown in global bundle adjustment. No significant difference can be measured in local bundle adjustment. I could also not find a significant difference in the obtained solution quality. This PR reverts back to using two cameras for fixing the gauge. If no such two cameras can be found, which can occasionally happen in case of configured non-trivial rigs, we fall back to fixing three points.

### DIFF
--- a/src/colmap/controllers/bundle_adjustment.cc
+++ b/src/colmap/controllers/bundle_adjustment.cc
@@ -91,6 +91,10 @@ void BundleAdjustmentController::Run() {
   for (const image_t image_id : reconstruction_->RegImageIds()) {
     ba_config.AddImage(image_id);
   }
+  // Fixing the gauge with two cameras leads to a more stable optimization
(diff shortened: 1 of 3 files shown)


sarlinpe (approved):
Wow nice find! Is the slow-down due to increased step time or increased number of iterations until convergence?

B1ueber2y (approved):

ahojnnes:
> Wow nice find! Is the slow-down due to increased step time or increased number of iterations until convergence?

Already discussed offline, but for the record here: primarily more steps to converge.

Issue #3529, opened 2025-08-04
the rig translation is scaled up compared to configuration


lacayqwq:
I recorded videos using a ZED stereo camera. First, I performed feature extraction using the accurate camera parameters provided by the ZED SDK. Then, I configured the rig according to the baseline length of the ZED. After that, I conducted feature matching and reconstruction. In the reconstruction results, the rotations of the two cameras in the same frame strictly follow the rig configuration, but the translation (about -0.162) is scaled up compared to the real baseline length (-0.12). This issue occurs consistently across multiple scenario tests. Why is this happening? (I have already disabled the refinement of camera parameters and rig parameters in the reconstruction settings.)

```
[
  {
    "cameras": [
      {
        "image_prefix": "rig1/left/",
        "ref_sensor": true,
        "camera_model_name": "PINHOLE",
        "camera_params": [524.7269897460938, 524.7269897460938, 634.0955810546875, 357.984375]
      },
      {
        "image_prefix": "rig1/right/",
        "cam_from_rig_rotation": [1.0, 0.0, 0.0, 0.0],
        "cam_from_rig_translation": [-0.12015447998046875, 0, 0],
        "camera_model_name": "PINHOLE",
        "camera_params": [524.7269897460938, 524.7269897460938, 634.0955810546875, 357.984375]
      }
    ]
  }
]
```

B1ueber2y:
Hey. There is normalization operation in the process of incremental mapping for numerical stability, so the reconstruction is not kept metric in the process. Nonetheless, it should work if you just rescale the reconstruction to match the scale of the original translation.

lacayqwq:
Thank you, this has been very helpful to me.

clemense:
@lacayqwq Can you share your workflow? Is it
`feature_extractor` --> `sequential_matcher` --> `rig_configurator` --> `point_triangulator` --> `bundle_adjuster` ? 

Because I'm always running into a segfault during point triangulation.

PR #3530, opened 2025-08-04
Revert scale changes for metric reconstruction in rig sfm.

B1ueber2y:
Fixes https://github.com/colmap/colmap/issues/3529

Can be also concerning for hierarchical pipeline https://github.com/colmap/colmap/blob/main/src/colmap/controllers/hierarchical_pipeline_test.cc#L116.


ahojnnes:
Thanks, I am a bit worried about numerical issues due to repeated multiplication of the relative scale factor. Could we alternatively find the average scale between optimized and original registered rigs?

B1ueber2y:
> Thanks, I am a bit worried about numerical issues due to repeated multiplication of the relative scale factor. Could we alternatively find the average scale between optimized and original registered rigs?

Done. This new approach will need access to the database cache to get the initial rig extrinsics.

src/colmap/scene/reconstruction.cc
```diff
@@ -524,6 +524,36 @@ void Reconstruction::Transform(const Sim3d& new_from_old_world) {
   }
 }
 
+void Reconstruction::RevertScaleChanges(const DatabaseCache& database_cache) {
```
ahojnnes: I suggest extracting this as a free-standing function and move it to `colmap/estimators/alignment.h`. It could have the following interface:
```
AlignReconstructionToOrigRigScales(
      const std::unordered_map<rig_t, Rig>& orig_rigs, Reconstruction* reconstruction);
```

B1ueber2y: Thanks. This is done now.

src/colmap/estimators/alignment.cc
```diff
@@ -537,4 +537,37 @@ bool MergeReconstructions(const double max_reproj_error,
...
+bool AlignReconstructionToOrigRigScales(
+    const std::unordered_map<rig_t, Rig>& orig_rigs,
+    Reconstruction* reconstruction) {
+  double scale_sum = 0;
+  int scale_count = 0;
+  for (const auto& [rig_id, rig_orig] : orig_rigs) {
+    double scale_sum_rig = 0;
+    int scale_count_rig = 0;
+    for (auto& [sensor_id, sensor_from_rig_orig] : rig_orig.Sensors()) {
+      if (!sensor_from_rig_orig.has_value()) continue;
```
ahojnnes: nit: let's please always use braces around if statement bodies.

src/colmap/estimators/alignment.cc
```diff
@@ -537,4 +537,37 @@ bool MergeReconstructions(const double max_reproj_error,
...
 
+bool AlignReconstructionToOrigRigScales(
+    const std::unordered_map<rig_t, Rig>& orig_rigs,
+    Reconstruction* reconstruction) {
+  double scale_sum = 0;
+  int scale_count = 0;
+  for (const auto& [rig_id, rig_orig] : orig_rigs) {
+    double scale_sum_rig = 0;
+    int scale_count_rig = 0;
+    for (auto& [sensor_id, sensor_from_rig_orig] : rig_orig.Sensors()) {
```
ahojnnes: nit
```suggestion
    for (auto& [sensor_id, sensor_from_orig_rig] : rig_orig.Sensors()) {
```

src/colmap/estimators/alignment.cc
```diff
@@ -537,4 +537,37 @@ bool MergeReconstructions(const double max_reproj_error,
...
+      if (!sensor_from_rig_orig.has_value()) continue;
+      // Here we do not include rigs that are panoramic.
+      if (sensor_from_rig_orig->translation.norm() < 1e-6) continue;
```
ahojnnes: Extract the norm as a temp variable and reuse below to avoid redundant computation?

src/colmap/estimators/alignment.cc
```diff
@@ -537,4 +537,37 @@ bool MergeReconstructions(const double max_reproj_error,
...
+    }
+    if (scale_count_rig > 0) {
+      scale_sum += scale_sum_rig / scale_count_rig;
+      ++scale_count;
+    }
+  }
+  if (scale_count == 0) return false;
+  double scale = scale_sum / scale_count;
+  Sim3d new_from_old_world;
+  new_from_old_world.scale = 1.0 / scale;
```
ahojnnes:
```suggestion
  Sim3d new_from_old_world;
  new_from_old_world.scale = scale_count / scale_sum;
```

src/colmap/estimators/alignment.cc
```diff
@@ -537,4 +537,37 @@ bool MergeReconstructions(const double max_reproj_error,
   return true;
 }
 
+bool AlignReconstructionToOrigRigScales(
```
ahojnnes, line 540: Let's add a unit test. Thank you.

src/colmap/controllers/incremental_pipeline.cc
```diff
@@ -582,6 +587,10 @@ void IncrementalPipeline::Reconstruct(
           reconstruction->UpdatePoint3DErrors();
           LOG(INFO) << "Keeping successful reconstruction";
           mapper.EndReconstruction(/*discard=*/false);
+          if (!options_->ba_refine_sensor_from_rig) {
```
ahojnnes: Why are we only doing this when not refining the sensor from rig? We could just always do it?

B1ueber2y: I was being a bit careful as our current alignment is not robust, but thinking of it this rescaling should not hurt anything. So i have changed to always using it. Thanks.

src/colmap/controllers/incremental_pipeline_test.cc
```diff
@@ -128,7 +132,9 @@ TEST(IncrementalPipeline, WithoutNoiseAndWithNonTrivialFrames) {
                                *reconstruction_manager->Get(0),
                                /*max_rotation_error_deg=*/1e-2,
                                /*max_proj_center_error=*/1e-3,
-                               /*num_obs_tolerance=*/0);
+                               /*num_obs_tolerance=*/0,
+                               /*align=*/true,
+                               /*check_scale=*/!refine_sensor_from_rig);
```
ahojnnes: Is this still necessary? We can always check the scale now?

src/colmap/controllers/incremental_pipeline_test.cc
```diff
@@ -43,7 +43,8 @@ void ExpectEqualReconstructions(const Reconstruction& gt,
                                 const double max_rotation_error_deg,
                                 const double max_proj_center_error,
                                 const double num_obs_tolerance,
-                                const bool align = true) {
+                                const bool align = true,
```
ahojnnes, line 46: Unused?

src/colmap/controllers/incremental_pipeline_test.cc
B1ueber2y: Here we are checking whether the scale is exactly 1.0. But if the sensor_from_rig is refined they wont match exactly?

src/colmap/controllers/incremental_pipeline_test.cc
B1ueber2y: Which one is unused here?

ahojnnes: align?

src/colmap/controllers/incremental_pipeline_test.cc
ahojnnes: OK, got it. I guess otherwise we could do EXPECT_NEAR to still check that the scale is roughly close to 1?

src/colmap/controllers/incremental_pipeline_test.cc
ahojnnes: Forget it. I thought you added the parameter but it was already there before. It just looks like it was changed because you added a comma....

B1ueber2y: It is used in PriorBasedSfMWithoutNoise?

src/colmap/controllers/incremental_pipeline_test.cc
B1ueber2y: Done. I used a loose threshold when sensor_from_rig is refined.

Issue #3537, opened 2025-08-05
Error when importing pycolmap


nushakrishnan:
Hello! I am working with pycolmap and pyceres and have pip installed the two libraries in a python environment. When I run -

```python
import pycolmap
```
I face the error 

```
Traceback (most recent call last):
  File "/local/home/akrishnan/research/envs/lamaria/lib/python3.10/site-packages/pycolmap/__init__.py", line 7, in <module>
    from . import _core
ImportError: generic_type: type "PositiveExponentialManifold" referenced unknown base type "ceres::Manifold"
```

What could the issue be? I would appreciate any help, thank you in advance!

**Environment:**
 - OS: Ubuntu 22.04
 - Pycolmap version: 3.12.3
 - Pyceres version: 2.4

B1ueber2y:
I can confirm that this is easily reproducible on ubuntu. It happens when pycolmap and pyceres are installed together from pypi. The reason is the inconsistency of cibuildwheel for the new releases.

B1ueber2y:
Relevant: https://github.com/cvg/pyceres/pull/52

B1ueber2y:
Also relevant: https://github.com/colmap/colmap/pull/3494

PR #3538, opened 2025-08-05
Change parameterization in cost function to log scale and remove custom manifold from colmap. 

B1ueber2y:
Fixes: https://github.com/colmap/colmap/issues/3537

Cherry-picked from https://github.com/colmap/colmap/pull/3494 since there were no consensus on changing the Sim3d structure due to numerical stability.


B1ueber2y:
Its a bit unfortunate that a release was made 44 minutes ago. We would have to patch another one...

B1ueber2y:
Maintaining cibuildtool consistency for the pycolmap and pyceres pypi releases is impractical since we update it a lot. Now that pycolmap 3.12.3, pyceres 2.4, and the heads of the two repos are using three different cibuildtool versions. So we should take action to remove the custom manifold from pycolmap, as it has brought us so much trouble.

PR #3540, opened 2025-08-05
Throw exception when image.SetFramePtr() sets a frame without the image listed as its data.

B1ueber2y:
Otherwise the reconstruction data structure is broken, and the round trip wont write out the images. 

Seems to catch issues in some of the test construction.

### DIFF
--- a/src/colmap/scene/reconstruction.cc
+++ b/src/colmap/scene/reconstruction.cc
@@ -253,9 +253,6 @@ void Reconstruction::AddImage(class Image image) {
   } else {
     image.SetFramePtr(&frame);
   }
-  THROW_CHECK(frame.HasDataId(image.DataId()))
-      << "Image " << image.ImageId() << " does not exist in frame "
-      << frame.FrameId();
   const image_t image_id = image.ImageId();
   THROW_CHECK(images_.emplace(image_id, std::move(image)).second);
 }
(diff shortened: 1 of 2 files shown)


src/colmap/scene/reconstruction.cc
```diff
@@ -253,6 +253,10 @@ void Reconstruction::AddImage(class Image image) {
   } else {
     image.SetFramePtr(&frame);
   }
+  if (!frame.HasDataId(image.DataId())) {
+    LOG(FATAL_THROW) << "Image " << image.ImageId()
+                     << " does not exist in frame " << frame.FrameId();
+  }
```
ahojnnes:
```suggestion
  THROW_CHECK(frame.HasDataId(image.DataId())) << "Image " << image.ImageId()
                     << " does not exist in frame " << frame.FrameId();
```

src/colmap/scene/reconstruction.cc
```diff
@@ -253,6 +253,10 @@ void Reconstruction::AddImage(class Image image) {
   } else {
     image.SetFramePtr(&frame);
   }
+  if (!frame.HasDataId(image.DataId())) {
```
ahojnnes: Let's add a test case with EXPECT_THROW.

B1ueber2y:
Thanks. Added an update here https://github.com/colmap/colmap/pull/3540/commits/9b4d3b8d8cface6b3e304fe361a3343ca140cfb0 to throw exception at ``image.SetFramePtr`` as well and updated the PR title. PTAL @ahojnnes 

Will follow up with the same safety check for ``Frame.SetRigPtr`` and ``reconstruction.AddFrame()`` later tonight in a separate PR.

B1ueber2y:
One thing that is still extremely unsafe for the reconstruction class is that, there is no guarantee that all images in ``frame.DataIds()`` exist in the reconstruction. This is conceptually fine, as we can accept that only partial set of associated data exist in certain reconstruction, but we would need to update a lot of logic for iterating on ``frame.ImageIds()`` to include a check with ``reconstruction.ExistsImage()``. For example, ``NumRegImages()`` will currently break when such case happens.

PR #3541, opened 2025-08-06
Abstract database interface and sqlite implementation

ahojnnes:
Introduces an abstract interface for the database, so we can support different implementations (e.g., Postgres/MySQL, key-value stores, etc.). New implementations can be registered statically.


ahojnnes:
Seems like the pybind smart_holder feature is only available in 3.0.0. I cannot make the python bindings work without it though. @sarlinpe FYI

src/pycolmap/scene/database.cc
```diff
@@ -100,6 +454,22 @@ void BindDatabase(py::module& m) {
...
+            all_pair_ids.reserve(pair_ids_and_num_matches.size());
+            std::vector<int> all_num_matches;
+            all_num_matches.reserve(pair_ids_and_num_matches.size());
+            for (auto& [pair_id, num_matches] : pair_ids_and_num_matches) {
+              all_pair_ids.push_back(pair_id);
+              all_num_matches.push_back(num_matches);
+            }
+            return std::make_pair(std::move(all_pair_ids),
+                                  std::move(all_num_matches));
+          })
```
paulinus, line 472: This does not seem to be called from any file or test. It was added here just for completeness of the interface, right?

paulinus (approved):
LGTM. It is large, i hope i did not miss errors, but i imagine most would already be caught by the compiler since a lot of it was variable to pointer changes.

On the wrapper side, is it the case that you finally did not need to use `py::trampoline_self_life_support` because the Database objects are returned using shared pointers rather than unique pointers?  I'm not familiar enough with wrapping abstract classes to understand if it is needed.

src/pycolmap/scene/database.cc
ahojnnes: Yes, this was just missing from the python bindings, otherwise no change.

ahojnnes:
> On the wrapper side, is it the case that you finally did not need to use py::trampoline_self_life_support because the Database objects are returned using shared pointers rather than unique pointers? I'm not familiar enough with wrapping abstract classes to understand if it is needed.

Yes, without the trampoline support (introduced by Pybind11 3.X), this was not easily possible.

PR #3542, opened 2025-08-06
Use pybind smart holder for virtual interfaces

ahojnnes:
Needs pybind11 3.0.0 or later.


ahojnnes:
@sarlinpe Thank you very much for fixing the pybind-stub issues. This seems to work now :-)

sarlinpe (approved):
You can alternatively use `py::classh` as shorthand to minimize diffs.

PR #3543, opened 2025-08-06
Avoid frame having data from a sensor that does not exist in its associated rig. 

B1ueber2y:
As a follow up to https://github.com/colmap/colmap/pull/3540 for safety improvement, this PR includes the following fixes:

* ``AddFrame`` should not add a frame with empty data as its never useful in the reconstruction. NumData() is added for the frame class.
* ``Frame.SetRigPtr`` should set a rig that has all sensors in the frame data supported.
* Conversely, when a rig pointer is available for the frame, ``Frame.AddDataId`` should add data that with the sensor available in the associated rig. 

Related tests are added.

### DIFF
--- a/src/colmap/scene/image_test.cc
+++ b/src/colmap/scene/image_test.cc
@@ -126,6 +126,7 @@ TEST(Image, FramePtr) {
   rig.AddRefSensor(sensor_t(SensorType::CAMERA, 1));
   Frame frame;
   frame.SetFrameId(1);
+  frame.SetRigId(1);
   frame.SetRigPtr(&rig);
   Image image;
   image.SetImageId(1);
@@ -142,6 +143,7 @@ TEST(Image, SetResetPose) {
   rig.AddRefSensor(sensor_t(SensorType::CAMERA, 1));
   Frame frame;
   frame.SetFrameId(1);
+  frame.SetRigId(1);
   frame.SetRigPtr(&rig);
   Image image;
   image.SetImageId(1);
@@ -167,6 +169,7 @@ TEST(Image, ConstructCopy) {
   rig.AddRefSensor(sensor_t(SensorType::CAMERA, 1));
   Frame frame;
   frame.SetFrameId(1);
+  frame.SetRigId(1);
   frame.SetRigFromWorld(Rigid3d());
   frame.SetRigPtr(&rig);
   Image image;
@@ -189,6 +192,7 @@ TEST(Image, AssignCopy) {
   rig.AddRefSensor(sensor_t(SensorType::CAMERA, 1));
   Frame frame;
   frame.SetFrameId(1);
+  frame.SetRigId(1);
   frame.SetRigFromWorld(Rigid3d());
   frame.SetRigPtr(&rig);
   Image image;
@@ -297,6 +301,7 @@ TEST(Image, ProjectionCenter) {
   rig.AddRefSensor(sensor_t(SensorType::CAMERA, 1));
   Frame frame;
   frame.SetFrameId(1);
+  frame.SetRigId(1);
   frame.SetRigFromWorld(Rigid3d());
   frame.SetRigPtr(&rig);
   Image image;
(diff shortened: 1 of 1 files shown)


src/colmap/scene/reconstruction_test.cc
```diff
@@ -241,12 +243,10 @@ TEST(Reconstruction, AddFrame) {
   EXPECT_EQ(reconstruction.NumFrames(), 1);
   EXPECT_EQ(reconstruction.NumRegFrames(), 0);
   EXPECT_EQ(reconstruction.NumImages(), 0);
-  EXPECT_EQ(reconstruction.NumRegImages(), 0);
   EXPECT_EQ(reconstruction.NumPoints3D(), 0);
   reconstruction.Frame(1).SetRigFromWorld(Rigid3d());
   reconstruction.RegisterFrame(1);
   EXPECT_EQ(reconstruction.NumRegFrames(), 1);
-  EXPECT_EQ(reconstruction.NumRegImages(), 0);
```
B1ueber2y, line 249: ``NumRegImages()`` behavior here is broken, as mentioned in the end of https://github.com/colmap/colmap/pull/3540. It should return 0 since no ``AddImage`` is called, but will actually return 1 as we do not check if the image exists in the reconstruction. This is a separate issue that should not be tested here, and will be fixed in another separate PR.

src/colmap/scene/frame.h
```diff
@@ -53,6 +53,7 @@ class Frame {
   inline std::set<data_t>& DataIds();
   inline const std::set<data_t>& DataIds() const;
   inline void AddDataId(data_t data_id);
+  inline size_t NumData() const;
```
ahojnnes: nit
```suggestion
  inline size_t NumDataIds() const;
```

ahojnnes: Maybe do a search & replace of existing DataIds().size().

src/colmap/scene/frame.h
```diff
@@ -128,7 +129,14 @@ std::set<data_t>& Frame::DataIds() { return data_ids_; }
 
 const std::set<data_t>& Frame::DataIds() const { return data_ids_; }
 
-void Frame::AddDataId(data_t data_id) { data_ids_.insert(data_id); }
+void Frame::AddDataId(data_t data_id) {
```
ahojnnes:
```suggestion
void Frame::AddDataId(const data_t& data_id) {
```

src/colmap/scene/frame.h
```diff
@@ -142,7 +150,20 @@ bool Frame::HasRigId() const { return rig_id_ != kInvalidRigId; }
...
 Rig* Frame::RigPtr() const { return THROW_CHECK_NOTNULL(rig_ptr_); }
 
-void Frame::SetRigPtr(class Rig* rig) { rig_ptr_ = rig; }
+void Frame::SetRigPtr(class Rig* rig) {
+  THROW_CHECK_NOTNULL(rig);
+  THROW_CHECK_NE(rig->RigId(), kInvalidRigId);
+  for (const auto& data_id : data_ids_) {
+    THROW_CHECK(rig->HasSensor(data_id.sensor_id));
+  }
+  if (!HasRigPtr()) {
```
ahojnnes: nit: maybe remove the "!" and swap the two cases.

src/colmap/scene/reconstruction_test.cc
ahojnnes: I assume you will follow up with this PR right after, so we don't forget? Thank you!

src/pycolmap/scene/frame.cc
```diff
@@ -29,6 +29,7 @@ void BindFrame(py::module& m) {
                     &Frame::SetRigId,
                     "Unique identifier of the rig.")
       .def("add_data_id", &Frame::AddDataId, "Associate data with frame.")
+      .def("num_data", &Frame::NumData, "Number of associated data in frame")
```
ahojnnes:
```suggestion
      .def("num_data_ids", &Frame::NumData, "Number of associated data items in frame.")
```

ahojnnes (approved):

src/colmap/scene/reconstruction_test.cc
B1ueber2y: Here is an example for the cases mentioned in https://github.com/colmap/colmap/pull/3546

Issue #3522, opened 2025-08-03
Is there any example of how to create pycolmap.Image and add to pycolmap.Reconstruction?


wzy-99:
Hello, the latest pycolmap is not able to set cam_from_world. The new freature supports Rig and Frame.

But I dont know how to use these, is there any example?

My scene has just a camera.

I have written some code:

```python
         camera = pycolmap.Camera(
                model=camera_type, width=image_size[0], height=image_size[1], params=pycolmap_intri, camera_id=fidx + 1
            )
        reconstruction.add_camera(camera)

        rig = pycolmap.Rig()
        rig.rig_id = fidx + 1
        sensor_t = pycolmap.sensor_t()
        sensor_t.id = camera.camera_id
        rig.add_sensor(sensor_t, pycolmap.Rigid3d())
        reconstruction.add_rig(rig)

        frame = pycolmap.Frame()
        frame.frame_id = fidx + 1
        frame.rig_id = rig.rig_id
        frame.rig = rig
        rotation = pycolmap.Rotation3d(extrinsics[fidx][:3, :3])
        translation = extrinsics[fidx][:3, 3]
        rigid3d = pycolmap.Rigid3d(rotation, translation)
        frame.set_cam_from_world(camera_id=camera.camera_id, cam_from_world=rigid3d)
        reconstruction.add_frame(frame)

        image = pycolmap.Image(
            image_id=fidx + 1, name=f"image_{fidx + 1}", camera_id=camera.camera_id, frame_id=frame.frame_id
        )
        reconstruction.add_image(image)
```

But it doest work

B1ueber2y:
Hey. I think the idea is correct and you just need to change to ``rig.add_ref_sensor(pycolmap.sensor_t(pycolmap.SensorType.CAMERA, camera.camera_id))``.

wzy-99:
@B1ueber2y, thank you for your response!

I solved the error of sensor_t, but encountered a new error:

```bash
    reconstruction.add_frame(frame)
ValueError: [reconstruction.cc:228] Check failed: frame.RigPtr() == &rig (0x55f2cc203b40 vs. 0x55f2cdcea690) 
```

This is my code:

```python
        camera = pycolmap.Camera(
                model=camera_type, width=image_size[0], height=image_size[1], params=pycolmap_intri, camera_id=fidx + 1
            )
        reconstruction.add_camera(camera)

        rig = pycolmap.Rig()
        rig.rig_id = fidx + 1
[...]

PR #3546, opened 2025-08-07
Throw exception at counting registered images when an image in the frame does not exist in the reconstruction. 

B1ueber2y:
Follow up to https://github.com/colmap/colmap/pull/3540 and https://github.com/colmap/colmap/pull/3543

### DIFF
--- a/src/colmap/scene/reconstruction.cc
+++ b/src/colmap/scene/reconstruction.cc
@@ -91,7 +91,7 @@ size_t Reconstruction::NumRegImages() const {
   for (const frame_t frame_id : reg_frame_ids_) {
     const class Frame& frame = Frame(frame_id);
     if (frame.HasPose()) {
-      for ([[maybe_unused]] const data_t& data_id : frame.ImageIds()) {
+      for (const data_t& data_id : frame.ImageIds()) {
         THROW_CHECK(ExistsImage(data_id.id))
             << "The reconstruction object is broken as image " << data_id.id
             << " in frame " << frame.FrameId()


ahojnnes:
I may have misunderstood the issue before, but I don't think we should support images in frames that are not part of the reconstruction. What's the use case for this?

B1ueber2y:
> I may have misunderstood the issue before, but I don't think we should support images in frames that are not part of the reconstruction. What's the use case for this?

There are no explicit use cases. The fix is more for the completeness of the reconstruction data structure, since there is no way to enforce that the images in the frames are all added in our current design. We don't want NumRegImages() and RegImageIds() to be broken in certain state of reconstruction object.

B1ueber2y:
So this is a safety improvement, similar to the previous two PRs.

ahojnnes:
I'd rather have an exception thrown when the images were not added rather than silently skipping over those images, as implemented in this PR... If we want to add some extra safety, we could do another loop in the Reconstruction::Load() method to check that all images in the frames actually exist?

B1ueber2y:
> I'd rather have an exception thrown when the images were not added rather than silently skipping over those images, as implemented in this PR... If we want to add some extra safety, we could do another loop in the Reconstruction::Load() method to check that all images in the frames actually exist?

Since we expose the Reconstruction class to public, ``Load()`` is not the only way to create a Reconstruction object. People can do it in the same way as how we do to create test cases, or construct it via pycolmap (this is how i found all these safety issues since some of the pycolmap construction code (e.g., https://github.com/colmap/colmap/issues/3522) can miss pieces but no error will be given).

B1ueber2y:
Because ``AddImage`` is not forced when we do ``AddFrame``, there is no way to enforce the images to be added, and there is no clear place to check and raise the exception. With the current design, I think we would need to live with such Reconstruction object.

ahojnnes:
Yeah, I understand, but then somebody is misusing the contract. Silently ignoring errors in the setup is not the right solution. Let's make sure, if somebody misconfigures the reconstruction, they will see a meaningful error message. I thought your previous PRs would address this?

B1ueber2y:
> Yeah, I understand, but then somebody is misusing the contract. Silently ignoring errors in the setup is not the right solution. Let's make sure, if somebody misconfigures the reconstruction, they will see a meaningful error message. I thought your previous PRs would address this?

Yes. Most of the errors are addressed by my previous PRs, except for that we cannot check whether all images in the frame are added, because ``AddImage`` happens after ``AddFrame``, which leads to the proposal in this PR.

ahojnnes:
> > Yeah, I understand, but then somebody is misusing the contract. Silently ignoring errors in the setup is not the right solution. Let's make sure, if somebody misconfigures the reconstruction, they will see a meaningful error message. I thought your previous PRs would address this?
> 
> Yes. Most of the errors are addressed by my previous PRs, except for that we cannot check whether all images in the frame are added, because `AddImage` happens after `AddFrame`, which leads to the proposal in this PR.

Thanks, I think that is fine as long as subsequent operations on the reconstruction / frame / image objects lead to obvious errors, so people can fix the issue. The constraints in the SQLite schema should already prevent this when people manually populate objects in the database.

B1ueber2y:
> > > Yeah, I understand, but then somebody is misusing the contract. Silently ignoring errors in the setup is not the right solution. Let's make sure, if somebody misconfigures the reconstruction, they will see a meaningful error message. I thought your previous PRs would address this?
> > 
> > 
> > Yes. Most of the errors are addressed by my previous PRs, except for that we cannot check whether all images in the frame are added, because `AddImage` happens after `AddFrame`, which leads to the proposal in this PR.
> 
> Thanks, I think that is fine as long as subsequent operations on the reconstruction / frame / image objects lead to obvious errors, so people can fix the issue. The constraints in the SQLite schema should already prevent this when people manually populate objects in the database.

Should I then change all the "continue;" in this PR to throwing exceptions?

ahojnnes:
I don't think that adds any value. The `reconstruction.Image(data_id.id);` will already throw with a meaningful error message when the image does not exist?

ahojnnes:
The only place where it could be useful is where we count/accum the registered images?

B1ueber2y:
Sure. Then i will just throw at ``NumRegImages`` and ``RegImageIds()``?

B1ueber2y:
> Sure. Then i will just throw at `NumRegImages` and `RegImageIds()`?

Done. PTAL : )

src/colmap/scene/reconstruction.cc
```diff
@@ -92,6 +92,11 @@ size_t Reconstruction::NumRegImages() const {
     const class Frame& frame = Frame(frame_id);
     if (frame.HasPose()) {
       for ([[maybe_unused]] const data_t& data_id : frame.ImageIds()) {
```
ahojnnes: The maybe_unused can now be removed.

PR #3553, opened 2025-08-11
Fix build with Boost 1.89.0

cho-m:
In the upcoming Boost 1.89.0 release, the Boost.System stub library introduced back in 1.69[^1] has been removed (https://github.com/boostorg/system/commit/7a495bb46d7ccd808e4be2a6589260839b0fd3a3), which causes a CMake error.

This PR change using OPTIONAL_COMPONENTS is based on upstream comment `https://github.com/boostorg/system/issues/132#issuecomment-3146378680` given that no minimum Boost is mentioned for project.

[^1]: https://www.boost.org/doc/libs/1_69_0/libs/system/doc/html/system.html#changes_in_boost_1_69


ahojnnes (approved):
Thank you.

PR #3557, opened 2025-08-11
Fix rig configuration with differing IDs between database and reconstruction

ahojnnes:
Revealed by recent stricter checks on consistency of IDs.

### DIFF
--- a/python/examples/custom_incremental_pipeline.py
+++ b/python/examples/custom_incremental_pipeline.py
@@ -198,7 +198,7 @@ def reconstruct_sub_model(controller, mapper, mapper_options, reconstruction):
     if (
         reconstruction.num_reg_frames() >= 2
         and reconstruction.num_reg_frames() != ba_prev_num_reg_frames
-        and reconstruction.num_points3D != ba_prev_num_points
+        and reconstruction.num_points3D() != ba_prev_num_points
     ):
         iterative_global_refinement(options, mapper_options, mapper)
     return pycolmap.IncrementalMapperStatus.SUCCESS
@@ -341,7 +341,7 @@ def main(
     )
 
     # main runner
-    num_images = pycolmap.Database(database_path).num_images
+    num_images = pycolmap.Database(database_path).num_images()
     with enlighten.Manager() as manager:
         with manager.counter(
             total=num_images, desc="Images registered:"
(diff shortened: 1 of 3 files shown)


src/colmap/scene/rig.cc
```diff
@@ -133,6 +134,123 @@ void UpdateRigAndCameraCalibFromReconstruction(
...
+  for (const auto& [_, image] : reconstruction->Images()) {
+    image_name_to_image.emplace(image.Name(), &image);
+  }
+
+  auto visit_frame_data =
+      [&database, &image_name_to_image, &database_frames](
+          std::function<void(const Frame&, const Image&, const Image&)>
+              visitor) {
+        for (const Frame& database_frame : database_frames) {
+          for (const data_t& data_id : database_frame.DataIds()) {
```
B1ueber2y: Should use ImageIds() here instead?

src/colmap/scene/rig_test.cc
```diff
@@ -192,6 +192,73 @@ TEST(ApplyRigConfig, WithReconstruction) {
...
+
+  std::vector<RigConfig> configs;
+  auto& config = configs.emplace_back();
+  auto& camera1 = config.cameras.emplace_back();
+  camera1.image_prefix = "camera000001_";
+  camera1.ref_sensor = true;
+  auto& camera2 = config.cameras.emplace_back();
+  camera2.image_prefix = "camera000002_";
+
+  ApplyRigConfig(configs, database, &differing_reconstruction);
```
B1ueber2y, line 272: Actually I did not fully understand this test case. From my understanding ``ApplyRigConfig`` does nothing here?

B1ueber2y: Since it was already 1 rig with 2 cameras, and 5 frames before applying the rig config?

ahojnnes: It makes sure it doesn't crash. Without the changes in this PR, this will crash in various places, where you added checks before. However, we could improve the test by generating the reconstruction with trivial frames.

B1ueber2y: I see. Thanks. Yes would be great to change into trivial frames for the initial input reconstruction.

B1ueber2y (approved):
Thank you!

src/colmap/scene/rig.cc
ahojnnes: Yes. Thanks!

PR #3558, opened 2025-08-11
Throw exception for Frame::SetRigId when rig pointer is available.

B1ueber2y:
To align with ``Image::SetCameraId`` and ``Image::SetFrameId``.

### DIFF
--- a/src/colmap/scene/frame.h
+++ b/src/colmap/scene/frame.h
@@ -144,7 +144,11 @@ bool Frame::HasDataId(data_t data_id) const {
 
 rig_t Frame::RigId() const { return rig_id_; }
 
-void Frame::SetRigId(rig_t rig_id) { rig_id_ = rig_id; }
+void Frame::SetRigId(const rig_t rig_id) {
+  THROW_CHECK_NE(rig_id, kInvalidRigId);
+  THROW_CHECK(!HasRigPtr());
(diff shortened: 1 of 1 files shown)


src/colmap/scene/frame.h
```diff
@@ -144,7 +144,11 @@ bool Frame::HasDataId(data_t data_id) const {
 
 rig_t Frame::RigId() const { return rig_id_; }
 
-void Frame::SetRigId(rig_t rig_id) { rig_id_ = rig_id; }
+void Frame::SetRigId(const rig_t rig_id) {
+  THROW_CHECK_NE(rig_id, kInvalidRigId);
+  THROW_CHECK(!HasRigPtr());
+  rig_id_ = rig_id;
```
ahojnnes, line 150: Alternatively, if the rig has a pointer, we could just check that its ID is consistent?

B1ueber2y: If the rig has a pointer, it means that the ID has been set before, since setting the pointer requires the ID to be set beforehand. So there are actually no practical cases where we want to set the ID after the pointer. 

Also, the current fix is consistent with ``Image::SetCamreaId`` and ``Image::SetFrameId``.

ahojnnes (approved):

PR #3564, opened 2025-08-14
Throw exception if any camera in the rig does not exist in the reconstruction.

B1ueber2y:


B1ueber2y:
Actually, should we enforce ``AddCamera`` to be always called before ``AddRig``? Fixing the order of adding objects is the only way that this consistency can be tested. Any opinions?

src/colmap/scene/reconstruction.cc
```diff
@@ -220,6 +220,22 @@ void Reconstruction::TearDown() {
...
 void Reconstruction::AddRig(class Rig rig) {
+  const sensor_t& ref_sensor_id = rig.RefSensorId();
+  if (ref_sensor_id.type == SensorType::CAMERA) {
+    THROW_CHECK(ExistsCamera(ref_sensor_id.id))
+        << "Camera " << ref_sensor_id.id << " from rig " << rig.RigId()
+        << " not found in the reconstruction. Note that AddCamera should be "
+           "called before AddRig.";
+  }
+  for (const auto& [sensor_id, _] : rig.Sensors()) {
+    if (sensor_id.type == SensorType::CAMERA) {
```
ahojnnes: I suggest making this a switch/case, so the compiler complains in the future, when introduce new types but forget to update the check here.

src/colmap/scene/reconstruction.cc
```diff
@@ -220,6 +220,22 @@ void Reconstruction::TearDown() {
 }
 
 void Reconstruction::AddRig(class Rig rig) {
+  const sensor_t& ref_sensor_id = rig.RefSensorId();
+  if (ref_sensor_id.type == SensorType::CAMERA) {
```
ahojnnes: Same here.

src/colmap/scene/reconstruction.cc
```diff
@@ -220,6 +220,22 @@ void Reconstruction::TearDown() {
...
+    THROW_CHECK(ExistsCamera(ref_sensor_id.id))
+        << "Camera " << ref_sensor_id.id << " from rig " << rig.RigId()
+        << " not found in the reconstruction. Note that AddCamera should be "
+           "called before AddRig.";
+  }
+  for (const auto& [sensor_id, _] : rig.Sensors()) {
+    if (sensor_id.type == SensorType::CAMERA) {
+      THROW_CHECK(ExistsCamera(sensor_id.id))
```
ahojnnes: Could create a lambda function and reuse in both instances.

src/colmap/scene/reconstruction.cc

src/colmap/scene/reconstruction.cc
B1ueber2y: Nice idea. Thanks.

PR #3565, opened 2025-08-14
Add option to keep specific cameras constant

ahojnnes:


src/colmap/controllers/option_manager.cc
```diff
@@ -630,6 +630,10 @@ void OptionManager::AddMapperOptions() {
   AddAndRegisterDefaultOption("Mapper.local_ba_min_tri_angle",
                               &mapper->mapper.local_ba_min_tri_angle);
 
+  AddDefaultOption("Mapper.image_list_path", &mapper_image_list_path_);
```
B1ueber2y, line 633: This change in the CLI can break quite some code: https://github.com/search?q=colmap+--image_list_path&type=code

src/pycolmap/sfm/incremental_mapper.cc
```diff
@@ -184,6 +184,11 @@ void BindIncrementalPipeline(py::module& m) {
                      &Opts::fix_existing_frames,
                      "If reconstruction is provided as input, fix the existing "
                      "frame poses.")
+      .def_readwrite("constant_cameras",
```
B1ueber2y, line 187: Need to update ``IncrementalMapperOptions`` binding as well?

src/colmap/controllers/option_manager.cc
ahojnnes: Thanks for double checking. I went through the entire list and actually couldn't find a single one that would be affected. Notice that this only concerns the mapper's image_list_path argument and not the feature_extractor, etc.

src/pycolmap/sfm/incremental_mapper.cc
ahojnnes: Thanks, done.

PR #3571, opened 2025-08-18
Expect error threshold for generalized relative pose in pixel space

ahojnnes:
The keypoint coordinates are given in pixel space, so it is more natural to also expect the error threshold to be in pixel space.

### DIFF
--- a/src/colmap/estimators/generalized_pose_test.cc
+++ b/src/colmap/estimators/generalized_pose_test.cc
@@ -277,7 +277,7 @@ TEST(EstimateGeneralizedRelativePose, Nominal) {
                 sensor_from_rig_translation_stddev);
 
         RANSACOptions ransac_options;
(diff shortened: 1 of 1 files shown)


src/colmap/estimators/generalized_pose_test.cc
```diff
@@ -277,7 +277,7 @@ TEST(EstimateGeneralizedRelativePose, Nominal) {
                 sensor_from_rig_translation_stddev);
 
         RANSACOptions ransac_options;
-        ransac_options.max_error = 1e-2;
+        ransac_options.max_error = 1e-4;
```
B1ueber2y: I think the direction is wrong here? this should become 1.0 rather than 1e-4?

ahojnnes: Good catch, thanks.

PR #3572, opened 2025-08-18
Add function to read number of matches

ahojnnes:
The equivalent of `ReadTwoViewGeometryNumInliers`. Needed for rig verification.


src/colmap/scene/database.cc
```diff
@@ -811,6 +811,22 @@ std::vector<std::pair<image_pair_t, FeatureMatches>> Database::ReadAllMatches()
   return all_matches;
 }
 
+std::vector<std::pair<image_pair_t, int>> Database::ReadNumMatches() const {
+  Sqlite3StmtContext context(sql_stmt_read_num_matches_);
+
+  std::vector<std::pair<image_pair_t, int>> num_inliers;
```
B1ueber2y: The naming here does not seem to be updated? This should be ``num_matches``?

PR #3574, opened 2025-08-18
Make use_log_scale an option for Point3DAlignmentCost.

B1ueber2y:
Log scale optimization has different behaviors from direct scale optimization and is not always better. In particular, direct scale optimization can slightly favor a scale that is close to 1.0. This PR adds back direct scale optimization support but still uses log scale optimization as default.

### DIFF
--- a/src/colmap/controllers/option_manager.h
+++ b/src/colmap/controllers/option_manager.h
@@ -214,19 +214,25 @@ template <typename T>
 void OptionManager::AddDefaultOption(const std::string& name,
                                      T* option,
                                      const std::string& help_text) {
-  desc_->add_options()(
-      name.c_str(),
-      boost::program_options::value<T>(option)->default_value(*option),
-      help_text.c_str());
+  if constexpr (std::is_floating_point<T>::value) {
+    desc_->add_options()(
+        name.c_str(),
+        boost::program_options::value<T>(option)->default_value(
+            *option, StringPrintf("%.3g", *option)),
+        help_text.c_str());
+  } else {
+    desc_->add_options()(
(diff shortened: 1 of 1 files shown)


src/colmap/estimators/cost_functions.h
```diff
@@ -453,16 +453,23 @@ struct RelativePosePriorCostFunctor
 struct Point3DAlignmentCostFunctor
     : public AutoDiffCostFunctor<Point3DAlignmentCostFunctor, 3, 3, 4, 3, 1> {
  public:
-  explicit Point3DAlignmentCostFunctor(const Eigen::Vector3d& point_in_b_prior)
-      : point_in_b_prior_(point_in_b_prior) {}
+  explicit Point3DAlignmentCostFunctor(const Eigen::Vector3d& point_in_b_prior,
```
ahojnnes, line 456: Since the whole cost function is already templated, we could just as well make the cost functor take a template, so the boolean switch becomes a compile-time operation?

B1ueber2y: Thanks. I have updated. This will however make the pybind a bit ugly. I dont think this boolean switch would be a bottleneck, so feel free to let me know if we want to roll back.

ahojnnes: It's indeed a bit complicated and I didn't think ahead with my suggestion. Feel free to revert (slight preference for your original version) or keep as is. Thank you!

B1ueber2y: Reverted. Thanks!

ahojnnes (approved):

PR #3585, opened 2025-08-19
Use Qt QSettings to remember last-used paths across file dialogs

MotivaCG:
Right now, every time Colmap is restarted, all file dialogs start from the default location. This forces the user to browse back to the project or dataset folder again and again, which is annoying.

This PR changes that behavior by using Qt QSettings to store the last-used directory for each dialog (open, save, import, export, etc.), with a global fallback. That way, the next time a dialog is opened it will point to the folder the user was previously working in.

### DIFF
--- a/src/colmap/ui/feature_extraction_widget.cc
+++ b/src/colmap/ui/feature_extraction_widget.cc
@@ -228,7 +228,7 @@ QGroupBox* FeatureExtractionWidget::CreateCameraModelBox() {
   SelectCameraModel(camera_model_cb_->currentIndex());
 
   connect(camera_model_cb_,
-          (void (QComboBox::*)(int))&QComboBox::currentIndexChanged,
+          (void (QComboBox::*)(int)) & QComboBox::currentIndexChanged,
           this,
           &FeatureExtractionWidget::SelectCameraModel);
   connect(camera_params_exif_rb_,
--- a/src/colmap/ui/feature_extraction_widget.cc
+++ b/src/colmap/ui/feature_extraction_widget.cc
@@ -228,7 +228,7 @@ QGroupBox* FeatureExtractionWidget::CreateCameraModelBox() {
   SelectCameraModel(camera_model_cb_->currentIndex());
 
   connect(camera_model_cb_,
-          (void(QComboBox::*)(int)) & QComboBox::currentIndexChanged,
+          (void (QComboBox::*)(int))&QComboBox::currentIndexChanged,
           this,
           &FeatureExtractionWidget::SelectCameraModel);
   connect(camera_params_exif_rb_,
--- a/src/colmap/ui/feature_extraction_widget.cc
+++ b/src/colmap/ui/feature_extraction_widget.cc
@@ -228,7 +228,7 @@ QGroupBox* FeatureExtractionWidget::CreateCameraModelBox() {
   SelectCameraModel(camera_model_cb_->currentIndex());
 
   connect(camera_model_cb_,
-          (void (QComboBox::*)(int)) & QComboBox::currentIndexChanged,
+          (void(QComboBox::*)(int)) & QComboBox::currentIndexChanged,
           this,
           &FeatureExtractionWidget::SelectCameraModel);
   connect(camera_params_exif_rb_,
--- a/src/colmap/ui/main_window.cc
+++ b/src/colmap/ui/main_window.cc
@@ -33,7 +33,55 @@
 #include "colmap/util/logging.h"
 #include "colmap/util/version.h"
 
+#include <QDir>
+#include <QFileInfo>
+#include <QSettings>
+#include <QStandardPaths>
 #include <clocale>
+
+namespace {
+// Get proper QSettings
+static QSettings GetQSettings() { return QSettings("Colmap", "ColmapUI"); }
+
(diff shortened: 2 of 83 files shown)


ahojnnes:
Thanks, this seems like a nice improvement. I don't know about QSettings. Where are these settings persisted?

MotivaCG:
We are using QSettings("Colmap", "UISettings") and Qt uses the native settings location of each platform, so the saved paths are persistent across application restarts:

**In Windows:**
Stored in the registry under
HKEY_CURRENT_USER\Software\Colmap\UISettings

**In macOS:**
Stored in ~/Library/Preferences/com.Colmap.UISettings.plist

**In Linux:**
Stored in a plain ini file under
~/.config/Colmap/UISettings.conf


Regards, Víctor

src/colmap/ui/main_window.cc
```diff
@@ -33,8 +33,56 @@
...
 
+#include <QDir>
+#include <QFileInfo>
+#include <QSettings>
+#include <QStandardPaths>
 #include <clocale>
 
+namespace {
+// Get proper QSettings
+static QSettings GetQSettings() { return QSettings("Colmap", "ColmapUI"); }
```
ahojnnes: Since the functions are defined in anonymous namespace, the "static" is not necessary.

src/colmap/ui/main_window.cc
```diff
@@ -682,6 +732,7 @@ bool MainWindow::ProjectOpen() {
       project_widget_->SetDatabasePath(*options_.database_path);
       project_widget_->SetImagePath(*options_.image_path);
       UpdateWindowTitle();
+      SetLastOpen("last_dir/project", QString::fromStdString(project_path));
```
ahojnnes: nit: it can be good to define a shared `const static char kLastDirProject[] = "last_dir/project";` variable and reuse the same in the multiple instances below.
```suggestion
      SetLastOpen(kLastDirProject, QString::fromStdString(project_path));
```

src/colmap/ui/render_options_widget.cc
```diff
@@ -124,7 +124,7 @@ RenderOptionsWidget::RenderOptionsWidget(QWidget* parent,
   HideOption(&point3D_colormap_max_q_);
   HideOption(&point3D_colormap_scale_);
   connect(point3D_colormap_cb_,
-          (void(QComboBox::*)(int)) & QComboBox::currentIndexChanged,
+          (void (QComboBox::*)(int))&QComboBox::currentIndexChanged,
```
ahojnnes: You will need to run scripts/format/c++.sh to fix the changed formatting.

ahojnnes (commented):
Thank you for your answer. Left a few minor comments, otherwise looks good to me.

src/colmap/ui/render_options_widget.cc
MotivaCG: That's weird that change was suggested by c++.sh ! I've reverted it anyway

src/colmap/ui/main_window.cc
MotivaCG: Done!

src/colmap/ui/main_window.cc
MotivaCG: Done!

src/colmap/ui/render_options_widget.cc
```diff
@@ -124,7 +124,7 @@ RenderOptionsWidget::RenderOptionsWidget(QWidget* parent,
   HideOption(&point3D_colormap_max_q_);
   HideOption(&point3D_colormap_scale_);
   connect(point3D_colormap_cb_,
-          (void(QComboBox::*)(int)) & QComboBox::currentIndexChanged,
+          (void (QComboBox::*)(int)) & QComboBox::currentIndexChanged,
```
ahojnnes:
```suggestion
          (void(QComboBox::*)(int)) & QComboBox::currentIndexChanged,
```

src/colmap/ui/render_options_widget.cc
```diff
@@ -164,7 +164,7 @@ RenderOptionsWidget::RenderOptionsWidget(QWidget* parent,
 
   HideLayout(image_colormap_name_filter_layout_);
   connect(image_colormap_cb_,
-          (void(QComboBox::*)(int)) & QComboBox::currentIndexChanged,
+          (void (QComboBox::*)(int)) & QComboBox::currentIndexChanged,
```
ahojnnes:
```suggestion
          (void(QComboBox::*)(int)) & QComboBox::currentIndexChanged,
```

src/colmap/ui/feature_extraction_widget.cc
```diff
@@ -228,7 +228,7 @@ QGroupBox* FeatureExtractionWidget::CreateCameraModelBox() {
   SelectCameraModel(camera_model_cb_->currentIndex());
 
   connect(camera_model_cb_,
-          (void(QComboBox::*)(int)) & QComboBox::currentIndexChanged,
+          (void (QComboBox::*)(int)) & QComboBox::currentIndexChanged,
```
ahojnnes:
```suggestion
          (void(QComboBox::*)(int)) & QComboBox::currentIndexChanged,
```

ahojnnes (approved):

Issue #3502, opened 2025-07-20
Request for easier reuse of camera poses across static captures


FrankEscobar:
Hi, thank you for your excellent work on COLMAP.

I’d like to suggest a feature or request clarification regarding the reuse of camera poses between different static captures, not animations or video sequences, but multiple independent captures on the same setup.

The goal is to reuse previously estimated camera poses as initial values for a new reconstruction, allowing for some refinement in case of slight changes (e.g., small shifts in tripod position or lens changes). The reused poses would serve as an informed prior, not as fixed final values.

Currently, possible approaches like directly editing the COLMAP database are error-prone and not user-friendly. The camera rig feature, while powerful, seems more geared toward animation or synchronized captures, and assumes images are organized in separate folders per camera. In typical static capture workflows, all images are often in a single folder (but with consistent alphabetical ordering), which makes rig-based approaches cumbersome.

An ideal solution might be a way to export and later re-import camera poses as initial estimates for new datasets.

If this functionality already exists, would it be possible to provide a tutorial or clearer documentation on how to achieve it?

This feature would be highly beneficial for scanning workflows that repeatedly capture with a consistent setup.

Thanks for considering this!

tsattler:
If you (approximately) know the camera poses and the camera intrinsics, you create an "empty" colmap model (one without 3D points). You can then use the `point_triangulator` to obtain the 3D points via triangulation (see here for details: https://colmap.github.io/faq.html#reconstruct-sparse-dense-model-from-known-camera-poses).

You can next use this PR #2145 to refine the camera poses and intrinsics, and the 3D points. 

The whole procedure assumes that for the first step, the known camera poses are reasonably accurate.

FrankEscobar:
Unfortunately the index of the images can change from one recontruction from another due to the multithread ordering

tsattler:
By default, the `point_triangulator` will ignore the image_ids that are in the "empty" reconstruction and will use the file names of the images to get their corresponding image_ids from the database.

FrankEscobar:
What means empty? id -1 maybe?

tsattler:
Empty as in does not contain any 3D points (see my first reply). I typically just use numbers 1 to N for the image_ids.

JulienStanguennec:
As a **workaround**, before the merge request is completed, could we do the following ? 
- setup colmap with one camera per rig (each camera as references in its own rig),
- run after ```point_triangulator``` from known poses
- then do bundle_adjstment with 
```
[...]

Issue #3433, opened 2025-06-30
Refine camera pose from known camera intrinsics and extrinsics with point_triangulator and bundle adjustment


Obliviate-git:
Hello, I am currently using COLMAP for sparse reconstruction and pose optimization. My scenario is as follows: a camera is mounted on a car to continuously capture image information. I have precise camera intrinsics, and GNSS and IMU provide highly accurate data that is time-stamped and aligned with the images. Following the webpage [,](https://colmap.github.io/faq.html#reconstruct-sparse-dense-model-from-known-camera-poses) I prepared cameras.txt, images.txt, and point3D.txt (empty), and ran the point_triangulator to generate a large number of 3D points (600 images, approximately 35,000 points).

However, another goal of mine is to optimize camera poses using point_triangulator and bundle adjustment in scenarios where GNSS is less accurate. I have read many related issues, including your provided refine.cc code in [1057](https://github.com/colmap/colmap/issues/1057) for alternating between triangulator, local BA, and global BA. In that issue, other responders shared a refine.cc adapted for COLMAP 3.9, but it is incompatible with the latest version (3.10.2). You also mentioned in this issue that "bundle adjuster resulting in a change in scale" and that "The sole purpose of the point_triangulator is to generate 3D points from known poses. It is not aimed at refining poses." I noticed that compared to version 3.9, the 3.10.2 version has removed the ba_refine_cam_extrinsics parameter.

I have the following questions:

1.For the current version (3.10.2), are there any mature techniques/parameters to optimize camera extrinsics within `point_triangulator`?
2.After reading the source code of incremental_pipeline.cc, bundle_adjustment.cc, etc., the term "rig" is mentioned multiple times. When calling point_triangulator via the command line with default parameters, will a rig be defined?

Obliviate-git:
@tsattler

tsattler:
You can try this pull request #2145

Obliviate-git:
> You can try this pull request [#2145](https://github.com/colmap/colmap/pull/2145)

[...]

Issue #2819, opened 2024-10-01
Is it possible to add a module to accommodate and adjust for the noise in fixed camera poses？


han-xiangyu:
Hi,

When I use the fix pose sparse reconstruction in autonomous driving datasets such as nuscenes, I always cannot get the good sparse reconstruction bacause of the inaccurate given poses. So I would like to ask if it is possible to have some error tolerance for the fix pose sparse reconstruction to allow the camera to adjust their poses a little bit in order to obtain a better sparse points cloud?

Thanks

ahojnnes:
By creating a sparse reconstruction, I assume you mean (1) populating an existing reconstruction with the poses, (2) running colmap feature extraction + matching, (3) running the point triangulator?

tsattler:
This will not solve the issue. Rather, after the point triangulator, you want to run a few iterations of: (4) refine the poses, and (5) complete and merge tracks. I opened a PR for this a while ago, but with outdated code and I never got around to fixing it.

han-xiangyu:
> By creating a sparse reconstruction, I assume you mean (1) populating an existing reconstruction with the poses, (2) running colmap feature extraction + matching, (3) running the point triangulator?

Yes, that's what I mean. I'm referring to the steps outlined here: https://colmap.github.io/faq.html#reconstruct-sparse-dense-model-from-known-camera-poses

han-xiangyu:
> This will not solve the issue. Rather, after the point triangulator, you want to run a few iterations of: (4) refine the poses, and (5) complete and merge tracks. I opened a PR for this a while ago, but with outdated code and I never got around to fixing it.

Do you mean PR #2145? It seems to be exactly what I’m looking for, and it appears to be in demand by some others as well.

ahojnnes:
This question has been addressed in #2821. Use `pose_prior_mapper` with per-frame covariance (low for known poses, high for uncertain ones), or use `Mapper.fix_existing_images=1` with an initial reconstruction containing fixed poses and triangulated points. Closing as answered.

PR #3589, opened 2025-08-20
Point_triangulator: optional intrinsics/extrinsics refinement + refine loop

MotivaCG:
**WHAT IS DONE?**

The intention is to make point_triangulator more useful when starting from approximate poses. It adds three optional flags:

--refine_intrinsics (default 0): refine focal length and extra params (principal point stays fixed).
--refine_extrinsics (default 0): allow pose refinement; otherwise poses are kept fixed.
--refine_loop (default 0): after initial triangulation, run a short loop of global BA → CompleteAndMergeTracks → FilterPoints with early stopping.

**WHY IS THIS USEFUL?**

Running a standalone bundle adjustment after triangulation does not add/merge tracks using the improved geometry. The refine loop interleaves bundle adjustment with track completion and filtering, which typically yields a cleaner and more complete sparse model without chaining multiple commands.

This could fit the required features in issues like
https://github.com/colmap/colmap/issues/3502
https://github.com/colmap/colmap/issues/2819
https://github.com/colmap/colmap/issues/3433

Meanwhile maintenance is easier than doing a new mode for Colmap CLI

**HOW IT WORKS?**

Initial triangulation respects refine_extrinsics (fix_existing_frames = !refine_extrinsics).
In the loop, poses are optimized only if --refine_extrinsics=1; otherwise a  bundle adjustment is run with all poses kept constant.
Defaults preserve current behavior (no loop, poses fixed), so it doesn't break compatibility at all!

### DIFF
--- a/src/colmap/exe/sfm.cc
+++ b/src/colmap/exe/sfm.cc
@@ -519,6 +519,7 @@ int RunPointTriangulator(int argc, char** argv) {
   bool clear_points = true;
   bool refine_intrinsics = false;
   bool refine_extrinsics = false;
+  bool refine_loop = false;
 
   OptionManager options;
   options.AddDatabaseOptions();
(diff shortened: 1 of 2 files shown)


DuVogel87:
Why wasnt this further checked and implemented?

MotivaCG:
Should I close this PR in case it won't be merged?

PR #3600, opened 2025-08-28
Share feature extraction max_image_size option

ahojnnes:
Towards the introduction of new feature extractors.

### DIFF
--- a/.github/workflows/build-mac.yml
+++ b/.github/workflows/build-mac.yml
@@ -48,7 +48,7 @@ jobs:
       - name: Setup Mac
         run: |
           # Fix `brew link` error.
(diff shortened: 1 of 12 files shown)


cr333:
FYI: This change just landed in pycolmap 3.13 and broke our existing tests. Maybe mention this under “breaking changes” in the Changelog?

It would've been nice to keep things backwards compatible, at least for a while, ideally with a deprecation warning about upcoming breaking changes in a future version. But I understand that things sometimes (need to) break to make progress.

ahojnnes:
Thanks @cr333 for the feedback. We usually trying to maintain backwards compatibility and here it would have not been possible due to the change to an abstract interface. That being said, we missed to call it out in the change log. Let me know if you need help upgrading. I am less than 1m away :-)

PR #3604, opened 2025-08-29
Add option to fix individual rigs

ahojnnes:


src/colmap/controllers/incremental_pipeline_test.cc
```diff
@@ -570,16 +623,18 @@ TEST(IncrementalPipeline, GPSPriorBasedSfMWithNoise) {
...
+
   const std::string database_path = CreateTestDir() + "/database.db";
 
   auto database = Database::Open(database_path);
   Reconstruction gt_reconstruction;
   SyntheticDatasetOptions synthetic_dataset_options;
   synthetic_dataset_options.num_rigs = 2;
   synthetic_dataset_options.num_cameras_per_rig = 1;
-  synthetic_dataset_options.num_frames_per_rig = 7;
+  synthetic_dataset_options.num_frames_per_rig = 5;
```
ahojnnes, line 635: Making this test a bit faster to run.

src/colmap/controllers/incremental_pipeline_test.cc
```diff
@@ -600,48 +655,53 @@ TEST(IncrementalPipeline, SfMWithRandomSeedStability) {
     return reconstruction_manager;
   };
 
-  // Single-thread execution
+  // Single-threaded execution.
```
ahojnnes, line 658: Needed to update the below logic to make the test results more reliable, because the newly added test was randomly breaking the below due to different random numbers in the generated synthetic scenes.

src/colmap/controllers/incremental_pipeline_test.cc
```diff
@@ -600,48 +655,53 @@ TEST(IncrementalPipeline, SfMWithRandomSeedStability) {
...
-    ExpectReconstructionsIdentical(*reconstruction_manager0->Get(0),
-                                   *reconstruction_manager1->Get(0));
-
-    // Different seed should produce different reconstructions
-    auto reconstruction_manager2 =
-        run_mapper(/*num_threads=*/1, /*random_seed=*/123);
-    ExpectReconstructionsDifferent(*reconstruction_manager0->Get(0),
-                                   *reconstruction_manager2->Get(0));
+        run_mapper(/*num_threads=*/1, /*random_seed=*/kRandomSeed);
+    EXPECT_TRUE(AreReconstructionsIdentical(*reconstruction_manager0->Get(0),
```
B1ueber2y, line 665: nit: ExpectReconstructionsIdentical?

B1ueber2y (approved):

src/colmap/controllers/incremental_pipeline_test.cc
ahojnnes: If you don't do the EXPECT_* call in the test body, you cannot easily trace the source of the expectation (when there are multiple calls to the function within the same test). This is why I moved the EXPECT_TRUE to the test function.

PR #3608, opened 2025-09-03
Cuda-Enabled Pip Package (Linux x86-64 only)

Tobias314:
Adds an additional job run to the `build-pycolmap.yml` github action to build a Linux x86-64 Python wheel with CUDA support named **pycolmap_cuda**

This is achieved by the following changes:
- use CUDA-enabled `sameli/manylinux_2_34_x86_64_cuda_12.8` container image for cibuildwheel
- modify repair wheel command to remove shared CUDA libraries from wheel
- dynamically modify the `pyproject.toml` file during the CUDA build to name the resulting package pycolmap_cuda to avoid name collision with the non-CUDA pycolmap package

## TODO:
- [ ] decide on name of CUDA pycolmap package for publishing on pypi


sarlinpe:
Thank you for your contribution. IIUC you suggest having a different package name for CUDA wheels. How would you then handle different versions of CUDA? PyTorch only uploads to PyPI a single CUDA version (12.8 currently) and hosts externally other versions and their dependencies (example: https://download.pytorch.org/whl/cu129). NVIDIA does have some CUDA shared libraries exposed as PyPI packages, are they useful for us?

Tobias314:
> How would you then handle different versions of CUDA?

That is not really tackled so far.
Currently, everything is built with CUDA version 12.8 (using the `sameli/manylinux_2_34_x86_64_cuda_12.8` container image).
Without further adjustments, cibuildwheel would use repairwheel to include required shared libraries into the wheel. Upon inspecting such a wheel I found this to include corresponding versions of `libcudart` and `libcurand`. 
Since including them in the wheel comes with several problems (e.g. wheel size exceeding PyPI limits), I modified the repairwheel command to not include those CUDA libraries. So currently, getting the right versions of those dependencies as well as a working CUDA runtime would probably be left to the user. That is not ideal.

One way to improve this might be to dynamically modify the `pyproject.toml` file while building the CUDA wheel to add CUDA dependencies via PyPI packages. There seem to exist [cuda-runtime](https://pypi.org/project/nvidia-cuda-runtime-cu12/) and [curand](https://pypi.org/project/nvidia-curand-cu12) packages on PyPI which include those shared libraries. The PyTorch wheel also seems to have them as dependencies.

Offering multiple CUDA versions would probably go even further than this. Maybe one option would be to build different wheels using container images with different CUDA versions and different versions of  those Nvidia PyPI packages specified as dependencies in the `pyproject.toml`? To differentiate the wheels on PyPI one could include the CUDA version in the package name (e.g. `pycolmap_cuda_12.8`).

PyTorch also seems to build multiple wheels for multiple CUDA versions. However, they do not differentiate them by name. Instead, it seems like they are hosting packages for different CUDA versions in different package indices. However, this probably requires self-hosting those alternative package indices.

I think an underlying problem is that wheel tags include only operating system and CPU architecture information. So dealing with things like different CUDA versions might always require getting a bit creative. 

However, I do not have much experience with building CUDA-enabled Python packages. So I would be more than happy to learn about some alternative ideas.

sarlinpe:
> I think an underlying problem is that wheel tags include only operating system and CPU architecture information. So dealing with things like different CUDA versions might always require getting a bit creative.

Yes, this is a long-standing issue. More info on it at https://pypackaging-native.github.io/key-issues/gpus/ Apparently, as of CUDA 11, CUDA promises binary compatibility across minor versions, so we'd only need one PyPI package per major version. I suggest to start with `pycolmap-cuda12`, and we can add other ones for 11 and 13 later.

Dynamically patching `pyproject.toml` to add the CUDA dependencies makes the most sense. It looks like https://github.com/Di-Is/faiss-gpu-wheels defines one `pyproject.toml`, which isn't nice to maintain.

Tobias314:
> I suggest to start with `pycolmap-cuda12`, and we can add other ones for 11 and 13 later.

Sounds good. Since all the changes are already targeting CUDA 12, I just changed the package name to `pycolmap-cuda12`.

> Dynamically patching `pyproject.toml` to add the CUDA dependencies makes the most sense.

The github action is now modified to dynamically add the [cuda-runtime](https://pypi.org/project/nvidia-cuda-runtime-cu12/) and [curand](https://pypi.org/project/nvidia-curand-cu12) packages as dependencies when building the CUDA-enabled package.

sarlinpe:
Nice, thank you! Would it make sense to update pyproject.toml file using Python instead of sed, to be more robust to changes of formatting? For example adding a script to `python/ci/update_pyproject_toml.py`:
```python
import argparse
from pathlib import Path
import tomlkit

# Set up command-line argument parser
parser = argparse.ArgumentParser(description="Modify pyproject.toml for a custom build.")
parser.add_argument("--name", required=True, help="The new package name for the wheel.")
parser.add_argument(
    "--add-deps",
    nargs="+",
    default=[],
    help="Space-separated list of Python dependencies to add.",
)
args = parser.parse_args()

# Modify the pyproject.toml file
pyproject_path = Path("pyproject.toml")
if not pyproject_path.exists():
    raise FileNotFoundError(pyproject_path)

with open(pyproject_path, "r", encoding="utf-8") as f:
    config = tomlkit.load(f)

config["project"]["name"] = args.name

if args.add_deps:
    if "dependencies" not in config["project"]:
        config["project"]["dependencies"] = []

    existing_deps = config["project"]["dependencies"]
    for req in args.add_deps:
        if req not in existing_deps:
            existing_deps.append(req)

with open(pyproject_path, "w", encoding="utf-8") as f:
    tomlkit.dump(config, f)
```

.github/workflows/build-pycolmap.yml
```diff
@@ -16,14 +17,15 @@ on:
 
 jobs:
   build:
-    name: ${{ matrix.config.os }} ${{ matrix.config.arch }}
+    name: ${{ matrix.config.os }} ${{ matrix.config.arch }} cuda_${{ matrix.config.build_with_cuda }}
```
sarlinpe: I recommend using a different syntax:
https://github.com/colmap/colmap/blob/fd25e1a42c2377446330bd7dfd9ee9d5e2f3278e/.github/workflows/build-ubuntu.yml#L19

Tobias314:
> Would it make sense to update pyproject.toml file using Python instead of sed

Sure, I have basically added the Python script that you suggested above.

python/ci/install-colmap-almalinux.sh
```diff
@@ -20,7 +20,8 @@ yum install -y \
     curl \
     zip \
     unzip \
-    tar
+    tar \
+    perl
```
sarlinpe, line 24: I guess that we don't need perl anymore?

Tobias314:
> The CUDA build increases the size of the pycolmap .so by 2x (54MB to 110MB), I guess that this is expected?

I guess an increase in size is expected. However, for me it looks like the .whl is only 62MB in size. I think this can be seen in  the summary at https://github.com/colmap/colmap/actions/runs/18450241678/attempts/2#summary-52636181721.
Where did you get the 110MB from?

Also, I think  PyPI has an upper limit of 100MB by default ([https://docs.pypi.org/project-management/storage-limits/](https://docs.pypi.org/project-management/storage-limits/)). So publishing something larger than this limit might require special permission from PyPI.

sarlinpe:
I was referring to the .so file (obtain by unpacking the wheel, which is just a zip file), the wheel size is indeed fine for PyPI (3x larger though).

python/ci/install-colmap-almalinux.sh
Tobias314: I have tried without perl but the action failed:
https://github.com/colmap/colmap/actions/runs/18574726052/job/52957266278?pr=3608
Therefore, I have added it again.

sarlinpe:
@ahojnnes @B1ueber2y any opinion against merging this?

.github/workflows/build-pycolmap.yml
```diff
@@ -151,28 +153,42 @@ jobs:
...
+
       - name: Build wheels
         uses: pypa/cibuildwheel@v3.1.2
         with:
           package-dir: ./
         env:
           CIBW_ARCHS_MACOS: ${{ matrix.config.arch }}
+          CIBW_ENVIRONMENT:  >
+            BUILD_CUDA_ENABLED=${{ matrix.config.cudaEnabled }}
+            VCPKG_COMMIT_ID="0cb95c860ea83aafc1b24350510b30dec535989a"
```
B1ueber2y: Maybe I overlooked something, but I thought vcpkg tag was centralized at https://github.com/colmap/colmap/blob/main/vcpkg-configuration.json?

.github/workflows/build-pycolmap.yml
```diff
@@ -151,28 +153,42 @@ jobs:
...
 
       - name: Archive wheels
         uses: actions/upload-artifact@v4
         with:
-          name: pycolmap-${{ matrix.config.os }}-${{ matrix.config.arch }}
-          path: wheelhouse/pycolmap-*.whl
+          name: pycolmap-${{ matrix.config.os }}-${{ matrix.config.arch }}${{ matrix.config.cudaEnabled && '-CUDA' || '' }}
+          path: wheelhouse/pycolmap*.whl
```
B1ueber2y, line 183: Is there a reason for this change?

sarlinpe: We need the artifacts of CUDA vs regular builds to be named differently otherwise they would overwrite each other.

B1ueber2y: Yes. I meant from wheelhouse/pycolmap-*.whl -> wheelhouse/pycolmap\*.whl.

sarlinpe: This is because the wheel is now named `pycolmap_cuda12-3.13.0.dev0-cp39-cp39-manylinux_2_34_x86_64.whl `. I don't know why since we rename the package to `pycolmap-cuda12`...

.github/workflows/build-pycolmap.yml
ahojnnes: Yes, we should remove this here.

.github/workflows/build-pycolmap.yml
```diff
@@ -151,28 +153,41 @@ jobs:
           CIBW_ENVIRONMENT_PASS_LINUX="VCPKG_TARGET_TRIPLET VCPKG_INSTALLATION_ROOT CMAKE_TOOLCHAIN_FILE VCPKG_BINARY_SOURCES CONTAINER_COMPILER_CACHE_DIR CCACHE_DIR CCACHE_BASEDIR"
           echo "CIBW_ENVIRONMENT_PASS_LINUX=${CIBW_ENVIRONMENT_PASS_LINUX}" >> "$GITHUB_ENV"
 
-          CIBW_MANYLINUX_X86_64_IMAGE="quay.io/pypa/manylinux_2_28_x86_64"
+          if [ "${{ matrix.config.cudaEnabled }}" = "true" ]; then
+            CIBW_MANYLINUX_X86_64_IMAGE="sameli/manylinux_2_34_x86_64_cuda_12.8"
```
ahojnnes:
```suggestion
            CIBW_MANYLINUX_X86_64_IMAGE="sameli/manylinux_2_34_x86_64_cuda_12.9"
```

sarlinpe:
@Tobias314 Thank you for your work, we now have pycolmap-cuda12 on PyPI: https://pypi.org/project/pycolmap-cuda12/ Would you mind removing https://pypi.org/project/pycolmap-cuda/ (I believe that it's owned by you) to prevent any confusion? thank you!

Tobias314:
> we now have pycolmap-cuda12 on PyPI: https://pypi.org/project/pycolmap-cuda12/ Would you mind removing https://pypi.org/project/pycolmap-cuda/ (I believe that it's owned by you) to prevent any confusion? thank you!

Great to hear! 
I have archived pycolmap-cuda so that it can no longer be found via PyPI.

PR #3611, opened 2025-09-04
Also replace mask file extension to .png instead of appending

Dawars:
Sometimes the input images are a mix of .png and other extensions.
To handle the mask filenames, which are always .png, it is cleaner to replace the image extension instead of appending.
This would avoid filenames such as **filename.jpg.png** and **filename.png.png**

Change:
If checking fails for appending .png to the mask path, I also check by replacing the file extension by .png.


Dawars:
Is there any reason against merging this feature?

src/colmap/controllers/image_reader.cc
```diff
@@ -156,6 +156,19 @@ ImageReader::Status ImageReader::Next(Rig* rig,
           exists_mask = true;
         }
       }
+      if (!exists_mask) {
```
ahojnnes: In theory I would have assumed the block above should have already addressed the same scenario?

ahojnnes:
Thanks and apologies for not looking at this earlier. I am surprised this doesn't already work and left a comment for you. If it's not already covered by a test, we should add a unit test to image_reader_test.cc to cover this new behavior. Thank you.

src/colmap/controllers/image_reader.cc
Dawars: The above block handles cases when the image is .png AND the mask is .png as well.
I removed this because my changes can handle those cases as well.

Dawars:
This wasn't discovered because the test is only testing for `.png` images.
- I added an option to set the image extension
- Now for the first image the mask path is appended and for the rest the extension is replaced.

When running the test on my machine I get the following error, not sure if it's relevant
`E0924 15:20:55.321195 1165799 image_reader.cc:103] Check failed: image_index_ <= options_.image_names.size() (6 vs. 5) 
`

Let me know if you need any changes, I don't have a lot of experience writing tests

src/colmap/controllers/image_reader.cc
```diff
@@ -148,9 +148,12 @@ ImageReader::Status ImageReader::Next(Rig* rig,
         JoinPaths(options_.mask_path, image->Name() + ".png");
     if (!ExistsFile(mask_path)) {
       bool exists_mask = false;
-      if (HasFileExtension(image->Name(), ".png")) {
-        std::string alt_mask_path =
-            JoinPaths(options_.mask_path, image->Name());
+      // Try replacing extension with .png
+      std::string base_name = image->Name();
+      size_t last_dot = base_name.find_last_of('.');
```
ahojnnes: nit

```suggestion
      const size_t last_dot = base_name.find_last_of('.');
```

ahojnnes (approved):
Thank you, LGTM.

Dawars:
The test fails due to jpg compression being lossy as far as I can tell, how should we handle it?
```
[ RUN      ] ImageReaderTests/ParameterizedImageReaderTests.Nominal/3
I20250924 16:50:43.943231 10168 testing.cc:51] Creating test directory: C:\Users\RUNNER~1\AppData\Local\Temp\colmap_test_data\ImageReaderTests/ParameterizedImageReaderTests.Nominal/3
..\src\colmap\controllers\image_reader_test.cc(131): error: Expected equality of these values:
  bitmap.ConvertToRowMajorArray()
    Which is: { '\0', '\0', '\0', '\0', '\0', '\0', '\0', '\0', '\0' }
  test_bitmap.ConvertToRowMajorArray()
    Which is: { '\x1' (1), '\x1' (1), '\x1' (1), '\0', '\0', '\0', '\0', '\0', '\0' }

..\src\colmap\controllers\image_reader_test.cc(131): error: Expected equality of these values:
  bitmap.ConvertToRowMajorArray()
    Which is: { '\0', '\0', '\0', '\0', '\0', '\0', '\0', '\0', '\0' }
  test_bitmap.ConvertToRowMajorArray()
    Which is: { '\x1' (1), '\x1' (1), '\x1' (1), '\0', '\0', '\0', '\0', '\0', '\0' }

..\src\colmap\controllers\image_reader_test.cc(131): error: Expected equality of these values:
  bitmap.ConvertToRowMajorArray()
    Which is: { '\0', '\0', '\0', '\0', '\0', '\0', '\0', '\0', '\0' }
  test_bitmap.ConvertToRowMajorArray()
    Which is: { '\x1' (1), '\x1' (1), '\x1' (1), '\0', '\0', '\0', '\0', '\0', '\0' }

..\src\colmap\controllers\image_reader_test.cc(131): error: Expected equality of these values:
  bitmap.ConvertToRowMajorArray()
    Which is: { '\0', '\0', '\0', '\0', '\0', '\0', '\0', '\0', '\0' }
  test_bitmap.ConvertToRowMajorArray()
    Which is: { '\x1' (1), '\x1' (1), '\x1' (1), '\0', '\0', '\0', '\0', '\0', '\0' }

..\src\colmap\controllers\image_reader_test.cc(131): error: Expected equality of these values:
  bitmap.ConvertToRowMajorArray()
    Which is: { '\0', '\0', '\0', '\0', '\0', '\0', '\0', '\0', '\0' }
  test_bitmap.ConvertToRowMajorArray()
    Which is: { '\x1' (1), '\x1' (1), '\x1' (1), '\0', '\0', '\0', '\0', '\0', '\0' }

E20250924 16:50:43.952133 10168 image_reader.cc:103] Check failed: image_index_ <= options_.image_names.size() (6 vs. 5) 
[  FAILED  ] ImageReaderTests/ParameterizedImageReaderTests.Nominal/3, where GetParam() = (5, true, false, true, ".jpg") (10 ms)
```

ahojnnes:
You could use another lossess image format like bmp instead of jpg or cross fingers that explicitly setting a jpeg quality of 100 through the "flags" parameter in Bitmap::Write() makes the roundtrip lossess.

ahojnnes:
Thank you!

PR #3612, opened 2025-09-04
Make an exception for images in "others" folder for per_folder shared camera

Dawars:
When working with mixed image sources (video, internet photography) it makes sense to create shared cameras for the videos and individual cameras for the pictures.

This addition would make an exception for images in "others" folder when per_folder shared cameras is turned on.
This makes life a lot easier when using pycolmap or hloc in Python.

Is this a good idea? Should I make a similar exception for "shared camera for all images"?


ahojnnes:
Thank you. This can already be achieved by running the extractor twice (+ image_list) or by re-configuring the cameras with pycolmap. I am admittedly not a fan of the proposed magical "others" folder, given that there are already existing solutions to the problem. Were you not aware of it or is something else not working for you with this?

Dawars:
I didn't want to change the hloc part of my pipeline which I'm calling from the terminal but you are right that changing it in python is a lot easier than recompiling colmap.

PR #3505, opened 2025-07-24
Add support for pose prior rotation in database and pose_piror_mapper

StonerLing:
- Extend pose prior format to include rotation as quaternion in database,
- support optional rotation covariance (axis-angle std dev)
- Update importer, database schema, and pose prior mapper to handle rotation priors

PR #3613, opened 2025-09-05
6-DoF Pose Priors BA

tavislocus:
Motivated by requiring rotation in pose priors for reconstructing repetitive feature sparse environments. 

- Added rotation and rotation_covariance to PosePrior
    - Locally I'm using combined 6x6 covariances but thought this is a good middle ground to later progress to a Matrix6D
    - If PosePriors have rotation -> covariances are combined into a 6d, just before BA
    - Decided on this for now as it leads to minimal changes required throughout existing codebase
- Rotation is a [w,x,y,z] quaternion
- Added into PosePrior, DB, BA, Python bindings 
- PosePrior tests updated to include rotation & rotation_covariance


ahojnnes:
Thank you. This is related to the work here: https://github.com/colmap/colmap/pull/3505

ahojnnes:
@sarlinpe @B1ueber2y FYI

nvb-ai:
@tavislocus If both position and rotation priors are used, then the position prior need to be in world coordinate and the rotation need to be (quaternion of) world-to-cam rotation matrix. Is my understanding correct?

tavislocus:
@nhatvu192 
So the rotation is also the world -> rig (I need to check what it's relative too), and then from this each sensor (camera) has it's own rig -> sensor trans + rotation transform.

tavisshore:
Hi - any updates on adding 6-DoF to the main branch?

PR #3617, opened 2025-09-08
Fix focal length extraction from 35mm equivalent

ahojnnes:
Testing currently not possible with FreeImage, to be done when migrating to OpenImageIO as part of: https://github.com/colmap/colmap/pull/3459

### DIFF
--- a/src/colmap/sensor/bitmap.cc
+++ b/src/colmap/sensor/bitmap.cc
@@ -416,7 +416,15 @@ bool Bitmap::ExifFocalLength(double* focal_length) const {
     if (std::regex_search(focal_length_35mm_str.c_str(), result, regex)) {
       const double focal_length_35 = std::stold(result[1]);
       if (focal_length_35 > 0) {
-        *focal_length = focal_length_35 / 35.0 * max_size;
+        // Based on https://en.wikipedia.org/wiki/35_mm_equivalent_focal_length
+        // According to CIPA guidelines, 35 mm equivalent focal length is to be
+        // calculated like this:
+        // "focal length in 35 mm camera" =
+        //   (Diagonal distance of image area in the 35 mm camera (43.27 mm) /
+        //    Diagonal distance of image area on the image sensor of the DSC)
+        //    * focal length of the lens of the DSC.
+        const double diagonal = std::sqrt(width_ * width_ + height_ * height_);
+        *focal_length = focal_length_35 / 43.27 * diagonal;
         return true;
       }
     }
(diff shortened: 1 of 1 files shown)


src/colmap/sensor/bitmap.cc
```diff
@@ -484,6 +482,19 @@ bool Bitmap::ExifFocalLength(double* focal_length) const {
           }
         }
       }
+
+      // Fall back to look up sensor width in database.
```
ahojnnes: Swapped the order where we prioritize EXIF over the sensor database.

src/colmap/sensor/bitmap.cc
```diff
@@ -473,7 +468,10 @@ bool Bitmap::ExifFocalLength(double* focal_length) const {
             // the image might have been resized, but the EXIF data preserved.
             const double ccd_width = x_res * pixel_x_dim;
             if (ccd_width > 0 && focal_length_mm > 0) {
-              if (res_unit_str == "cm") {
+              if (res_unit_str == "mm") {
```
paulinus: nit: could define `ccd_width_mm` and factor the focal length equation out of the if statements

src/colmap/sensor/bitmap.cc
```diff
@@ -473,7 +468,10 @@ bool Bitmap::ExifFocalLength(double* focal_length) const {
...
+                *focal_length = focal_length_mm / (ccd_width * 1.0) * max_size;
```
paulinus: aside: Why is the focal length divided by the width in mm but multiplied by the max_size, which may be either width or height? Should it not match which axis divides and which axis multiply?

B1ueber2y: My understanding is that this was due to the presence of exif orientation. And because the exif orientation is not always reliable (due to inconsistent convention in different software) it is not safe to choose from width_ and height_ accordingly. 

Here COLMAP makes an assumption that the native orientation for most existing common sensors are landscape, where PixelXDimension corresponds to the long side.

src/colmap/sensor/bitmap.cc
ahojnnes: Done

src/colmap/sensor/bitmap.cc
ahojnnes: I think we can completely factor out the image size from the equation. As a matter of fact, I cannot find an image with the FocalPlane EXIF fields set, so cannot verify, but the latest commit should be correct based on reading the specs of the EXIF fields.

PR #3621, opened 2025-09-10
 Add bindings for mvs::model for covisibility support in pycolmap

B1ueber2y:
A step to make covisibility support available in pycolmap, for easier pair generation. For example, this could speed up the python for loop from hloc here: https://github.com/cvg/Hierarchical-Localization/blob/master/hloc/pairs_from_covisibility.py#L16-L47 and also simplify the code from the python side. 

Ideally it might be better to move the whole covisibility logic to the reconstruction utility, but this will require adding some code to convert PMVS to colmap format, so is only left to future work for now.

### DIFF
--- a/src/pycolmap/mvs/bindings.cc
+++ b/src/pycolmap/mvs/bindings.cc
@@ -2,6 +2,6 @@
 
 namespace py = pybind11;
 
-void BindMvsModel(py::module& m);
+void BindMVSModel(py::module& m);
 
-void BindMvs(py::module& m) { BindMvsModel(m); }
+void BindMvs(py::module& m) { BindMVSModel(m); }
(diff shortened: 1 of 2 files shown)


ahojnnes:
nit: "Mvs" -> "MVS".

ahojnnes (approved):

B1ueber2y:
> nit: "Mvs" -> "MVS".

Sure. I renamed the class name to ``MVSModel``. However, the internal function ``BindMvs`` cannot be renamed due to a name collapsing with https://github.com/colmap/colmap/blob/main/src/pycolmap/pipeline/mvs.cc#L87. This seems to be already the case for SfM and Sfm so I just follow it: https://github.com/colmap/colmap/blob/main/src/pycolmap/pipeline/mvs.cc#L87 and https://github.com/colmap/colmap/blob/main/src/pycolmap/sfm/bindings.cc#L9

PR #3627, opened 2025-09-15
Improve the Gauge logic for fixing two views.

B1ueber2y:
An improvement for the case when the reference sensor is not available in the BA problem.

### DIFF
--- a/src/colmap/estimators/bundle_adjustment_test.cc
+++ b/src/colmap/estimators/bundle_adjustment_test.cc
@@ -1012,7 +1012,7 @@ TEST(DefaultBundleAdjuster, FixGaugeWithTwoCamsFromWorldNoReferenceSensor) {
   SynthesizeDataset(synthetic_dataset_options, &reconstruction);
   const Reconstruction orig_reconstruction = reconstruction;
 
-  // delete observations from the two reference images
+  // Delete observations from the two reference images.
   THROW_CHECK(reconstruction.Image(1).HasTrivialFrame());
   THROW_CHECK(reconstruction.Image(3).HasTrivialFrame());
   for (point2D_t i = 0; i < reconstruction.Image(1).NumPoints2D(); ++i) {
@@ -1026,7 +1026,7 @@ TEST(DefaultBundleAdjuster, FixGaugeWithTwoCamsFromWorldNoReferenceSensor) {
     }
   }
 
-  // only add two non-reference image
+  // Only add two non-reference images.
   BundleAdjustmentOptions options;
   BundleAdjustmentConfig config;
   config.AddImage(2);


src/colmap/estimators/bundle_adjustment.cc
```diff
@@ -494,9 +494,9 @@ void FixGaugeWithTwoCamsFromWorld(
   // First, search through the already fixed cameras in the problem.
   for (const image_t image_id : image_ids) {
     Image& image = reconstruction.Image(image_id);
-    if (image.FramePtr()->RigPtr()->IsRefSensor(
```
B1ueber2y, line 497: No need to be a reference sensor here because the check is done on ``image.FrameId()``.

src/colmap/estimators/bundle_adjustment.cc
```diff
@@ -506,21 +506,26 @@ void FixGaugeWithTwoCamsFromWorld(
...
   }
 
-  auto IsParameterizedRefSensor = [&problem](const Image& image) {
-    return image.FramePtr()->RigPtr()->IsRefSensor(
-               image.CameraPtr()->SensorId()) &&
-           problem.HasParameterBlock(
-               image.FramePtr()->RigFromWorld().translation.data());
+  // Check if a sensor is either a reference sensor, or a non-reference sensor
+  // with sensor_from_rig fixed.
+  auto IsParameterizedRefSensor = [&problem](
```
ahojnnes: nit, we could capture the config in the lambda instead of the argument?

src/colmap/estimators/bundle_adjustment.cc
ahojnnes: On my phone. Not sure I understand? We still want this to be a reference sensor, otherwise the camera from world pose would not be fixed?

B1ueber2y:
> Thanks, so this improves the edge case where a camera is set as constant but not added to the problem or where a camera doesn’t have any residuals? Out of curiosity, did you see this happen in practice? To be extra precise, do we also need to check for existence of the rotation parameters in the problem?

This improves the case when the reference sensor is not in the problem, but the other sensor in the rig is in the problem and its ``SensorFromRig`` is set constant.

src/colmap/estimators/bundle_adjustment.cc
B1ueber2y: In the whole Gauge logic we in the end fix the ``image.FramePtr()->RigFromWorld()``, and here in this part of the logic we check ``config.HasConstantRigFromWorld(image.FrameId())``, so whether the image is a reference sensor or not does not matter here.

In the other when the reference sensor is not in the problem (no residuals or no associated data), we will skip this frame which was however set constant in the config.

ahojnnes: Yes, but here we look for two cameras in the problem that already fix the Gauge, so we can skip the search for fixing additional parameters. I don't think the new logic satisfies fixing the Gauge, if it's not a reference sensor, because we can arbitrarily move the sensor from rig pose? The only improvement here can be that we accept a non ref camera if its sensor from rig is fixed?

B1ueber2y: Indeed. Thanks! This is done by using the same lambda ``IsParameterizedRefSensor``.

src/colmap/estimators/bundle_adjustment.cc
B1ueber2y: Done.

src/colmap/estimators/bundle_adjustment.cc
```diff
@@ -491,12 +491,23 @@ void FixGaugeWithTwoCamsFromWorld(
   Image* image1 = nullptr;
   Image* image2 = nullptr;
 
+  // Check if a sensor is either a reference sensor, or a non-reference sensor
+  // with sensor_from_rig fixed.
+  auto IsParameterizedRefSensor = [&config, &problem](const Image& image) {
```
ahojnnes: nit: IsParameterizedConstSensor?

ahojnnes:
If meaningful, we could cover the new case with an additional unit test as well?

src/colmap/estimators/bundle_adjustment.cc
```diff
@@ -493,21 +493,34 @@ void FixGaugeWithTwoCamsFromWorld(
...
+    Rigid3d& sensor_from_rig = image.FramePtr()->RigPtr()->SensorFromRig(
+        image.CameraPtr()->SensorId());
+    if (problem.HasParameterBlock(sensor_from_rig.rotation.coeffs().data()) &&
+        problem.IsParameterBlockConstant(
+            sensor_from_rig.rotation.coeffs().data()) &&
+        problem.HasParameterBlock(sensor_from_rig.translation.data()) &&
+        problem.IsParameterBlockConstant(sensor_from_rig.translation.data())) {
+      return true;
+    }
+    // Cover corner case when ReprojErrorConstantPoseCostFunctor is used
```
B1ueber2y, line 511: Note: the parameter block wont exist in the problem when both rig_from_world and sensor_from_rig are fixed.

B1ueber2y:
> If meaningful, we could cover the new case with an additional unit test as well?

Added two related tests that caught quite some issues in the existing logic. Fixed in the new commit. PTAL.

PR #3628, opened 2025-09-15
Throw exception when the rig configurations of the database and existing reconstruction are inconsistent. 

B1ueber2y:
This improves safety at the triangulation and will throw exceptions for all legacy databases that do not go through rig_configurator while corresponding to a rig configured reconstruction. 

Note that the current hloc triangulation breaks for reconstruction with non-trivial frames due to legacy database interfaces not updated here: https://github.com/cvg/Hierarchical-Localization/blob/master/hloc/triangulation.py#L29. Will send in a fix later in hloc today.


B1ueber2y:
also cc: @Phil26AT

sarlinpe:
> Note that the current hloc triangulation breaks 

Thank you for flagging this! Alternatively, would it make sense for COLMAP to provide a function to populate the cameras and images of an empty database given a reconstruction?

B1ueber2y:
> > Note that the current hloc triangulation breaks
> 
> Thank you for flagging this! Alternatively, would it make sense for COLMAP to provide a function to populate the cameras and images of an empty database given a reconstruction?

After migrating to ``pycolmap.Database``, the implementation of creating an empty database from a reconstruction is extremely straightforward: https://github.com/cvg/Hierarchical-Localization/blob/features/pycolmap_database/hloc/triangulation.py#L28-L44, so I think we do not need this function from COLMAP anymore?

src/colmap/scene/reconstruction.cc
```diff
@@ -135,23 +135,36 @@ void Reconstruction::Load(const DatabaseCache& database_cache) {
...
   }
 
   // Add rigs.
   rigs_.reserve(database_cache.NumRigs());
   for (const auto& [rig_id, rig] : database_cache.Rigs()) {
-    if (!ExistsRig(rig_id)) {
+    if (ExistsRig(rig_id)) {
+      class Rig& existing_rig = Rig(rig_id);
+      THROW_CHECK(existing_rig.RefSensorId() == rig.RefSensorId());
+      THROW_CHECK(existing_rig.SensorIds() == rig.SensorIds());
```
ahojnnes, line 154: Unrelated to this PR, but the current naming of ref vs non ref sensors is a bit ambiguous, where it's not clear whether "sensors" includes the reference sensor or not. Applies to most of the methods in the rig class. It's also a bit inconsistent where SensorsFromRig() doesn't include the ref sensor but the new SensorIds() does.

src/colmap/sensor/rig.h
```diff
@@ -70,6 +71,9 @@ class Rig {
   inline bool IsRefSensor(sensor_t sensor_id) const;
   inline bool HasSensorFromRig(sensor_t sensor_id) const;
 
+  // Get all sensor ids in the rig.
+  inline std::set<sensor_t> SensorIds() const;
```
ahojnnes, line 75: Let's add a test for this.

src/colmap/scene/reconstruction.cc
```diff
@@ -135,23 +135,36 @@ void Reconstruction::Load(const DatabaseCache& database_cache) {
...
     }
   }
 
   // Add rigs.
   rigs_.reserve(database_cache.NumRigs());
   for (const auto& [rig_id, rig] : database_cache.Rigs()) {
-    if (!ExistsRig(rig_id)) {
+    if (ExistsRig(rig_id)) {
+      class Rig& existing_rig = Rig(rig_id);
+      THROW_CHECK(existing_rig.RefSensorId() == rig.RefSensorId());
```
ahojnnes, line 153: Superfluous because SensorIds() already includes the ref sensor?

B1ueber2y: But we would want to check if the ref sensor is the same, since SensorIds() is sorted?

src/colmap/scene/reconstruction.cc
B1ueber2y: How about we do ``AllSensorIds()`` and ``SensorIds()``, with the former including the ref sensor and the latter being consistent with sensor_from_rig?

B1ueber2y: > How about we do `AllSensorIds()` and `SensorIds()`, with the former including the ref sensor and the latter being consistent with sensor_from_rig?

Done with this. PTAL : )

src/colmap/sensor/rig.h
```diff
@@ -136,6 +143,23 @@ bool Rig::HasSensorFromRig(sensor_t sensor_id) const {
...
+std::set<sensor_t> Rig::SensorIds() const {
+  std::set<sensor_t> sensor_ids;
+  for (const auto& [sensor_id, _] : sensors_from_rig_) {
+    sensor_ids.insert(sensor_id);
+  }
+  return sensor_ids;
+}
+
+std::set<sensor_t> Rig::AllSensorIds() const {
+  std::set<sensor_t> sensor_ids;
```
ahojnnes: How about reusing the function above, inserting the ref sensor and returning the set?

src/colmap/scene/reconstruction.cc
ahojnnes: Thanks. For full consistency, we could have a NumSensors and NumAllSensors? WDYT?

B1ueber2y: > Thanks. For full consistency, we could have a NumSensors and NumAllSensors? WDYT?

Nice catch of inconsistency. But ``NumSensors()`` and ``NumAllSensors()`` feel really confusing for me. Thinking about it, I will instead remove ``AllSensorIds()`` but rename ``Sensors()`` to ``NonRefSensors()``, to be semantically less confusing.

B1ueber2y: > > Thanks. For full consistency, we could have a NumSensors and NumAllSensors? WDYT?
> 
> Nice catch of inconsistency. But `NumSensors()` and `NumAllSensors()` feel really confusing for me. Thinking about it, I will instead remove `AllSensorIds()` but rename `Sensors()` to `NonRefSensors()`, to be semantically less confusing.

Done: https://github.com/colmap/colmap/pull/3628/commits/1154bdf7188af17cbe9c252186bd4c9f73c0997f. What do you think?

ahojnnes: SGTM.

PR #3631, opened 2025-09-16
Cherry pick changes for 3.12.6 release.

B1ueber2y:


B1ueber2y:
The database abstraction update is deeply coupled with the abstraction of feature extraction and matching, and it might not make sense to cherry pick these abstraction out since they are huge updates. 

A small note is that, if we leave these abstration out, this unfortunately means that for hloc fix with pycolmap 3.12.6 we would have to stick to the interface ``pycolmap.Database(db_path)``, which is different from that of the colmap head, where we have the nice scope feature ``with pycolmap.Database.open(db_path) as db:``. This will only be available when these abstractions are merged, probably only in the long future until the 3.13 release, upon when we have to update hloc again if we want to use newer pycolmap as this interface change will not be backwards compatible.

ahojnnes:
> The database abstraction update is deeply coupled with the abstraction of feature extraction and matching, and it might not make sense to cherry pick these abstraction out since they are huge updates. 
> 
> 
> 
> A small note is that, if we leave these abstration out, this unfortunately means that for hloc fix with pycolmap 3.12.6 we would have to stick to the interface ``pycolmap.Database(db_path)``, which is different from that of the colmap head, where we have the nice scope feature ``with pycolmap.Database.open(db_path) as db:``. This will only be available when these abstractions are merged, probably only in the long future until the 3.13 release, upon when we have to update hloc again if we want to use newer pycolmap as this interface change will not be backwards compatible. 

Thanks, how about an if-condition on the pycolmap version in hloc for the time being? This should be minimal code duplication?

B1ueber2y:
>> Thanks, how about an if-condition on the pycolmap version in hloc for the time being? This should be minimal code duplication?

Good idea. Added a custom context manager here: https://github.com/cvg/Hierarchical-Localization/blob/features/pycolmap_database/hloc/utils/database.py

PR #3634, opened 2025-09-17
Fix broken logic of removing cameras for Reconstruction::TearDown.

B1ueber2y:
With the new abstraction of rigs and frames, at ``Reconstruction::TearDown()``, the removal of cameras should correspond to the removal of rigs. Otherwise the reconstruction object will be broken after TearDown().

In practice, This PR that only fixes corner cases when no associated images are available for a camera in the rig.

### DIFF
--- a/src/colmap/scene/reconstruction.cc
+++ b/src/colmap/scene/reconstruction.cc
@@ -190,16 +190,7 @@ void Reconstruction::Load(const DatabaseCache& database_cache) {
 void Reconstruction::TearDown() {
   // Remove all non-registered frames/images.
   std::unordered_set<rig_t> keep_rig_ids;
-  std::unordered_set<camera_t> keep_camera_ids;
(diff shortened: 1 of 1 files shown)


src/colmap/scene/reconstruction.cc
```diff
@@ -208,24 +205,20 @@ void Reconstruction::TearDown() {
     }
   }
 
-  // Remove all unused rigs.
+  // Remove all unused rigs and corresponding sensors.
   for (auto it = rigs_.begin(); it != rigs_.end();) {
     if (keep_rig_ids.count(it->first) == 0) {
+      for (const sensor_t& sensor_id : it->second.SensorIds()) {
+        if (sensor_id.type == SensorType::CAMERA) {
```
ahojnnes: Suggest to make this a switch/case, so the compiler complains when we don't handle specific sensor types in the future.

PR #3644, opened 2025-09-25
Remove plyfile python package copy due to GNU license

ahojnnes:
plyfile comes with a GNU license and can be easily installed from PyPI if needed.


B1ueber2y:
Is this other file useful anywhere (it used plyfile): https://github.com/colmap/colmap/blob/main/scripts/python/merge_ply_files.py? Also it seems that many of the files in the scripts folder is a bit outdated.

ahojnnes:
> Is this other file useful anywhere (it used plyfile): https://github.com/colmap/colmap/blob/main/scripts/python/merge_ply_files.py? Also it seems that many of the files in the scripts folder is a bit outdated.

Yes, we can clean it up a bit. At the moment, there is no harm in keeping this specific file. Maybe somebody is using it and they continue to do so through a `pip install plyfile`.

PR #3645, opened 2025-09-26
rig_from_world and sensor_from_rig in pycolmap should return reference rather than copy.

B1ueber2y:
This is a crucial fix. Without this the python side problem construction with pyceres will never work. Broken since the rig abstraction in colmap 3.12. 

Equivalent reference in ``image.cam_from_world`` from COLMAP 3.11: https://github.com/colmap/colmap/blob/release/3.11/src/pycolmap/scene/image.cc#L81-L87

To reproduce:
```
import pycolmap

recon = pycolmap.Reconstruction("example/sfm/0")
rig_from_world_1 = recon.frames[1].rig_from_world
rig_from_world_2 = recon.frames[1].rig_from_world
print(id(rig_from_world_1))
print(id(rig_from_world_2))
assert(id(rig_from_world_1) == id(rig_from_world_2))
```

The assertion fails without this fix.

### DIFF
--- a/src/pycolmap/scene/frame.cc
+++ b/src/pycolmap/scene/frame.cc
@@ -55,16 +55,9 @@ void BindFrame(py::module& m) {
            "Make the rig pointer a nullptr.")
       .def_property(
           "rig_from_world",
-          [](Frame& self) -> py::typing::Optional<Rigid3d> {
-            if (self.HasPose()) {
-              return py::cast(self.RigFromWorld());
-            } else {
-              return py::none();
-            }
-          },
-          [](Frame& self, const Rigid3d& rig_from_world) {
-            self.SetRigFromWorld(rig_from_world);
-          },
+          py::overload_cast<>(&Frame::MaybeRigFromWorld),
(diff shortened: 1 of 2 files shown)


B1ueber2y:
Same applies to ``rig.sensor_from_rig(sensor_id)``.

ahojnnes (approved):

robinlindh:
Ahhhh yes, I've been debugging exactly this for 2 days now after I finally managed to get our custom SfM pipeline upgraded from colmap 3.10 to 3.12.6. Was a pretty big step to get everything running again after all python changes but then I ended up getting just VERY few 3D points after my pipeline ran through.
I am indeed passing in the rig_from_world into a pyceres problem (based loosely off of the custom bundle adjuster python example). The bug you mention here would explain exactly the results I see. I will let you know how it works with the fix. Thank you a lot in advance <3 and fingers crossed!

robinlindh:
This fixed my issue indeed. Awesome, thank you very much!

PR #3652, opened 2025-10-03
Expose image_ids filter view in frame in pycolmap

ahojnnes:


src/pycolmap/scene/frame.cc
```diff
@@ -39,6 +39,16 @@ void BindFrame(py::module& m) {
...
           [](const Frame& self) { return self.DataIds(); },
           "The associated data.")
+      .def_property_readonly(
+          "image_ids",
+          [](const Frame& self) {
+            std::vector<data_t> image_ids;
+            for (const auto& image_id : self.ImageIds()) {
+              image_ids.push_back(image_id);
+            }
+            return image_ids;
```
sarlinpe:
```suggestion
            const auto image_ids = self.ImageIds();
            return std::vector<data_t>(image_ids.begin(), image_ids.end());
```

PR #3656, opened 2025-10-07
Disable auto vacuum to speed up CloseImpl

whuaegeanse:
After calling the ApplyRigConfig function, the exe may experience long I/O cycles (90% utilization). Investigation revealed that the Database destructor was particularly time-consuming. Further investigation revealed that this was caused by setting auto_vacuum to 1. If auto_vacuum is 0, SQLite does not clean up deleted space when deleting data, allowing it to reuse the deleted space when adding data later.
The ApplyRigConfig function primarily deletes frames and rigs. However, frames and rigs occupy very little storage space, so clearing them would not yield significant benefits.

Changing from
SQLITE3_EXEC(database->database_, "PRAGMA auto_vacuum=1", nullptr);
to
SQLITE3_EXEC(database->database_, "PRAGMA auto_vacuum=0", nullptr);
reduces the time it takes to destroy the database, especially when the SQLite database size reaches 1-10GB.

### DIFF
--- a/src/colmap/exe/database.cc
+++ b/src/colmap/exe/database.cc
@@ -153,7 +153,9 @@ int RunRigConfigurator(int argc, char** argv) {
     reconstruction->Read(input_path);
   }
 
-  auto database = Database::Open(database_path);
+  Database::Options database_options;
+  database_options["disable_auto_vacuum"] = "1";
+  auto database = Database::Open(database_path, database_options);
 
   ApplyRigConfig(
       ReadRigConfig(rig_config_path),
--- a/src/colmap/scene/database.cc
+++ b/src/colmap/scene/database.cc
@@ -41,10 +41,11 @@ void Database::Register(Factory factory) {
 
 Database::~Database() = default;
 
-std::shared_ptr<Database> Database::Open(const std::string& path) {
+std::shared_ptr<Database> Database::Open(const std::string& path,
+                                         const Options& options) {
   for (auto it = factories_.rbegin(); it != factories_.rend(); ++it) {
     try {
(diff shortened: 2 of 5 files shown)


ahojnnes:
It also means that storage is not automatically freed up when data is deleted from the database?

whuaegeanse:
> It also means that storage is not automatically freed up when data is deleted from the database?

Yes.
When auto_vacuum=0, SQLite will reuse storage marked for deletion. This should be sufficient for colmap, especially for feature extraction and feature matching.

ahojnnes:
What's the runtime savings and corresponding stats (#images/frames/rigs)?

whuaegeanse:
> What's the runtime savings and corresponding stats (#images/frames/rigs)?

Sorry for the late reply.
I used two datasets for experiments, with the following results:
Data 1, with 309 x 2 images (1 rig, 309 frames, 718 images), a database size of 0.93 GB, reduced ApplyRigConfig time from 30.3 seconds to 0.02 seconds.
Data 2, with 1630 x 2 images (1 rig, 1630 frames, 3260 images), a database size of 6.60 GB, reduced ApplyRigConfig time from 557.5 seconds to 0.02 seconds.

sarlinpe:
> The ApplyRigConfig function primarily deletes frames and rigs. However, frames and rigs occupy very little storage space, so clearing them would not yield significant benefits.

Ideally we would set `auto_vacuum=0` only when deleting frames/rigs. There are other use cases in which one might want to remove features from the database - then `auto_vacuum=1` helps reduce the database size by a lot.

PR #3658, opened 2025-10-08
Adjust reconstruction consistency checks

ahojnnes:
It's hugely inconvenient to require data to already be in the frame when a frame is added to the reconstruction and it broke a bunch of internal code. Instead, I am tightening some of the other checks, where I am now checking that when adding an image, the corresponding frame has the image and the equivalent for the sensors in frames.


B1ueber2y:
We can do the same update for rigs and cameras. Now that camera needs to be inserted before rig which is with a different order from images and frames. We can switch the order and check if the rig exists and if the camera is in the rig.

PR #3663, opened 2025-10-14
Expose comparison operators in Python bindings

jhacsonmeza:
This pull request exposes to Python the `operator==` and `operator<` from multiple C++ classes and structs that are not being exposed with the pybind11 definitions.

I noticed this problem after trying to compare two identical cameras in Python:
```python
cam1 = pycolmap.Camera.create(1, pycolmap.CameraModelId.PINHOLE, 7446, 6205, 4134)
cam2 = pycolmap.Camera.create(1, pycolmap.CameraModelId.PINHOLE, 7446, 6205, 4134)
cam1 == cam2 # This produces False
```
In this case the output should be True.

These are all the classes/structs I identified with missing comparison operators in Python: `Camera`, `TrackElement`, `Track`, `Point2D`, `Point3D`, `PosePrior`, `Frame`, `Image`, `Sim3d`, `Rigid3d`, `sensor_t`, `data_t`, and `Rig`.

### DIFF
--- a/src/pycolmap/helpers.h
+++ b/src/pycolmap/helpers.h
@@ -9,6 +9,7 @@
 #include <regex>
 #include <sstream>
 #include <string>
+#include <type_traits>
 
 #include <Eigen/Core>
 #include <glog/logging.h>
@@ -292,6 +293,31 @@ void AddDefaultsToDocstrings(py::classh<T, options...> cls) {
   }
 }
 
+template <typename T, typename = void>
+struct has_equality_operator : std::false_type {};
+
+template <typename T>
+struct has_equality_operator<
+    T,
+    std::void_t<decltype(std::declval<T&>() == std::declval<T&>())>>
+    : std::true_type {};
+
+template <typename T, typename = void>
+struct has_less_than_operator : std::false_type {};
+
+template <typename T>
+struct has_less_than_operator<
+    T,
+    std::void_t<decltype(std::declval<T&>() < std::declval<T&>())>>
+    : std::true_type {};
+
+template <typename T, typename = void>
+struct is_hashable : std::false_type {};
+
+template <typename T>
+struct is_hashable<T, std::void_t<decltype(std::hash<T>{}(std::declval<T>()))>>
+    : std::true_type {};
+
 template <typename T, typename... options>
 void MakeDataclass(py::classh<T, options...> cls,
                    const std::vector<std::string>& attributes = {}) {
@@ -335,6 +361,29 @@ void MakeDataclass(py::classh<T, options...> cls,
         self.attr("mergedict").attr("__call__")(dict);
         return self.cast<T>();
       }));
+
+  if constexpr (has_equality_operator<T>::value) {
+    cls.def(py::self == py::self);
+    if constexpr (is_hashable<T>::value) {
+      cls.def("__hash__", [](const T& self) { return std::hash<T>()(self); });
+    } else {
+      cls.attr("__hash__") = py::none();
+    }
+  } else {
+    cls.def("__eq__", [attributes](const T& self, const py::object& other) {
+      if (!py::isinstance<T>(other)) {
+        return false;
+      }
+      py::dict self_dict = ConvertToDict(self, attributes, true);
+      py::dict other_dict = ConvertToDict(other.cast<T>(), attributes, true);
+      return self_dict.equal(other_dict);
+    });
+    cls.attr("__hash__") = py::none();
+  }
+
+  if constexpr (has_less_than_operator<T>::value) {
+    cls.def(py::self < py::self);
+  }
 }
 
 // Catch python keyboard interrupts


sarlinpe:
Thank you! Should these be added to MakeDataclass instead?
https://github.com/colmap/colmap/blob/7a896e19d818b89fa342ea64ed44c75ada933aca/src/pycolmap/helpers.h#L296

jhacsonmeza:
Hi,
Do you mean to add something like `cls.def(py::self == py::self);` in `MakeDataclass`? If that's the case, the problem is that not all the classes have `operator==`. Also, `sensor_t` and `data_t` are the only ones that define a `operator<` function.

I just quickly tried to add `cls.def(py::self == py::self);` to `MakeDataclass` and the compilation fails because `MakeDataclass` is used with many classes that don't have a `operator==`. A potential solution I was thinking was to use a helper function to detect if the class has a `operator==` function:
```
template<typename T>
struct has_equality_operator {
    template<typename U>
    static auto test(U* u) -> decltype(*u == *u, std::true_type{});
    
    template<typename U>
    static auto test(...) -> std::false_type;
    
    static constexpr bool value = decltype(test<T>(nullptr))::value;
};
```
which can be used inside `MakeDataclass` in this way:
```
if constexpr (has_equality_operator<T>::value) {
    cls.def(py::self == py::self);
}
```
I tried this and it worked.

However, I don't know if I'm overlooking something and the solution is actually easier. I would like to know what do you think.

sarlinpe:
Thank you for the detailed reply! 

First, it looks like the tests are failing even with the current state. In Python, when you define an equality method but not a hash method, the object automatically becomes unhashable. So we'd need to define `__hash__` for sensor_t and data_t, e.g.
```c++
    .def("__hash__",
         [](const sensor_t& a) { return std::hash<colmap::sensor_t>()(a); });
```

As for the MakeDataclass, which classes do not have an equality operator? We could have a fallback that checks equality of the python dict (slower but ok):
```c++
#include <type_traits>

template<typename T, typename = void>
struct has_equality_operator : std::false_type {};

template<typename T>
struct has_equality_operator<T, std::void_t<decltype(std::declval<T&>() == std::declval<T&>())>> : std::true_type {};

template<typename T, typename = void>
struct has_less_than_operator : std::false_type {};

template<typename T>
struct has_less_than_operator<T, std::void_t<decltype(std::declval<T&>() < std::declval<T&>())>> : std::true_type {};

template<typename T, typename = void>
struct is_hashable : std::false_type {};

template<typename T>
struct is_hashable<T, std::void_t<decltype(std::hash<T>{}(std::declval<T>()))>> : std::true_type {};

template <typename T, typename... options>
void MakeDataclass(py::classh<T, options...> cls,
                   const std::vector<std::string>& attributes = {}) {
    // ...

    if constexpr (has_equality_operator<T>::value) {
        cls.def(py::self == py::self);
        if constexpr (is_hashable<T>::value) {
            cls.def("__hash__", [](const T& self) { return std::hash<T>()(self); });
        } else {
            cls.attr("__hash__") = py::none();
        }
    } else {
        cls.def("__eq__", [attributes](const T& self, const py::object& other) {
            if (!py::isinstance<T>(other)) {
                return false;
            }
            py::dict self_dict = ConvertToDict(self, attributes, true);
            py::dict other_dict = ConvertToDict(other.cast<T>(), attributes, true);
            return self_dict.equal(other_dict);
        });
        cls.attr("__hash__") = py::none();
    }

    if constexpr (has_less_than_operator<T>::value) {
        cls.def(py::self < py::self);
    }
}
```
(not tested)

jhacsonmeza:
Thank you for pointing that out! I didn't know we needed the definition of a hash method. However, I don't understand why `data_t` is the only one with this problem. The other classes don't seem to have this hashing problem in the tests (if I understand correctly). Is it something with the design of the test that somehow only checks for `data_t`?

These are examples of classes that use `MakeDataclass` and don't have an `operator==`: [`mvs::PatchMatchOptions`](https://github.com/colmap/colmap/blob/7a896e19d818b89fa342ea64ed44c75ada933aca/src/colmap/mvs/patch_match_options.h#L37) ([binding](https://github.com/colmap/colmap/blob/7a896e19d818b89fa342ea64ed44c75ada933aca/src/pycolmap/pipeline/mvs.cc#L175)), [`mvs::StereoFusionOptions`](https://github.com/colmap/colmap/blob/7a896e19d818b89fa342ea64ed44c75ada933aca/src/colmap/mvs/fusion.h#L53) ([binding](https://github.com/colmap/colmap/blob/7a896e19d818b89fa342ea64ed44c75ada933aca/src/pycolmap/pipeline/mvs.cc#L237)), [`internal::PoseParam`](https://github.com/colmap/colmap/blob/7a896e19d818b89fa342ea64ed44c75ada933aca/src/colmap/estimators/covariance.h#L136) ([binding](https://github.com/colmap/colmap/blob/7a896e19d818b89fa342ea64ed44c75ada933aca/src/pycolmap/estimators/covariance.cc#L74)), [`BACovarianceOptions`](https://github.com/colmap/colmap/blob/7a896e19d818b89fa342ea64ed44c75ada933aca/src/colmap/estimators/covariance.h#L93) ([binding](https://github.com/colmap/colmap/blob/7a896e19d818b89fa342ea64ed44c75ada933aca/src/pycolmap/estimators/covariance.cc#L96)), and many more that I saw when I got the compilation error.

I tested the code you provided and it compiles correctly. This time I also ran the test and it's passing successfully. So, if you agree I can make a new commit with these changes. Do you prefer to have the helper functions just before `MakeDataclass`?

sarlinpe:
In the tests (and likely in COLMAP in general) data_t and sensor_t are the only types that we try to hash in sets or as map keys. I think that it's fine to fallback to dict equality for Options objects, since we rarely need to compare them. You can add this code as is, thank you!

jhacsonmeza:
Thank you for clarifying! I already added the changes.

sarlinpe (approved):

PR #3664, opened 2025-10-17
Compute alignment RANSAC max_error from RMS stddev and chi-square

StonerLing:
The previous method is sensitive to priors with extreme covariances, producing unstable RANSAC thresholds. 
This PR computes the mean RMS stddev from valid covariances and scale it by sqrt(chi-square 95% quantile, dof=3)
 to yield an approximate 95% 3D confidence radius.

### DIFF
--- a/src/colmap/estimators/bundle_adjustment_test.cc
+++ b/src/colmap/estimators/bundle_adjustment_test.cc
@@ -130,9 +130,9 @@ void AddNoiseToReconstruction(Reconstruction& reconstruction,
   // Perturb 3D points
   for (const point3D_t point3D_id : reconstruction.Point3DIds()) {
     Point3D& point = reconstruction.Point3D(point3D_id);
-    point.xyz += Eigen::Vector3d(RandomGaussian<double>(0, translation_stddev),
-                                 RandomGaussian<double>(0, translation_stddev),
-                                 RandomGaussian<double>(0, translation_stddev));
+    point.xyz += Eigen::Vector3d(RandomGaussian<double>(0, point_stddev),
+                                 RandomGaussian<double>(0, point_stddev),
+                                 RandomGaussian<double>(0, point_stddev));
   }
 }
 


src/colmap/estimators/bundle_adjustment.cc
```diff
@@ -1049,26 +1049,28 @@ class PosePriorBundleAdjuster : public BundleAdjuster {
...
-                            .eigenvalues()
-                            .maxCoeff());
-          max_stddev_sum += max_stddev;
-          ++num_valid_covs;
+          const double trace = pose_prior.position_covariance.trace();
+          if (trace > 0.0) {
+            const double rms_stddev = std::sqrt(trace / 3.0);
+            sum_rms_stddev += rms_stddev;
+            ++num_valid_covs;
+          }
```
ahojnnes: What do you think about the following? This would be more robust to outliers?
```
std::vector<double> rms_vars;
for (const auto& [_, pose_prior] : pose_priors_) {
  if (pose_prior.IsCovarianceValid()) {
    const double trace = pose_prior.position_covariance.trace();
          if (trace > 0.0) {
            const double rms_stddev = trace / 3.0;
            rms_vars.push_back(rms_var);
          }
    
  }
}
ransac_options.max_error = std::sqrt(kChiSquare95ThreeDof * Median(rms_vars));

```

StonerLing: Good point, using the median makes it more robust to outliers.

src/colmap/estimators/bundle_adjustment.cc
```diff
@@ -1049,42 +1049,44 @@ class PosePriorBundleAdjuster : public BundleAdjuster {
...
       }
-      // Set max error at the 3 sigma confidence interval. Assumes no
-      // outliers.
-      ransac_options.max_error = 3 * max_stddev_sum / num_valid_covs;
+
+      // Set max error using the median RMS variance of valid pose priors.
+      // Scaled by sqrt(chi-square 95% quantile, 3 DOF) to approximate a 95%
+      // confidence radius.
+      ransac_options.max_error =
+          std::sqrt(kChiSquare95ThreeDof * Median(rms_vars));
```
ahojnnes, line 1096: Does it make sense to make this consistent?

```suggestion
          std::sqrt(options_.prior_position_loss_scale * Median(rms_vars));
```

StonerLing: I think only the default `kChiSquare95ThreeDof` makes sense. Using a user-specified `prior_position_loss_scale` to scale the threshold would break its statistical meaning: a large user value could make RANSAC classify outliers as inliers, which is unreliable

StonerLing: Alternatively, it might make more sense to use `sqrt(kChiSquare95ThreeDof * Median(rms_vars))` as the default for both the RANSAC threshold and the robust loss scale. This keeps the defaults consistent and statistically meaningful

ahojnnes: Yes, I agree with the latter proposal.

StonerLing: Rethinking this, `options_.prior_position_loss_scale` should default to `sqrt(kChiSquare95ThreeDof)` , since the prior residuals in Ceres are already scaled by their covariance, i.e., normalized

ahojnnes:
Last request: Can we adjust existing tests or create a new one to cover this new feature? Thanks.

StonerLing:
Added two simple tests.

src/colmap/estimators/bundle_adjustment_test.cc
```diff
@@ -103,6 +106,71 @@ constexpr double kConstantPoseVarEps = 1e-9;
...
+                        RandomGaussian<double>(0, translation_stddev),
+                        RandomGaussian<double>(0, translation_stddev));
+  }
+
+  // Perturb 3D points
+  for (const point3D_t point3D_id : reconstruction.Point3DIds()) {
+    Point3D& point = reconstruction.Point3D(point3D_id);
+    point.xyz += Eigen::Vector3d(RandomGaussian<double>(0, translation_stddev),
+                                 RandomGaussian<double>(0, translation_stddev),
+                                 RandomGaussian<double>(0, translation_stddev));
```
ahojnnes: point_stddev?

PR #3670, opened 2025-10-24
Extract synthetic reconstruction noise functionality into separate utility

ahojnnes:
…ility

### DIFF
--- a/src/colmap/scene/synthetic.cc
+++ b/src/colmap/scene/synthetic.cc
@@ -433,52 +433,48 @@ void SynthesizeDataset(const SyntheticDatasetOptions& options,
 
 void SynthesizeNoise(const SyntheticNoiseOptions& options,
                      Reconstruction* reconstruction) {
-  THROW_CHECK_GE(options.rig_from_world_translation_noise_stddev, 0.);
-  THROW_CHECK_GE(options.rig_from_world_rotation_noise_stddev, 0.);
-  THROW_CHECK_GE(options.point3D_noise_stddev, 0.);
-  THROW_CHECK_GE(options.point2D_noise_stddev, 0.);
+  THROW_CHECK_GE(options.rig_from_world_translation_stddev, 0.);
+  THROW_CHECK_GE(options.rig_from_world_rotation_stddev, 0.);
+  THROW_CHECK_GE(options.point3D_stddev, 0.);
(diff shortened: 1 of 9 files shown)


src/colmap/scene/synthetic.h
```diff
@@ -73,4 +72,14 @@ void SynthesizeDataset(const SyntheticDatasetOptions& options,
                        Reconstruction* reconstruction,
                        Database* database = nullptr);
 
+struct SyntheticNoiseOptions {
+  double rig_from_world_translation_noise_stddev = 0.0;
+  double rig_from_world_rotation_noise_stddev = 0.0;  // in degrees
```
B1ueber2y: Should we add a comment on the rotation axis here?

src/colmap/scene/synthetic.cc
```diff
@@ -438,4 +431,58 @@ void SynthesizeDataset(const SyntheticDatasetOptions& options,
   reconstruction->UpdatePoint3DErrors();
 }
 
+void SynthesizeNoise(const SyntheticNoiseOptions& options,
+                     Reconstruction* reconstruction) {
+  THROW_CHECK_GE(options.rig_from_world_translation_noise_stddev, 0.);
```
B1ueber2y: nit: Move to Check() and use CHECK_OPTION_GE?

B1ueber2y (approved):

src/pycolmap/scene/synthetic.cc
```diff
@@ -61,4 +59,21 @@ void BindSynthetic(py::module& m) {
...
       "database"_a = py::none());
+
+  auto PySyntheticNoiseOptions =
+      py::classh<SyntheticNoiseOptions>(m, "SyntheticNoiseOptions")
+          .def(py::init<>())
+          .def_readwrite(
+              "rig_from_world_translation_noise_stddev",
+              &SyntheticNoiseOptions::rig_from_world_translation_noise_stddev)
+          .def_readwrite(
+              "rig_from_world_rotation_noise_stddev",
```
B1ueber2y: Same here? Maybe comment on the rotation axis.

src/colmap/scene/synthetic.cc
ahojnnes: The CHECK_OPTION_GE is primarily meant to be used with the OptionManager. I don't think this options struct will make it to there.

PR #3673, opened 2025-10-27
Add gmock matcher for checking reconstruction equality

ahojnnes:


ahojnnes:
> Thank you! Alternatively, does it make sense to implement operator overloading for == and != of Reconstruction?

We can have it, but the matcher has the benefit that it can explain the difference with more detail.

PR #3676, opened 2025-10-28
Add timeout option for incremental mapper

ahojnnes:


src/colmap/controllers/incremental_pipeline.cc
```diff
@@ -201,16 +201,16 @@ IncrementalPipeline::IncrementalPipeline(
...
   THROW_CHECK(options_->Check());
   RegisterCallback(INITIAL_IMAGE_PAIR_REG_CALLBACK);
   RegisterCallback(NEXT_IMAGE_REG_CALLBACK);
   RegisterCallback(LAST_IMAGE_REG_CALLBACK);
 }
 
 void IncrementalPipeline::Run() {
-  Timer run_timer;
-  run_timer.Start();
+  total_run_timer_->Start();
```
B1ueber2y, line 213: Nit: to match the previous logic we should call total_run_timer_->Reset() here, though looking at the current logic of Reset() it does not matter at all.

src/colmap/controllers/incremental_pipeline.h
```diff
@@ -158,6 +160,10 @@ struct IncrementalPipelineOptions {
   // of refine_focal_length, refine_principal_point, and refine_extra_params.
   std::unordered_set<camera_t> constant_cameras;
 
+  // Maximum runtime in seconds for the reconstruction process.
+  // If set to a negative value, the process will run until completion.
```
B1ueber2y: Non positive?

src/pycolmap/sfm/incremental_mapper.cc
```diff
@@ -189,6 +189,11 @@ void BindIncrementalPipeline(py::module& m) {
                      "List of cameras for which to fix the camera parameters "
                      "independent of refine_focal_length, "
                      "refine_principal_point, and refine_extra_params.")
+      .def_readwrite(
+          "max_runtime_seconds",
+          &Opts::max_runtime_seconds,
+          "Maximum runtime in seconds for the reconstruction process. If set "
+          "to a negative value, the process will run until completion.")
```
B1ueber2y: Non positive?

src/colmap/controllers/incremental_pipeline.cc
ahojnnes: The reset is actually unnecessary. Calling start will reset the timer.

PR #3677, opened 2025-10-28
Fix BA convergence setting bug during initialization

ahojnnes:


src/colmap/sfm/incremental_mapper.cc
```diff
@@ -777,9 +777,9 @@ bool IncrementalMapper::AdjustGlobalBundle(
   BundleAdjustmentOptions custom_ba_options = ba_options;
   // Use stricter convergence criteria for first registered images.
   constexpr size_t kMinNumRegFramesForFastBA = 10;
-  const bool use_fast_ba =
+  const bool sufficient_reconstruction_size =
```
sarlinpe: How about calling this more informatively `is_small_reconstruction`? (and reverse the condition)

ahojnnes: Done. Had to invert the other conditions though now :-)

PR #3680, opened 2025-10-30
Fix noise synthesis to consistently update database keypoints

ahojnnes:


src/colmap/controllers/incremental_pipeline_test.cc
```diff
@@ -690,23 +674,6 @@ TEST(IncrementalPipeline, PriorBasedSfMWithRandomSeedStability) {
         run_mapper(/*num_threads=*/1, /*random_seed=*/kRandomSeed);
     EXPECT_THAT(*reconstruction_manager0->Get(0),
                 ReconstructionEq(*reconstruction_manager1->Get(0)));
-
-    // Different seed should produce different reconstructions. Notice that, for
```
ahojnnes, line 694: It seems like our pipeline is less sensitive to the random seed and it becomes harder to get different results for different seeds. Removing this check, as it now fails almost every other PR.

PR #3681, opened 2025-11-03
Update the logic of duplicate correspondence to support one-to-many matches.

B1ueber2y:


B1ueber2y:
Looking at the failing test I realized that these one-to-many matches are filtered out by design. I am not a big fan of this, since these one-to-many matches can potentially help two-view geometry to find better models, and can easily come from one-way nearest neighbor matching. 

The greedy logic employed in the current code is also not very ideal, since it is dependent on the order of input matches. Consider the following minimal case where we have matches [0, 0], [1, 1], [0, 1]. The number of correspondences can be either 1 or 2 depending on the ordering. The warning of the current logic is also a bit confusing as these are not really "duplicate" matches. 

I suggest to keep the one-to-many matches and only filter out the completely duplicate ones. I understand that this might be designed for keeping a good count of correspondences for NBV selection etc., but I believe that this can be handled from the matching side outside ``CorrespondenceGraph``, without sacrificing the possibility to use such one-to-many matches.

ahojnnes:
Thanks, I generally agree but we need to check whether this violates some assumptions in the current implementation of keeping track of number of visible points for NBV, etc. I vaguely remember something about this.

ahojnnes:
@B1ueber2y I checked the observation manager implementation and couldn't find anything that would prevent us from merging your suggested change.

B1ueber2y:
> @B1ueber2y I checked the observation manager implementation and couldn't find anything that would prevent us from merging your suggested change.

Thanks for checking!

src/colmap/scene/correspondence_graph.cc
```diff
@@ -126,20 +126,15 @@ void CorrespondenceGraph::AddCorrespondences(const image_t image_id1,
...
-      const bool duplicate2 =
-          std::find_if(corrs2.begin(),
-                       corrs2.end(),
-                       [image_id1](const Correspondence& corr) {
-                         return corr.image_id == image_id1;
-                       }) != corrs2.end();
-
-      if (duplicate1 || duplicate2) {
+
+      if (duplicate) {
```
ahojnnes, line 139: This seems wrong. Shouldn't we instead just change this to be `duplicate1 && duplicate2` instead of `duplicate1 || duplicate2`?

B1ueber2y: Why is this wrong?

ahojnnes (changes requested):
I have a concern with the current change.

src/colmap/scene/correspondence_graph.cc
B1ueber2y: duplicate1 && duplicate2 is to say that both keypoints have a second match, while the current duplicate is to say that there are duplicate matches.

ahojnnes: Didn't see we are adding the correspondences bi-directionally, which makes this correct. Maybe add a short comment here to explain why checking it in one direction is sufficient?

B1ueber2y: Sure. This is done: https://github.com/colmap/colmap/pull/3681/commits/61d1a26440a5dee4fb8f674c13f68aa22d0c55cb

PR #3690, opened 2025-11-04
Update deprecated Eigen JacobiSVD usage

ahojnnes:


sarlinpe:
The compile-time options were only added in v5.0.0, right? https://gitlab.com/libeigen/eigen/-/merge_requests/826 Should we instead add a check on the Eigen version?

ahojnnes:
@sarlinpe Thanks, didn't realize. Done.

PR #3696, opened 2025-11-05
Add function to synthesize images

ahojnnes:
Creates simple routine for generating synthetic images that can be successfully matched and reconstructed. This is in preparation for a synthetic end-to-end test of the automatic reconstruction controller.

Keypoints:
keypoints

Inlier matches:
inliers

Reconstruction:
model

### DIFF
--- a/src/colmap/scene/synthetic.h
+++ b/src/colmap/scene/synthetic.h
@@ -93,7 +93,8 @@ struct SyntheticImageOptions {
 
 // Generates patches with a dark background and a bright feature peak for each
 // 2D point in an image. The color of the peak and the pattern of the background
-// is unique per 3D point.
+// is unique per 3D point. Notice that this approach does not result in perfect
+// feature detections and matches due to overlapping patches, etc.
 void SynthesizeImages(const SyntheticImageOptions& options,
                       const Reconstruction& reconstruction,
                       const std::string& image_path);


B1ueber2y:
Thanks for the nice feature! However, I have the following two concerns:

1. Some of the feature patches are occluded by other peaks, depending on the ordering of the synthesis. 
2. Not sure how reliable the matches are, especially given that many descriptors work on gray-scaled images. I suggest to create a test on the resulting matches.

ahojnnes:
> Thanks for the nice feature! However, I have the following two concerns:
> 
> 1. Some of the feature patches are occluded by other peaks, depending on the ordering of the synthesis.
> 2. Not sure how reliable the matches are, especially given that many descriptors work on gray-scaled images. I suggest to create a test on the resulting matches.

Thanks, these are valid concerns, but the goal is not for producing perfect matches, which requires a much more sophisticated approach and likely isn't even possible in the first place. It also depends on the specific feature detector and matcher implementation / parameters. As per my limited testing, the resulting images produce consistent feature detections/matches and reconstructions. If you have specific suggestions for how to make this better without boiling the ocean, then I am more than happy to improve upon the current implementation.

B1ueber2y:
> > Thanks for the nice feature! However, I have the following two concerns:
> > 
> > 1. Some of the feature patches are occluded by other peaks, depending on the ordering of the synthesis.
> > 2. Not sure how reliable the matches are, especially given that many descriptors work on gray-scaled images. I suggest to create a test on the resulting matches.
> 
> Thanks, these are valid concerns, but the goal is not for producing perfect matches, which requires a much more sophisticated approach and likely isn't even possible in the first place. It also depends on the specific feature detector and matcher implementation / parameters. As per my limited testing, the resulting images produce consistent feature detections/matches and reconstructions. If you have specific suggestions for how to make this better without boiling the ocean, then I am more than happy to improve upon the current implementation.

To improve matchability, does it make sense to use a coded pattern for the peak rather than use one single color? We can benefit from these gabor bases as in calibration and structured light, e.g., https://github.com/opencv/opencv_contrib/blob/4.x/modules/structured_light/src/sinusoidalpattern.cpp.

The code looks long and complicated, but with the help of ChatGPT i got something quite minimal (xD): https://chatgpt.com/share/690b6054-da6c-8010-83fa-2e75c93b6e4a

I assume classical feature descriptors would work very well on such patterns.

B1ueber2y:
> > > Thanks for the nice feature! However, I have the following two concerns:
> > > 
> > > 1. Some of the feature patches are occluded by other peaks, depending on the ordering of the synthesis.
> > > 2. Not sure how reliable the matches are, especially given that many descriptors work on gray-scaled images. I suggest to create a test on the resulting matches.
> > 
> > 
> > Thanks, these are valid concerns, but the goal is not for producing perfect matches, which requires a much more sophisticated approach and likely isn't even possible in the first place. It also depends on the specific feature detector and matcher implementation / parameters. As per my limited testing, the resulting images produce consistent feature detections/matches and reconstructions. If you have specific suggestions for how to make this better without boiling the ocean, then I am more than happy to improve upon the current implementation.
> 
> To improve matchability, does it make sense to use a coded pattern for the peak rather than use one single color? We can benefit from these gabor bases as in calibration and structured light, e.g., https://github.com/opencv/opencv_contrib/blob/4.x/modules/structured_light/src/sinusoidalpattern.cpp.
> 
> The code looks long and complicated, but with the help of ChatGPT i got something quite minimal (xD): https://chatgpt.com/share/690b6054-da6c-8010-83fa-2e75c93b6e4a
> 
> I assume classical feature descriptors would work very well on such patterns.

Well I realized that if we generate such patterns we would ideally also need to handle the homography, then it can get quite complicated.

ahojnnes:
Thanks, I think these are good ideas and an opportunity to improve upon what I hacked together in 1h here. I'll go ahead with what we have now, as it satisfies the use case of creating some simple end-to-end tests without real data (and having synthetic GT camera poses).

PR #3697, opened 2025-11-05
Create unit test for automatic reconstruction controller

ahojnnes:


src/colmap/controllers/CMakeLists.txt
```diff
@@ -59,6 +59,17 @@ COLMAP_ADD_LIBRARY(
         Boost::boost
 )
 
+COLMAP_ADD_TEST(
+    NAME automatic_reconstruction_test
+    SRCS automatic_reconstruction_test.cc
+    LINK_LIBS colmap_controllers
+)
+if(CUDA_ENABLED AND TESTS_ENABLED)
+    target_link_libraries(colmap_controllers_automatic_reconstruction_test
```
B1ueber2y: Dense and use_gpu are disabled in the test, so this block of code does not do anything?

ahojnnes: It needs to be linked anyway in the CUDA CI jobs, otherwise there are missing symbols during the linking stage.

ahojnnes: Maybe it is better to link CUDA into the library instead of the test, though...

B1ueber2y (approved):

PR #3701, opened 2025-11-06
Update to latest vcpkg

ahojnnes:


B1ueber2y:
Can we merge this only after the new release? Since pyceres release was already made with https://github.com/cvg/pyceres/pull/79, I am slightly afraid of new incompatibiliy issues.

ahojnnes:
> Can we merge this only after the new release? Since pyceres release was already made with https://github.com/cvg/pyceres/pull/79, I am slightly afraid of new incompatibiliy issues. 

Good callout. Didn't think about it.

PR #3714, opened 2025-11-10
Add CI workflows for tagging breaking changes. 

B1ueber2y:
### Does this PR introduce breaking changes to **colmap/pycolmap public APIs**?

> - [x] No — no breaking changes to colmap/pycolmap APIs.
> - [ ] Yes — this PR breaks colmap/pycolmap APIs.


B1ueber2y:
The label update wont work for PRs created from fork repo, and is thus not very practical. I suggest to tag "breaking" manually if we go with this PR. The label update is removed for now.

PR #3717, opened 2025-11-11
Normalize quaternions when reading from stream to ensure unit length.

B1ueber2y:
Found by @vlarsson


ahojnnes:
This makes the IO non idempotent. Why do we need this? I'd rather generate a warning when reading a model if a quaternion is not normalized?

PR #3740, opened 2025-11-20
Add unit tests for controller thread

ahojnnes:


src/colmap/util/controller_thread_test.cc
```diff
@@ -0,0 +1,225 @@
...
+  Barrier paused_barrier;
+  Barrier resumed_barrier;
+  Barrier end_barrier;
+
+  void Run() override {
+    start_barrier.Wait();
+    pause_barrier.Wait();
+    paused_barrier.Wait();
+    // This will block if paused
+    CheckIfStopped();
```
B1ueber2y: I dont think this line does what the comment says. And I am confused on what the logic here was supposed to achieve. 

My guess is that the original incentive here was to do paused_barrier.Wait() inside a CheckIfStopped if branch?

ahojnnes: I don't know what I wanted to do there :-) I removed this test.

PR #3741, opened 2025-11-20
Inline SQLite3 utils in database implementation

ahojnnes:
Only used in one place. No need to pollute the utils target with the SQLite3 dependency.

### DIFF
--- a/src/colmap/scene/database_sqlite.cc
+++ b/src/colmap/scene/database_sqlite.cc
@@ -75,7 +75,9 @@ struct Sqlite3StmtContext {
 };
 
 void SwapFeatureMatchesBlob(FeatureMatchesBlob* matches) {
-  matches->col(0).swap(matches->col(1));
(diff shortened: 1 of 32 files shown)


src/colmap/scene/database_sqlite.cc
```diff
@@ -29,12 +29,42 @@
...
+    case SQLITE_DONE:
+      return result_code;
+    default:
+      LogMessageFatalThrow<std::runtime_error>(filename.c_str(), line).stream()
+          << "SQLite error: " << sqlite3_errstr(result_code);
+      return result_code;
+  }
+}
+
+#define SQLITE3_CALL(func) colmap::SQLite3CallHelper(func, __FILE__, __LINE__)
```
B1ueber2y: If we make it inline, the ``colmap::`` namespace here is not necessary anymore.

PR #3744, opened 2025-11-21
Add drag-and-drop folder support for model import in the GUI

StonerLing:
Make frequent model visualization a bit faster and more convenient.


ahojnnes (approved):
Great improvement, thank you!

PR #3746, opened 2025-11-22
Update Windows build instructions for PyCOLMAP

JCRPaquin:
The current build instructions for PyCOLMAP don't work on Windows. This PR adds some additional steps and context that should resolve the problem.

### DIFF
--- a/python/README.md
+++ b/python/README.md
@@ -35,8 +35,7 @@ python -m pip install . `
     -C skbuild.cmake.define.VCPKG_TARGET_TRIPLET="x64-windows"
 ```
 
-> [!NOTE]
-> If you get linker errors when building PyCOLMAP on Windows, be sure that the repository version matches the version you installed via VCPKG.
+If you get linker errors when building PyCOLMAP on Windows, be sure that the repository version matches the version you installed via VCPKG.
 
 
 </details>


python/README.md
```diff
@@ -20,18 +20,24 @@ To benefit from GPU acceleration, wheels built for CUDA 12 (only for Linux - for
...
 1. Install COLMAP from source following [the official guide](https://colmap.github.io/install.html).
 
-3. Build PyCOLMAP:
+1. Build PyCOLMAP from the root of the repository:
   - On Linux and macOS:
 ```bash
 python -m pip install .
 ```
-  - On Windows, after installing COLMAP [via VCPKG](https://colmap.github.io/install.html#id3), run in powershell:
+  - On Windows, after installing COLMAP [via VCPKG](https://colmap.github.io/install.html#id3):
```
ahojnnes, line 28: nit: not your fault but this link is actually not persistent, and we should link using a markdown reference?

python/README.md
```diff
@@ -20,18 +20,24 @@ To benefit from GPU acceleration, wheels built for CUDA 12 (only for Linux - for
...
+    1. Determine the installed COLMAP version via: `<VCPKG-ROOT>\packages\colmap_<TRIPLET>\tools\colmap\colmap.exe help`
```
ahojnnes, line 29: nit: The easier path would be to just run vcpkg install from the root of the colmap repo. Then the version is guaranteed to be consistent.

JCRPaquin: Ah, noted. That does seem much easier. :)

I'll go verify, then make the edit.

python/README.md
JCRPaquin: It looks like the Windows source-build section anchor is named that way because the section header is a duplicate? Maybe it would be reasonable to add an explicitly named anchor.

Or the link could be removed entirely, maybe? Maybe change this to:
```suggestion
  - On Windows, after installing COLMAP via VCPKG per the installation instructions above:
```

JCRPaquin: ~~Created a PR to add a permanent section anchor: https://github.com/colmap/colmap.github.io/pull/19~~

Accidentally updated the mirror of the site instead of the actual docs. I'll add the anchor to this PR shortly.

sarlinpe:
Please also update the instructions at https://github.com/colmap/colmap/blob/main/doc/pycolmap/index.rst

python/README.md
JCRPaquin: @ahojnnes So I tried `vcpkg install` from the root of the repo and all it did was install the dependencies. Is there another command I should try?

I also tried doing a cmake build using the vcpkg toolchain file, then the pip command in the install docs + some changes to resolve pybind11. That built successfully, but was pretty convoluted and I ultimately can't get the resulting package to load because it fails to resolve DLL dependencies in the `vcpkg_installed` folder.

For what it's worth, I also couldn't load the resulting package when building the python bindings against the `vcpkg install colmap:x64-windows` version. It's the same DLL resolution issue caused by vcpkg-built shared libraries living in their own prefix.

JCRPaquin: Looks like there are 3 issues causing the loading problem:
1. vcpkg shared libraries can't be found
2. CUDA libraries can't be found
3. OpenEXR 3.4 is crashing during initialization (something about a null pointer dereference while locking a critical section)

I'll update the build instructions based on the method I originally proposed in my PR, then open a separate issue to track this loading issue.

ahojnnes:
Thanks for your contribution. Due to an upcoming major release, I took over and made the remaining changes myself: https://github.com/colmap/colmap/pull/4084

PR #3748, opened 2025-11-23
Import of GLOMAP source code

ahojnnes:
Disabled by default, passes all tests, follow-up with further refactoring steps.

### DIFF
--- a/src/glomap/controllers/global_mapper.cc
+++ b/src/glomap/controllers/global_mapper.cc
@@ -10,8 +10,8 @@
 #include "glomap/processors/track_filter.h"
 #include "glomap/processors/view_graph_manipulation.h"
 
-#include <colmap/util/file.h>
-#include <colmap/util/timer.h>
+#include "colmap/util/file.h"
+#include "colmap/util/timer.h"
 
 namespace glomap {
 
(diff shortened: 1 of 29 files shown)


CMakeLists.txt
```diff
@@ -257,6 +258,9 @@ include_directories(src)
 link_directories(${COLMAP_LINK_DIRS})
 
 add_subdirectory(src/colmap)
```
B1ueber2y, line 260: Realizing that, here by concept thirdparty should be put before colmap, although in the current design it does not matter since we dont define any target in the thirdparty.

src/glomap/CMakeLists.txt
```diff
@@ -0,0 +1,91 @@
...
+        PoseLib::PoseLib
+        Boost::boost
+)
+
+if(MSVC)
+    target_compile_options(colmap_glomap PRIVATE /bigobj)
+else()
+    target_compile_options(colmap_glomap PRIVATE
+        -Wall
+        -Wno-sign-compare
```
B1ueber2y, line 66: Disabling sign-compare and unused-variable is never a good practice? But I understand that this might be a temporary solution for the migration.

B1ueber2y (approved):

PR #3751, opened 2025-11-23
Remove custom glomap constants

ahojnnes:


src/glomap/math/two_view_geometry.cc
```diff
@@ -71,8 +71,9 @@ double SampsonError(const Eigen::Matrix3d& E,
 double SampsonError(const Eigen::Matrix3d& E,
                     const Eigen::Vector3d& x1,
                     const Eigen::Vector3d& x2) {
-  Eigen::Vector3d Ex1 = E * x1 / (EPS + x1[2]);
-  Eigen::Vector3d Etx2 = E.transpose() * x2 / (EPS + x2[2]);
+  constexpr double kEps = 1e-12;
```
B1ueber2y: These kEps in this function are not really meaningful and should be removed. The base values are not necessarily positive and this can even flip sign.

src/glomap/processors/image_pair_inliers.cc
```diff
@@ -100,7 +100,8 @@ double ImagePairInliers::ScoreErrorFundamental() {
 
   bool status = false;
   for (auto i = 0; i < 3; i++) {
-    if ((epipole(i) > EPS) || (epipole(i) < -EPS)) {
+    constexpr double kEps = 1e-12;
+    if ((epipole(i) > kEps) || (epipole(i) < -kEps)) {
```
B1ueber2y: nit: here one can do std::abs?

src/glomap/processors/track_filter.cc
```diff
@@ -17,7 +17,8 @@ int TrackFilter::FilterTracksByReprojection(
     for (auto& [image_id, feature_id] : track.observations) {
       const Image& image = images.at(image_id);
       Eigen::Vector3d pt_calc = image.CamFromWorld() * track.xyz;
-      if (pt_calc(2) < EPS) continue;
+      constexpr double kEps = 1e-12;
```
B1ueber2y, line 20: The kEps in this file also looks unnecessary. I guess one can compare with zero and remove kEps for the reprojection error.

src/glomap/math/two_view_geometry.cc
ahojnnes: Agreed, this doesn't make sense. The goal here was to refactor without change of behavior. The end goal is to replace this function with what we have in colmap/estimators/utils.cc.

src/glomap/processors/track_filter.cc
ahojnnes: Done

PR #3752, opened 2025-11-23
Add support for geometric verification given known reconstruction.

B1ueber2y:
This PR attempts to support geometric verification given a known colmap reconstruction, which was needed for the potential hloc migration. The following interface is supported:
```
colmap guided_geometric_verifier --input_path ${MODEL_PATH} --database_path ${DB_PATH} \
              --TwoViewGeometry.min_num_inliers 15 --TwoViewGeometry.max_error 4.0
```

To better achieve this, I also did some refactorization on the existing ``FeatureMatcherThread``. In particular, ``only_verification`` introduced in https://github.com/colmap/colmap/pull/3581 was only used in ``RunGeometricVerifier`` and was moved to a separate class ``FeatureMatchesVerifierThread`` (which does not require ``FeatureMatchingOptions``). Rig verification is supported in both classes. 

TODOs:

* [x] Add tests.
* [x] Add pybind.
* [x] Aligning with bidirectional epipolar errors in hloc and add some benchmarking.

### DIFF
--- a/src/colmap/exe/feature.cc
+++ b/src/colmap/exe/feature.cc
@@ -414,7 +414,7 @@ int RunGeometricVerifier(int argc, char** argv) {
 }
 
 void RunGuidedGeometricVerifierImpl(
-    const std::shared_ptr<Reconstruction>& reconstruction,
+    const Reconstruction& reconstruction,
     const std::string& database_path,
     const ExistingMatchedPairingOptions& pairing_options,
     const TwoViewGeometryOptions& geometry_options,
@@ -430,12 +430,12 @@ void RunGuidedGeometricVerifierImpl(
       continue;
     }
     const auto [image_id1, image_id2] = PairIdToImagePair(pair_id);
-    if (!reconstruction->ExistsImage(image_id1) ||
-        !reconstruction->ExistsImage(image_id2)) {
+    if (!reconstruction.ExistsImage(image_id1) ||
+        !reconstruction.ExistsImage(image_id2)) {
       continue;
     }
-    Image& image1 = reconstruction->Image(image_id1);
-    Image& image2 = reconstruction->Image(image_id2);
+    const Image& image1 = reconstruction.Image(image_id1);
+    const Image& image2 = reconstruction.Image(image_id2);
     if (!image1.HasPose() || !image2.HasPose()) {
       continue;
     }
@@ -481,8 +481,8 @@ int RunGuidedGeometricVerifier(int argc, char** argv) {
     return EXIT_FAILURE;
   }
 
-  auto reconstruction = std::make_shared<Reconstruction>();
-  reconstruction->Read(input_path);
+  Reconstruction reconstruction;
+  reconstruction.Read(input_path);
 
   RunGuidedGeometricVerifierImpl(reconstruction,
                                  *options.database_path,
--- a/src/colmap/exe/feature.h
+++ b/src/colmap/exe/feature.h
@@ -78,7 +78,7 @@ bool VerifyCameraParams(const std::string& camera_model,
                         const std::string& params);
 
 void RunGuidedGeometricVerifierImpl(
-    const std::shared_ptr<Reconstruction>& reconstruction,
+    const Reconstruction& reconstruction,
     const std::string& database_path,
     const ExistingMatchedPairingOptions& pairing_options,
     const TwoViewGeometryOptions& geometry_options,
--- a/src/pycolmap/pipeline/match_features.cc
+++ b/src/pycolmap/pipeline/match_features.cc
@@ -83,7 +83,7 @@ void GeometricVerification(const std::string& database_path,
 }
 
 void GuidedGeometricVerification(
-    const std::shared_ptr<Reconstruction>& reconstruction,
+    const Reconstruction& reconstruction,
     const std::string& database_path,
     const ExistingMatchedPairingOptions& pairing_options,
     const TwoViewGeometryOptions& geometry_options,


src/colmap/estimators/two_view_geometry.cc
```diff
@@ -830,4 +830,72 @@ void FilterStationaryMatches(double max_error,
...
+        cam_point2) {
+      matched_cam_rays2[i] = cam_point2->homogeneous().normalized();
+    } else {
+      matched_cam_rays2[i].setZero();
+    }
+  }
+
+  // For now, we use the average threshold from cameras following the design of
+  // EstimateCalibratedTwoViewGeometry.
+  double max_error_in_cam = (camera1.CamFromImgThreshold(max_error) +
```
B1ueber2y: Here I followed the convention used in ``EstimateCalibratedTwoViewGeometry``, which is different from the hloc implementation that uses bidirectional epipolar errors with keypoint noises included: https://github.com/cvg/Hierarchical-Localization/blob/master/hloc/triangulation.py#L146-L153

src/colmap/controllers/feature_matching.h
```diff
@@ -153,11 +153,18 @@ std::unique_ptr<Thread> CreateFeaturePairsFeatureMatcher(
...
+};
+
 // Perform geometric verification of existing matched image pairs.
 std::unique_ptr<Thread> CreateGeometricVerifier(
     const ExistingMatchedPairingOptions& pairing_options,
-    const FeatureMatchingOptions& matching_options,
     const TwoViewGeometryOptions& geometry_options,
-    const std::string& database_path);
+    const std::string& database_path,
+    const VerifierOptions& verifier_options);
```
ahojnnes: nit: swap to have all options first.

src/colmap/controllers/feature_matching.h
```diff
@@ -153,11 +153,18 @@ std::unique_ptr<Thread> CreateFeaturePairsFeatureMatcher(
     const TwoViewGeometryOptions& geometry_options,
     const std::string& database_path);
 
+// Verifier options for CreateGeometricVerifier
+struct VerifierOptions {
```
ahojnnes: nit: GeometricVerifierOptions?

src/colmap/controllers/feature_matching_utils.cc
```diff
@@ -224,19 +242,18 @@ class VerifierWorker : public Thread {
  private:
   const TwoViewGeometryOptions options_;
   std::shared_ptr<FeatureMatcherCache> cache_;
+  bool use_existing_relative_pose_;
```
ahojnnes: const

src/colmap/controllers/feature_matching_utils.cc
```diff
@@ -213,8 +215,24 @@ class VerifierWorker : public Thread {
         const std::vector<Eigen::Vector2d> points2 =
             FeatureKeypointsToPointsVector(*keypoints2);
 
-        data.two_view_geometry = EstimateTwoViewGeometry(
-            camera1, points1, camera2, points2, data.matches, options_);
+        if (use_existing_relative_pose_ &&
+            data.two_view_geometry.config !=
+                TwoViewGeometry::ConfigurationType::DEGENERATE &&
+            data.two_view_geometry.config !=
+                TwoViewGeometry::ConfigurationType::UNDEFINED) {
```
ahojnnes, line 226: Also skip when MULTIPLE, WATERMARK?

src/colmap/controllers/feature_matching_utils.cc
```diff
@@ -492,9 +499,138 @@ void FeatureMatcherController::Match(
...
-      cache_->WriteMatches(output.image_id1, output.image_id2, output.matches);
+    cache_->WriteMatches(output.image_id1, output.image_id2, output.matches);
+    cache_->WriteTwoViewGeometry(
+        output.image_id1, output.image_id2, output.two_view_geometry);
+  }
+
+  THROW_CHECK_EQ(output_queue_.Size(), 0);
+}
+
+FeatureMatchesVerifierController::FeatureMatchesVerifierController(
```
ahojnnes: GeometricVerifierController?

src/colmap/controllers/feature_matching_utils.h
```diff
@@ -108,4 +107,44 @@ class FeatureMatcherController {
...
 
+class FeatureMatchesVerifierController {
+ public:
+  FeatureMatchesVerifierController(
+      const TwoViewGeometryOptions& geometry_options,
+      std::shared_ptr<FeatureMatcherCache> cache,
+      const VerifierOptions& options);
+
+  inline const VerifierOptions& Options() const;
+  inline VerifierOptions& Options();
```
ahojnnes: Not sure if worth inlining? Would just move it to the cc file.

src/colmap/estimators/two_view_geometry.cc
```diff
@@ -830,4 +830,72 @@ void FilterStationaryMatches(double max_error,
...
+                             camera2.CamFromImgThreshold(max_error)) /
+                            2;
+
+  Eigen::Matrix3d E = EssentialMatrixFromPose(cam2_from_cam1);
+  std::vector<double> residuals(matches.size());
+  EssentialMatrixFivePointEstimator::Residuals(
```
sarlinpe: Use ComputeSquaredSampsonError instead?
```suggestion
  ComputeSquaredSampsonError(
```

src/colmap/estimators/two_view_geometry.cc
sarlinpe: This is a problem when the two cameras have very different focal lengths.

src/colmap/exe/feature.cc
```diff
@@ -390,22 +390,89 @@ int RunVocabTreeMatcher(int argc, char** argv) {
...
+    Image& image2 = reconstruction->Image(image_id2);
+    if (!image1.HasPose() || !image2.HasPose()) {
+      database->WriteTwoViewGeometry(image_id1, image_id2, two_view_geometry);
+      continue;
+    }
+    const Rigid3d cam1_from_world = image1.CamFromWorld();
+    const Rigid3d cam2_from_world = image2.CamFromWorld();
+
+    TwoViewGeometry two_view_geometry_copy = two_view_geometry;
+    two_view_geometry_copy.cam2_from_cam1 =
```
sarlinpe: Shouldn't you clear out the inlier matches here, otherwise the verifier isn't ran again?

sarlinpe (commented):
Thank you! Could you please run some benchmark to check that this works as expected? Mihai had ran Aachen https://github.com/cvg/Hierarchical-Localization/pull/178

src/colmap/exe/feature.cc
B1ueber2y: Nice catch. However, realizing that this was not possible without https://github.com/colmap/colmap/pull/3754.

B1ueber2y:
> Thank you! Could you please run some benchmark to check that this works as expected? Mihai had ran Aachen [cvg/Hierarchical-Localization#178](https://github.com/cvg/Hierarchical-Localization/pull/178)

Here are some numbers on ETH3D triangulation (averaged over all 13 DSLR scenes) with aliked + lightglue + netvlad top 5 retrieval:

```
aliked n-16 & aliked+lightglue hloc no geometric verification
Metric          |    1cm (%) |    2cm (%) |    5cm (%)
-------------------------------------------------------
Completeness    |       0.27 |       1.15 |       5.43
Accuracy        |      57.31 |      70.57 |      85.32

aliked n-16 & aliked+lightglue hloc estimation and geometric verification
Metric          |    1cm (%) |    2cm (%) |    5cm (%)
-------------------------------------------------------
Completeness    |       0.28 |       1.21 |       5.52
Accuracy        |      57.70 |      71.05 |      85.71

aliked n-16 & aliked+lightglue hloc guided geometric verification
Metric          |    1cm (%) |    2cm (%) |    5cm (%)
-------------------------------------------------------
Completeness    |       0.27 |       1.16 |       5.41
Accuracy        |      57.70 |      71.11 |      85.70

aliked n-16 & aliked+lightglue pycolmap guided geometric verification
Metric          |    1cm (%) |    2cm (%) |    5cm (%)
-------------------------------------------------------
Completeness    |       0.28 |       1.20 |       5.52
Accuracy        |      57.73 |      71.16 |      85.69
```

The pycolmap implementation was using the new interface ``pycolmap.guided_geometric_verification(reference, database)``. Relevant code block: https://github.com/cvg/Hierarchical-Localization/blob/master/hloc/triangulation.py#L223-L230. 

Edit: the "pycolmap guided geometric verification" result is broken and fixed below.

sarlinpe:
Thank you! This looks sufficient, though looking at the number of 3D points and track length might be more infomative (especially for scenes with symmetries = more incorrect matches).

One thing: it looks like this requires existing two-view geometries as input, so one still needs to run the RANSAC verification even though it's not needed, right? should we use the raw matches as input instead, such that we can skip the RANSAC verification? (like hloc does).

B1ueber2y:
> Thank you! This looks sufficient, though looking at the number of 3D points and track length might be more infomative (especially for scenes with symmetries = more incorrect matches).
> 
> One thing: it looks like this requires existing two-view geometries as input, so one still needs to run the RANSAC verification even though it's not needed, right? should we use the raw matches as input instead, such that we can skip the RANSAC verification? (like hloc does).

Amazing catch! Indeed, the previous implementation relies on existing two view geometries and i have updated it.  Actually you caught an undefined behavior in the previous test caused by https://github.com/colmap/colmap/pull/3761. Because of this the previous benchmarking actually falls back to geometric verification with ransac so does not really work. 

Here are the new numbers that make more sense (the hloc results are not affected):
```
aliked n-16 & aliked+lightglue hloc no geometric verification (deterministic)
Metric          |    1cm (%) |    2cm (%) |    5cm (%)
-------------------------------------------------------
Completeness    |       0.27 |       1.15 |       5.43
Accuracy        |      57.31 |      70.57 |      85.32

aliked n-16 & aliked+lightglue hloc estimation and geometric verification (run 1)
Metric          |    1cm (%) |    2cm (%) |    5cm (%)
-------------------------------------------------------
Completeness    |       0.28 |       1.21 |       5.52
Accuracy        |      57.70 |      71.05 |      85.71

aliked n-16 & aliked+lightglue hloc estimation and geometric verification (run 2)
Metric          |    1cm (%) |    2cm (%) |    5cm (%)
Completeness    |       0.28 |       1.21 |       5.51
Accuracy        |      57.84 |      71.17 |      85.78

aliked n-16 & aliked+lightglue hloc estimation and geometric verification (run 3)
Metric          |    1cm (%) |    2cm (%) |    5cm (%)
Completeness    |       0.28 |       1.20 |       5.51
Accuracy        |      57.70 |      71.16 |      85.72

aliked n-16 & aliked+lightglue hloc guided geometric verification (deterministic)
Metric          |    1cm (%) |    2cm (%) |    5cm (%)
-------------------------------------------------------
Completeness    |       0.27 |       1.16 |       5.41
Accuracy        |      57.70 |      71.11 |      85.70

aliked n-16 & aliked+lightglue pycolmap guided geometric verification (after the fix, deterministic)
Metric          |    1cm (%) |    2cm (%) |    5cm (%)
-------------------------------------------------------
Completeness    |       0.29 |       1.22 |       5.58
Accuracy        |      57.95 |      71.53 |      85.93
```

So the geometric verification with known poses indeed consistently improves the results. This makes sense to me since we are using exactly the same scoring function as the original geometric verification with ransac. 

I rerun the hloc numbers to make sure that they were not affected by this fix, and got the same numbers except for "estimation and geometric verification" due to its non-deterministic behavior (i added results from 3 different runs). Note that the original hloc implementation of geometric verification with known poses does not work as well by default on ETH3D. 

This is due to its overestimated noise (~6 pixels) for aliked keypoints (COLMAP always assumes unit pixel noise), resulting in its adaptive threshold being too large. 

```
aliked n-16 & aliked+lightglue hloc guided geometric verification (noise = 1.0px, max_error = 4.0, deterministic)
Metric          |    1cm (%) |    2cm (%) |    5cm (%)
-------------------------------------------------------
Completeness    |       0.27 |       1.16 |       5.31
Accuracy        |      58.97 |      72.85 |      86.73

aliked n-16 & aliked+lightglue hloc guided geometric verification (noise = 1.5px, max_error = 4.0, deterministic)
Metric          |    1cm (%) |    2cm (%) |    5cm (%)
-------------------------------------------------------
Completeness    |       0.29 |       1.22 |       5.53
Accuracy        |      58.26 |      71.92 |      86.15

aliked n-16 & aliked+lightglue hloc guided geometric verification (noise = 2.0px, max_error = 4.0, deterministic)
Metric          |    1cm (%) |    2cm (%) |    5cm (%)
-------------------------------------------------------
Completeness    |       0.29 |       1.22 |       5.58
Accuracy        |      57.57 |      71.16 |      85.70

aliked n-16 & aliked+lightglue hloc guided geometric verification (noise = 3.0px, max_error = 4.0, deterministic)
Metric          |    1cm (%) |    2cm (%) |    5cm (%)
-------------------------------------------------------
Completeness    |       0.28 |       1.20 |       5.53
Accuracy        |      57.70 |      71.16 |      85.65

aliked n-16 & aliked+lightglue hloc guided geometric verification (default (noise = ~6px), max_error = 4.0, deterministic)
Metric          |    1cm (%) |    2cm (%) |    5cm (%)
-------------------------------------------------------
Completeness    |       0.27 |       1.16 |       5.41
Accuracy        |      57.70 |      71.11 |      85.70
```

Comparing with the curve of pycolmap guided geometric verification below.
```
aliked n-16 & aliked+lightglue pycolmap guided geometric verification (max_error = 2.0, deterministic)
Metric          |    1cm (%) |    2cm (%) |    5cm (%)
-------------------------------------------------------
Completeness    |       0.27 |       1.12 |       5.17
Accuracy        |      59.27 |      73.23 |      87.09

aliked n-16 & aliked+lightglue pycolmap guided geometric verification (max_error = 3.0, deterministic)
Metric          |    1cm (%) |    2cm (%) |    5cm (%)
-------------------------------------------------------
Completeness    |       0.28 |       1.20 |       5.48
Accuracy        |      58.57 |      72.25 |      86.46

aliked n-16 & aliked+lightglue pycolmap guided geometric verification (default, max_error = 4.0, deterministic)
Metric          |    1cm (%) |    2cm (%) |    5cm (%)
-------------------------------------------------------
Completeness    |       0.29 |       1.22 |       5.58
Accuracy        |      57.95 |      71.53 |      85.93

aliked n-16 & aliked+lightglue pycolmap guided geometric verification (max_error = 5.0, deterministic)
Metric          |    1cm (%) |    2cm (%) |    5cm (%)
-------------------------------------------------------
Completeness    |       0.29 |       1.22 |       5.58
Accuracy        |      57.60 |      71.17 |      85.67
```

sarlinpe:
Thank you for running the tests, so the two approaches are indeed mostly equivalent, at least when all images have the same intrinsics (which holds on ETH3D). We don't yet have an option to skip RANSAC verification in the matching stage, right?

B1ueber2y:
> Thank you for running the tests, so the two approaches are indeed mostly equivalent, at least when all images have the same intrinsics (which holds on ETH3D). We don't yet have an option to skip RANSAC verification in the matching stage, right?

In HLoc we have such option, but we do not have it in the colmap native feature matching afaik.

src/colmap/exe/feature.cc
```diff
@@ -390,22 +390,105 @@ int RunVocabTreeMatcher(int argc, char** argv) {
...
-                                         *options.database_path);
-  matcher->Start();
-  matcher->Wait();
+  if (!ExistsDir(input_path)) {
+    LOG(ERROR) << "`input_path` is not a directory";
+    return EXIT_FAILURE;
+  }
+
+  auto reconstruction = std::make_shared<Reconstruction>();
+  reconstruction->Read(input_path);
```
ahojnnes: No need for shared_ptr here?

src/colmap/estimators/two_view_geometry.cc
```diff
@@ -830,4 +831,74 @@ void FilterStationaryMatches(double max_error,
...
+    const std::vector<Eigen::Vector2d>& points1,
+    const Camera& camera2,
+    const std::vector<Eigen::Vector2d>& points2,
+    const Rigid3d& cam2_from_cam1,
+    const FeatureMatches& matches,
+    int min_num_inliers,
+    double max_error) {
+  TwoViewGeometry geometry;
+
+  if (matches.size() < static_cast<size_t>(min_num_inliers)) {
```
ahojnnes: nit: could extract `const size_t num_matches = matches.size()` and reuse here and below.

src/colmap/estimators/two_view_geometry.cc
```diff
@@ -830,4 +831,74 @@ void FilterStationaryMatches(double max_error,
...
+        cam_point2) {
+      matched_cam_rays2[i] = cam_point2->homogeneous().normalized();
+    } else {
+      matched_cam_rays2[i].setZero();
+    }
+  }
+
+  // For now, we use the average threshold from cameras following the design of
+  // EstimateCalibratedTwoViewGeometry.
+  double max_error_in_cam = (camera1.CamFromImgThreshold(max_error) +
```
ahojnnes:
```suggestion
  const double max_error_in_cam = (camera1.CamFromImgThreshold(max_error) +
```

src/colmap/estimators/two_view_geometry.cc
```diff
@@ -830,4 +831,74 @@ void FilterStationaryMatches(double max_error,
...
+                             camera2.CamFromImgThreshold(max_error)) /
+                            2;
+
+  Eigen::Matrix3d E = EssentialMatrixFromPose(cam2_from_cam1);
+  std::vector<double> residuals(matches.size());
+  ComputeSquaredSampsonError(
+      matched_cam_rays1, matched_cam_rays2, E, &residuals);
+  FeatureMatches inlier_matches;
+  double squared_max_error_in_cam = max_error_in_cam * max_error_in_cam;
```
ahojnnes:
```suggestion
  const double squared_max_error_in_cam = max_error_in_cam * max_error_in_cam;
```

src/colmap/estimators/two_view_geometry.cc
```diff
@@ -830,4 +831,74 @@ void FilterStationaryMatches(double max_error,
...
+    }
+  }
+
+  // For now, we use the average threshold from cameras following the design of
+  // EstimateCalibratedTwoViewGeometry.
+  double max_error_in_cam = (camera1.CamFromImgThreshold(max_error) +
+                             camera2.CamFromImgThreshold(max_error)) /
+                            2;
+
+  Eigen::Matrix3d E = EssentialMatrixFromPose(cam2_from_cam1);
```
ahojnnes:
```suggestion
  const Eigen::Matrix3d E = EssentialMatrixFromPose(cam2_from_cam1);
```

src/colmap/estimators/two_view_geometry.cc
```diff
@@ -830,4 +831,74 @@ void FilterStationaryMatches(double max_error,
...
+  TwoViewGeometry geometry;
+
+  if (matches.size() < static_cast<size_t>(min_num_inliers)) {
+    geometry.config = TwoViewGeometry::ConfigurationType::DEGENERATE;
+    return geometry;
+  }
+
+  // Extract corresponding points.
+  std::vector<Eigen::Vector2d> matched_img_points1(matches.size());
+  std::vector<Eigen::Vector2d> matched_img_points2(matches.size());
```
ahojnnes: These are unused? Were they meant to be used in the calibrated case with F? I guess we could have an if-else branch here and compute the sampson error in normalized vs. pixel space?

src/colmap/estimators/two_view_geometry.cc
```diff
@@ -830,4 +831,74 @@ void FilterStationaryMatches(double max_error,
...
 
+TwoViewGeometry TwoViewGeometryFromKnownRelativePose(
+    const Camera& camera1,
+    const std::vector<Eigen::Vector2d>& points1,
+    const Camera& camera2,
+    const std::vector<Eigen::Vector2d>& points2,
+    const Rigid3d& cam2_from_cam1,
+    const FeatureMatches& matches,
+    int min_num_inliers,
+    double max_error) {
```
ahojnnes, line 842: THROW_CHECK on these two parameters?

src/colmap/controllers/feature_matching_test.cc
```diff
@@ -317,14 +317,13 @@ TEST(CreateGeometricVerifier, Nominal) {
 
   ExistingMatchedPairingOptions pairing_options;
 
-  FeatureMatchingOptions matching_options;
-  matching_options.use_gpu = false;
-  matching_options.num_threads = 1;
+  GeometricVerifierOptions verifier_options;
```
ahojnnes, line 320: Can we add a test that checks the new functionality?

src/colmap/controllers/feature_matching.h
```diff
@@ -153,10 +153,17 @@ std::unique_ptr<Thread> CreateFeaturePairsFeatureMatcher(
     const TwoViewGeometryOptions& geometry_options,
     const std::string& database_path);
 
+// Options for CreateGeometricVerifier
+struct GeometricVerifierOptions {
+  int num_threads = -1;
+  bool rig_verification = false;
+  bool use_existing_relative_pose = false;
```
ahojnnes, line 168: Short comments on what these settings to will be great.

src/colmap/controllers/feature_matching_test.cc
B1ueber2y: Done.

src/colmap/estimators/two_view_geometry.cc
B1ueber2y: I removed these variables, since the sampson error in pixel space cannot handle radial distortion. I think it might not be practical in this case, where we have prior intrinsics and poses.

src/colmap/exe/feature.cc
B1ueber2y: The use of shared_ptr is mainly for the pybind interface, similar to ``RunPointTriangulatorImpl``.

B1ueber2y:
This is ready to merge from my side (feel free to go ahead). I can add the ``skip_geometric_verification`` option for feature matching later in a separate PR.

src/colmap/exe/feature.cc
ahojnnes: Doesn't pybind11 3.X solve this automatically with smart holders? I thought it should be able to seamlessly convert between different "holders"?

B1ueber2y: Good point. This is updated : )

PR #3754, opened 2025-11-24
Support two view geometry with empty inlier matches in the database.

B1ueber2y:
The previous naming of ``ExistsInlierMatches`` and ``DeleteInlierMatches`` are confusing as it actually check directly the two_view_geometry. 

Also added an explicit check on whether the two_view_geometry already exists at ``WriteTwoViewGeometry``.

### DIFF
--- a/src/pycolmap/scene/database.cc
+++ b/src/pycolmap/scene/database.cc
@@ -335,10 +335,16 @@ class PyDatabaseImpl : public Database, py::trampoline_self_life_support {
     PYBIND11_OVERRIDE_PURE(void, Database, UpdateKeypoints, image_id, blob);
   }
 
-  void UpdateTwoViewGeometry(image_t image_id,
-                             const FeatureKeypointsBlob& blob) override {
-    PYBIND11_OVERRIDE_PURE(
-        void, Database, UpdateTwoViewGeometry, image_id, blob);
+  void UpdateTwoViewGeometry(
+      image_t image_id1,
+      image_t image_id2,
+      const TwoViewGeometry& two_view_geometry) override {
+    PYBIND11_OVERRIDE_PURE(void,
+                           Database,
+                           UpdateTwoViewGeometry,
+                           image_id1,
+                           image_id2,
+                           two_view_geometry);
   }
 
   void DeleteMatches(image_t image_id1, image_t image_id2) override {


src/colmap/scene/database_sqlite.cc
```diff
@@ -1369,6 +1381,17 @@ class SqliteDatabase : public Database {
...
 
+  void DeleteInlierMatches(const image_t image_id1,
+                           const image_t image_id2) override {
+    if (!ExistsTwoViewGeometry(image_id1, image_id2)) {
+      return;
+    }
+    TwoViewGeometry geom = ReadTwoViewGeometry(image_id1, image_id2);
+    geom.inlier_matches.clear();
+    DeleteTwoViewGeometry(image_id1, image_id2);
+    WriteTwoViewGeometry(image_id1, image_id2, geom);
```
sarlinpe: Do we want to instead add a method UpdateTwoViewGeometry?

B1ueber2y:
Note that this PR potentially breaks the backwards compatibility of pycolmap, where the previous ``db.exists_inlier_matches`` and ``db.delete_inlier_matches`` would need to change to ``db.exists_two_view_geometry`` and ``db.delete_two_view_geometry``. The changing of deletion behavior is significant since ``db.delete_inlier_matches`` will no longer delete the two view geometry object.

Nonetheless, this interface was not used anywhere on github: https://github.com/search?q=delete_inlier_matches&type=code.

src/colmap/scene/database_sqlite.cc
```diff
@@ -541,12 +541,21 @@ class SqliteDatabase : public Database {
...
 
-  bool ExistsInlierMatches(const image_t image_id1,
-                           const image_t image_id2) const override {
+  bool ExistsTwoViewGeometry(const image_t image_id1,
+                             const image_t image_id2) const override {
     return ExistsRowId(sql_stmt_exists_two_view_geometry_,
                        ImagePairToPairId(image_id1, image_id2));
   }
 
+  bool ExistsInlierMatches(const image_t image_id1,
```
ahojnnes: Is this used anywhere?

ahojnnes: If not, I suggest to just rename and not keep this functionality?

B1ueber2y: No. But this is useful in the other pending PR.

B1ueber2y: Okay this logic can also be done outside. Let me remove it.

B1ueber2y: Done.

src/colmap/scene/database_sqlite.cc
B1ueber2y: Done.

PR #3762, opened 2025-11-25
Replace custom glomap Track with colmap Point3D

ahojnnes:
The Track struct is pretty much identical to colmap's Point3D struct.


src/glomap/scene/types.h
```diff
@@ -29,15 +29,13 @@ using colmap::frame_t;
...
 
 // Each image pair gets a unique ID, see `Database::ImagePairToPairId`.
-typedef uint64_t image_pair_t;
-
-// Index per image, i.e. determines maximum number of 2D points per image.
-typedef uint32_t feature_t;
+using colmap::image_pair_t;
+using colmap::kInvalidImagePairId;
 
 // Unique identifier per added 3D point. Since we add many 3D points,
```
B1ueber2y: Update (or remove) this comment?

B1ueber2y: Clean up this comment?

src/glomap/estimators/global_positioning.cc
```diff
@@ -256,49 +257,49 @@ void GlobalPositioner::AddPointToCameraConstraints(
...
     const Eigen::Vector3d translation =
         image.CamFromWorld().rotation.inverse() *
-        image.features_undist[observation.second];
+        image.features_undist[observation.point2D_idx];
 
     double& scale = scales_.emplace_back(1);
 
-    if (!options_.generate_scales && tracks[track_id].is_initialized) {
+    if (!options_.generate_scales &&
+        tracks[track_id].xyz != Eigen::Vector3d::Zero()) {
```
B1ueber2y: Small risk of changing behavior. I assume the global positioning can work with some points initialized at origin.

B1ueber2y: I do not have a better idea to cleanly address this without adding an additional field to Point3D.

ahojnnes: I guess we could check whether track length is zero? A 3d point won't be initialized unless it has observations?

lpanaf: I think this is not a viable solution? The is_initialized flag is used for the global positioning step, so the track should not have 0 length. Potentially, we can initialize it to be NaN? I used this for the rigged rotation averaging.

src/glomap/io/colmap_converter.cc
```diff
@@ -81,17 +82,17 @@ void ConvertGlomapToColmap(
     colmap::Point3D colmap_point;
     colmap_point.xyz = track.xyz;
     colmap_point.color = track.color;
-    colmap_point.error = 0;
+    colmap_point.error = track.error;
```
lpanaf, line 85: If the track got editted, maybe the error need to be recomputed.

src/glomap/estimators/global_positioning.cc
lpanaf: Emm, thinking again about the flag, maybe this is also not a good idea. This was initially to distinguish points where the initial positions are estimated. I guess from random initialization, the xyz would not be NaN anymore.

B1ueber2y: Can we add an additional argument ``tracks_randomly_initialized`` (equal to ``options_.optimize_points && options_.generate_random_points``) to this ``AddTrackToProblem``and just check whether the track length >= min_num_view_per_track? This should be sufficient to keep the logic unchanged.

B1ueber2y: Actually, we can also just do the same logic again in the function, and check ``options_.optimize_points && options_.generate_random_points && track.observations.size() >= options_.min_num_view_per_track``. No need to add any new argument.

ahojnnes: I went with a variant of @B1ueber2y's proposal.

src/glomap/io/colmap_converter.cc
ahojnnes: Sure, but this is a simple conversion function that should not tamper with what is stored in the reconstruction inputs. We should address this wherever we think that the error should be updated instead.

PR #3769, opened 2025-11-26
Allow running glomap from colmap main executable

ahojnnes:
@lpanaf What is the purpose of the global_mapper_resume command? Is the main purpose to use camera calibrations and colmap as a pose prior/initialization?


lpanaf:
The major purpose is to be able to run only global positioning or only BA, or other steps when some steps failed in the main mapping function. Potentially, it can be replaced by several sub functions which only run a single step.

PR #3773, opened 2025-11-27
Add utility function to compute Rigid3d origin

ahojnnes:


src/glomap/estimators/global_positioning.cc
```diff
@@ -576,8 +570,7 @@ void GlobalPositioner::ConvertResults(
     std::unordered_map<frame_t, Frame>& frames) {
   // translation now stores the camera position, needs to convert back
   for (auto& [frame_id, frame] : frames) {
-    frame.RigFromWorld().translation =
-        frame.RigFromWorld().rotation * -frame.RigFromWorld().translation;
+    frame.RigFromWorld().translation = OriginBInA(frame.RigFromWorld());
```
ahojnnes: I am a bit confused. The previous code was not taking the inverse of the rotation here. Was this a mistake? @lpanaf

src/glomap/estimators/global_positioning.cc
```diff
@@ -349,8 +345,6 @@ void GlobalPositioner::AddTrackToProblem(
 
       if (!cam_from_rig_translation.hasNaN()) {
         const Eigen::Vector3d translation_rig =
-            // image.cam_from_world.rotation.inverse() *
-            // cam_from_rig.translation;
             image.CamFromWorld().rotation.inverse() * cam_from_rig_translation;
```
ahojnnes, line 349: @lpanaf Is this right? Is there a "-" missing on the translation?

src/colmap/geometry/rigid3.h
```diff
@@ -98,6 +98,10 @@ inline Rigid3d Inverse(const Rigid3d& b_from_a) {
   return a_from_b;
 }
 
+inline Eigen::Vector3d OriginBInA(const Rigid3d& b_from_a) {
```
B1ueber2y: There is no sense of A and B in Rigid3d. How about ``RefOriginInSource()`` or ``OriginRefInSource()``?

B1ueber2y: Or, trying to be less ambiguous: TranslationOfInverse()?

ahojnnes: The input transformation is called "b_from_a" ?

ahojnnes: TgtOriginInSrc() Is also fine with me.

B1ueber2y: But when using it as ``OriginalBInA(cam_from_world)``, it s unclear which is A and which is B. Its a bit unideal to refer to the variable names of header definitions. Maybe this is overthinking : )

src/colmap/geometry/rigid3_test.cc
```diff
@@ -83,6 +83,12 @@ TEST(Rigid3d, Inverse) {
   }
 }
 
+TEST(Rigid3d, TgtOriginInSrc) {
+  const Rigid3d b_from_a = TestRigid3d();
+  const Eigen::Vector3d origin_in_b = TgtOriginInSrc(b_from_a);
+  EXPECT_LT((b_from_a * origin_in_b - Eigen::Vector3d::Zero()).norm(), 1e-6);
```
sarlinpe: Shouldn't this be origin_in_a?
```suggestion
  const Eigen::Vector3d origin_in_a = TgtOriginInSrc(b_from_a);
  EXPECT_LT((b_from_a * origin_in_a - Eigen::Vector3d::Zero()).norm(), 1e-6);
```

sarlinpe (approved):
Thank you!

src/colmap/geometry/rigid3_test.cc
ahojnnes: origin_b_in_a

src/glomap/estimators/global_positioning.cc
lpanaf: I'm not sure I understand your questions fully. The translation here is actually storing the centers of the camera, so t = -Rc, and it should be correct?

src/glomap/estimators/global_positioning.cc
lpanaf: The minus sign is not needed here. image

src/colmap/geometry/rigid3.h
```diff
@@ -98,6 +98,10 @@ inline Rigid3d Inverse(const Rigid3d& b_from_a) {
   return a_from_b;
 }
 
+inline Eigen::Vector3d TgtOriginInSrc(const Rigid3d& b_from_a) {
+  return b_from_a.rotation.inverse() * -b_from_a.translation;
```
lpanaf: I kind of agree that the naming is not very understandable. Would something like poseCenter more understandable?

src/glomap/estimators/global_positioning.cc
ahojnnes: Yes, but before this change, it was `R * -t` and not `R^T * -t` ?

ahojnnes: OK, looking at it again, it was correct. We only updated the translation and the rotation stays the original rig_from_world transformation.

src/colmap/geometry/rigid3.h
ahojnnes: I think it's quite clear now with the latest name change.

src/pycolmap/geometry/rigid3.cc
```diff
@@ -39,6 +39,8 @@ void BindRigid3(py::module& m) {
                     t.translation.transpose();
            })
       .def("inverse", static_cast<Rigid3d (*)(const Rigid3d&)>(&Inverse))
+      .def("tgt_origin_in_src",
+           static_cast<Rigid3d (*)(const Rigid3d&)>(&TgtOriginInSrc))
```
sarlinpe:
```suggestion
           static_cast<Eigen::Vector3d (*)(const Rigid3d&)>(&TgtOriginInSrc))
```

src/colmap/geometry/rigid3.h
```diff
@@ -98,6 +98,11 @@ inline Rigid3d Inverse(const Rigid3d& b_from_a) {
   return a_from_b;
 }
 
+// Return the origin position of the target frame in the source frame.
+inline Eigen::Vector3d TgtOriginInSrc(const Rigid3d& b_from_a) {
```
sarlinpe: Why not make this a class method? Having a regular function makes the binding code more complex, what is the rationale behind it?

PR #3776, opened 2025-11-27
Add convenience method for filtering data by sensor type

ahojnnes:
Make it easier to filter by IMUs or PosePriors.


src/pycolmap/scene/frame.cc
```diff
@@ -39,6 +39,14 @@ void BindFrame(py::module& m) {
           "data_ids",
           [](const Frame& self) { return self.DataIds(); },
           "The associated data.")
+      .def_property_readonly(
```
B1ueber2y: Should just be ".def" rather than property.

PR #3777, opened 2025-11-27
Associate pose priors to generic sensor measurement data

ahojnnes:


src/colmap/util/types.h
```diff
@@ -154,17 +162,18 @@ struct sensor_t {
 };
 
 constexpr sensor_t kInvalidSensorId =
-    sensor_t(SensorType::INVALID, std::numeric_limits<uint32_t>::max());
+    sensor_t(SensorType::INVALID, sensor_t::kInvalidId);
 
 struct data_t {
+  constexpr static uint32_t kInvalidId = std::numeric_limits<uint32_t>::max();
```
B1ueber2y, line 166: Should bind this out as well.

src/pycolmap/pipeline/images.cc
```diff
@@ -59,12 +59,16 @@ void ImportImages(const std::string& database_path,
...
-        database->WritePosePrior(image.ImageId(), pose_prior);
-      }
+
       Frame frame;
       frame.SetRigId(rig.RigId());
       frame.AddDataId(image.DataId());
+
+      if (pose_prior.IsValid()) {
+        pose_prior.pose_prior_id = database->WritePosePrior(pose_prior);
+        frame.AddDataId(pose_prior.DataId());
```
B1ueber2y: I am a bit confused. If the pose prior sensor is not in the rig, what can we benefit by adding it to the frame? I think we would need to put it in the rig as well, and thereby also store the transformation?

src/colmap/estimators/bundle_adjustment.h
```diff
@@ -235,7 +235,7 @@ std::unique_ptr<BundleAdjuster> CreatePosePriorBundleAdjuster(
     BundleAdjustmentOptions options,
     PosePriorBundleAdjustmentOptions prior_options,
     BundleAdjustmentConfig config,
-    std::unordered_map<image_t, PosePrior> pose_priors,
```
B1ueber2y, line 238: How do we associate pose_priors with reconstruction? If we dont have them in the frames?

B1ueber2y (commented):
Thank you! Left some comments. I think with the new abstraction it makes sense to populate pose priors in rigs, frames, and reconstruction.

src/colmap/estimators/bundle_adjustment.h
B1ueber2y: We would need to store frame_id and frame_ptr in the PosePrior class? Similar to Image as a measurement.

src/pycolmap/pipeline/images.cc
ahojnnes: Yeah, you are right... One more round of refactor :-)

src/colmap/estimators/bundle_adjustment.h
ahojnnes: I would not do the explicit association and carry around pose priors everywhere. They are used very rarely and we don't need the extra convenience of being able to directly hop from one object to another (as we need for images/cameras/frames/rigs).

ahojnnes: I wouldn't even store these in the reconstruction.

B1ueber2y: The problem is that, in this function, pose_prior_t actually has one-to-one correspondence to image_t (such that we can find the corresponding pose priors in the logic of this BA construction), which makes the definition of new class ``pose_prior_t`` meaningless.  

Given that for now we assume each pose prior is on the image, I have a temporary proposal until we want to support arbitrary transformation of the pose prior: how about that in the rig, when indexing the transformation of a pose prior, we take the transformation of the corresponding image with the same id?

B1ueber2y: I agree that we can exclude pose priors from the reconstruction.

src/colmap/estimators/alignment.cc
```diff
@@ -239,7 +239,7 @@ bool AlignReconstructionToLocations(
 
 bool AlignReconstructionToPosePriors(
     const Reconstruction& src_reconstruction,
-    const std::unordered_map<image_t, PosePrior>& tgt_pose_priors,
+    const std::unordered_map<pose_prior_t, PosePrior>& tgt_pose_priors,
```
ahojnnes: Todo

src/colmap/estimators/bundle_adjustment.cc
```diff
@@ -961,11 +961,12 @@ class DefaultBundleAdjuster : public BundleAdjuster {
 
 class PosePriorBundleAdjuster : public BundleAdjuster {
  public:
-  PosePriorBundleAdjuster(BundleAdjustmentOptions options,
-                          PosePriorBundleAdjustmentOptions prior_options,
-                          BundleAdjustmentConfig config,
-                          std::unordered_map<image_t, PosePrior> pose_priors,
-                          Reconstruction& reconstruction)
+  PosePriorBundleAdjuster(
```
ahojnnes: Todo

src/colmap/feature/matcher.cc
```diff
@@ -346,13 +347,14 @@ void FeatureMatcherCache::MaybeLoadPosePriors() {
...
 
+  std::vector<PosePrior> pose_priors = database_->ReadAllPosePriors();
   pose_priors_cache_ =
-      std::make_unique<std::unordered_map<image_t, PosePrior>>();
-  pose_priors_cache_->reserve(database_->NumPosePriors());
-  for (const auto& image : *images_cache_) {
-    if (database_->ExistsPosePrior(image.first)) {
-      pose_priors_cache_->emplace(image.first,
-                                  database_->ReadPosePrior(image.first));
+      std::make_unique<std::unordered_map<pose_prior_t, PosePrior>>();
```
ahojnnes: image_t

src/colmap/geometry/pose_prior_test.cc
```diff
@@ -36,6 +36,7 @@ namespace {
 
 TEST(PosePrior, Equals) {
   PosePrior prior;
+  prior.pose_prior_id = 0;
```
ahojnnes, line 39: Todo

src/colmap/geometry/pose_prior_test.cc
```diff
@@ -49,14 +50,16 @@ TEST(PosePrior, Equals) {
 
 TEST(PosePrior, Print) {
   PosePrior prior;
+  prior.pose_prior_id = 0;
```
ahojnnes, line 72: Todo

src/colmap/feature/matcher.cc
```diff
@@ -346,13 +347,14 @@ void FeatureMatcherCache::MaybeLoadPosePriors() {
...
+  pose_priors_cache_->reserve(pose_priors.size());
+  for (PosePrior& pose_prior : pose_priors) {
+    if (pose_prior.corr_data_id.sensor_id.type == SensorType::CAMERA) {
+      pose_priors_cache_->emplace(pose_prior.corr_data_id.id,
```
B1ueber2y: I dont think the current design supports multiple priors for one single image. Should we throw at emplace?

src/colmap/scene/database.h
```diff
@@ -85,7 +85,8 @@ class Database {
   virtual bool ExistsFrame(frame_t frame_id) const = 0;
   virtual bool ExistsImage(image_t image_id) const = 0;
   virtual bool ExistsImageWithName(const std::string& name) const = 0;
-  virtual bool ExistsPosePrior(image_t image_id) const = 0;
+  virtual bool ExistsPosePrior(pose_prior_t pose_prior_id,
```
B1ueber2y, line 88: Does it make sense to still support ``ExistPosePriorWithImageId``,  ``ReadPosePriorWithImageId``, "ReadPosePriorWithImageName" etc, with tailored sql statements?

B1ueber2y (commented):
Looks great to me on a high-level pass. Thank you!!

src/colmap/scene/database.h
B1ueber2y: Maybe also throw at ``WritePosePrior`` and ``UpdatePosePrior`` when the corr_data_id is duplicate?

ahojnnes: I created a unique index on the pose priors table, which will throw on duplicate pose priors for a data measurement.

ahojnnes: Let's only create the ExistsPosePriorForImage if a need arises. These are cheap enough to read in bulk from the database.

src/colmap/estimators/bundle_adjustment.cc
```diff
@@ -1027,56 +1055,54 @@ class PosePriorBundleAdjuster : public BundleAdjuster {
...
-                             const PosePrior& prior,
-                             Reconstruction& reconstruction) {
-    if (!prior.IsValid() || !prior.IsCovarianceValid()) {
-      LOG(ERROR) << "Could not add prior for image #" << image_id;
-      return;
-    }
-
+  void AddImagePosePriorToProblem(image_t image_id,
+                                  const PosePrior& pose_prior,
+                                  Reconstruction& reconstruction) {
```
B1ueber2y, line 1060: Nit: let s add a check on pose_prior.corr_data_id.id in this function for completeness?

src/colmap/estimators/bundle_adjustment.cc
```diff
@@ -1027,56 +1055,54 @@ class PosePriorBundleAdjuster : public BundleAdjuster {
...
+        config_.HasConstantRigFromWorldPose(image.FrameId());
+    if (constant_sensor_from_rig && constant_rig_from_world) {
       return;
     }
 
-    // cam_from_world.rotation is normalized in AddImageToProblem()
-    double* cam_from_world_rotation = cam_from_world.rotation.coeffs().data();
+    ceres::Problem& problem = *default_bundle_adjuster_->Problem();
+    Frame& frame = *image.FramePtr();
+    const Eigen::Vector3d position_in_world_prior =
```
B1ueber2y: Just realizing that, here the position covariance should be scaled as well? Maybe should be addressed in a different PR since this was an existing bug non-relevant to the updates here.

src/colmap/scene/database_sqlite.cc
```diff
@@ -1921,6 +2006,33 @@ class SqliteDatabase : public Database {
...
+      // Migrate existing data to frame_data table.
+      for (Frame& frame : ReadAllFrames()) {
+        for (const auto& data_id : frame.ImageIds()) {
+          // Note that in the old schema pose_prior_id == image_id.
+          if (ExistsPosePrior(data_id.id,
+                              /*is_deprecated_image_prior=*/false)) {
+            PosePrior pose_prior =
+                ReadPosePrior(data_id.id, /*is_deprecated_image_prior=*/false);
+            pose_prior.corr_data_id = data_id;
+            UpdatePosePrior(pose_prior);
```
B1ueber2y, line 2029: If my understanding is correct, here the update is probably broken, since pose_prior_id is not set?

src/colmap/estimators/bundle_adjustment.cc
ahojnnes: Right, this was a bug. PTAL.

src/colmap/scene/database_sqlite.cc
ahojnnes: No, it should work, because ReadPosePrior sets the ID.

src/colmap/estimators/bundle_adjustment.cc
B1ueber2y: Actually, mathematically it would make sense to do R * \Sigma * R^T as well, although it does not matter here as normalized_from_metric_ always has rotation as identity.

ahojnnes: OK, let's do the proper thing, otherwise this will easily lead to errors in the future.

src/colmap/scene/database_sqlite.cc
```diff
@@ -526,8 +556,10 @@ class SqliteDatabase : public Database {
     return ExistsRowString(sql_stmt_exists_image_name_, name);
   }
 
-  bool ExistsPosePrior(const image_t image_id) const override {
-    return ExistsRowId(sql_stmt_exists_pose_prior_, image_id);
+  bool ExistsPosePrior(const pose_prior_t pose_prior_id,
+                       bool is_deprecated_image_prior) const override {
```
sarlinpe, line 560: This is called is_legacy_image_prior in database.h

src/colmap/scene/database_sqlite.cc
```diff
@@ -440,6 +457,19 @@ void WriteFrameData(const frame_t frame_id,
   }
 }
 
+// TODO(jsch): Change is_deprecated_image_prior default to true after next
```
sarlinpe: This PR already makes the default to true, is this comment stale?

PR #3785, opened 2025-12-01
Add convenience Python constructors for data_t/sensor_t

ahojnnes:


B1ueber2y:
Isn't this already supported with MakeDataClass?

ahojnnes:
Doesn't work for me without this explicit constructor. Maybe because of the already defined init method?

ahojnnes:
Checked again and couldn't make it work without these custom constructors. Going ahead and merging.

sarlinpe:
Without this PR, the following code works fine on my end:
```python
sensor = pycolmap.sensor_t(type=pycolmap.SensorType.CAMERA, id=1)  # or type='CAMERA'
data = pycolmap.data_t(sensor_id=sensor, id=1)
```

B1ueber2y:
The init from MakeDataClass is keyword only, so it also does not hurt to add a new positional constructor if really needed.

ahojnnes:
Yep, thanks, that was the difference. I didn't specify the args as keywords.

PR #3787, opened 2025-12-02
Add gravity to pose prior

ahojnnes:
In preparation for replacing glomap's custom gravity fields.

### DIFF
--- a/src/colmap/ui/database_management_widget.cc
+++ b/src/colmap/ui/database_management_widget.cc
@@ -757,7 +757,8 @@ void PosePriorsTab::Reload(const std::shared_ptr<Database>& database) {
   database_ = database;
 
   QString info;
-  info += QString("Pose Priors: ") + QString::number(database_->NumPosePriors());
+  info +=
+      QString("Pose Priors: ") + QString::number(database_->NumPosePriors());
   info_label_->setText(info);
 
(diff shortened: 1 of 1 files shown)


src/colmap/geometry/pose_prior_test.cc
```diff
@@ -41,6 +41,8 @@ TEST(PosePrior, Equals) {
   prior.position = Eigen::Vector3d::Zero();
   prior.position_covariance = Eigen::Matrix3d::Identity();
   prior.coordinate_system = PosePrior::CoordinateSystem::CARTESIAN;
+  prior.gravity = Eigen::Vector3d::UnitZ();
+  prior.gravity_covariance = Eigen::Matrix3d::Identity();
```
B1ueber2y: This is not very ideal in my opinion. Here I assume the gravity refers to the gravity direction, which is in SphereManifold<3>. The covariance in this case should ideally be defined on its tangent space, which should be 2x2. Having a 3x3 covariance for a unit vector does not make too much sense to me. 

On the other hand, is this covariance used anywhere in glomap? If not I suggest to remove it and support later when needed.

ahojnnes: Good callout. Agreed that 3x3 doesn't make sense. Let's remove it as we likely don't have a use case for it.

PR #3790, opened 2025-12-02
Defer OpenImageIO initialization to pybind module loading time.  

B1ueber2y:
An attempt to fix pycolmap on mac and windows.

### DIFF
--- a/src/colmap/sensor/bitmap.cc
+++ b/src/colmap/sensor/bitmap.cc
@@ -158,7 +158,6 @@ Bitmap::Bitmap(const int width,
 }
 
 Bitmap::Bitmap(const Bitmap& other) {
-  EnsureOpenImageIOInitialized();
   width_ = other.width_;
   height_ = other.height_;
   channels_ = other.channels_;
@@ -168,7 +167,6 @@ Bitmap::Bitmap(const Bitmap& other) {
 }
 
 Bitmap::Bitmap(Bitmap&& other) noexcept {
-  EnsureOpenImageIOInitialized();
   width_ = other.width_;
   height_ = other.height_;
   channels_ = other.channels_;
@@ -181,7 +179,6 @@ Bitmap::Bitmap(Bitmap&& other) noexcept {
 }
 
 Bitmap& Bitmap::operator=(const Bitmap& other) {
-  EnsureOpenImageIOInitialized();
   width_ = other.width_;
   height_ = other.height_;
   channels_ = other.channels_;
@@ -192,7 +189,6 @@ Bitmap& Bitmap::operator=(const Bitmap& other) {
 }
 
 Bitmap& Bitmap::operator=(Bitmap&& other) noexcept {
-  EnsureOpenImageIOInitialized();
   if (this != &other) {
     width_ = other.width_;
     height_ = other.height_;


src/colmap/sensor/bitmap.cc
```diff
@@ -97,6 +90,7 @@ std::vector<uint8_t> ConvertColorSpace(const uint8_t* src_data,
 
 void SetImageSpecColorSpace(OIIO::ImageSpec& image_spec,
                             const OIIO::string_view& colorspace) {
+  const OIIOInitializer initializer;
```
sarlinpe: Why does this need to be defined in every single function? isn't it enough at the file level as done previously?

python/CMakeLists.txt
```diff
@@ -32,7 +32,7 @@ find_package(pybind11 2.13.0 REQUIRED)
 file(GLOB_RECURSE SOURCE_FILES "${PROJECT_SOURCE_DIR}/../src/pycolmap/*.cc")
 pybind11_add_module(_core ${SOURCE_FILES})
 target_include_directories(_core PRIVATE ${PROJECT_SOURCE_DIR}/../src/)
-target_link_libraries(_core PRIVATE colmap::colmap glog::glog Ceres::ceres)
+target_link_libraries(_core PRIVATE colmap::colmap glog::glog Ceres::ceres OpenImageIO::OpenImageIO)
```
sarlinpe: Is this needed? pycolmap doesn't include OIOI directly.

src/pycolmap/main.cc
```diff
@@ -39,6 +40,11 @@ PYBIND11_MODULE(_core, m) {
   m.attr("COLMAP_version") = py::str(GetVersionInfo());
   m.attr("COLMAP_build") = py::str(GetBuildInfo());
 
+  // OpenImageIO initialization
+  m.def("_init_image_io",
+        &InitializeOpenImageIO,
+        "Explicitly initialize OpenImageIO resources.");
```
sarlinpe: Could you directly call InitializeOpenImageIO in this function? Similarly to how InitGoogleLogging is called in BindLogging. Then we don't need the additional call on the Python side.

src/colmap/sensor/bitmap.cc
B1ueber2y: The static global constructor in C++ file level caused the linking issues for pycolmap on mac and windows.

B1ueber2y: The definition in every single function is only for safety purpose. The one init at the top exe level should be enough to initialize the std::call_once.

B1ueber2y: Okay. Since we initialize OIIO in both colmap exe and pycolmap main, I think its safe enough to remove all these. I will move forward and remove them.

B1ueber2y: > Okay. Since we initialize OIIO in both colmap exe and pycolmap main, I think its safe enough to remove all these. I will move forward and remove them.

This, however, will require InitializeImageIO() to be called prior to the usage of colmap::Bitmap. This feels extremely unsafe when colmap is used as a third party in C++. Let me know if we want to revert this change.

src/colmap/util/oiio_utils.cc
```diff
@@ -0,0 +1,21 @@
...
+
+#include <cstdlib>
+#include <mutex>
+
+namespace colmap {
+
+static std::once_flag oiio_setup_flag;
+
+void InitializeOpenImageIO() {
+  std::call_once(oiio_setup_flag, []() {
```
ahojnnes, line 11: fyi, static initialization is also thread-safe as per C++ standard.

ahojnnes (approved):
Great, thank you very much for root-causing the issue.

src/colmap/sensor/bitmap.cc
ahojnnes: Can you elaborate a bit more why this issue happened before?

B1ueber2y: The issue happened because of the static global constructor, which was initialized at the library loading for pybind11. This was, on mac and windows, somehow before the resources being fully initialized for OpenImageIO and caused the crush. Now that we moved the initialization to the pybind module loading time, which is after the whole library loading is fully completed.

B1ueber2y: So now the solution is to avoid static global initialization of OpenImageIO and moved it to main() of exe/colmap.cc. However, this will make the usage of colmap::Bitmap unsafe without prior OIIO initialization, for example, when colmap is used as a third party. 

```
#include <colmap/util/bitmap.h>

# InitializeOpenImageIO() needs to be called somewhere in the program. 
Bitmap bitmap(h, w, true);
bitmap.Read(...);
```

Adding the initialization scope to every OIIO related operations in Bitmap class can solve this issue but can be considered cluttering the codebase, as was undone here: https://github.com/colmap/colmap/pull/3790/commits/bafd1e8d7fd75c48e54521a2c7b307933d468def.

src/colmap/util/oiio_utils.cc
B1ueber2y: Updated the PR naming. Thanks!

src/colmap/sensor/bitmap.cc
ahojnnes: Ok, can we also do the static initialization and call into your new function? It should be a noop when it was already called sometime before (eg in the main function) but be guaranteed to be called when entering the bitmap translation unit for the first time?

B1ueber2y: > Ok, can we also do the static initialization and call into your new function? It should be a noop when it was already called sometime before (eg in the main function) but be guaranteed to be called when entering the bitmap translation unit for the first time?

Not sure if you meant this: https://github.com/colmap/colmap/pull/3790/commits/eb1113f59e5ce16c1ec551afdf56da0dcf9b807a? Likely this will cause the same issues on mac and windows as before (lets wait and see).

ahojnnes: I see, I understand the problem now after this failed, sorry for wasting time with the suggestion.

Could it be enough to add this initializer to the constructors and the static methods? The remaining methods should be covered then, because the object cannot be constructed otherwise in a normal way?

B1ueber2y: Nice idea. Updated: https://github.com/colmap/colmap/pull/3790/commits/f67401f019c21ed9ca4d43cc866c97e68975d082. Also added it to three relevant stand-alone functions. PTAL.

ahojnnes: I guess the move/copy assignment operators and constructors are not strictly necessary because the "other" object would have already been constructed and initialized OIIO?

ahojnnes: Otherwise LGTM.

B1ueber2y: Done.

PR #3794, opened 2025-12-03
Replace glomap gravity info with pose prior

ahojnnes:


ahojnnes:
The tests are all passing now. @lpanaf Do you have a real dataset on which you can verify that this still works as intended after the refactoring?

src/glomap/controllers/rotation_averager_test.cc
```diff
@@ -1,58 +1,26 @@
...
-    }
-
-    if (outlier_ratio > 0.0 &&
+void SynthesizeGravityOutliers(std::vector<colmap::PosePrior>& pose_priors,
+                               double outlier_ratio = 0.0) {
+  for (auto& pose_prior : pose_priors) {
+    if (pose_prior.HasGravity() &&
         colmap::RandomUniformReal<double>(0, 1) < outlier_ratio) {
-      gravityInRig = CreateRandomAxis(rng);
+      pose_prior.gravity = Eigen::Vector3d::Random();
```
B1ueber2y: normalized()?

src/glomap/scene/image.h
```diff
@@ -44,11 +42,6 @@ struct Image {
   // Check if cam_from_world needs to be composed with sensor_from_rig pose.
   inline bool HasTrivialFrame() const;
```
B1ueber2y: A lot of methods in this class have already been present in colmap::Image?

src/glomap/estimators/global_rotation_averaging.cc
```diff
@@ -44,51 +45,97 @@ double RelAngleError(std::mt19937& rng,
...
+  for (const auto& [image_id, image] : images) {
+    image_to_frame[image_id] = image.frame_id;
+  }
+
+  std::unordered_map<frame_t, colmap::PosePrior*> frame_to_pose_prior;
+  for (auto& pose_prior : pose_priors) {
+    if (pose_prior.corr_data_id.sensor_id.type == SensorType::CAMERA) {
+      const image_t image_id = pose_prior.corr_data_id.id;
+      const Image& image = images.at(image_id);
+      if (image.HasTrivialFrame()) {
```
B1ueber2y: I am confused here. So for now this rotation averaging does not support the handling of either of the following:

1) multiple gravity priors in one single frame. 
2) a gravity prior from non-reference sensor (this is contrary to other comments in this file).

src/glomap/estimators/global_rotation_averaging.cc
```diff
@@ -44,51 +45,97 @@ double RelAngleError(std::mt19937& rng,
...
+  auto it = frame_to_pose_prior.find(frame_id);
+  if (it == frame_to_pose_prior.end() || !it->second->HasGravity()) {
+    return nullptr;
+  }
+  return &it->second->gravity;
+}
+
 }  // namespace
 
 bool RotationEstimator::EstimateRotations(
```
B1ueber2y, line 102: Why not use const for rigs, images, and pose_priors? Its confusing now what the output is for this function.

src/glomap/estimators/gravity_refinement.cc
```diff
@@ -6,25 +6,78 @@
 #include "glomap/math/gravity.h"
 
 namespace glomap {
-void GravityRefiner::RefineGravity(const ViewGraph& view_graph,
-                                   std::unordered_map<frame_t, Frame>& frames,
-                                   std::unordered_map<image_t, Image>& images) {
+namespace {
+
+Eigen::Matrix3d GetImageAlignRot(const Image& image,
```
B1ueber2y, line 11: Can we improve the naming of this function with some comments? 

Related to it, the functions in glomap/math/gravity.h really need to be improved IMO. The namings are not very clear and the assumption on the input gravity is a bit shaky: some of them assume non-normalized gravity and some of them (e.g., average) assume normalized.

src/glomap/io/pose_io.cc
```diff
@@ -159,20 +173,31 @@ void ReadGravity(const std::string& gravity_path,
...
-        images[ite->second].frame_ptr->gravity_info.SetGravity(gravity);
-        Rigid3d& cam_from_world = images[ite->second].frame_ptr->RigFromWorld();
+    auto ite = image_name_to_id.find(name);
+    if (ite != image_name_to_id.end()) {
+      auto& image = images[ite->second];
+      if (image.HasTrivialFrame()) {
+        counter++;
+        auto& pose_prior = pose_priors.emplace_back();
+        pose_prior.pose_prior_id = ite->second;
+        pose_prior.corr_data_id.id = ite->second;
```
B1ueber2y: pose_prior.corr_data_id = image.DataId();

src/glomap/scene/image.h
ahojnnes: Yes, I am working towards eliminating this class. Removing the gravity was step 1.

src/glomap/estimators/global_rotation_averaging.cc
ahojnnes: @lpanaf Can you clarify? Maybe I misunderstood.

src/glomap/estimators/global_rotation_averaging.cc
ahojnnes: Thanks, I made it a bit more restrictive. Only rigs and frames are modified.

src/glomap/estimators/gravity_refinement.cc
ahojnnes: Fully agreed. Was already the plan in a follow up PR later.

src/glomap/estimators/global_rotation_averaging.cc
lpanaf: The current implementation only support 1 gravity prior per frame, which corresponds to the reference frame. @B1ueber2y 's understanding is correct here. In this case, the other comments are incorrect. 

The reason for such a design is that the alignment function would be complicated for a non-reference sensor, and having several priors associated with the same frame can cause potential conflicts.

lpanaf:
> The tests are all passing now. @lpanaf Do you have a real dataset on which you can verify that this still works as intended after the refactoring?

I will need to dig into my computer to find some. Will look into this :)

src/glomap/estimators/global_rotation_averaging.cc
B1ueber2y: The gravity would anyway cause conflicts, and was currently fixed with GetAligendRot (nit: I think GetAlignedRotation would be a better name). In this case, getting an average gravity from multiple priors for the reference would be quite natural.

lpanaf: Yes, I agree with this. The original plan includes this as well, but it was still a TODO, not finished.

ahojnnes: I assume this closes this discussion. We can address these limitations in other PRs, but this one here is just meant to move us to use colmap's PosePrior instead of the custom GravityInfo. The intention is to have no change in behavior. I'll wait for @lpanaf to confirm on a real dataset before merging.

src/glomap/estimators/global_rotation_averaging.cc
```diff
@@ -44,51 +45,98 @@ double RelAngleError(std::mt19937& rng,
...
-                     << sensor_id.id
-                     << ", but the gravity aligned rotation is "
-                        "requested. Please add the rig calibration.";
-          return false;
-        }
-      }
-    }
+    const std::unordered_map<image_t, Image>& images,
+    const std::vector<colmap::PosePrior>& pose_priors) {
+  // Now, for the gravity aligned case, we only support trivial rigs only or
```
B1ueber2y: Can we update the comments here, according to our discussions?

B1ueber2y: Or maybe in the header?

ahojnnes: Done

B1ueber2y (approved):

lpanaf:
Hey @ahojnnes, I have verified with real world data that the updated version works well now. Thanks!

PR #3795, opened 2025-12-03
Synthesize gravities and move pose prior noise synthesis

ahojnnes:
* Synthesize prior gravities.
* Move pose prior noise synthesis to appropriate function.

### DIFF
--- a/src/pycolmap/scene/synthetic.cc
+++ b/src/pycolmap/scene/synthetic.cc
@@ -80,9 +80,9 @@ void BindSynthetic(py::module& m) {
           .def_readwrite("point2D_stddev",
                          &SyntheticNoiseOptions::point2D_stddev)
           .def_readwrite("prior_position_stddev",
-                         &SyntheticDatasetOptions::prior_position_stddev)
+                         &SyntheticNoiseOptions::prior_position_stddev)
           .def_readwrite("prior_gravity_stddev",
-                         &SyntheticDatasetOptions::prior_gravity_stddev);
+                         &SyntheticNoiseOptions::prior_gravity_stddev);
   MakeDataclass(PySyntheticNoiseOptions);
 
   m.def("synthesize_noise",


src/colmap/scene/synthetic.cc
```diff
@@ -502,6 +520,38 @@ void SynthesizeNoise(const SyntheticNoiseOptions& options,
...
+            RandomGaussian<double>(0, options.prior_position_stddev));
+        pose_prior.position_covariance = options.prior_position_stddev *
+                                         options.prior_position_stddev *
+                                         Eigen::Matrix3d::Identity();
+        if (prior_in_wgs84) {
+          PosePriorPositionCartesianToWGS84(pose_prior);
+        }
+      }
+      if (options.prior_gravity_stddev > 0.) {
+        pose_prior.gravity =
```
B1ueber2y, line 550: I think this is not the correct way in SphereManifold<3> to produce uniform tilt noise around the original gravity. One should project the rotation axis to the perpendicular direction of the original gravity (e.g., using cross product). Something like:

```
axis = pose.gravity.cross(Eigen::Vector3d::Random().normalized()).normalized();
```

src/colmap/scene/synthetic.cc
```diff
@@ -297,32 +315,29 @@ void SynthesizeDataset(const SyntheticDatasetOptions& options,
...
-                RandomGaussian<double>(0, options.prior_position_stddev));
-            pose_prior.position_covariance = options.prior_position_stddev *
-                                             options.prior_position_stddev *
-                                             Eigen::Matrix3d::Identity();
-          } else {
-            pose_prior.position_covariance = Eigen::Matrix3d::Identity();
+
+          if (options.prior_position) {
+            pose_prior.position = cam_from_world.TgtOriginInSrc();
+            pose_prior.position_covariance = 1e-3 * Eigen::Matrix3d::Identity();
```
B1ueber2y: There is a behavior change here. Before it was identity but now it becomes 1e-3 * identity.

src/colmap/scene/synthetic.cc
```diff
@@ -502,6 +520,38 @@ void SynthesizeNoise(const SyntheticNoiseOptions& options,
...
+        const bool prior_in_wgs84 =
+            pose_prior.coordinate_system == PosePrior::CoordinateSystem::WGS84;
+        if (prior_in_wgs84) {
+          PosePriorPositionWGS84ToCartesian(pose_prior);
+        }
+        pose_prior.position += Eigen::Vector3d(
+            RandomGaussian<double>(0, options.prior_position_stddev),
+            RandomGaussian<double>(0, options.prior_position_stddev),
+            RandomGaussian<double>(0, options.prior_position_stddev));
+        pose_prior.position_covariance = options.prior_position_stddev *
```
B1ueber2y: Should be += rather than =?

src/colmap/scene/synthetic.cc
ahojnnes: You are right, thanks. Minor nit: we want a Gaussian and not a uniform distribution.

src/colmap/scene/synthetic.cc
ahojnnes: The value before didn't really make much sense, so I changed it to a smaller value, which is a bit more meaningful but still not correct :-). We could leave it unset but then the prior bundle adjustment term doesn't have a weight and we need to special case the implementation there.

B1ueber2y: Does it make sense to make 1e-3 an option then? e.g., default_prior_position_stddev. Also, it would be nice to use 1e-4 / -6 (with an even number) to make it geometrically more meaningful (1cm / 1mm)?

ahojnnes: PTAL, I now kept it undefined and handle the case more robustly in the prior bundler.

src/colmap/estimators/bundle_adjustment.cc
```diff
@@ -1075,11 +1082,18 @@ class PosePriorBundleAdjuster : public BundleAdjuster {
 
     const Eigen::Vector3d normalized_position =
         normalized_from_metric_ * pose_prior.position;
-    const Eigen::Matrix3d sR =
+    const Eigen::Matrix3d normalized_from_metric_scaled_rotation =
         normalized_from_metric_.scale *
         normalized_from_metric_.rotation.toRotationMatrix();
+    const Eigen::Matrix3d position_cov =
```
ahojnnes, line 1088: Could also remember the Median(rms_var) from the alignment and use it here or depend on the scene scale?

B1ueber2y: I think we should not promote the use case of no valid covariances. The fallback should be enough : )

PR #3799, opened 2025-12-04
Relax CUDA requirements in Python wheels

sarlinpe:
Switching the Docker image for CUDA-enabled wheels to `pytorch/manylinux2_28-builder:cuda12.9`, which requires a lower glibc version, identical to the non-CUDA wheels. This new Docker image doesn't have CUDA wheels in the global path by default.

This uncovered that we need specific logic in `__init__.py` to load the libs from NVIDIA pip packages, as done in Pytorch's https://github.com/pytorch/pytorch/blob/main/torch/__init__.py

To handle CUDA 13 (which requires knowing which cuda version to load), we might need to store the CUDA version in a py file, similar to https://github.com/pytorch/pytorch/blob/main/tools/generate_torch_version.py


ducha-aiki:
Could you please release new wheels? :)

Issue #3797, opened 2025-12-04
Init pycolmap.Image with cam_from_world (read-only)


Dawars:
**Is your feature request related to a problem? Please describe.**
I'm trying to init a reconstruction from known camera poses according to this:
https://colmap.github.io/faq.html#reconstruct-sparse-dense-model-from-known-camera-poses

It would be a lot easier/cleaner if I could construct such an empty reconstruction using pycolmap instead of writing txt files.

Currently, **cam_from_world** is read-only, but it shouldn't be a problem to set it in the constructor.
AttributeError: 'pycolmap._core.Image' object attribute 'cam_from_world' is read-only

**Describe the solution you'd like**
Make it possible to init an Image object with **cam_from_world**

**Additional context**
```python
pycolmap.Image(
    name="name.jpg",
    image_id=1,
    cam_from_world=pycolmap.Rigid3d()
)
```

sarlinpe:
Yes, this unfortunately got complicated when we added support for multi-camera rigs. You actually need to set the pose through the frame object:
```python
import pycolmap
rec = pycolmap.Reconstruction()

camera_id = 1
image_id = 1
image_name = '1.jpg'
cam_from_world = pycolmap.Rigid3d()
# For simplicity, use rig_id == camera_id and frame_id == image_id for single-camera rigs.

camera = pycolmap.Camera.create(camera_id=camera_id, model=pycolmap.CameraModelId.SIMPLE_PINHOLE, focal_length=256, width=256, height=256)
rec.add_camera(camera)

rig = pycolmap.Rig(rig_id=camera_id)
rig.add_ref_sensor(camera.sensor_id)  # The reference sensor has identity transform to the rig.
rec.add_rig(rig)

image = pycolmap.Image(image_id=image_id, camera_id=camera_id, frame_id=image_id, name=image_name)
# Use rig_from_world == cam_from_world since the image is the reference sensor.
# Alternatively, set the pose later via reconstruction.frame(frame_id).set_cam_from_world
frame = pycolmap.Frame(frame_id=image.frame_id, rig_id=rig.rig_id, rig_from_world=cam_from_world)
frame.add_data_id(image.data_id)  # Add the image to the frame.
rec.add_frame(frame)
rec.add_image(image)
```

@B1ueber2y is this the simplest way to add an image to a reconstruction? could we provide helpers to simplify this for the regular case with only trivial frames?

B1ueber2y:
[...]

PR #3800, opened 2025-12-04
Add support on adding camera/image with trivial rig/frame to reconstruction.

B1ueber2y:
Fixes: https://github.com/colmap/colmap/issues/3797

### DIFF
--- a/src/colmap/scene/reconstruction.cc
+++ b/src/colmap/scene/reconstruction.cc
@@ -261,6 +261,10 @@ void Reconstruction::AddCamera(struct Camera camera) {
 }
 
 void Reconstruction::AddCameraWithTrivialRig(struct Camera camera) {
+  THROW_CHECK(!ExistsRig(camera.camera_id))
+      << "AddCameraWithTrivialRig tried to add a rig with the same id as the "
+         "camera, but failed because Rig "
+      << camera.camera_id << "already exists in the reconstruction. ";
   class Rig rig;
   rig.SetRigId(camera.camera_id);
   rig.AddRefSensor(camera.SensorId());
@@ -319,6 +323,10 @@ void Reconstruction::AddImage(class Image image) {
 }
 
 void Reconstruction::AddImageWithTrivialFrame(class Image image) {
+  THROW_CHECK(!ExistsFrame(image.ImageId()))
+      << "AddImageWithTrivialFrame tried to add a frame with the same id as "
+         "the image, but failed because Frame "
+      << image.ImageId() << "already exists in the reconstruction.";
   class Frame frame;
   frame.SetFrameId(image.ImageId());
   frame.SetRigId(image.CameraId());


src/colmap/scene/reconstruction.cc
```diff
@@ -310,6 +318,28 @@ void Reconstruction::AddImage(class Image image) {
...
+  AddImage(std::move(image));
+}
+
+void Reconstruction::AddImageWithTrivialFrame(class Image image,
+                                              const Rigid3d& cam_from_world) {
+  const frame_t frame_id = image.ImageId();
+  AddImageWithTrivialFrame(std::move(image));
+  Frame(frame_id).SetRigFromWorld(cam_from_world);
+  RegisterFrame(frame_id);
+}
```
sarlinpe, line 349: How about having a single function with an optional argument cam_from_world? This would also simplify the bindings.

src/colmap/scene/reconstruction.cc
```diff
@@ -260,6 +260,14 @@ void Reconstruction::AddCamera(struct Camera camera) {
   THROW_CHECK(cameras_.emplace(camera_id, std::move(camera)).second);
 }
 
+void Reconstruction::AddCameraWithTrivialRig(struct Camera camera) {
+  class Rig rig;
+  rig.SetRigId(camera.camera_id);
```
sarlinpe, line 269: This won't work if users call this function on a reconstruction with existing rigs. Consider checking that there is no rig with the same rig_id, or make the error message in AddRig more informative.

src/colmap/scene/reconstruction.cc
B1ueber2y: The behavior is different in whether to register the frame, so I made it two different methods.

src/colmap/scene/reconstruction.cc
B1ueber2y: Thanks. I added some more explicit error messages for both methods.

src/colmap/scene/reconstruction.cc
```diff
@@ -310,6 +322,32 @@ void Reconstruction::AddImage(class Image image) {
...
+  }
+  AddFrame(frame);
+  AddImage(std::move(image));
+}
+
+void Reconstruction::AddImageWithTrivialFrame(class Image image,
+                                              const Rigid3d& cam_from_world) {
+  const frame_t frame_id = image.ImageId();
+  AddImageWithTrivialFrame(std::move(image));
+  Frame(frame_id).SetRigFromWorld(cam_from_world);
```
sarlinpe, line 347: Here, to be safe, should we check that this camera is the reference sensor of the rig?

B1ueber2y: Indeed. I think we should check that the rig has only one sensor to be safe.

B1ueber2y: Added here: https://github.com/colmap/colmap/pull/3802.

PR #3801, opened 2025-12-04
Rename HasTrivialFrame to IsRefInFrame.

B1ueber2y:
The previous naming can be confusing.

### DIFF
--- a/src/colmap/scene/reconstruction.cc
+++ b/src/colmap/scene/reconstruction.cc
@@ -260,6 +260,18 @@ void Reconstruction::AddCamera(struct Camera camera) {
   THROW_CHECK(cameras_.emplace(camera_id, std::move(camera)).second);
 }
 
+void Reconstruction::AddCameraWithTrivialRig(struct Camera camera) {
+  THROW_CHECK(!ExistsRig(camera.camera_id))
+      << "AddCameraWithTrivialRig tried to add a rig with the same id as the "
+         "camera, but failed because Rig "
+      << camera.camera_id << "already exists in the reconstruction. ";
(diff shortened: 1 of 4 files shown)


src/glomap/scene/image.h
```diff
@@ -42,7 +42,7 @@ struct Image {
   inline int ClusterId() const;
 
   // Check if cam_from_world needs to be composed with sensor_from_rig pose.
-  inline bool HasTrivialFrame() const;
+  inline bool IsReferenceInFrame() const;
```
ahojnnes: nit: I suggest `IsRefInFrame()` for brevity.

ahojnnes:
Needs an equivalent update in pycolmap? If not, then we should add a binding for it?

B1ueber2y:
> Needs an equivalent update in pycolmap? If not, then we should add a binding for it?

Done. PTAL : )

src/pycolmap/scene/image.cc
```diff
@@ -152,6 +152,10 @@ void BindSceneImage(py::module& m) {
       .def("reset_frame_ptr",
            &Image::ResetFramePtr,
            "Make the frame pointer a nullptr.")
+      .def("is_ref_in_frame",
+           &Image::IsRefInFrame,
+           "Check if cam_from_world needs to be composed with sensor_from_rig "
```
ahojnnes: nit: This comment is a bit limited. Maybe better "Check if the image was captured by the reference sensor in the rig." Same applies on the C++ side.

ahojnnes (approved):

PR #3816, opened 2025-12-10
Fix stand-alone python visualizer.

B1ueber2y:
The python visualizer has been broken for a while, since the rig support.


B1ueber2y:
On the other hand, should we completely drop ``read_write_model.py`` from the repo, as it is broken since the rig support?

This code is actually something that has been used extensively in the community: https://github.com/search?q=read_write_model&type=code. Dropping it would encourage the migration to pycolmap.

ahojnnes:
Yes, let's kill the read write model.

B1ueber2y:
> Yes, let's kill the read write model.

Done: https://github.com/colmap/colmap/pull/3817

PR #3817, opened 2025-12-10
Remove the deprecated read_write_model.py from colmap.

B1ueber2y:
Encouraging the users to use pycolmap instead.


B1ueber2y:
Also removed the clang-format-code.py which is no longer useful with the presence of ``scripts/format/c++.sh``. And moved two needed functions to ``read_write_fused_vis.py``.

Issue #3809, opened 2025-12-08
issues with freeimageio and dense reconstruction


plusk01:
**Describe the bug**
Undistortion of single-channel greyscale images (both jpg / png) results in incorrect output. Bug introduced by #3459 (works fine on prior commit). Likely a memory buffer / stride issue.

Unrelated to greyscale imagery (tried on fresh south building dataset), attempting to generate depth maps by clicking 'Stereo' yields
```
==============================================================================
Processing view 127 / 128 for P1180219.JPG
==============================================================================
E20251208 02:07:54.936579 17538 bitmap.cc:50] 'dynamic_cast<OIIOMetaData*>(meta_data)' Must be non NULL
I20251208 02:07:54.936589 17538 misc.cc:44] 
==============================================================================
Processing view 128 / 128 for P1180198.JPG
==============================================================================
E20251208 02:07:54.936592 17538 bitmap.cc:50] 'dynamic_cast<OIIOMetaData*>(meta_data)' Must be non NULL
I20251208 02:07:54.936600 17527 timer.cc:90] Elapsed time: 0.004 [minutes]
```

**To Reproduce**
1. Download south building dataset
2. Use GIMP to convert an image to greyscale
3. Start colmap gui
4. Import sparse reconstruction model
5. Double check Inspect Database to ensure that input image is greyscale.
6. Dense reconstruction, select model
7. Run Undistortion
8. View undistorted image

**Screenshots**
_Input_
![Image](https://github.com/user-attachments/assets/5171b1b0-019a-4dec-8248-7db61d4f1879)

_Undistorted_
![Image](https://github.com/user-attachments/assets/414dafdb-9d01-4392-a4c4-307d21887485)

**Environment:**
 - OS: Ubuntu
 - COLMAP Version 145baccaace9b8a46df8ac0cb9ca353585484d5a

Parskatt:
Can reproduce this on most recent `pycolmap` wheels with CUDA.

ahojnnes:
Thank you for reporting the issue. Should be fixed with: https://github.com/colmap/colmap/pull/3812

ahojnnes:
Appreciate if you can confirm and close in case it successfully fixes the issue also on your end.

plusk01:
thx - greyscale undistortion issue resolved

However, still see the following when trying to patch match on [e117f9c](https://github.com/colmap/colmap/commit/e117f9c1ab2d60a8a9c40145d0ca86fb1e8583b6)

```
[...]

PR #3823, opened 2025-12-12
Fix copying behaviors of empty bitmap.

B1ueber2y:
Fixes https://github.com/colmap/colmap/issues/3809

### DIFF
--- a/src/colmap/sensor/bitmap_test.cc
+++ b/src/colmap/sensor/bitmap_test.cc
@@ -152,7 +152,7 @@ TEST(Bitmap, MoveAssign) {
 
 TEST(Bitmap, ConstructCopyEmpty) {
   Bitmap bitmap;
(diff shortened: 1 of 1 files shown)


src/colmap/mvs/image.h
```diff
@@ -66,8 +66,8 @@ class Image {
 
  private:
   std::string path_;
-  size_t width_;
-  size_t height_;
+  size_t width_ = 0;
+  size_t height_ = 0;
```
sarlinpe, line 70: Is this intended?

B1ueber2y: Yes. There is a default constructor of mvs::Image so i added a default value here. There is unfortunately no test coverage in the current codebase for the mvs::Image class.

ahojnnes (approved):

ahojnnes:
Thank you for fixing this!

Issue #3667, opened 2025-10-23
`CreateSiftFeatureMatcher` in python interface


senyai:
**Is your feature request related to a problem? Please describe.**

Currently, I can extract Sift features

```python
sift = pycolmap.Sift(feature_extracting_options)
for img in _read_images(image_directory):
    keypoints, descriptors = sift.extract(img)
```

but what do I do next? I think I want to create a `Matcher` using `CreateSiftFeatureMatcher` and call Matcher's `match` method, but it is not exposed in python interface.

**Describe the solution you'd like**

Add `CreateSiftFeatureMatcher` method, and maybe add an example file.

**Describe alternatives you've considered**

I know, I can call `CreateSiftFeatureMatcher` implicitly and use database, but it is inconvenient.

PR #3827, opened 2025-12-14
Expose raw SIFT descriptor matching in pycolmap via SiftMatcher

dlichy:
## Summary

This PR adds a small Python wrapper to match SIFT descriptors directly (without a database).

Related to #3667.

## What’s included

- New `pycolmap.SiftMatcher` wrapper around COLMAP `FeatureMatcher`
- GPU matching and CPU brute-force matching supported
- Added docstring explained how to cache descriptors on GPU with `image_id` and warning that reusing an `image_id` for different descriptors can produce incorrect results.

## Limitation

- CPU FAISS matching (`cpu_brute_force_matcher=False`) is not supported. See follow-up question.

## Example

```python
sift = pycolmap.Sift()
k1, d1 = sift.extract(img1)
k2, d2 = sift.extract(img2)

matcher = pycolmap.SiftMatcher(device=pycolmap.Device.cuda)  # or Device.cpu
matches = matcher.match(d1, d2, image_id1=1, image_id2=2)
```

## Follow-ups

- If this looks good,  I can go ahead and implement the `MatchGuided` function as part of this class.
- For CPU FAISS, do you have any ideas or comments about integrating index caching?

### DIFF
--- a/src/pycolmap/feature/matching.cc
+++ b/src/pycolmap/feature/matching.cc
@@ -9,9 +9,79 @@
 #include <pybind11/eigen.h>
 #include <pybind11/pybind11.h>
 
+inline static constexpr int kDescDim = 128;
+
 using namespace colmap;
 using namespace pybind11::literals;
 namespace py = pybind11;
+
+class SiftMatcher {
+ public:
+  SiftMatcher(std::optional<FeatureMatchingOptions> options, Device device)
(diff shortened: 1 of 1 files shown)


ahojnnes:
This was implemented in another PR and is available in the main branch now and the upcoming release.

PR #3834, opened 2025-12-20
Add integration tests for meshing, update poisson recon library to 18.75

ahojnnes:
* Simple integration tests to ensure the meshing doesn't crash and produces *some* mesh.
* Updates poisson recon library to 18.75 to fix address sanitizer issues.


B1ueber2y:
Relevant for the CI failure: https://github.com/colmap/colmap/pull/1491

B1ueber2y:
Does it make sense to use submodule instead?

ahojnnes:
> Does it make sense to use submodule instead?

I would prefer no submodule but instead use CMake's "Fetch" commands. It is possible but requires some more effort, because PoissonRecon needs some code modifications.

PR #3835, opened 2025-12-20
Remove custom glomap image class

ahojnnes:


src/glomap/io/colmap_converter.cc
```diff
@@ -210,6 +213,7 @@ void ConvertDatabaseToGlomap(
     Image& glomap_image = images[image_id];
     glomap_image.SetImageId(image_id);
     glomap_image.SetCameraId(image.CameraId());
+    glomap_image.SetCameraPtr(&cameras.at(image.CameraId()));
```
B1ueber2y: The naming and logic of this code block can be greatly simplified.

PR #3836, opened 2025-12-21
Consolidate calculation of angle between vectors

ahojnnes:


src/colmap/geometry/triangulation.cc
```diff
@@ -156,56 +156,38 @@ bool TriangulateOptimalPoint(const Eigen::Matrix3x4d& cam1_from_world_mat,
...
-        baseline_length_squared, proj_center1, proj_center2, points3D[i]);
+    angles[i] =
+        CalculateTriangulationAngle(proj_center1, proj_center2, points3D[i]);
   }
   return angles;
 }
 
+double CalculateAngleBetweenVectors(const Eigen::Vector3d& v1,
+                                    const Eigen::Vector3d& v2) {
+  const double norm1 = v1.squaredNorm();
```
B1ueber2y: Nit: sq_norm1?

PR #3839, opened 2025-12-21
Improve glomap cost function, add tests

ahojnnes:
* Improved notation and some other minor refactoring.
* Some unit tests with work left for the Fetzer local length cost functions.

### DIFF
--- a/src/glomap/estimators/cost_function.h
+++ b/src/glomap/estimators/cost_function.h
@@ -11,8 +11,9 @@ namespace glomap {
 // from two positions such that: t_ij - scale * (p_j - p_i) is minimized.
 // The positions can either be two camera centers or one camera center and one
 // 3D point.
-struct BATAPairwiseDirectionError {
-  explicit BATAPairwiseDirectionError(const Eigen::Vector3d& pos2_from_pos1_dir)
+struct BATAPairwiseDirectionCostFunctor {
+  explicit BATAPairwiseDirectionCostFunctor(
+      const Eigen::Vector3d& pos2_from_pos1_dir)
       : pos2_from_pos1_dir_(pos2_from_pos1_dir) {}
 
   template <typename T>
@@ -30,8 +31,9 @@ struct BATAPairwiseDirectionError {
   static ceres::CostFunction* Create(
       const Eigen::Vector3d& pos2_from_pos1_dir) {
     return (
-        new ceres::AutoDiffCostFunction<BATAPairwiseDirectionError, 3, 3, 3, 1>(
-            new BATAPairwiseDirectionError(pos2_from_pos1_dir)));
+        new ceres::
+            AutoDiffCostFunction<BATAPairwiseDirectionCostFunctor, 3, 3, 3, 1>(
+                new BATAPairwiseDirectionCostFunctor(pos2_from_pos1_dir)));
   }
 
   const Eigen::Vector3d pos2_from_pos1_dir_;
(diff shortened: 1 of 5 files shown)


src/glomap/estimators/cost_function.h
```diff
@@ -7,117 +7,105 @@
...
+// from two positions such that t_ij - scale * (c_j - c_i + rig_scale * t_rig)
 // is minimized.
 struct RigBATAPairwiseDirectionError {
-  RigBATAPairwiseDirectionError(const Eigen::Vector3d& translation_obs,
-                                const Eigen::Vector3d& translation_rig)
-      : translation_obs_(translation_obs), translation_rig_(translation_rig) {}
+  RigBATAPairwiseDirectionError(const Eigen::Vector3d& cam_from_point3D_dir,
+                                const Eigen::Vector3d& cam_from_rig_dir)
+      : cam_from_point3D_dir_(cam_from_point3D_dir),
+        cam_from_rig_dir_(cam_from_rig_dir) {}
```
B1ueber2y: cam_from_rig is not a direction, but instead a metric translation.

src/glomap/estimators/cost_function.h
```diff
@@ -7,117 +7,105 @@
...
+                  const T* rig_in_world,
+                  const T* cam_in_rig,
                   const T* scale,
                   T* residuals) const {
-    Eigen::Map<Eigen::Matrix<T, 3, 1>> residuals_vec(residuals);
-
-    Eigen::Matrix<T, 3, 1> translation_rig =
-        rig_from_world_rot_.toRotationMatrix().transpose() *
-        Eigen::Map<const Eigen::Matrix<T, 3, 1>>(cam_from_rig_center);
+    const Eigen::Matrix<T, 3, 1> cam_from_rig_dir =
```
B1ueber2y: Same here. This is not a direction, but a metric translation.

src/glomap/estimators/cost_function.h
```diff
@@ -309,32 +287,27 @@ class FetzerFocalLengthSameCameraCost {
   Eigen::Vector4d d_12;
 };
 
-// ----------------------------------------
-// GravError
-// ----------------------------------------
-struct GravError {
-  explicit GravError(const Eigen::Vector3d& grav_obs) : grav_obs_(grav_obs) {}
+struct GravityError {
```
B1ueber2y: How about adding a general cost similar to https://github.com/cvg/pyceres/blob/main/_pyceres/factors/bindings.h#L17 in a new general cost function utility file. We can also move the wrappers (e.g. covariance weighted wrapper) to the same file as well.

src/glomap/estimators/cost_function.h
B1ueber2y: By metric here I mean that the scale matters.

ahojnnes: Right, I guess cam_from_rig_translation_in_rig_scale_ would be more appropriate but is a mouthful... maybe just cam_from_rig_translation_

src/glomap/estimators/cost_function.h
ahojnnes: Agreed. Let's do this as a follow-up to this.

B1ueber2y (approved):

src/glomap/estimators/cost_function.h
```diff
@@ -7,117 +7,106 @@
...
 
-// ----------------------------------------
-// BATAPairwiseDirectionError
-// ----------------------------------------
 // Computes the error between a translation direction and the direction formed
-// from two positions such that t_ij - scale * (c_j - c_i) is minimized.
+// from two positions such that: t_ij - scale * (p_j - p_i) is minimized.
+// The positions can either be two camera centers or one camera center and one
+// 3D point.
 struct BATAPairwiseDirectionError {
```
B1ueber2y: Should we align the naming with colmap and add Functor at the end of the naming of those cost function structs?

B1ueber2y: Done here: https://github.com/colmap/colmap/pull/3842.

ahojnnes: Thanks!

PR #3843, opened 2025-12-24
Refactor general cost function utils and add NormalPriorCost and NormalErrorCost.

B1ueber2y:
Dependent on https://github.com/colmap/colmap/pull/3839

As a result, ``GravityErrorFunctor`` in glomap becomes ``colmap::NormalPriorCostFunctor<3>``.


src/colmap/estimators/cost_function_utils_test.cc
```diff
@@ -0,0 +1,120 @@
...
+
+TEST(NormalPriorCostFunctor, Nominal) {
+  const Eigen::Vector3d prior(1, 2, 3);
+
+  std::unique_ptr<ceres::CostFunction> cost_function(
+      NormalPriorCostFunctor<3>::Create(prior));
+  ASSERT_NE(cost_function, nullptr);
+  EXPECT_EQ(cost_function->num_residuals(), 3);
+
+  double residuals[3];
```
ahojnnes: nit: could use Eigen::Vector3d residuals here and then use the EigenMatrixNear matcher below.

PR #3844, opened 2025-12-24
Various improvements and fixes on glomap gravity impl.

B1ueber2y:
* Enforce all functions in glomap/math/gravity.h to take normalized gravity.
* Apply several fixes in the implementation as pinpointed. 
* Minor improvements on namings and comments.

### DIFF
--- a/src/colmap/estimators/CMakeLists.txt
+++ b/src/colmap/estimators/CMakeLists.txt
@@ -37,6 +37,7 @@ COLMAP_ADD_LIBRARY(
         alignment.h alignment.cc
         bundle_adjustment.h bundle_adjustment.cc
         coordinate_frame.h coordinate_frame.cc
+        cost_function_utils.h
         cost_functions.h
         manifold.h
         covariance.h covariance.cc
@@ -104,6 +105,11 @@ COLMAP_ADD_TEST(
 COLMAP_ADD_TEST(
     NAME cost_functions_test
     SRCS cost_functions_test.cc
+    LINK_LIBS colmap_estimators
+)
+COLMAP_ADD_TEST(
+    NAME cost_function_utils_test
+    SRCS cost_function_utils_test.cc
     LINK_LIBS colmap_estimators
 )
 COLMAP_ADD_TEST(
--- a/src/colmap/estimators/cost_function_utils.h
+++ b/src/colmap/estimators/cost_function_utils.h
@@ -1,0 +1,172 @@
+// Copyright (c), ETH Zurich and UNC Chapel Hill.
+// All rights reserved.
+//
+// Redistribution and use in source and binary forms, with or without
+// modification, are permitted provided that the following conditions are met:
+//
+//     * Redistributions of source code must retain the above copyright
+//       notice, this list of conditions and the following disclaimer.
+//
+//     * Redistributions in binary form must reproduce the above copyright
+//       notice, this list of conditions and the following disclaimer in the
+//       documentation and/or other materials provided with the distribution.
+//
+//     * Neither the name of ETH Zurich and UNC Chapel Hill nor the names of
+//       its contributors may be used to endorse or promote products derived
+//       from this software without specific prior written permission.
+//
+// THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS "AS IS"
+// AND ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO, THE
+// IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS FOR A PARTICULAR PURPOSE
+// ARE DISCLAIMED. IN NO EVENT SHALL THE COPYRIGHT HOLDERS OR CONTRIBUTORS BE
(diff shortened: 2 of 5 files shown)


src/glomap/estimators/gravity_refinement.h
```diff
@@ -40,7 +40,7 @@ class GravityRefiner {
       const std::unordered_map<frame_t, Frame>& frames,
       const std::unordered_map<image_t, Image>& images,
       std::unordered_map<image_t, colmap::PosePrior*>& image_to_pose_prior,
-      std::unordered_set<image_t>& error_prone_images);
```
B1ueber2y, line 43: The definition in the header was not aligned with the implementation.

src/glomap/estimators/gravity_refinement.cc
```diff
@@ -8,19 +8,6 @@
 namespace glomap {
 namespace {
 
-Eigen::Matrix3d GetImageAlignRot(const Image& image,
```
B1ueber2y, line 11: All the call points of this function are broken since the gravity is already on the image coordinate. They are working only because only the reference sensor is supported. This function can be simply removed.

src/glomap/math/gravity.cc
```diff
@@ -24,66 +24,42 @@ Eigen::Matrix3d GetAlignRot(const Eigen::Vector3d& gravity) {
...
+    A += g * g.transpose();
+  }
   A /= gravities.size();
 
-  // Compute the SVD of this 3x3 matrix
-  Eigen::JacobiSVD<Eigen::MatrixXd> svd(
-      A, Eigen::ComputeThinU | Eigen::ComputeThinV);
+  // The first singular vector corresponds to the principal direction.
+  Eigen::JacobiSVD<Eigen::Matrix3d> svd(A, Eigen::ComputeFullU);
+  Eigen::Vector3d average = svd.matrixU().col(0);
```
B1ueber2y, line 53: The first eigenvalue is always the largest from JacobiSVD.

src/glomap/math/gravity.cc
```diff
@@ -24,66 +24,42 @@ Eigen::Matrix3d GetAlignRot(const Eigen::Vector3d& gravity) {
...
+  return colmap::AngleAxisToRotationMatrix(Eigen::Vector3d(0, yaw, 0));
 }
 
-// Code adapted from
-// https://gist.github.com/PeteBlackerThe3rd/f73e9d569e29f23e8bd828d7886636a0
-Eigen::Vector3d AverageGravity(const std::vector<Eigen::Vector3d>& gravities) {
-  if (gravities.size() == 0) {
-    std::cerr
-        << "Error trying to calculate the average gravities of an empty set!\n";
+Eigen::Vector3d AverageGravityDirection(
```
ahojnnes: Maybe we generalize this function and call it ComputeAverageDirection or similar. We also already have ComputeAverageQuaternion on the colmap side, which does essentially the same thing but in 4d. Something for a follow-up PR.

src/glomap/math/gravity.cc
ahojnnes: Yes, I also already noticed this but left it untouched. See other comment above.

src/glomap/math/gravity.h
```diff
@@ -1,20 +1,23 @@
...
+#include <vector>
+
 #include <Eigen/Core>
 
 namespace glomap {
 
-// Get the aligment rotation matrix by QR decomposition
-// The second col of output is gravity direction
-Eigen::Matrix3d GetAlignRot(const Eigen::Vector3d& gravity);
+// Compute rotation matrix from gravity direction via Householder QR.
```
ahojnnes: nit: I guess we could be a bit more precise in the description and the name in the sense that this rotation transforms from a coordinate frame that has gravity aligned with the Y axis. Cannot come up with something great but e.g. `GravityFromUnitYRotation()` or similar.

src/glomap/math/gravity.h
```diff
@@ -1,20 +1,23 @@
...
 
-// Get the upright rotation matrix from a rotation angle
-Eigen::Matrix3d AngleToRotUp(double angle);
+// Construct rotation matrix from yaw angle (rotation about Y-axis).
+Eigen::Matrix3d RotationFromYaw(double yaw);
 
-// Estimate the average gravity direction from a set of gravity directions
-Eigen::Vector3d AverageGravity(const std::vector<Eigen::Vector3d>& gravities);
+// Compute the average direction from normalized gravity direction vectors.
+Eigen::Vector3d AverageGravityDirection(
```
ahojnnes: A unit test will be great for this, so we can be sure the refactoring didn't break the logic.

ahojnnes (approved):
Thanks, a few minor comments.

src/glomap/math/gravity.h
B1ueber2y: Also cannot come up with anything great and concise. I renamed this to GravityAlignedRotation().

src/glomap/math/gravity.cc
B1ueber2y: Done in a follow-up PR: https://github.com/colmap/colmap/pull/3845

src/glomap/math/gravity.h
B1ueber2y: Done in a follow-up PR: https://github.com/colmap/colmap/pull/3845

PR #3845, opened 2025-12-25
Migrate glomap gravity utilities into colmap and add tests. 

B1ueber2y:


ahojnnes:
nit: For averaging the unit vectors, you could consider removing the template and moving the implementation to the cc file by taking a MatrixXd or vector of VectorXd.

ahojnnes (approved):

B1ueber2y:
> cr.

Done: https://github.com/colmap/colmap/pull/3845/commits/7446a0a40e57a162db679221e01a237b5c28ac17. PTAL : )

B1ueber2y:
> > cr.
> 
> Done: [7446a0a](https://github.com/colmap/colmap/commit/7446a0a40e57a162db679221e01a237b5c28ac17). PTAL : )

I m not sure if this is the best way since this will incur copying at ``AverageDirections``. Let me know if we want to revert this change.

ahojnnes:
LGTM. Now that we have a matrix and if we also take the weights as an eigen vector, we could probably rewrite the for loops with pure Eigen expressions.

B1ueber2y:
> LGTM. Now that we have a matrix and if we also take the weights as an eigen vector, we could probably rewrite the for loops with pure Eigen expressions.

Nice idea! This is done: https://github.com/colmap/colmap/pull/3845/commits/f0817c49fd5d848a45a17cb064edec9a2f95288f

PR #3846, opened 2025-12-25
Pass database as pointer instead of reference in global mapper. Drop GlobalMapperResume.

B1ueber2y:
There is no database in global_mapper_resume. 

On the other hand, there is no test coverage for this interface, and I personally do not see why we need this. I am personally strongly into removing this interface.

### DIFF
--- a/src/colmap/scene/database_sqlite.cc
+++ b/src/colmap/scene/database_sqlite.cc
@@ -53,16 +53,18 @@ inline int SQLite3CallHelper(int result_code,
 
 #define SQLITE3_CALL(func) SQLite3CallHelper(func, __FILE__, __LINE__)
 
-#define SQLITE3_EXEC(database, sql, callback)                             \
-  {                                                                       \
-    char* err_msg = nullptr;                                              \
-    const int result_code = sqlite3_exec(                                 \
-        THROW_CHECK_NOTNULL(database), sql, callback, nullptr, &err_msg); \
-    if (result_code != SQLITE_OK) {                                       \
-      LOG(ERROR) << "SQLite error [" << __FILE__ << ", line " << __LINE__ \
-                 << "]: " << err_msg;                                     \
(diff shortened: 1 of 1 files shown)


B1ueber2y:
> I think the idea was to use a colmap reconstruction as a pose prior and initialize glomap from the given poses. @lpanaf to comment. I am fine with removing it.

In this function there is no database and thus the matches only come from the existing colmap reconstruction. I dont see why this can make the results better.

B1ueber2y:
Made an update to at least make sure that the global mapper resume does not use nullptr in its logic. The code is fixed now, ofc if we really want to keep this ``RunGlobalMapperResume`` interface (again, there is no test coverage for this and I personally do not see why this is something useful as a public interface).

src/glomap/sfm/global_mapper.h
```diff
@@ -48,7 +48,8 @@ class GlobalMapper {
   explicit GlobalMapper(const GlobalMapperOptions& options)
       : options_(options) {}
 
-  bool Solve(const colmap::Database& database,
+  // database can be nullptr if skip_retriangulation is true
+  bool Solve(const colmap::Database* database,
```
B1ueber2y, line 52: The database is only used in track retriangulation.

src/glomap/exe/global_mapper.cc
```diff
@@ -161,7 +160,7 @@ int RunGlobalMapperResume(int argc, char** argv) {
   colmap::Timer run_timer;
   run_timer.Start();
   std::unordered_map<frame_t, int> cluster_ids;
-  global_mapper.Solve(*database,
+  global_mapper.Solve(nullptr,
```
B1ueber2y: There is no sense of database in this ``RunGlobalMapperResume`` interface.

ahojnnes:
I suggest, let's remove it. We can bring it back, if we find a need for it later.

PR #3847, opened 2025-12-25
Improvements and fixes on glomap two view geometry.

B1ueber2y:
* Expose per-point sampson error and homography error from colmap and use it from glomap
* Cleanup unnecessary included (2 places) in glomap for glomap/math/two_view_geometry.h
* Move the two functions from PoseLib and GCRANSAC into empty namespace of image_pair_inliers.cc
* Remove glomap/math/two_view_geometry.h / cc


src/glomap/math/two_view_geometry.cc
```diff
@@ -1,81 +0,0 @@
...
-
-  return r2;
-}
-
-double SampsonError(const Eigen::Matrix3d& E,
-                    const Eigen::Vector3d& x1,
-                    const Eigen::Vector3d& x2) {
-  const Eigen::Vector3d Ex1 = E * x1 / x1[2];
-  const Eigen::Vector3d Etx2 = E.transpose() * x2 / x2[2];
-  const double C = Ex1.dot(x2);
```
B1ueber2y, line 62: This 3D function was buggy. x2 is normalized in the first two lines but directly used here. The 2D function was correct and equivalent to the colmap one.

PR #3850, opened 2025-12-25
Move cluster id logic to top level and remove custom glomap Frame class.

B1ueber2y:
Now that the cluster_ids stay at the top level, and we can use colmap::Frame instead.


B1ueber2y:
After this the base types of colmap and glomap are fully aligned. We can replace those glomap scattered objects with colmap::Reconstruction.

ahojnnes (approved):
Thanks, LGTM.

PR #3863, opened 2025-12-26
[Improve RA] Move stratified logic into estimators and only keep one option class.

B1ueber2y:
Dependent on https://github.com/colmap/colmap/pull/3876

This PR includes:
* move stratified logic into estimators
* use one single option class for rotation averaging and propagate it to the codebase
* largely simplify and improve the namings. Rename "trivial reconstruction" to expanded reconstruction, improve logic and logging. 
* Fix the messed up logic of has_unknown_cameras in the sfm/rotation_averaging.cc
* move rotation initializer into estimators/rotation_averaging.h

### DIFF
--- a/src/glomap/controllers/option_manager.cc
+++ b/src/glomap/controllers/option_manager.cc
@@ -287,7 +287,7 @@ void OptionManager::Parse(const int argc, char** argv) {
     config::store(config::parse_command_line(argc, argv, *desc_), vmap);
 
     if (vmap.count("help")) {
-      std::cout << "The following options can be specified via command-line:\n"
+      LOG(INFO) << "The following options can be specified via command-line:\n"
                 << *desc_;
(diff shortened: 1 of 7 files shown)


B1ueber2y:
This PR does not necessarily keep the logic since the rotation averaging code handles cameras in the rigs by converting it to trivial rigs first: https://github.com/colmap/colmap/blob/c41bfd1ee4100097730fe464e2c88df666d4f38e/src/glomap/sfm/rotation_averager.cc#L101-L197 (or in official glomap: https://github.com/colmap/glomap/blob/main/glomap/controllers/rotation_averager.cc#L80-L158).

This looks extremely fishy and should be fixed first, so marking this PR as draft for now.

B1ueber2y:
This PR is ready for review. next step would be to move rotation averaging from sfm to controller, in colmap style.

PR #3864, opened 2025-12-26
Move relative pose filtering utilities into ViewGraph.

B1ueber2y:


src/glomap/scene/view_graph.h
```diff
@@ -32,6 +32,18 @@ struct ViewGraph {
   int MarkConnectedComponents(const colmap::Reconstruction& reconstruction,
                               std::unordered_map<frame_t, int>& cluster_ids,
                               int min_num_images = -1);
+
+  // Mark image pairs as invalid if their relative rotation differs from the
+  // reconstructed rotation by more than max_angle_deg.
```
ahojnnes, line 37: Maybe add: `// Keeps existing invalid edges as invalid.`

Same for other 2 methods.

B1ueber2y:
> Thanks, ideally, we can also add some unit tests, unless we plan to remove/refactor further later.

Added some naive unit tests for ``ViewGraph``.

PR #3865, opened 2025-12-26
Fix database race conditions in mapping pipelines

ahojnnes:
Opening a database is not thread-safe due to the incurred database schema migration and PRAGMA operations, which modify the database. In particular, the concurrent incremental mappers in the hierarchical mapper currently sometimes lead to errors when opening the database access at the same time with an error "database is locked". This then silently fails the reconstruction of the leaf node cluster (as the exception is swallowed by the thread pool, which will be addressed in a separate PR).

For consistency across all mapping pipelines, we now take the database object to construct the pipelines. Likewise, the image_path is now consistently specified via the pipeline options.

### DIFF
--- a/python/examples/custom_incremental_pipeline.py
+++ b/python/examples/custom_incremental_pipeline.py
@@ -91,7 +91,7 @@ def initialize_reconstruction(
     ):
         return pycolmap.IncrementalMapperStatus.BAD_INITIAL_PAIR
     if options.extract_colors:
-        reconstruction.extract_colors_for_all_images(controller.image_path)
+        reconstruction.extract_colors_for_all_images(options.image_path)
     return pycolmap.IncrementalMapperStatus.SUCCESS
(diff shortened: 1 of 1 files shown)


src/pycolmap/sfm/incremental_mapper.cc
```diff
@@ -237,17 +241,13 @@ void BindIncrementalPipeline(py::module& m) {
 
   py::classh<IncrementalPipeline>(m, "IncrementalPipeline")
       .def(py::init<std::shared_ptr<const IncrementalPipelineOptions>,
-                    const std::string&,
-                    const std::string&,
+                    std::shared_ptr<Database>,
                     std::shared_ptr<ReconstructionManager>>(),
            "options"_a,
            "image_path"_a,
```
B1ueber2y: arguments need update.

PR #3861, opened 2025-12-26
[Improve RA] Abstract RotationAveragingProblem and RotationAveragingSolver. 

B1ueber2y:
Logics unchanged except for a bug fix.

PR #3866, opened 2025-12-26
Fix incorrect registered frame check in rotation averaging

rohangarg27:
This PR fixes an incorrect condition in `global_rotation_averaging.cc` where registered frames were skipped instead of unregistered ones during average step size computation.

**What was wrong**
The code skipped registered frames:
```cpp
if (frame.HasPose()) continue;

**FIX:**
if (!frame.HasPose()) continue;


B1ueber2y:
Thanks! This has been addressed in https://github.com/colmap/colmap/pull/3861 (will be merged soon).

rohangarg27:
Got it, thanks for confirming! I’ll close this PR then.

PR #3867, opened 2025-12-26
Fix 2D association issue at AddPoint3D with point3D_id.

B1ueber2y:
Crucial bug fix. Surprised that this stays in colmap for so long.

### DIFF
--- a/src/colmap/retrieval/visual_index_test.cc
+++ b/src/colmap/retrieval/visual_index_test.cc
@@ -46,7 +46,7 @@ class ParameterizedVisualIndexTests
 TEST_P(ParameterizedVisualIndexTests, Nominal) {
   const auto [desc_dim, embedding_dim] = GetParam();
 
-  SetPRNGSeed(0);
+  SetPRNGSeed(1);
 
   {
     auto visual_index = VisualIndex::Create(desc_dim, embedding_dim);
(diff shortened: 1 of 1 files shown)


src/colmap/scene/reconstruction.cc
```diff
@@ -359,6 +359,18 @@ void Reconstruction::AddImageWithTrivialFrame(class Image image,
...
   max_point3D_id_ = std::max(max_point3D_id_, point3D_id);
+
+  for (const auto& track_el : point3D.track.Elements()) {
+    class Image& image = Image(track_el.image_id);
+    const Point2D& point2D = image.Point2D(track_el.point2D_idx);
+    if (point2D.HasPoint3D()) {
+      THROW_CHECK_EQ(point2D.point3D_id, point3D_id);
+    } else {
+      image.SetPoint3DForPoint2D(track_el.point2D_idx, point3D_id);
+    }
```
ahojnnes, line 370: Should we do the same for the other AddPoint3D method below or ideally even consolidate the two? I guess the bug was introduced due to not realizing that there are two parallel methods almost doing the same thing?

B1ueber2y: Indeed. I consolidated the two methods. PTAL : )

PR #3869, opened 2025-12-26
Propagate exceptions from ThreadPool::Wait

ahojnnes:
In most call sites of ThreadPool, we only await all tasks to finish through the Wait() method. However, the Wait() method previously silently swallowed exceptions as opposed to `std::future::get()`. This is not intended and leads to undetected errors, e.g. see https://github.com/colmap/colmap/pull/3865.

### DIFF
--- a/python/examples/custom_incremental_pipeline.py
+++ b/python/examples/custom_incremental_pipeline.py
@@ -91,7 +91,7 @@ def initialize_reconstruction(
     ):
         return pycolmap.IncrementalMapperStatus.BAD_INITIAL_PAIR
     if options.extract_colors:
-        reconstruction.extract_colors_for_all_images(controller.image_path)
+        reconstruction.extract_colors_for_all_images(options.image_path)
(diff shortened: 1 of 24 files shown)


src/colmap/util/threading.cc
```diff
@@ -222,6 +223,13 @@ void ThreadPool::Wait() {
     finished_condition_.wait(
         lock, [this]() { return tasks_.empty() && num_active_workers_ == 0; });
   }
+
+  // Check remaining futures to propagate any exceptions.
+  while (!future_checkers_.empty()) {
+    auto checker = std::move(future_checkers_.back());
+    future_checkers_.pop_back();
+    checker();
```
B1ueber2y, line 242: Should release the mutex here?

ahojnnes: Not sure why but I probably don't see the issue? We need exclusive access here to make sure no other thread is adding a new task.

PR #3871, opened 2025-12-26
Use colmap ObservationManager for track filtering in glomap.

B1ueber2y:
The previous glomap logic only used ``SetElements``, which does not handle the cleanup of observations. Switching to the use of ObservationManager is smooth and naturally solves the issue. The track_filter.h/cc in glomap is removed after the switching. The logic except for the observation handling issue stays unchanged.

### DIFF
--- a/src/colmap/sfm/observation_manager.cc
+++ b/src/colmap/sfm/observation_manager.cc
@@ -410,6 +410,10 @@ size_t ObservationManager::FilterPoints3DWithLargeReprojectionError(
     const ReprojectionErrorType error_type) {
   size_t num_filtered_observations = 0;
 
+  // Precompute squared/converted thresholds to avoid redundant computations.
+  const double max_squared_error = max_error * max_error;
+  const double max_angular_error_rad = DegToRad(max_error);
+
   for (const auto point3D_id : point3D_ids) {
     if (!reconstruction_.ExistsPoint3D(point3D_id)) {
       continue;
@@ -438,7 +442,7 @@ size_t ObservationManager::FilterPoints3DWithLargeReprojectionError(
         case ReprojectionErrorType::PIXEL: {
           const double squared_error = CalculateSquaredReprojectionError(
               point2D.xy, point3D.xyz, image.CamFromWorld(), camera);
-          should_filter = squared_error > max_error * max_error;
+          should_filter = squared_error > max_squared_error;
           observation_error = std::sqrt(squared_error);
           break;
         }
@@ -460,14 +464,14 @@ size_t ObservationManager::FilterPoints3DWithLargeReprojectionError(
               point3D_in_cam.hnormalized().head<2>();
           const double squared_error =
               (reproj_point - *cam_point).squaredNorm();
-          should_filter = squared_error > max_error * max_error;
+          should_filter = squared_error > max_squared_error;
           observation_error = std::sqrt(squared_error);
           break;
         }
         case ReprojectionErrorType::ANGULAR: {
           const double error = CalculateAngularReprojectionError(
               point2D.xy, point3D.xyz, image.CamFromWorld(), camera);
-          should_filter = error > DegToRad(max_error);
+          should_filter = error > max_angular_error_rad;
           observation_error = RadToDeg(error);
           break;
         }


src/colmap/sfm/observation_manager.h
```diff
@@ -105,9 +105,30 @@ class ObservationManager {
...
       double min_tri_angle, const std::unordered_set<point3D_t>& point3D_ids);
+  // Filter observations with large reprojection error.
+  //
+  // @param max_reproj_error     Maximum reprojection error threshold.
+  // @param point3D_ids          The points to be filtered.
+  // @param use_normalized_error If true, compute error in normalized camera
+  //                             coordinates; otherwise in pixel coordinates.
+  //
+  // @return                     The number of filtered observations.
   size_t FilterPoints3DWithLargeReprojectionError(
```
ahojnnes, line 123: How about keeping only one method with three filtering types switched by: `enum FilterPointsErrorType { PIXEL, NORMALIZED, ANGULAR }`? Then, we can reuse the same logic except for the error computation?

ahojnnes (approved):
Thank you! Ideally, we cover the new filtering functionality with minimal tests in the observation manager.

B1ueber2y:
Realizing that this is not the only place where glomap does not respect observation handling. For example, ``SetElements`` without setting Point3D for 2d points is also the case in ``TrackEstablishment``. This is probably due to the previous usage of scattered variables, and let s fix it in the follow-ups.

src/colmap/sfm/observation_manager.h
B1ueber2y: Done.

src/glomap/sfm/global_mapper.cc
```diff
@@ -232,14 +234,15 @@ bool GlobalMapper::Solve(const colmap::Database* database,
...
+      colmap::ObservationManager obs_manager(reconstruction);
       bool status = true;
       size_t filtered_num = 0;
       while (status && ite < options_.num_iteration_bundle_adjustment) {
         double scaling = std::max(3 - ite, 1);
-        filtered_num += TrackFilter::FilterTracksByReprojection(
-            view_graph,
-            reconstruction,
-            scaling * options_.inlier_thresholds.max_reprojection_error);
+        filtered_num += obs_manager.FilterPoints3DWithLargeReprojectionError(
```
B1ueber2y, line 272: There is a behavior change here. In glomap the function returns number of filtered points, while in colmap we return number of filtered observations. This only makes the stopping criteria tighter and should not an issue in practice.

src/colmap/sfm/observation_manager.h
```diff
@@ -105,9 +110,22 @@ class ObservationManager {
...
   size_t FilterPoints3DWithSmallTriangulationAngle(
       double min_tri_angle, const std::unordered_set<point3D_t>& point3D_ids);
+
+  // Filter observations with large reprojection error.
+  //
+  // @param max_error       Maximum error threshold. For PIXEL and NORMALIZED,
+  //                        this is the reprojection error. For ANGULAR, this
+  //                        is the angular error in degrees.
+  // @param point3D_ids     The points to be filtered.
+  // @param error_type      Type of error metric to use. For ANGULAR, uses a
```
B1ueber2y: This relaxation for angular error is quite dirty, but I dont see a cleaner alternative to keep the glomap logic unchanged.

ahojnnes (approved):
Thanks, you can probably avoid sqrt'ing and the acos by only computing it when not filtering. Otherwise LGTM.

src/colmap/sfm/observation_manager.h
B1ueber2y: I made an update to move this relaxation logic to the top level of glomap: https://github.com/colmap/colmap/pull/3871/commits/908a11bc0d48745227a7bcd9df5fd7d80034e12f. Also switched to CalculateAngularError from colmap. PTAL

src/colmap/sfm/observation_manager.cc
```diff
@@ -425,20 +423,60 @@ size_t ObservationManager::FilterPoints3DWithLargeReprojectionError(
...
       const Image& image = reconstruction_.Image(track_el.image_id);
-      const struct Camera& camera = *image.CameraPtr();
+      const Camera& camera = *image.CameraPtr();
       const Point2D& point2D = image.Point2D(track_el.point2D_idx);
-      const double squared_reproj_error = CalculateSquaredReprojectionError(
-          point2D.xy, point3D.xyz, image.CamFromWorld(), camera);
-      if (squared_reproj_error > max_squared_reproj_error) {
+
+      bool should_filter = false;
+      double observation_error = 0.0;
```
ahojnnes, line 439: You can extract and perform pure the max error here to avoid redundant DegToRad etc.

B1ueber2y: Done.

PR #3875, opened 2025-12-27
Catch AggregateException for ThreadPool.

B1ueber2y:
To make the naive try catch work again. The current PR version wont drain all errors at try catch.

### DIFF
--- a/src/colmap/util/threading.h
+++ b/src/colmap/util/threading.h
@@ -186,8 +186,8 @@ class AggregateException : public std::exception {
   explicit AggregateException(std::vector<std::exception_ptr> excs)
       // NOLINTNEXTLINE(bugprone-throw-keyword-missing)
       : exceptions_(std::move(excs)) {
-    message_ = std::to_string(exceptions_.size()) +
-               " task(s) threw exception(s):\n";
+    message_ =
+        std::to_string(exceptions_.size()) + " task(s) threw exception(s):\n";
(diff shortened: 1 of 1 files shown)


src/colmap/util/threading.cc
```diff
@@ -194,6 +194,9 @@ ThreadPool::ThreadPool(const int num_threads)
 ThreadPool::~ThreadPool() {
   try {
```
ahojnnes, line 196: nit: I should have done this already but a short comment that the destructor shouldn't throw and that's why we do a fatal log.

src/colmap/util/threading.h
```diff
@@ -178,6 +179,28 @@ class Thread {
...
 };
 
+// Exception that aggregates multiple exceptions from parallel tasks.
+// Thrown by ThreadPool::Wait() when one or more tasks threw exceptions.
+class AggregateException : public std::exception {
+ public:
+  explicit AggregateException(std::vector<std::exception_ptr> excs)
+      // NOLINTNEXTLINE(bugprone-throw-keyword-missing)
+      : exceptions_(std::move(excs)) {
+    message_ = "One or more tasks threw exceptions (" +
```
ahojnnes: We could concatenate all the exception messages separated by a new line?

ahojnnes (approved):
Great improvement!

PR #3876, opened 2025-12-27
Improve the horrendous logic and naming in rotation initializer.

B1ueber2y:
As a individual substep of https://github.com/colmap/colmap/pull/3863. Will be moved inside the rotation averaging logic in the [Improve RA] PR. 

This function was written in a quite crazy way that this has to be a stand-alone PR.

### DIFF
--- a/src/colmap/util/threading.cc
+++ b/src/colmap/util/threading.cc
@@ -192,8 +192,12 @@ ThreadPool::ThreadPool(const int num_threads)
 }
 
 ThreadPool::~ThreadPool() {
+  // Destructors should not throw, so we catch and log fatal instead.
   try {
     Stop();
+  } catch (const AggregateException& e) {
+    LOG(FATAL) << "Uncaught exceptions in thread pool destructor: "
+               << e.exceptions().size() << " exception(s) not handled";
   } catch (...) {
     LOG(FATAL) << "Uncaught exception in thread pool destructor";
   }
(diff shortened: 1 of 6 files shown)


src/glomap/estimators/rotation_initializer.cc
```diff
@@ -4,124 +4,106 @@
 
 namespace glomap {
 
-bool ConvertRotationsFromImageToRig(
+bool InitializeRigRotationsFromImages(
```
ahojnnes, line 7: Some of this is similar to UpdateRigAndCameraCalibsFromReconstruction. Not too important to consolidate. Just FYI.

src/glomap/estimators/rotation_initializer.h
```diff
@@ -6,9 +6,11 @@
 
 namespace glomap {
 
-// Initialize the rotations of the rigs from the images
-bool ConvertRotationsFromImageToRig(
-    const std::unordered_map<image_t, Rigid3d>& cam_from_worlds,
+// Initialize rig rotations by averaging per-image rotations.
+// Estimates cam_from_rig for cameras with unknown calibration,
+// then computes rig_from_world for each frame.
+bool InitializeRigRotationsFromImages(
```
ahojnnes, line 12: Maybe a simple test can be added, where we synthesize a reconstruction and check that the initialized rotations match the synthesized ones?

B1ueber2y: Lets add the test in a later PR, as this file is planned to be moved into rotation averaging in https://github.com/colmap/colmap/pull/3863. Trying to make the merging easier.

PR #3878, opened 2025-12-27
Add Reconstruction::IsValid to check if a reconstruction object is not broken.

B1ueber2y:
In such significant refactorization of glomap, I believe we have reached a point where we keep hitting broken reconstruction (e.g., https://github.com/colmap/colmap/pull/3867 and https://github.com/colmap/colmap/pull/3877). I think its right time to add a method with good verbal messages. This can be very helpful for improving code safety and easy debugging. A small test with three nominal wrong operations are included.

### DIFF
--- a/src/colmap/scene/reconstruction.cc
+++ b/src/colmap/scene/reconstruction.cc
@@ -131,17 +131,23 @@ std::unordered_set<point3D_t> Reconstruction::Point3DIds() const {
   return point3D_ids;
 }
 
-bool Reconstruction::CheckIsValid() const {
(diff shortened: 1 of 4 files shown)


B1ueber2y:
The check mainly includes three parts with all clear verbal messages:

1) reference consistency in cameras-rigs-images-frames.
2) 2d-3d point track consistency.
3) registered frames must have poses.

Not sure if i missed some other constraints but these are already good to defend/debug against most cases.

ahojnnes (approved):

PR #3868, opened 2025-12-26
Directly use COLMAP bundle adjustment interface for glomap BA.

B1ueber2y:
Dependent on https://github.com/colmap/colmap/pull/3867

This PR includes:
* add Huber loss to colmap BA option
* add an additional option ``constant_rig_from_world_rotation`` to colmap BA option
* Simplify glomap BA to use colmap BA while keeping the original glomap interfaces for now.
* rename track_id, track to point3D_id, point3D as pinpointed in https://github.com/colmap/colmap/pull/3853.

PR #3881, opened 2025-12-27
Abstract base option manager and consolidate usage.

B1ueber2y:
In preparation to share more options, such as bundle adjustment (https://github.com/colmap/colmap/pull/3868).


B1ueber2y:
> Maybe move/split the existing tests to base_option_manager_test.cc?

Thanks. This is done: https://github.com/colmap/colmap/pull/3881/commits/e9082de9241a140248579143427cf55f4f3569e7

PR #3882, opened 2025-12-28
ImagePair inherits colmap::TwoViewGeometry. Remove image id fields.

B1ueber2y:
Big first step towards eliminating the redundancy of glomap::ImagePair and ultimately making ViewGraph (or maybe better named PoseGraph) into colmap. The next step will be gradually removing those fields similar to what we did before for Image and Frame. 

The logic is kept fully unchanged.


src/glomap/io/pose_io.cc
```diff
@@ -95,14 +95,13 @@ void ReadRelPose(const std::string& file_path,
       pose_rel.translation[i] = std::stod(item);
     }
 
-    if (view_graph.image_pairs.find(pair_id) == view_graph.image_pairs.end()) {
-      view_graph.image_pairs.insert(
```
B1ueber2y, line 99: This looks like a bug? The view_graph.config is by default UNDEFINED and is not set here. I kept the logic unchanged in this PR but got confused here.

ahojnnes (approved):

src/glomap/io/pose_io.cc
B1ueber2y: Looking more closely, this should indeed be a bug. But in practice this branch is never reached in glomap codebase. Fixing it anyway for safety: https://github.com/colmap/colmap/pull/3882/commits/dd2f546d95087ca6a36bfd784f8bf5f935c39e1f

PR #3883, opened 2025-12-28
Abstract general minimum spanning tree algorithm in colmap.

B1ueber2y:
* Abstract general minimum spanning tree algorithm in colmap. Add small nominal tests.
* Inline the remaining glomap MST logic to rotation averaging (the only call site)
* remove glomap/math and colmap_glomap_math target.

### DIFF
--- a/src/colmap/math/spanning_tree.cc
+++ b/src/colmap/math/spanning_tree.cc
@@ -42,7 +42,7 @@ using BoostGraph =
                           boost::vecS,
                           boost::undirectedS,
                           boost::no_property,
-                          boost::property<boost::edge_weight_t, double>>;
+                          boost::property<boost::edge_weight_t, float>>;
 using EdgeDescriptor = boost::graph_traits<BoostGraph>::edge_descriptor;
 
 // Build parent pointers from adjacency list using BFS from root.
@@ -54,7 +54,7 @@ void BuildParentsFromAdjacencyList(
   parents.assign(num_nodes, -1);
   parents[root] = root;
 
-  std::vector<bool> visited(num_nodes, false);
+  std::vector<char> visited(num_nodes, false);
   visited[root] = true;
 
   std::queue<int> queue;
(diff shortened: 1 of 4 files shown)


src/colmap/math/spanning_tree.cc
```diff
@@ -0,0 +1,147 @@
...
+// Build parent pointers from adjacency list using BFS from root.
+void BuildParentsFromAdjacencyList(
+    const std::vector<std::vector<int>>& adjacency_list,
+    int root,
+    std::vector<int>& parents) {
+  const int num_nodes = static_cast<int>(adjacency_list.size());
+  parents.assign(num_nodes, -1);
+  parents[root] = root;
+
+  std::vector<bool> visited(num_nodes, false);
```
ahojnnes: Bool is a bitset, maybe make it char?

src/colmap/math/spanning_tree.h
```diff
@@ -0,0 +1,80 @@
...
+// - root: The root node for the resulting tree (default 0)
+//
+// Returns a SpanningTree with parent pointers rooted at the specified root.
+// If the graph is disconnected, only the component containing root is included.
+//
+// Uses Kruskal's algorithm with negated weights to find maximum spanning tree.
+SpanningTree ComputeMaximumSpanningTree(
+    int num_nodes,
+    const std::vector<std::pair<int, int>>& edges,
+    const std::vector<double>& weights,
```
ahojnnes: nit: double is maybe a bit overkill, but then we are using double all over the place in colmap where we would be fine with float instead. We could make a start and use float here 😃

src/glomap/estimators/rotation_averaging.cc
```diff
@@ -29,6 +29,59 @@ bool AllSensorsFromRigKnown(
...
+// Returns the root image_id and populates the parents map.
+image_t ComputeMaximumSpanningTree(
+    const ViewGraph& view_graph,
+    const std::unordered_map<image_t, Image>& images,
+    std::unordered_map<image_t, image_t>& parents) {
+  // Build mapping between image_id and contiguous indices.
+  std::unordered_map<image_t, int> image_id_to_idx;
+  std::vector<image_t> idx_to_image_id;
+  image_id_to_idx.reserve(images.size());
+  idx_to_image_id.reserve(images.size());
```
ahojnnes, line 42: nit: image_idx_to_id (for consistency in other code)

B1ueber2y: Other code also used ``image_id_to_idx``? ``image_idx_to_id`` seems never used in the codebase.

PR #3886, opened 2025-12-28
Simplify the logic and naming of reconstruction pruning. 

B1ueber2y:
After all skip_pruning is by default set to true. I wonder when this block of code was tested last time. 

The logic change of this PR should be equivalent. The follow up would be to move both "reconstruction pruning" and "view graph manipulation" to sth like "global_mapper_impl".


src/glomap/processors/reconstruction_pruning.cc
```diff
@@ -1,128 +1,233 @@
...
-  // Establish the visibility graph
-  size_t counter = 0;
-  ViewGraph visibility_graph_frame;
-  std::vector<int> pair_count;
-  for (auto& [pair_id, count] : pair_covisibility_count) {
-    // since the relative pose is only fixed if there are more than 5 points,
-    // then require each pair to have at least 5 points
-    if (count >= 5) {
-      counter++;
-      const auto [image_id1, image_id2] = colmap::PairIdToImagePair(pair_id);
```
B1ueber2y, line 46: Note: this is actually meant to be frame_id1 and frame_id2 rather than image_ids, since this visibility graph is on frames. The update in this PR removes the usage of ViewGraph throughout this function.

src/glomap/processors/reconstruction_pruning.cc
```diff
@@ -1,128 +1,233 @@
...
-      break;
-    }
+    edge_weights[pair_id] = count;
+    weight_values.push_back(count);
   }
+  LOG(INFO) << "Established visibility graph with " << edge_weights.size()
+            << " pairs";
 
-  ViewGraph visibility_graph;
-  for (auto& [pair_id, frame_image_pair] : visibility_graph_frame.image_pairs) {
```
B1ueber2y: Note: here, a new visibility graph is built on top of images rather than frames. And the logic is to connect the first image in frames if the two frames are connected. This is totally unnecessary since the final UnionFind works on frames by converting this back...

src/glomap/processors/view_graph_manipulation.cc
```diff
@@ -1,119 +1,10 @@
...
-    ViewGraph& view_graph,
-    colmap::Reconstruction& reconstruction,
-    std::unordered_map<frame_t, int>& cluster_ids,
-    StrongClusterCriteria criteria,
-    double min_thres,
-    int min_num_images) {
-  view_graph.KeepLargestConnectedComponents(reconstruction);
-
-  // Construct the initial cluster by keeping the pairs with weight > min_thres
-  colmap::UnionFind<image_pair_t> uf;
```
B1ueber2y, line 19: Note: this is actually UnionFind<frame_t> rather than image_pair_t...

src/glomap/processors/reconstruction_pruning.h
```diff
@@ -6,11 +6,23 @@
...
+// The clustering uses an adaptive threshold (median - MAD) to handle varying
+// covisibility distributions. Frames are grouped using union-find, with
+// iterative refinement to merge clusters connected by multiple weaker edges.
+//
+// Args:
+//   reconstruction: The reconstruction containing frames and 3D points.
+//
+// Returns:
+//   Map from frame_id to cluster_id.
+std::unordered_map<frame_t, int> PruneWeaklyConnectedFrames(
```
B1ueber2y, line 34: Not sure if this is a good name. Probably ``ClusterFramesByCovisibility`` is better?

B1ueber2y:
I think the purpose of this function is still quite convoluted. I added a detailed description in the header. Probably in the future this should be further refactored, but this PR should be a good step towards it.

src/glomap/sfm/global_mapper.cc
```diff
@@ -341,8 +341,8 @@ bool GlobalMapper::Solve(const colmap::Database* database,
     colmap::Timer run_timer;
     run_timer.Start();
 
-    // Prune weakly connected images
-    PruneWeaklyConnectedImages(reconstruction, cluster_ids);
+    // Prune weakly connected frames
```
ahojnnes:
```suggestion
```

src/glomap/processors/reconstruction_pruning.h
ahojnnes: Maybe it makes sense to decompose the steps into individual sub-functions. Then it also becomes easier to find a good name?

src/glomap/processors/reconstruction_pruning.cc
```diff
@@ -1,128 +1,234 @@
...
     if (track.track.Length() <= 2) continue;
 
     for (size_t i = 0; i < track.track.Length(); i++) {
       const image_t image_id1 = track.track.Element(i).image_id;
       const frame_t frame_id1 = reconstruction.Image(image_id1).FrameId();
 
-      frame_observation_count[frame_id1]++;
+      nodes.insert(frame_id1);
       for (size_t j = i + 1; j < track.track.Length(); j++) {
         const image_t image_id2 = track.track.Element(j).image_id;
```
ahojnnes: This will over-count frames with multiple images in the same track or maybe this is desired?

src/glomap/processors/reconstruction_pruning.cc
```diff
@@ -1,128 +1,234 @@
...
-      pair_count.push_back(count);
-    }
-  }
-  LOG(INFO) << "Established visibility graph with " << counter << " pairs";
+  // Step 2: Filter edges to keep only reliable connections.
+  // Require at least 5 shared points (needed for stable relative pose).
+  std::unordered_map<frame_pair_t, int> edge_weights;
+  std::vector<int> weight_values;
+  for (const auto& [pair_id, count] : frame_covisibility_count) {
+    if (count < 5) continue;
```
ahojnnes: Suggest extracting these magical thresholds into constexpr at the beginning or even make them part of an options struct?

src/glomap/processors/reconstruction_pruning.cc
```diff
@@ -1,128 +1,234 @@
...
+      reconstruction.DeRegisterFrame(frame_id);
+    }
+  }
+  LOG(INFO) << "Kept " << largest_cc.size() << " frames in largest component";
+
+  // Filter to keep only edges within the largest component.
+  for (auto it = edge_weights.begin(); it != edge_weights.end();) {
+    const auto [frame_id1, frame_id2] = colmap::PairIdToImagePair(it->first);
+    if (largest_cc.count(frame_id1) == 0 || largest_cc.count(frame_id2) == 0) {
+      it = edge_weights.erase(it);
```
ahojnnes, line 190: Shouldn't we also filter the weight_values here? In other words, should we instead extract the weight_values vector after this filtering?

src/glomap/processors/reconstruction_pruning.cc
```diff
@@ -1,128 +1,234 @@
...
 
-  // sort the pair count
-  std::sort(pair_count.begin(), pair_count.end());
-  double median_count = pair_count[pair_count.size() / 2];
+  // Step 4: Compute adaptive threshold using median minus median absolute
+  // deviation (MAD).
+  std::sort(weight_values.begin(), weight_values.end());
+  const double median = weight_values[weight_values.size() / 2];
+  const double threshold =
+      std::max(median - ComputeMedianAbsoluteDeviation(weight_values), 20.0);
```
ahojnnes: We have a Median function in colmap/math and we could create a MedianAbsoluteDeviation function there (which simultaneously computes median and MAD).

src/glomap/processors/reconstruction_pruning.cc
```diff
@@ -1,128 +1,234 @@
...
-void PruneWeaklyConnectedImages(colmap::Reconstruction& reconstruction,
-                                std::unordered_map<frame_t, int>& cluster_ids,
-                                int min_num_images,
-                                int min_num_observations) {
-  // Prepare the 2d-3d correspondences
-  std::unordered_map<image_pair_t, int> pair_covisibility_count;
-  std::unordered_map<frame_t, int> frame_observation_count;
+namespace {
+
+// Alias for clarity: we're working with frame pairs, not image pairs.
```
ahojnnes, line 10: Great!

src/glomap/processors/reconstruction_pruning.cc
```diff
@@ -1,128 +1,234 @@
...
+
+// Computes the Median Absolute Deviation of a sorted vector.
+double ComputeMedianAbsoluteDeviation(const std::vector<int>& sorted_values) {
+  const double median = sorted_values[sorted_values.size() / 2];
+  std::vector<int> abs_deviations(sorted_values.size());
+  for (size_t i = 0; i < sorted_values.size(); i++) {
+    abs_deviations[i] = std::abs(sorted_values[i] - static_cast<int>(median));
+  }
+  std::sort(abs_deviations.begin(), abs_deviations.end());
+  return abs_deviations[abs_deviations.size() / 2];
```
ahojnnes: Could use twice the existing Median function?

src/glomap/processors/reconstruction_pruning.cc
```diff
@@ -1,128 +1,234 @@
...
+}
+
+// Finds connected components and returns the largest one.
+std::unordered_set<frame_t> FindLargestConnectedComponent(
```
ahojnnes: Maybe move as a helper to colmap/graph with some basic tests?

src/glomap/processors/reconstruction_pruning.cc
B1ueber2y: This was consistent with the previous code. I see a good argument in either way, so am not sure if this is intentional.

src/glomap/processors/reconstruction_pruning.cc
B1ueber2y: I moved five magic numbers as constexpr at the start of the file.

src/glomap/processors/reconstruction_pruning.cc
B1ueber2y: Done here: https://github.com/colmap/colmap/pull/3886/commits/e76b1e91f7dfd6eb54789f8907d82d29b0d9432c. With this new feature from colmap math we could also simplify the ViewGraph a bit.

B1ueber2y: PTAL.

src/glomap/processors/reconstruction_pruning.h
B1ueber2y: Yes. Lets keep this for future PR. This PR is already getting a bit huge and suffice in unblocking the removal of weight from glomap::ImagePair : )

src/glomap/processors/reconstruction_pruning.cc
ahojnnes: OK, maybe add a TODO?

src/colmap/math/connected_components.h
```diff
@@ -0,0 +1,104 @@
...
+  size_t largest_size = 0;
+  for (const auto& [root, members] : components) {
+    if (members.size() > largest_size) {
+      largest_size = members.size();
+      largest_root = root;
+    }
+  }
+
+  const auto& largest = components[largest_root];
+  return std::unordered_set<T>(largest.begin(), largest.end());
```
ahojnnes: Suggest returning a vector here for consistency with the function above. If the caller needs a set, we should let them convert it.

src/glomap/processors/reconstruction_pruning.h
B1ueber2y: Added a TODO here.

src/glomap/processors/reconstruction_pruning.cc
B1ueber2y: Done.

src/colmap/math/connected_components.h
B1ueber2y: Indeed. Done.

PR #3888, opened 2025-12-28
Add rotation averaging controller in colmap and consolidate glomap exe functions.

B1ueber2y:
Final PR of the rotation averaging update serie. 

This PR includes:
* Add rotation averaging controller in glomap, which takes in a database and runs view graph calibration, relative pose estimation, and rotation averaging. Add nominal test. 
* Move rotation averaging function from sfm to estimators. 
* Consolidate the two functions in glomap/exe into glomap.h/cc, in preparation for being moved to colmap/exe. 

The glomap rotation averaging test is kept in the sfm folder, because it actually uses global mapper and skips the steps after rotation averaging. To be able to really achieve a stand-alone case of rotation averaging, we would need to handle the IO. For now this is done by reading in the files with glomap/pose_io.h but in the future this can be done better by writing a solid I/O for ViewGraph (or probably renamed PoseGraph in the future and live in colmap).

### DIFF
--- a/src/colmap/controllers/rotation_averaging.cc
+++ b/src/colmap/controllers/rotation_averaging.cc
@@ -54,7 +54,7 @@ void RotationAveragingController::Run() {
   // Read pose priors from database.
   std::vector<PosePrior> pose_priors = database_->ReadAllPosePriors();
 
-  if (view_graph.image_pairs.empty()) {
+  if (view_graph.Empty()) {
     LOG(ERROR) << "Cannot continue without image pairs";
     return;
   }
--- a/src/glomap/exe/glomap.cc
+++ b/src/glomap/exe/glomap.cc
@@ -99,7 +99,7 @@ int RunGlobalMapper(int argc, char** argv) {
 
   std::vector<colmap::PosePrior> pose_priors = database->ReadAllPosePriors();
 
-  if (view_graph.image_pairs.empty()) {
+  if (view_graph.Empty()) {
     LOG(ERROR) << "Can't continue without image pairs";
     return EXIT_FAILURE;
   }
@@ -245,7 +245,7 @@ int RunRotationAverager(int argc, char** argv) {
   colmap::Timer run_timer;
   run_timer.Start();
   if (!SolveRotationAveraging(
-          view_graph, reconstruction, pose_priors, rotation_averager_options)) {
+          rotation_averager_options, view_graph, reconstruction, pose_priors)) {
     LOG(ERROR) << "Failed to solve global rotation averaging";
     return EXIT_FAILURE;
   }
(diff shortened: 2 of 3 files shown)


B1ueber2y:
In the future my idea is to use the corresponding colmap::Controller in the exe/glomap.cc that stays in colmap folder.

src/glomap/estimators/rotation_averaging.h
```diff
@@ -106,4 +106,13 @@ bool InitializeRigRotationsFromImages(
...
     colmap::Reconstruction& reconstruction);
 
+// High-level rotation averaging solver that handles rig expansion.
+// For cameras with unknown cam_from_rig, first estimates their orientations
+// independently using an expanded reconstruction, then initializes the
+// cam_from_rig and runs rotation averaging on the original reconstruction.
+bool SolveRotationAveraging(ViewGraph& view_graph,
+                            colmap::Reconstruction& reconstruction,
+                            std::vector<colmap::PosePrior>& pose_priors,
+                            const RotationEstimatorOptions& options);
```
ahojnnes: nit: suggest moving the options first

src/colmap/controllers/rotation_averaging.h
```diff
@@ -0,0 +1,77 @@
...
+class RotationAveragingController : public BaseController {
+ public:
+  RotationAveragingController(const RotationAveragingControllerOptions& options,
+                              std::shared_ptr<Database> database,
+                              std::shared_ptr<Reconstruction> reconstruction);
+
+  void Run() override;
+
+  // Access the result after Run() completes.
+  const std::shared_ptr<Reconstruction>& GetReconstruction() const {
```
ahojnnes: Why do we need this? The injected reconstruction is also the output?

ahojnnes (approved):

src/glomap/sfm/rotation_averaging_test.cc
```diff
@@ -1,4 +1,4 @@
-#include "glomap/sfm/rotation_averager.h"
+#include "glomap/estimators/rotation_averaging.h"
```
ahojnnes, line 1: This should go into estimators instead?

B1ueber2y: Despite its name, this test is not only about rotation averaging. Internally it uses global mapper with all the steps after rotation averaging skipped, so I think it belongs more to the sfm. I think to be able to test the rotation estimator module alone we would need to write a view graph synthesizing module, which can benefit from the ongoing ViewGraph migration (first replace ImagePair with TwoViewGeometry, and then move ViewGraph, or renamed PoseGraph, into colmap scene).

PR #3889, opened 2025-12-28
Add type annotations to python example scripts

ahojnnes:


python/examples/custom_bundle_adjustment.py
```diff
@@ -7,11 +7,17 @@
 import collections
 import copy
 
+import pyceres
```
B1ueber2y: pycolmap does not list pyceres as an explicit dependency: https://github.com/colmap/colmap/pull/2985, so ``pyceres.SolverSummary`` would be ``pycolmap.pyceres.SolverSummary``?

python/examples/custom_bundle_adjustment.py
```diff
@@ -11,22 +11,29 @@
...
     )
     summary = bundle_adjuster.solve()
     # Alternatively, you can customize the existing problem or options as:
-    # import pyceres
     # solver_options = ba_options.create_solver_options(
     #     ba_config, bundle_adjuster.problem
     # )
-    # summary = pyceres.SolverSummary()
-    # pyceres.solve(solver_options, bundle_adjuster.problem, summary)
+    # summary = pycolmap.pyceres.SolverSummary()
```
B1ueber2y, line 27: These two commented lines should not be updated, since we did not bind pyceres.Problem and pyceres.Solve in pycolmap. Here the logic was that we can either use the colmap internal solve to return the summary (binded), or alternatively rely on pyceres.Solve directly on the auto binded problem from bundle adjuster.

PR #3893, opened 2025-12-29
Add Python type checking for all code

ahojnnes:


python/examples/custom_incremental_pipeline.py
```diff
@@ -389,10 +389,10 @@ def main(
         logging.fatal(f"Image path does not exist: {image_path}")
     output_path.mkdir(exist_ok=True, parents=True)
     reconstruction_manager = pycolmap.ReconstructionManager()
-    if input_path is not None and input_path != "":
-        reconstruction_manager.read(input_path)
+    if input_path:
+        reconstruction_manager.read(str(input_path))
 
-    with pycolmap.Database.open(database_path) as database:
+    with pycolmap.Database.open(str(database_path)) as database:
```
sarlinpe, line 395: The cast to str really isn't nice, I think that we should fix the stubs instead - let me look into this.

benchmark/reconstruction/evaluation/utils.py
```diff
@@ -753,8 +777,8 @@ def compute_avg_auc(scene_metrics: dict[str, SceneMetrics]) -> list[float]:
 
 
 def diff_metrics(
-    metrics_a: dict[str, dict[str, SceneMetrics]],
-    metrics_b: dict[str, dict[str, SceneMetrics]],
+    metrics_a: dict[str, dict[str, dict[str, SceneMetrics]]],
```
sarlinpe: Consider defining type aliases to increase readability, e.g.
```python
from typing import TypeAlias
MetricsByScene: TypeAlias = dict[str, SceneMetrics]
MetricsByCatByScene: TypeAlias = dict[str, MetricsByScene]
MetricsByDatasetByCatByScene: TypeAlias = dict[str, MetricsByCatByScene]
```
And maybe rename `SceneMetrics` to `Metrics`?

python/examples/panorama_sfm.py
```diff
@@ -304,7 +315,7 @@ def run(args: argparse.Namespace) -> None:
     pycolmap.extract_features(
         database_path,
         image_dir,
-        reader_options={"mask_path": mask_dir},
+        reader_options=pycolmap.ImageReaderOptions({"mask_path": mask_dir}),
```
sarlinpe: Consider using kwargs:
```suggestion
        reader_options=pycolmap.ImageReaderOptions(mask_path=mask_dir),
```

benchmark/reconstruction/evaluation/utils.py
ahojnnes: Thanks, done.

pyproject.toml
```diff
@@ -41,8 +41,14 @@ wheel.packages = ["python/pycolmap"]
 [tool.cibuildwheel]
 build = "cp3{10,11,12,13,14}-{macosx,manylinux,win}*"
 archs = ["auto64"]
-test-requires = "pytest mypy==1.17.0 enlighten==1.13.0"
-test-command = "python -c \"import pycolmap; print(pycolmap.__version__)\" &&  python -m mypy --package pycolmap --implicit-optional && pytest {project}/python/examples/custom_incremental_pipeline_test.py"
+test-requires = "pytest>=8.0 mypy==1.19.1 enlighten==1.13.0"
+test-command = """\
+    python -c "import pycolmap; print(pycolmap.__version__)" && \
+    cd {project} && \
+    mypy --package pycolmap \
```
ahojnnes:
```suggestion
    mypy --package pycolmap && \
```

python/examples/panorama_sfm.py
```diff
@@ -20,6 +22,8 @@
 import pycolmap
 from pycolmap import logging
 
+NDArrayNx2 = Annotated[npt.NDArray[np.float64], Literal[Never, 2]]
```
sarlinpe: `Never` usually means unreachable code, and is introduced only in 3.11. Better use string litterals:
```suggestion
NDArrayNx2 = Annotated[npt.NDArray[np.float64], Literal['N', 2]]
```

ahojnnes: Somehow this didn't work with 3.9 and now also not with 3.10. I found another way.

ahojnnes: We should probably move away from overloading bindings, because it requires explicit typing on the caller side.

sarlinpe: Wasn't aware of this, good point.

sarlinpe (approved):

PR #3898, opened 2025-12-29
Improve and refactor rotation averaging I/O functions.

B1ueber2y:
I feel that this file should also be renamed. But rotation_averaging_io.h/cc also does not sound great. Let me know if there are good ideas. 

For each file format i disentangled the usage of ViewGraph to top level, added missing read/write functions and added round trip tests.

### DIFF
--- a/src/colmap/optim/least_absolute_deviations.cc
+++ b/src/colmap/optim/least_absolute_deviations.cc
@@ -114,7 +114,7 @@ LeastAbsoluteDeviationSolver::LeastAbsoluteDeviationSolver(
   THROW_CHECK_GE(options_.absolute_tolerance, 0);
   THROW_CHECK_GE(options_.relative_tolerance, 0);
   if (A.rows() < A.cols()) {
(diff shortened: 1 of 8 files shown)


ahojnnes:
Alternatively, we just drop support for this type of input and just require this information to be stored in the database through the pose priors table?

B1ueber2y:
Apparently there are indeed users for this type of input files: https://github.com/colmap/colmap/issues/3924.

leonngl:
Hi! I actually stumbled upon my issue (https://github.com/colmap/colmap/issues/3924) while trying to get the global_mapper with gravity-input to work. I raised an issue for the rotation_averager, because I thought that (specifically SolveRotationAveraging) is the main problem. I noticed the I/O problem of the rotation_averager additionally in the meantime. So from my side, no need to keep the text-file input, but thanks for considering! :)
Sorry about that, should have stated that more clearly in my issue.

B1ueber2y:
Closing as done alternatively in https://github.com/colmap/colmap/pull/3990.

PR #3853, opened 2025-12-25
Use colmap::Reconstruction instead of scattered variables in glomap.

B1ueber2y:
Reopening https://github.com/colmap/colmap/pull/3852 from an official branch. 

This is a huge major update that includes the following changes:

* Use colmap::Reconstruction throughout glomap instead of the scattered variables, whenever two of the {cameras, rigs, images, frames} are needed. This largely improves safety as these variables are coupled together.
* As a result, remove conversion functions between colmap and glomap, refactor I/O, and merge colmap_converter.h/cc and colmap_io.h/cc together.
* Make ReadGravity only read, and update rotation to gravity aligned at the top level.
* Use colmap Reconstruction::Normalize and remove reconstruction_normalizer.h/cc from glomap.
* Some minor improvements on naming and comments.

PR #3900, opened 2025-12-29
Simplify colmap_io in glomap thanks to the unification.

B1ueber2y:


src/glomap/io/colmap_io.cc
```diff
@@ -169,22 +144,13 @@ void InitializeGlomapFromDatabase(const colmap::Database& database,
...
 
-    // Collect the fundamental matrices
-    if (two_view.config == colmap::TwoViewGeometry::UNCALIBRATED) {
-      image_pair.F = two_view.F;
-    } else if (two_view.config == colmap::TwoViewGeometry::CALIBRATED) {
+    // For calibrated pairs, recompute F from the relative pose
+    if (image_pair.config == colmap::TwoViewGeometry::CALIBRATED) {
       image_pair.F = colmap::FundamentalFromEssentialMatrix(
           reconstruction.Camera(image2.CameraId()).CalibrationMatrix(),
           colmap::EssentialMatrixFromPose(image_pair.cam2_from_cam1),
```
B1ueber2y, line 151: This is a bug (should be two_view rather than image_pair) from https://github.com/colmap/colmap/pull/3853, but does not hurt the regression, probably because there are another decomposition and a re-estimation in the following steps : )

PR #3901, opened 2025-12-29
Switch paths from std::string to std::filesystem::path

sarlinpe:
Pybind11 can automatically type paths as `os.PathLike | str | bytes` for arguments and `pathlib.Path` for return values if paths are of type `std::filesystem::path`. I started by changing only Python-facing interfaces but ended up updating ~80% of paths throughout the codebase. Since this PR is already huge, I will migrate the remaining ones in a follow-up PR (removing any remaining cast `path::string()`).

The only pain point is that Unix accepts implicit conversion from `fs::path` to `std::string` but Windows does not, which makes local debugging difficult. Implicit conversion from `std::string` to `fs::path` is always allowed though.

### DIFF
--- a/src/colmap/controllers/global_pipeline_test.cc
+++ b/src/colmap/controllers/global_pipeline_test.cc
@@ -128,8 +128,7 @@ TEST(GlobalPipeline, SfMWithRandomSeedStability) {
 }
 
 TEST(GlobalPipeline, WithExistingRelativePoses) {
-  const std::string database_path = CreateTestDir() + "/database.db";
-
+  const auto database_path = CreateTestDir() / "database.db";
   auto database = Database::Open(database_path);
   Reconstruction gt_reconstruction;
   SyntheticDatasetOptions synthetic_dataset_options;
@@ -157,9 +156,7 @@ TEST(GlobalPipeline, WithExistingRelativePoses) {
 // To test relative pose re-estimation from view graph calibration.
 TEST(GlobalPipeline, WithNoisyExistingRelativePoses) {
   SetPRNGSeed(1);
-
-  const std::string database_path = CreateTestDir() + "/database.db";
-
+  const auto database_path = CreateTestDir() / "database.db";
   auto database = Database::Open(database_path);
   Reconstruction gt_reconstruction;
   SyntheticDatasetOptions synthetic_dataset_options;


src/colmap/scene/reconstruction.cc
```diff
@@ -1164,15 +1168,15 @@ void Reconstruction::ExtractColorsForAllImages(const std::string& path) {
   }
 }
 
-void Reconstruction::CreateImageDirs(const std::string& path) const {
-  std::unordered_set<std::string> image_dirs;
+void Reconstruction::CreateImageDirs(const std::filesystem::path& path) const {
+  std::set<std::filesystem::path> image_dirs;
```
sarlinpe, line 1172: This change is intentional, since older versions of GCC do not implement a hash for `std::filesystem::path`.

src/colmap/exe/mvs.cc
```diff
@@ -229,7 +229,7 @@ Reconstruction RunStereoFuserImpl(const std::string& output_path,
     reconstruction.WriteText(output_path);
   } else if (output_type == "ply") {
     WriteBinaryPlyPoints(output_path, fuser.GetFusedPoints());
-    mvs::WritePointsVisibility(output_path + ".vis",
+    mvs::WritePointsVisibility(AddFileExtension(output_path, ".vis"),
```
ahojnnes, line 232: Cannot be just `output_path + ".vis"`?

src/colmap/ui/dense_reconstruction_widget.cc
```diff
@@ -547,7 +543,7 @@ void DenseReconstructionWidget::RefreshWorkspace() {
         image_button, &QPushButton::released, [this, image_name, image_path]() {
           image_viewer_widget_->setWindowTitle(
               QString("Image for %1").arg(image_name.c_str()));
-          image_viewer_widget_->ReadAndShow(image_path);
+          image_viewer_widget_->ReadAndShow(image_path.string());
```
ahojnnes: Can probably also change ReadAndShow to take a path? Same for ReadAndShowWithKeypoints, ReadAndShowWithMatches?

src/colmap/util/file.cc
```diff
@@ -67,19 +67,26 @@ std::string EnsureTrailingSlash(const std::string& str) {
...
+bool HasFileExtension(const std::filesystem::path& file_name,
+                      const std::string& ext) {
   THROW_CHECK(!ext.empty());
   THROW_CHECK_EQ(ext.at(0), '.');
   std::string ext_lower = ext;
   StringToLower(&ext_lower);
-  if (file_name.size() >= ext_lower.size() &&
-      file_name.substr(file_name.size() - ext_lower.size(), ext_lower.size()) ==
-          ext_lower) {
+  if (file_name.string().size() >= ext_lower.size() &&
```
ahojnnes: filename.extension() ?

src/colmap/util/file.cc
```diff
@@ -162,12 +169,12 @@ std::string NormalizePath(const std::filesystem::path& path) {
...
 
-std::string GetNormalizedRelativePath(const std::string& full_path,
-                                      const std::string& base_path) {
+std::string GetNormalizedRelativePath(const std::filesystem::path& full_path,
+                                      const std::filesystem::path& base_path) {
   return NormalizePath(std::filesystem::relative(full_path, base_path));
 }
 
-std::vector<std::string> GetFileList(const std::string& path) {
+std::vector<std::string> GetFileList(const std::filesystem::path& path) {
```
ahojnnes, line 172: Return list of paths? Same below for other similar instances.

src/colmap/util/file.h
```diff
@@ -70,61 +70,67 @@ MAKE_ENUM_CLASS(FileCopyType, 0, COPY, HARD_LINK, SOFT_LINK);
 std::string EnsureTrailingSlash(const std::string& str);
 
 // Check whether file name has the file extension (case insensitive).
-bool HasFileExtension(const std::string& file_name, const std::string& ext);
+bool HasFileExtension(const std::filesystem::path& file_name,
+                      const std::string& ext);
+
+// Add a file extension.
+std::filesystem::path AddFileExtension(std::filesystem::path path,
```
ahojnnes, line 77: As mentioned above, this feels a bit unnecessary, as we can just use "file" + ".ext" ?

src/colmap/exe/mvs.cc
sarlinpe: `fs::path` defines `operator+=` but not `operator+` (likely for safety). We can define it ourselves (below), but I thought that having an explicit function is better. Wdyt?

```c++
template <typename T>
std::filesystem::path operator+(std::filesystem::path path, T&& other) {
    path += std::forward<T>(other);
    return path;
}
```

ahojnnes: OK, I thought I saw that you used this in other places but maybe those were strings. I guess one workaround would be to do `path.string() + ".ext"`. No strong opinions, feel free to leave as is.

sarlinpe: Leaving as is then.

src/colmap/util/file.cc
sarlinpe: Done.

src/colmap/util/file.cc
sarlinpe: Will do in the follow-up PR.

src/colmap/util/file.h
sarlinpe: See discussion above.

src/colmap/ui/dense_reconstruction_widget.cc
sarlinpe: Done.

src/colmap/util/file.cc
ahojnnes: SGTM, thanks.

PR #3465, opened 2025-07-09
Refactor feature extraction/matching to support other features

ahojnnes:

PR #3903, opened 2025-12-29
Implement AddAndRegisterDefaultEnumOption and propagate to codebase.

B1ueber2y:
This feature was added in https://github.com/colmap/colmap/pull/3465 but never implemented.

### DIFF
--- a/src/glomap/sfm/global_mapper.cc
+++ b/src/glomap/sfm/global_mapper.cc
@@ -115,8 +115,7 @@ bool GlobalMapper::GlobalPositioning(const GlobalPositionerOptions& options,
                                      double max_angle_error,
                                      double max_reprojection_error,
                                      double min_triangulation_angle) {
-  if (options.constraint_type !=
-      GlobalPositionerOptions::ConstraintType::ONLY_POINTS) {
+  if (options.constraint_type != GlobalPositioningConstraintType::ONLY_POINTS) {
     LOG(ERROR) << "Only points are used for solving camera positions";
     return false;
   }


src/colmap/controllers/base_option_manager.h
```diff
@@ -176,4 +212,26 @@ void BaseOptionManager::RegisterOption(const std::string& name,
   }
 }
 
+template <typename EnumT>
+void BaseOptionManager::AddAndRegisterDefaultEnumOption(
+    const std::string& name,
+    EnumT* option,
+    std::string_view (*to_string_fn)(EnumT),
+    EnumT (*from_string_fn)(std::string_view),
```
ahojnnes, line 220: Why not `std::function<EnumT(std::string_view)>`?

src/colmap/controllers/base_option_manager.h
```diff
@@ -176,4 +212,26 @@ void BaseOptionManager::RegisterOption(const std::string& name,
...
+  auto info = std::make_unique<EnumOptionInfo>();
+  info->value = std::string(to_string_fn(*option));
+
+  EnumOptionInfo* info_ptr = info.get();
+  info->apply = [info_ptr, option, from_string_fn]() {
+    *option = from_string_fn(info_ptr->value);
+  };
+
+  // Register as a string option pointing to our storage
+  AddDefaultOption(name, &info->value, help_text);
```
ahojnnes, line 232: Wondering whether we can use the existing enum macros to automatically compile a help text with all the available enum values as `"{VALUE_A, VALUE_B, ...}"`?

ahojnnes (approved):

src/colmap/controllers/base_option_manager.h
B1ueber2y: This is possible by adding another macro (e.g., ``xxxTypeValuesString) in enum_utils. However, it s a bit tricky what we should do for UNDEFINED and INVALID. It s unclear to me whether just filtering them is the right approach because they may be useful.

src/colmap/controllers/base_option_manager.h
B1ueber2y: In enum_utils.h we have overloaded the EnumTypeToString function with both int and EnumType, and because of this using std::function always throws.

To make it work we need to do ``static_cast<std::string_view(*)(FeatureExtractorType)>(FeatureExtractorTypeToString)`` at each call, which is very inconvenient.

ahojnnes: Maybe we should not overload but use a different name? E.g. `FeatureExtractorTypeToString` and `FeatureExtractorIntTypeToString` ?

B1ueber2y: I think the overloading is quite conveient in practice. Lets keep it as it is now?

PR #3902, opened 2025-12-29
Use colmap IterativeGlobalRefinement for retriangulation and refinement in glomap.

B1ueber2y:

PR #3906, opened 2025-12-29
Add single-thread deterministic test for global mapper and rotation averaging controller.

B1ueber2y:
Reopening https://github.com/colmap/colmap/pull/3905 from an official repo branch.

### DIFF
--- a/src/colmap/controllers/global_pipeline_test.cc
+++ b/src/colmap/controllers/global_pipeline_test.cc
@@ -119,8 +119,8 @@ TEST(GlobalPipeline, SfMWithRandomSeedStability) {
     // in optimization.
     EXPECT_THAT(*reconstruction_manager0->Get(0),
                 ReconstructionNear(*reconstruction_manager1->Get(0),
-                                   /*max_rotation_error_deg=*/1e-10,
-                                   /*max_proj_center_error=*/1e-10,
+                                   /*max_rotation_error_deg=*/2e-10,
+                                   /*max_proj_center_error=*/2e-10,
                                    /*max_scale_error=*/std::nullopt,
                                    /*num_obs_tolerance=*/0.01,
                                    /*align=*/false));


B1ueber2y:
The deterministic test is fixed via https://github.com/colmap/colmap/pull/3908. Relevant: https://groups.google.com/g/ceres-solver/c/2ITKHCi3-oU/m/OlxDBAq6BYAJ

src/glomap/processors/view_graph_manipulation.cc
```diff
@@ -81,7 +83,7 @@ void ViewGraphManipulator::DecomposeRelPose(
   const int64_t num_image_pairs = image_pair_ids.size();
   LOG(INFO) << "Decompose relative pose for " << num_image_pairs << " pairs";
 
-  colmap::ThreadPool thread_pool(colmap::ThreadPool::kMaxNumThreads);
+  colmap::ThreadPool thread_pool(colmap::GetEffectiveNumThreads(num_threads));
```
ahojnnes: I think the threadpool itself already calls the effective num threads function.

src/glomap/estimators/global_positioning.h
```diff
@@ -58,6 +58,10 @@ struct GlobalPositionerOptions {
   // Scaling factor for the loss function
   double loss_function_scale = 0.1;
 
+  // Whether to use custom parameter block ordering for Schur-based solvers.
+  // Disable for deterministic behavior when using a fixed random seed.
```
ahojnnes, line 62: It's unclear to me why a fixed parameter block ordering will make the tests non-deterministic? Sounds like a bug on the ceres side? Wouldn't we expect it to be the vice versa?

src/colmap/controllers/global_pipeline_test.cc
```diff
@@ -68,6 +68,18 @@ TEST(GlobalPipeline, Nominal) {
                                  /*max_proj_center_error=*/1e-4));
 }
 
+void ExpectExactEqualReconstructions(const Reconstruction& reconstruction1,
```
ahojnnes: We have a `ReconstructionEq` matcher. Does it not work here?

src/glomap/estimators/global_positioning.h
B1ueber2y: Yes it's an issue on the ceres side: https://groups.google.com/g/ceres-solver/c/2ITKHCi3-oU/m/OlxDBAq6BYAJ. There is not much we can do here.

B1ueber2y:
Fixing a potentially deterministic issue from https://github.com/colmap/colmap/pull/3902: https://github.com/colmap/colmap/pull/3906/commits/d40083e0faed537dda6c509b9456074289a9b3c2

PR #3907, opened 2025-12-30
Deterministic global positioning with parameter block ordering enabled.

B1ueber2y:
This feels quite an overkill to just fix the determinism.

### DIFF
--- a/src/glomap/estimators/global_positioning.cc
+++ b/src/glomap/estimators/global_positioning.cc
@@ -7,6 +7,8 @@
 
 #include "glomap/estimators/cost_functions.h"
 
+#include <algorithm>
+
 namespace glomap {
 namespace {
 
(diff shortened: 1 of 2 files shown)


B1ueber2y:
The culprit of non-deterministic single-thread behavior of global positioning is the parameter block ordering (see https://github.com/colmap/colmap/pull/3908). Dropping the block https://github.com/colmap/colmap/blob/5c87ba90129d277ce5e6062baa83073bfba5891a/src/glomap/estimators/global_positioning.cc#L363-L402 would get the deterministic behavior back. 

This PR is an alternative fix to this issue. The key problem is that our data (including 3D points) are stored in unordered map, which results in non-deterministic address behavior in ceres internal parameter block ordering logic. To be able to fix it we need to use continuous vector as this PR does.

Relevant: https://groups.google.com/g/ceres-solver/c/2ITKHCi3-oU/m/OlxDBAq6BYAJ

B1ueber2y:
Edit: while this fix worked locally in my machine, the CI does not seem to agree.

PR #3909, opened 2025-12-30
Replace the usage of inliers in ImagePair with inlier_matches.

B1ueber2y:
Also includes some cosmetic improvement in track establishment on the way.


src/glomap/sfm/track_establishment.h
```diff
@@ -53,8 +53,9 @@ class TrackEngine {
   const ViewGraph& view_graph_;
   const std::unordered_map<image_t, colmap::Image>& images_;
 
-  // Internal structure used for concatenating tracks
-  colmap::UnionFind<image_pair_t> uf_;
+  // Internal structure used for concatenating tracks.
+  // Key is (image_id, point2D_idx) pair identifying a 2D observation.
+  colmap::UnionFind<std::pair<image_t, colmap::point2D_t>> uf_;
```
ahojnnes: nit: move the "Observation" type alias to the header to use here as well?

src/glomap/processors/image_pair_inliers.cc
```diff
@@ -252,8 +250,8 @@ void ImagePairsInlierCount(ViewGraph& view_graph,
                            const InlierThresholdOptions& options,
                            bool clean_inliers) {
   for (auto& [pair_id, image_pair] : view_graph.ImagePairs()) {
-    if (!clean_inliers && image_pair.inliers.size() > 0) continue;
-    image_pair.inliers.clear();
+    if (!clean_inliers && !image_pair.inlier_matches.empty()) continue;
+    image_pair.inlier_matches.clear();
```
ahojnnes: If we want to save memory, we could do `image_pair.inlier_matches = {}` here.

ahojnnes: Same in other places.

src/glomap/scene/image_pair.h
```diff
@@ -21,9 +21,6 @@ struct ImagePair : public colmap::TwoViewGeometry {
   // Second column is the index of the feature in the second image.
   Eigen::MatrixXi matches;
```
ahojnnes, line 22: Use FeatureMatches instead of Eigen::MatrixXi? Eigen::MatrixXi probably also has an inconvenient memory layout due to column major default in Eigen.

B1ueber2y: This field will be dropped with the new regression test we discussed offline, and we will fully switch to TwoViewGeometry.

PR #3913, opened 2025-12-30
Add support for computing recall scores

ahojnnes:
* In addition to computing AUC scores, also support recall scores.
* Fix and remove magic 1.1 factor on computed AUC scores.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -15,8 +15,8 @@ build*/
 install-debug/
 install-release/
 data/
-benchmark/reconstruction/evaluation/data/
-benchmark/reconstruction/evaluation/runs/
+benchmark/reconstruction/data/
+benchmark/reconstruction/runs/
 doc/_build
 vcpkg_installed/
 


benchmark/reconstruction/evaluation/utils.py
```diff
@@ -703,53 +714,68 @@ def compute_abs_errors(
...
 
     aucs = np.zeros(len(thresholds), dtype=np.float64)
     for i, t in enumerate(thresholds):
         last_index = np.searchsorted(errors, t, side="right")
-        r = np.r_[recall[:last_index], recall[last_index - 1]]
+        r = np.r_[recalls[:last_index], recalls[last_index - 1]]
         e = np.r_[errors[:last_index], t]
         auc = np.trapezoid(r, x=e) / t
         aucs[i] = auc * 100
-    return aucs / 1.1
```
ahojnnes, line 737: No idea how this factor ended up here. It should have no effect on previous relative regression tests but be careful about future absolute comparisons.

.gitignore
```diff
@@ -15,8 +15,8 @@ build*/
 install-debug/
 install-release/
 data/
-benchmark/reconstruction/data/
-benchmark/reconstruction/runs/
+benchmark/reconstruction/evaluation/data/
```
B1ueber2y: Can we do ``**/runs/`` and ``**/data/``. This ignored directory is not how i set up my directory structure and I guess it differs across different users. I generally runs at the top level so have a directory runs there.

ahojnnes: It's the default directory for the data and the runs, if you don't specify a custom one on the command line. Nevertheless, I reverted the change, because the directories were inconsistent between the download and evaluate scripts, as we moved the argument parser to the evaluation sub-folder.

PR #3918, opened 2025-12-31
Use DatabaseCache to load reconstruction in glomap and remove the custom loader. 

B1ueber2y:


B1ueber2y:
We will need to use DatabaseCache instead of Database as the signature in global mapper, to avoid the current status where the database needs to be kept open and stored as a shared pointer in the global mapping process. 

This can be done by 1) migrating view graph calibration to the top level 2) change ViewGraph to PoseGraph 3) add std::unordered_map<image_pair_t, Rigid3d> in the DatabaseCache class.

src/glomap/sfm/global_mapper.cc
```diff
@@ -50,16 +58,29 @@ bool GlobalMapper::ReestimateRelativePoses(
   view_graph_->FilterByNumInliers(inlier_thresholds.min_inlier_num);
   view_graph_->FilterByInlierRatio(inlier_thresholds.min_inlier_ratio);
 
-  if (view_graph_->KeepLargestConnectedComponents(*reconstruction_) == 0) {
```
B1ueber2y, line 53: This was misplaced in glomap and should be moved to the rotation averaging block.

B1ueber2y:
Found a temporary solution to remove database_ from the member of GlobalMapper: https://github.com/colmap/colmap/pull/3918/commits/764ed88d6d3e6d3b65ed12224d3303783578fb0b.

ahojnnes:
I have no particular objection to the changes but I am not sure I understand why we cannot hold on to the shared pointer of the database in the mapper. For me to understand , can you elaborate a bit more? Thanks!

src/glomap/sfm/global_mapper.cc
```diff
@@ -15,18 +15,26 @@
 
 namespace glomap {
 
-GlobalMapper::GlobalMapper(std::shared_ptr<const colmap::Database> database)
-    : database_(std::move(THROW_CHECK_NOTNULL(database))) {}
+GlobalMapper::GlobalMapper(std::shared_ptr<const colmap::Database> database) {
+  THROW_CHECK_NOTNULL(database);
+  // TODO: Directy use DatabaseCache in the signature and make min_num_matches
```
ahojnnes:
```suggestion
  // TODO: Directly use DatabaseCache in the signature and make min_num_matches
```

src/glomap/sfm/rotation_averaging_test.cc
```diff
@@ -117,6 +117,13 @@ TEST(RotationEstimator, WithoutNoise) {
   std::unordered_map<frame_t, int> cluster_ids;
   global_mapper.Solve(CreateMapperTestOptions(), cluster_ids);
 
+  for (const auto& [frame_id, frame] : reconstruction->Frames()) {
```
ahojnnes, line 125: Maybe add a todo here and below?

ahojnnes (approved):

B1ueber2y:
> I have no particular objection to the changes but I am not sure I understand why we cannot hold on to the shared pointer of the database in the mapper. For me to understand , can you elaborate a bit more? Thanks!

On one hand, I believe we don't want the SfM mapper logic to be able to update database, for the modularity. And on the other hand, holding a database pointer would make things trickier when two different SfM mapper operates on the same database file, which can incur safety issues in the future.

I thought the DatabaseCache class in colmap was particularly designed for this purpose, but maybe there were actually different intentions?

ahojnnes:
> > I have no particular objection to the changes but I am not sure I understand why we cannot hold on to the shared pointer of the database in the mapper. For me to understand , can you elaborate a bit more? Thanks!
> 
> On one hand, I believe we don't want the SfM mapper logic to be able to update database, for the modularity. And on the other hand, holding a database pointer would make things trickier when two different SfM mapper operates on the same database file, which can incur safety issues in the future.
> 
> I thought the DatabaseCache class in colmap was particularly designed for this purpose, but maybe there were actually different intentions?

Thanks, makes sense. Reading from the database has no issue in terms of concurrency. See my recent changes to the hierarchical mapper., where we may also want to use the cache instead of the database?

Independent of that, we can probably also do a better job in labeling write/update methods as non-const in the database interface.

PR #3921, opened 2025-12-31
Drop experimental support for post pruning in global mapper.

B1ueber2y:


B1ueber2y:
> I think the pruning was useful for internet photo collections, where not all images are from the same connected component or when there are complete outliers. @lpanaf ?

IIUC the pruning is performed on the largest component, which was already selected in the rotation averaging stage. So multiple components should not be a concern in my opinion.

ahojnnes:
You are right, it doesn't concern the issue of multiple components but does remove weakly connected frames. Let's see what @lpanaf says.

lpanaf:
For this step, as mentioned by @ahojnnes, it is useful for getting a "clean" reconstruction. The reason for this is that for some images, though they are included in the maximum connected components, they are linked to others by outlier edges. The final reconstruction we would get in the end would not be several connected components mistakenly joined together.

B1ueber2y:
> For this step, as mentioned by @ahojnnes, it is useful for getting a "clean" reconstruction. The reason for this is that for some images, though they are included in the maximum connected components, they are linked to others by outlier edges. The final reconstruction we would get in the end would not be several connected components mistakenly joined together.

Sure. In this case, it would be nice to document when this step is useful, as it is disabled by default. Why not just always enable it? 

On the other hand, I think this complicated function needs to be further improved and broken into sub-steps, as in the TODO here: https://github.com/colmap/colmap/blob/main/src/glomap/processors/reconstruction_pruning.cc#L123. We would also need to find a good place for these utilities of cluster_ids and WriteReconstructionByClusters. Should these clustering just live in the global pipeline and handled by reconstruction manager instead of cluster ids?

lpanaf:
This step is not enabled by default to reproduce numbers in the paper. I agree that some documentation would be helpful. For the cluster id, it is only used for deciding which image to output when writing to disk. In this case, it should be possible to use reconstruction manager to manage them. Ideally, if multi-reconstruction is supported in the mapper, it should be managed in a similar way.

B1ueber2y:
> This step is not enabled by default to reproduce numbers in the paper. I agree that some documentation would be helpful. For the cluster id, it is only used for deciding which image to output when writing to disk. In this case, it should be possible to use reconstruction manager to manage them. Ideally, if multi-reconstruction is supported in the mapper, it should be managed in a similar way.

Should we just always enable it in this case? Or does it negatively affect the AUC a lot? But if this was for only filtering outliers, why would it hurt the numbers?

lpanaf:
Yes, it would negatively affect numbers, since sometimes, the filtering can be too aggressive. And also, outlier images can contribute to AUC scores. This module is a bit heuristic, so my general suggestion was to first try without, and run this step only if the reconstruction is not clean.

B1ueber2y:
I think multiple components can be supported by doing rotation averaging recursively, such that the mapper recursively finds the largest connected components in the remaining images. One can also just do the global mapper recursively with this pruning. 

I personally find the logic (as I documented here: https://github.com/colmap/colmap/blob/53af102ce2d398c861a0dee49e124cbc54903be3/src/glomap/processors/reconstruction_pruning.h#L17-L26) of this pruning really artificial, but improving it might take quite some efforts on further benchmarking.

B1ueber2y:
> Yes, it would negatively affect numbers, since sometimes, the filtering can be too aggressive. And also, outlier images can contribute to AUC scores. This module is a bit heuristic, so my general suggestion was to first try without, and run this step only if the reconstruction is not clean.

I dont like this idea of human-in-the-loop checking. It would be nice if we could improve the heuristic to be more robust, ideally not hurting the AUCs : )

lpanaf:
I think it can potentially be a seperate module, but it should not be removed. Users should be provided a way for clean up a noisy reconstruction. The current results is from a lot of trial and error, and it was the best working version. Though it is not beautiful, it generally works well. Of course, if a better algorithm can be developped, then it should be replaced.

B1ueber2y:
> I think it can potentially be a seperate module, but it should not be removed. Users should be provided a way for clean up a noisy reconstruction. The current results is from a lot of trial and error, and it was the best working version. Though it is not beautiful, it generally works well. Of course, if a better algorithm can be developped, then it should be replaced.

Sure. What was the metric on this working well, if it is hurting the AUC scores? Or, in other words, what was the metric you were using in the trials and errors when developing this module? 

It would be nice to quantify it to benefit further regression tests. We currently do not have any test coverage for this module, and it is disabled by default, so understanding the metric can also help on this. Is there a chance that you can add some meaningful tests for it, on top of which we can potentially try to iterate on a better algorithm? : ) Thank you!

lpanaf:
It was checked by visual quality and AUC scores, but nothing systematic. I think some tests should be good. I will try to write some test cases later :)

B1ueber2y:
Given that this function works on a single reconstruction, pulling it out of the global mapper and having it as a stand-alone CLI sounds the best to me.

lpanaf:
I added some tests cases in #4003.

B1ueber2y:
Closing as this is alternatively addressed in https://github.com/colmap/colmap/pull/4008.

PR #3929, opened 2026-01-02
Remove Database::Clone method

ahojnnes:
After adding it a few commits earlier together with making the database non-copyable, removing the method again as we found a better way to deal with concurrent access in the mappers.


B1ueber2y (approved):

B1ueber2y:
Actually, should we just drop this method? This is not used anywhere and can cause safety issues from the python side?

ahojnnes:
Yes, right, after your database cache changes. Sounds good to me.

PR #3932, opened 2026-01-02
Use std::optional for cam2_from_cam1 in TwoViewGeometry.

B1ueber2y:
The breaking usage includes:
* TwoViewGeometry will initialize with cam2_from_cam1 as None rather than identity matrix, which is a clear improvement. 
* **[IMPORTANT] When loading from a legacy database, a perfect identity cam2_from_cam1 will not be loaded as a valid relative pose.** 
* Forward compatibility of the database reading. The old code cannot read in the new database. 

This update can also largely improve the safety of the whole global pipeline, which heavily relies on this field. This will explicitly avoid un-decomposed relative poses in the ViewGraph.

### DIFF
--- a/src/colmap/util/version.h
+++ b/src/colmap/util/version.h
@@ -29,8 +29,6 @@
 
 #pragma once
 
-#include "colmap/util/logging.h"
-
 #include <string>
 
 namespace colmap {
@@ -42,15 +40,7 @@ std::string GetBuildInfo();
 // Computes database version number from version components.
 // Format: major * 1000000 + minor * 10000 + patch * 100 + revision
 // This gives 100 migration slots between consecutive releases.
-inline int MakeDatabaseVersionNumber(int major,
-                                     int minor,
-                                     int patch,
-                                     int revision) {
-  THROW_CHECK_LT(minor, 100);
-  THROW_CHECK_LT(patch, 100);
-  THROW_CHECK_LT(revision, 100);
-  return major * 1000000 + minor * 10000 + patch * 100 + revision;
-}
+int MakeDatabaseVersionNumber(int major, int minor, int patch, int revision);
 
 // Returns the database version number for the current COLMAP version.
 int GetDatabaseVersionNumber();
--- a/src/colmap/util/version.cc.in
+++ b/src/colmap/util/version.cc.in
@@ -30,6 +30,7 @@
 
 #include "colmap/util/version.h"
 
+#include "colmap/util/logging.h"
 #include "colmap/util/string.h"
 
 namespace colmap {
@@ -47,6 +48,13 @@ constexpr int kVersionPatch = ${COLMAP_VERSION_PATCH};
 constexpr int kDatabaseSchemaRevision = 0;
 
 }  // namespace
+
+int MakeDatabaseVersionNumber(int major, int minor, int patch, int revision) {
+  THROW_CHECK_LT(minor, 100);
+  THROW_CHECK_LT(patch, 100);
+  THROW_CHECK_LT(revision, 100);
+  return major * 1000000 + minor * 10000 + patch * 100 + revision;
+}
 
 int GetDatabaseVersionNumber() {
   return MakeDatabaseVersionNumber(


src/glomap/processors/view_graph_manipulation.cc
```diff
@@ -71,10 +71,11 @@ void ViewGraphManipulator::DecomposeRelPose(
 
       image_pair.config = two_view_geometry.config;
       image_pair.cam2_from_cam1 = two_view_geometry.cam2_from_cam1;
+      THROW_CHECK(image_pair.cam2_from_cam1.has_value());
 
-      if (image_pair.cam2_from_cam1.translation.norm() > 1e-12) {
-        image_pair.cam2_from_cam1.translation =
-            image_pair.cam2_from_cam1.translation.normalized();
+      if (image_pair.cam2_from_cam1->translation.norm() > 1e-12) {
```
ahojnnes: nit: store norm and replace normalized with division?

B1ueber2y:
Had to move pose decomposition into ViewGraph::LoadFromDatabase to ensure that all image pairs in ViewGraph has a valid pose. In the future, this decomposition logic can either be moved completely outside global mapper logic, or in the DatabaseCache::Create method.

B1ueber2y:
Regression:
```
I20260102 17:57:46.800717 169307 compare.py:main:58] Results A - B:

botanical_garden  -1.02  -0.84  -0.20  -0.10      0     0    0       0
boulders           0.73   0.38  -0.04  -1.07      0     0    0       0
bridge             0.63   0.44   0.25   0.21      0     0    0       0
courtyard          1.40   1.82   0.52   0.26      0     0    0       0
delivery_area     -1.03   2.64   9.69   7.64      0     0    0       0
door               0.00   0.00   0.00   0.00      0     0    0       0
electro            0.95   2.14   2.15   1.32      0     0    0       0
exhibition_hall   -0.18  -0.12  -0.03  -0.00      0     0    0       0
facade             0.02   0.00   0.01   0.00      0     0    0       0
kicker             1.36   4.58  10.29  11.22      0     0    0       0
lecture_room       8.38   6.91   2.71   1.36      0     0    0       0
living_room       -0.32  -0.23  -0.11  -0.07      0     0    0       0
lounge            -0.12   0.12   0.03   0.01      0     0    0       0
meadow            -2.28  -3.30 -10.54 -13.34      0     0    0       0
observatory        0.16   0.10   0.01   0.00      0     0    0       0
office            -0.21   0.44   0.65   0.99      0     0    0       0
old_computer       0.74   0.72   0.14   0.07      0     0    0       0
pipes              1.75   1.93   0.43   0.22      0     0    0       0
playground        -0.13  -0.08  -0.01  -0.01      0     0    0       0
relief            -4.95 -11.06 -17.89 -18.27      0     0    0       0
relief_2           0.02  -0.01  -0.01  -0.00      0     0    0       0
statue            -0.01  -0.00  -0.00  -0.00      0     0    0       0
terrace           -0.10  -0.11  -0.02  -0.01      0     0    0       0
terrace_2          0.05   0.01   0.00   0.00      0     0    0       0
terrains           1.17   2.65   4.27   3.42      0     0    0       0
overall            0.26   0.42   0.57   0.38      0     0    0       0
average            0.28   0.36   0.09  -0.25      0     0    0       0
```

src/colmap/estimators/two_view_geometry_test.cc
```diff
@@ -743,8 +745,9 @@ TEST(EstimateRigTwoViewGeometries, Nominal) {
   for (const auto& [image_pair, geometry] : geometries) {
     EXPECT_EQ(geometry.config,
               TwoViewGeometry::ConfigurationType::CALIBRATED_RIG);
+    EXPECT_TRUE(geometry.cam2_from_cam1.has_value());
```
ahojnnes: ASSERT?

src/colmap/scene/database_sqlite.cc
```diff
@@ -999,13 +1006,20 @@ class SqliteDatabase : public Database {
...
-      two_view_geometry.cam2_from_cam1.rotation = Eigen::Quaterniond(
-          quat_wxyz(0), quat_wxyz(1), quat_wxyz(2), quat_wxyz(3));
-      two_view_geometry.cam2_from_cam1.translation =
-          ReadStaticMatrixBlob<Eigen::Vector3d>(
-              sql_stmt_read_two_view_geometries_, rc, 9);
+      // Only populate cam2_from_cam1 for configs where pose was estimated.
+      if (two_view_geometry.config != TwoViewGeometry::UNDEFINED &&
+          two_view_geometry.config != TwoViewGeometry::DEGENERATE &&
+          two_view_geometry.config != TwoViewGeometry::WATERMARK &&
+          two_view_geometry.config != TwoViewGeometry::MULTIPLE) {
```
ahojnnes: Suggest to explicitly handle null values here and only keep the config based selection for old data.

src/colmap/scene/database_sqlite.cc
```diff
@@ -1284,19 +1298,20 @@ class SqliteDatabase : public Database {
     const Eigen::Matrix3d Ft = two_view_geometry_ptr->F.transpose();
     const Eigen::Matrix3d Et = two_view_geometry_ptr->E.transpose();
     const Eigen::Matrix3d Ht = two_view_geometry_ptr->H.transpose();
-    const Eigen::Vector4d quat_wxyz(
-        two_view_geometry_ptr->cam2_from_cam1.rotation.w(),
-        two_view_geometry_ptr->cam2_from_cam1.rotation.x(),
-        two_view_geometry_ptr->cam2_from_cam1.rotation.y(),
-        two_view_geometry_ptr->cam2_from_cam1.rotation.z());
+    // Write identity if cam2_from_cam1 is not set (for backwards compat).
```
ahojnnes: See previous comment above. How about new data setting a null value here? Also suggest keeping code for qvec/tvec columns together.

src/glomap/scene/view_graph.cc
```diff
@@ -1,14 +1,32 @@
...
                                  bool allow_duplicate) {
+  // TODO: Move relative pose decomposition logic to the top level.
   auto all_matches = database.ReadAllMatches();
+
+  // Read cameras and images upfront for pose decomposition.
+  const std::vector<colmap::Camera> cameras = database.ReadAllCameras();
+  std::unordered_map<colmap::camera_t, colmap::Camera> camera_map;
+  for (const auto& camera : cameras) {
+    camera_map.emplace(camera.camera_id, camera);
+  }
```
ahojnnes:
```suggestion
  std::unordered_map<colmap::camera_t, colmap::Camera> cameras;
  for (auto& camera : database.ReadAllCameras()) {
    camera_map.emplace(camera.camera_id, std::move(camera));
  }
```

src/glomap/scene/view_graph.cc
```diff
@@ -1,14 +1,32 @@
...
+
+  const std::vector<colmap::Image> images = database.ReadAllImages();
```
ahojnnes: Same as for cameras here.

ahojnnes (approved):

src/colmap/scene/database_sqlite.cc
B1ueber2y: Thanks! Made an update here: https://github.com/colmap/colmap/pull/3932/commits/7559a53202b82081d997210d5a9360790e9d6212 and also updated the PR description. PTAL.

B1ueber2y: I want to explicitly raise attention for the third breakage: "[IMPORTANT] When loading from a legacy database, a perfect identity cam2_from_cam1 will not be loaded as a valid relative pose." 

I dont think this is an issue though. Let me know if there are discrepancy on this.

src/colmap/scene/database_sqlite.cc
```diff
@@ -946,13 +978,22 @@ class SqliteDatabase : public Database {
...
-            sql_stmt_read_two_view_geometry_, rc, 8);
+    cam2_from_cam1.translation = ReadStaticMatrixBlob<Eigen::Vector3d>(
+        sql_stmt_read_two_view_geometry_, rc, 8);
+    const bool has_cam2_from_cam1_is_null =
+        sqlite3_column_type(sql_stmt_read_two_view_geometry_, 9) == SQLITE_NULL;
+    two_view_geometry.cam2_from_cam1 = MaybeReadCam2FromCam1(
+        cam2_from_cam1,
+        two_view_geometry.config,
+        has_cam2_from_cam1_is_null,
+        sqlite3_column_int64(sql_stmt_read_two_view_geometry_, 9));
```
sarlinpe: nit: For readability, consider assigning `has_cam2_from_cam1` explicitly, and similarly below.

src/colmap/scene/database_sqlite.cc
```diff
@@ -1962,7 +2019,8 @@ class SqliteDatabase : public Database {
           "    E        BLOB,"
           "    H        BLOB,"
           "    qvec     BLOB,"
-          "    tvec     BLOB);";
+          "    tvec     BLOB,"
+          "    has_cam2_from_cam1  INTEGER);";
```
ahojnnes: I don't think this new column is necessary. You can bind a null value when writing and check for a null value when reading?

src/colmap/scene/database_sqlite.cc
ahojnnes: I think we can solve the edge case by automatically migrating data using the `PostMigrateTables()` logic?

src/colmap/scene/database_sqlite.cc
B1ueber2y: Indeed. I made an update: https://github.com/colmap/colmap/pull/3932/commits/0c97de33890500689fd893e1c4040690636782c5. However, to be able to accept perfectly identity poses from non-legacy databases, we would need to make use of the version number.

src/colmap/scene/database_sqlite.cc
```diff
@@ -2094,7 +2103,7 @@ class SqliteDatabase : public Database {
     // Update user version number.
     std::unique_lock<std::mutex> lock(update_schema_mutex_);
     const std::string update_user_version_sql =
-        StringPrintf("PRAGMA user_version = 3900;");
```
B1ueber2y, line 2032: Is 3900 confusing btw? I checked through the history and this version was hardcoded in Nov 2023 https://github.com/colmap/colmap/pull/2275 (probably corresponding to colmap 3.9?) and has never been used. Not sure if incrementing it to 3901 in this PR makes it even more confusing : )

src/colmap/scene/database_sqlite.cc
B1ueber2y: Not used anymore as has_cam2_from_cam1 column is removed in the new update.

src/colmap/scene/database_sqlite.cc
B1ueber2y: Note that this removal of has_cam2_from_cam1 column breaks forward compatibility, in the sense that old code cannot read new database.

src/colmap/scene/database_sqlite.cc
```diff
@@ -946,13 +946,19 @@ class SqliteDatabase : public Database {
...
-    const Eigen::Vector4d quat_wxyz = ReadStaticMatrixBlob<Eigen::Vector4d>(
-        sql_stmt_read_two_view_geometry_, rc, 7);
-    two_view_geometry.cam2_from_cam1.rotation = Eigen::Quaterniond(
-        quat_wxyz(0), quat_wxyz(1), quat_wxyz(2), quat_wxyz(3));
-    two_view_geometry.cam2_from_cam1.translation =
-        ReadStaticMatrixBlob<Eigen::Vector3d>(
-            sql_stmt_read_two_view_geometry_, rc, 8);
+
+    // Read the pose data if present (NULL means not set).
+    if (sqlite3_column_type(sql_stmt_read_two_view_geometry_, 7) !=
```
ahojnnes: Maybe add a check that column 8 is also not null? Not sure what happens otherwise (crash?). Same below.

src/colmap/scene/database_sqlite.cc
ahojnnes: Yes, it is confusing and maybe we can inject a modified version of `COLMAP_VERSION` from colmap/util/version.cc automatically here?

src/colmap/scene/database_sqlite.cc
```diff
@@ -1997,6 +2019,62 @@ class SqliteDatabase : public Database {
...
+          &update_stmt,
+          nullptr));
+
+      while (SQLITE3_CALL(sqlite3_step(read_stmt)) == SQLITE_ROW) {
+        const Eigen::Vector4d qvec =
+            ReadStaticMatrixBlob<Eigen::Vector4d>(read_stmt, SQLITE_ROW, 1);
+        const Eigen::Vector3d tvec =
+            ReadStaticMatrixBlob<Eigen::Vector3d>(read_stmt, SQLITE_ROW, 2);
+        const Eigen::Quaterniond rotation(qvec(0), qvec(1), qvec(2), qvec(3));
+        const bool is_identity =
```
ahojnnes, line 2075: Isn't checking for the config the safer check here? This has the tiny chance that the pose was actually meant to be identity?

src/colmap/scene/database_sqlite.cc
ahojnnes: Something like 3140000 in the format MAJOR_MINOR_PATCH_X, where MAJOR_MINOR_PATCH are simply our semantic version and X gives us 100 times the chance to migrate data between consecutive releases.

src/glomap/scene/view_graph.cc
```diff
@@ -1,14 +1,30 @@
...
 void ViewGraph::LoadFromDatabase(const colmap::Database& database,
                                  bool allow_duplicate) {
+  // TODO: Move relative pose decomposition logic to the top level.
   auto all_matches = database.ReadAllMatches();
+
+  // Read cameras and images upfront for pose decomposition.
+  std::unordered_map<colmap::camera_t, colmap::Camera> cameras;
+  for (const auto& camera : database.ReadAllCameras()) {
+    cameras.emplace(camera.camera_id, camera);
+  }
```
ahojnnes:
```suggestion
  for (auto& camera : database.ReadAllCameras()) {
    cameras.emplace(camera.camera_id, std::move(camera));
  }
```

src/glomap/scene/view_graph.cc
```diff
@@ -1,14 +1,30 @@
...
+
+  std::unordered_map<image_t, colmap::Image> images;
+  for (const auto& image : database.ReadAllImages()) {
+    images.emplace(image.ImageId(), image);
```
ahojnnes: Same here.

src/glomap/scene/view_graph.cc
```diff
@@ -51,20 +67,62 @@ void ViewGraph::LoadFromDatabase(const colmap::Database& database,
         }
       }
       image_pair.matches.conservativeResize(count, 2);
+
+      // Decompose relative pose if not already present.
+      if (!image_pair.cam2_from_cam1.has_value()) {
```
ahojnnes, line 84: Can this ever happen?

src/glomap/scene/view_graph.cc
```diff
@@ -51,20 +67,62 @@ void ViewGraph::LoadFromDatabase(const colmap::Database& database,
...
+        const colmap::Image& image1 = images.at(image_id1);
+        const colmap::Image& image2 = images.at(image_id2);
+        const colmap::Camera& camera1 = cameras.at(image1.CameraId());
+        const colmap::Camera& camera2 = cameras.at(image2.CameraId());
+
+        const colmap::FeatureKeypoints keypoints1 =
```
ahojnnes: Could extract this into a small helper function and reuse here and below.

src/glomap/scene/view_graph_test.cc
```diff
@@ -38,9 +38,16 @@
 namespace glomap {
 namespace {
 
+colmap::Rigid3d TestCam2FromCam1() {
```
ahojnnes: Only used once, maybe just inline below?

ahojnnes (approved):

src/colmap/scene/database_sqlite.cc
B1ueber2y: For a legacy database, an unset relative pose is written as identity. Not checking for identity would load every undecomposed CALIBRATED two view geometry with identity pose, which is buggy.

src/glomap/scene/view_graph.cc
B1ueber2y: This can happen if the input database to the global pipeline has already been decomposed?

src/colmap/scene/database_sqlite.cc
ahojnnes: I see, this happens when `compute_relative_pose == false`.

src/glomap/scene/view_graph.cc
ahojnnes: Yes, similar to the other thread, I did not consider the case where `compute_relative_pose == false`.

src/colmap/sfm/incremental_mapper_impl.cc
```diff
@@ -751,13 +751,13 @@ bool IncrementalMapperImpl::EstimateInitialTwoViewGeometry(
...
           << " inlier matches, "
-          << two_view_geometry.cam2_from_cam1.translation.z()
+          << two_view_geometry.cam2_from_cam1->translation.z()
           << " z translation, " << RadToDeg(two_view_geometry.tri_angle)
           << " deg triangulation angle";
 
   if (static_cast<int>(two_view_geometry.inlier_matches.size()) <
           options.init_min_num_inliers ||
-      std::abs(two_view_geometry.cam2_from_cam1.translation.z()) >=
+      std::abs(two_view_geometry.cam2_from_cam1->translation.z()) >=
```
ahojnnes, line 760: In theory, we could now just use the cam2_from_cam1 in case the relative pose is already stored in the database?

src/colmap/scene/database_sqlite.cc
B1ueber2y: Done: https://github.com/colmap/colmap/pull/3932/commits/5a03c107ff8152421a6c3485ceb1603df1d9f855. PTAL.

src/colmap/sfm/incremental_mapper_impl.cc
B1ueber2y: In this block the two_view_geometry is freshly calculated from ``EstimateCalibratedTwoViewGeometry`` with init_max_error rather than max_error: https://github.com/colmap/colmap/blob/3b6c36ee10b6bc665ade65f2b8726f18a745def8/src/colmap/sfm/incremental_mapper_impl.cc#L739-L749, so the relative pose from the database does not apply here.

src/colmap/scene/database_sqlite.cc
B1ueber2y: > Note that this removal of has_cam2_from_cam1 column breaks forward compatibility, in the sense that old code cannot read new database.

Make an update on the PR description about forward compatibility breakage.

src/colmap/sfm/incremental_mapper_impl.cc
ahojnnes: Got it, missed that. Thanks.

src/colmap/util/version.h
```diff
@@ -37,4 +39,20 @@ std::string GetVersionInfo();
 
 std::string GetBuildInfo();
 
+// Computes database version number from version components.
+// Format: major * 1000000 + minor * 10000 + patch * 100 + revision
+// This gives 100 migration slots between consecutive releases.
+inline int MakeDatabaseVersionNumber(int major,
```
ahojnnes: Can we move this to the .cc file?

src/colmap/scene/database_sqlite.cc
ahojnnes: Thanks, LGTM.

src/colmap/util/version.h
B1ueber2y: This function is used here for better readability: https://github.com/colmap/colmap/pull/3932/files#diff-4d87fc25e3b40780d29e62fd42559925e07c2c748622ec65897c84e0bc703dfeR2050

ahojnnes: Yes, not suggesting to move it to anonymous namespace, but just not inline and only declare in the header and define in the .cc file.

PR #3941, opened 2026-01-03
Move and type annotate flickr downloader

ahojnnes:


python/util/flickr_downloader.py
```diff
@@ -32,30 +32,30 @@
...
+import urllib.request
 import xml.etree.ElementTree as ElementTree
 
 import urllib2
 import urlparse
 
-PER_PAGE = 500
-SORT = "date-posted-desc"
-URL = (
+PER_PAGE: int = 500
```
sarlinpe: I suggest marking these as `Final`, i.e. 
```python
from typing import Final
PER_PAGE: Final[int] = 500
```

python/util/flickr_downloader.py
```diff
@@ -103,26 +107,28 @@ def parse_page(page, api_key, text, min_date, max_date):
     root = ElementTree.fromstring(response)
 
     if root.attrib["stat"] != "ok":
-        raise IOError
+        raise OSError
 
-    photos = []
+    photos: list[dict[str, str]] = []
     for photo in root.iter("photo"):
         photos.append(photo.attrib)
```
sarlinpe: Suggest using list comprehension, making the type annotation unnecessary:
```suggestion
photos = tuple(photo.attrib for photo in root.iter("photo"))
```

PR #3943, opened 2026-01-04
Use custom two view options for global pipeline in automatic reconstructor.

B1ueber2y:
With this update we are finally able to keep the same (or maybe even slightly better because view graph calibration attempts to refine the GT intrinsics) accuracy in calibrated case (sfm with camera priors) when dropping view graph calibration, which makes us ready to disentangle this step to top level.


src/colmap/controllers/automatic_reconstruction.cc
```diff
@@ -92,8 +92,33 @@ AutomaticReconstructionController::AutomaticReconstructionController(
   option_manager_.mapper->num_threads = options_.num_threads;
   option_manager_.poisson_meshing->num_threads = options_.num_threads;
 
-  option_manager_.two_view_geometry->ransac_options.random_seed =
-      options_.random_seed;
+  // Apply mapper-appropriate two-view geometry defaults.
+  // Global uses stricter thresholds; Incremental/Hierarchical use standard.
+  TwoViewGeometryOptions& tvg_options = *option_manager_.two_view_geometry;
```
ahojnnes: nit: two_view_geometry_options for consistency

src/colmap/controllers/automatic_reconstruction.h
```diff
@@ -120,6 +120,14 @@ class AutomaticReconstructionController : public Thread {
     // comma, e.g., "0,1,2,3". For single-GPU stages only the first GPU will be
     // used. By default, all available GPUs will be used in all stages.
     std::string gpu_index = "-1";
+
+    // Two-view geometry options for feature matching.
+    // Use -1 to indicate "use mapper-appropriate default".
+    // Incremental: max_error=4.0, min_num_inliers=15, min_inlier_ratio=0.0
+    // Global: max_error=1.0, min_num_inliers=30, min_inlier_ratio=0.25
```
ahojnnes: * Do we really have a use case for these? Otherwise, I suggest to remove them from the interface. If it's just for our own use case of doing a parameter sweep with ETH3D, I suggest to not commit the change to the main branch though.
* If there is a genuine use case, I suggest to remove the specific values here, as it will be difficult to keep in sync in the future.

src/colmap/controllers/automatic_reconstruction.cc
```diff
@@ -92,8 +92,33 @@ AutomaticReconstructionController::AutomaticReconstructionController(
...
+  tvg_options.ransac_options.random_seed = options_.random_seed;
+  if (options_.mapper == Mapper::GLOBAL) {
+    tvg_options.ransac_options.max_error = 1.0;
+    tvg_options.min_num_inliers = 30;
+    tvg_options.min_inlier_ratio = 0.25;
+    // Disable guided matching for GLOBAL mapper because guided matching adds
```
ahojnnes: As per discussion offline, I am not sure this is the case and we don't instead need to adjust the min_num_inliers or min_inlier_ratio values in this case. Maybe reformulate and make this a TODO, so we do a parameter sweep in the future?

src/colmap/controllers/automatic_reconstruction.h
B1ueber2y: Sure. This is removed.

src/colmap/controllers/automatic_reconstruction.cc
B1ueber2y: Updated the comments and added a TODO. PTAL.

PR #3951, opened 2026-01-05
Make view graph calibration a stand-alone estimator module on colmap::Database and add CLI. 

B1ueber2y:
This PR includes:
* Make view graph calibration work on database instead of reconstruction and ViewGraph. Add CLI.
* Move view graph calibration into colmap/estimators. Add nominal tests that can pass with the current cost functor. 
* Move view graph calibration out of the global mapper into GlobalPipeline, **to be further moved out as a stand-alone module inside automatic reconstructor in the future.** 
* With the simplified global mapper, we can remove the dependency of global mapper in the tests of rotation averaging and gravity refinement and move them to glomap/estimators.

### DIFF
--- a/.github/workflows/install-ccache.ps1
+++ b/.github/workflows/install-ccache.ps1
@@ -10,6 +10,8 @@ $url = "https://github.com/ccache/ccache/releases/download/v$version/$folder.zip
 $expectedSha256 = "98AEA520D66905B8BA7A8E648A4CC0CA941D5E119D441F1E879A4A9045BF18F6"
 
 $ErrorActionPreference = "Stop"
+Set-StrictMode -Version Latest
+$PSNativeCommandUseErrorActionPreference = $true
 
 try {
     New-Item -Path "$Destination" -ItemType Container -ErrorAction SilentlyContinue
--- a/python/ci/install-colmap-windows.ps1
+++ b/python/ci/install-colmap-windows.ps1
@@ -1,3 +1,7 @@
+$ErrorActionPreference = "Stop"
+Set-StrictMode -Version Latest
+$PSNativeCommandUseErrorActionPreference = $true
+
 $CURRDIR = $PWD
 
 $COMPILER_TOOLS_DIR = "${env:COMPILER_CACHE_DIR}/bin"
--- a/src/colmap/exe/image.cc
+++ b/src/colmap/exe/image.cc
@@ -510,12 +510,11 @@ int RunImageUndistorterStandalone(int argc, char** argv) {
     }
   }
 
-  std::unique_ptr<BaseController> undistorter;
-  undistorter.reset(new PureImageUndistorter(undistort_camera_options,
+  auto undistorter =
+      std::make_unique<PureImageUndistorter>(undistort_camera_options,
                                              *options.image_path,
                                              output_path,
-                                             image_names_and_cameras));
-
+                                             image_names_and_cameras);
   undistorter->Run();
 
   return EXIT_SUCCESS;
--- a/src/colmap/mvs/gpu_mat_ref_image.cu
+++ b/src/colmap/mvs/gpu_mat_ref_image.cu
@@ -83,11 +83,11 @@ __global__ void FilterKernel(const cudaTextureObject_t image_texture,
 }  // namespace
 
 GpuMatRefImage::GpuMatRefImage(const size_t width, const size_t height)
-    : height_(height), width_(width) {
-  image.reset(new GpuMat<uint8_t>(width, height));
-  sum_image.reset(new GpuMat<float>(width, height));
-  squared_sum_image.reset(new GpuMat<float>(width, height));
-}
+    : height_(height),
+      width_(width),
+      image(std::make_unique<GpuMat<uint8_t>>(width, height)),
+      sum_image(std::make_unique<GpuMat<float>>(width, height)),
+      squared_sum_image(std::make_unique<GpuMat<float>>(width, height)) {}
 
 void GpuMatRefImage::Filter(const uint8_t* image_data,
                             const size_t window_radius,
--- a/src/colmap/mvs/patch_match_cuda.cu
+++ b/src/colmap/mvs/patch_match_cuda.cu
@@ -1436,9 +1436,10 @@ void PatchMatchCuda::RunWithWindowSizeAndStep() {
 
       if (last_sweep) {
         if (options_.filter) {
-          consistency_mask_.reset(new GpuMat<uint8_t>(cost_map_->GetWidth(),
-                                                      cost_map_->GetHeight(),
-                                                      cost_map_->GetDepth()));
+          consistency_mask_ =
+              std::make_unique<GpuMat<uint8_t>>(cost_map_->GetWidth(),
+                                                cost_map_->GetHeight(),
+                                                cost_map_->GetDepth());
           consistency_mask_->FillWithScalar(0);
         }
         if (options_.geom_consistency) {
@@ -1538,7 +1539,7 @@ void PatchMatchCuda::InitRefImage() {
   ref_height_ = ref_image.GetHeight();
 
   // Upload to device and filter.
-  ref_image_.reset(new GpuMatRefImage(ref_width_, ref_height_));
+  ref_image_ = std::make_unique<GpuMatRefImage>(ref_width_, ref_height_);
   ref_image_->Filter(ref_image.GetBitmap().RowMajorData().data(),
                      options_.window_radius,
                      options_.window_step,
@@ -1749,9 +1750,9 @@ void PatchMatchCuda::InitTransforms() {
 }
 
 void PatchMatchCuda::InitWorkspaceMemory() {
-  rand_state_map_.reset(new GpuMatPRNG(ref_width_, ref_height_));
+  rand_state_map_ = std::make_unique<GpuMatPRNG>(ref_width_, ref_height_);
 
-  depth_map_.reset(new GpuMat<float>(ref_width_, ref_height_));
+  depth_map_ = std::make_unique<GpuMat<float>>(ref_width_, ref_height_);
   if (options_.geom_consistency) {
     const DepthMap& init_depth_map =
         problem_.depth_maps->at(problem_.ref_image_idx);
@@ -1762,27 +1763,27 @@ void PatchMatchCuda::InitWorkspaceMemory() {
         options_.depth_min, options_.depth_max, *rand_state_map_);
   }
 
-  normal_map_.reset(new GpuMat<float>(ref_width_, ref_height_, 3));
+  normal_map_ = std::make_unique<GpuMat<float>>(ref_width_, ref_height_, 3);
 
   // Note that it is not necessary to keep the selection probability map in
   // memory for all pixels. Theoretically, it is possible to incorporate
   // the temporary selection probabilities in the global_workspace_.
   // However, it is useful to keep the probabilities for the entire image
   // in memory, so that it can be exported.
-  sel_prob_map_.reset(new GpuMat<float>(
-      ref_width_, ref_height_, problem_.src_image_idxs.size()));
-  prev_sel_prob_map_.reset(new GpuMat<float>(
-      ref_width_, ref_height_, problem_.src_image_idxs.size()));
+  sel_prob_map_ = std::make_unique<GpuMat<float>>(
+      ref_width_, ref_height_, problem_.src_image_idxs.size());
+  prev_sel_prob_map_ = std::make_unique<GpuMat<float>>(
+      ref_width_, ref_height_, problem_.src_image_idxs.size());
   prev_sel_prob_map_->FillWithScalar(0.5f);
 
-  cost_map_.reset(new GpuMat<float>(
-      ref_width_, ref_height_, problem_.src_image_idxs.size()));
+  cost_map_ = std::make_unique<GpuMat<float>>(
+      ref_width_, ref_height_, problem_.src_image_idxs.size());
 
   const int ref_max_dim = std::max(ref_width_, ref_height_);
-  global_workspace_.reset(
-      new GpuMat<float>(ref_max_dim, problem_.src_image_idxs.size(), 2));
+  global_workspace_ = std::make_unique<GpuMat<float>>(
+      ref_max_dim, problem_.src_image_idxs.size(), 2);
 
-  consistency_mask_.reset(new GpuMat<uint8_t>(0, 0, 0));
+  consistency_mask_ = std::make_unique<GpuMat<uint8_t>>(0, 0, 0);
 
   ComputeCudaConfig();
 
@@ -1849,11 +1850,11 @@ void PatchMatchCuda::Rotate() {
   }
 
   // Rotate selection probability map.
-  prev_sel_prob_map_.reset(
-      new GpuMat<float>(width, height, problem_.src_image_idxs.size()));
+  prev_sel_prob_map_ = std::make_unique<GpuMat<float>>(
+      width, height, problem_.src_image_idxs.size());
   sel_prob_map_->Rotate(prev_sel_prob_map_.get());
-  sel_prob_map_.reset(
-      new GpuMat<float>(width, height, problem_.src_image_idxs.size()));
+  sel_prob_map_ = std::make_unique<GpuMat<float>>(
+      width, height, problem_.src_image_idxs.size());
 
   // Rotate cost map.
   {
--- a/src/colmap/util/version_test.cc
+++ b/src/colmap/util/version_test.cc
@@ -29,6 +29,7 @@
 
 #include "colmap/util/version.h"
 
+#include <gmock/gmock.h>
 #include <gtest/gtest.h>
 
 namespace colmap {
@@ -44,6 +45,18 @@ TEST(MakeDatabaseVersionNumber, Nominal) {
   EXPECT_ANY_THROW(MakeDatabaseVersionNumber(3, 100, 0, 0));
   EXPECT_ANY_THROW(MakeDatabaseVersionNumber(3, 14, 100, 0));
   EXPECT_ANY_THROW(MakeDatabaseVersionNumber(3, 14, 0, 100));
+}
+
+TEST(GetVersionInfo, Nominal) {
+  const std::string version = GetVersionInfo();
+  EXPECT_FALSE(version.empty());
+  EXPECT_THAT(version, testing::HasSubstr("COLMAP"));
+}
+
+TEST(GetBuildInfo, Nominal) {
+  const std::string build_info = GetBuildInfo();
+  EXPECT_FALSE(build_info.empty());
+  EXPECT_THAT(build_info, testing::HasSubstr("Commit"));
 }
 
 }  // namespace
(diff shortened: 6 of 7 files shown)


src/colmap/controllers/global_pipeline.cc
```diff
@@ -46,6 +46,26 @@ GlobalPipeline::GlobalPipeline(
...
+    LOG(INFO) << "----- Running view graph calibration -----";
+    Timer run_timer;
+    run_timer.Start();
+    ViewGraphCalibrationOptions vgc_options = options_.view_graph_calibration;
+    if (options_.random_seed >= 0) {
+      vgc_options.random_seed = options_.random_seed;
+    }
+    if (options_.num_threads > 0) {
+      vgc_options.solver_options.num_threads = options_.num_threads;
+    }
```
ahojnnes: I think we can omit the if conditions here and just copy through the values, as the ViewGraphCalib code handles any special values.

src/colmap/controllers/rotation_averaging.cc
```diff
@@ -49,10 +49,7 @@ void RotationAveragingController::Run() {
   RotationAveragingControllerOptions options = options_;
   if (options.random_seed >= 0) {
```
ahojnnes: Same here. Do we need the if condition?

src/colmap/estimators/view_graph_calibration.cc
```diff
@@ -0,0 +1,458 @@
...
+  camera_t camera_id2;
+};
+
+// Result of focal length calibration.
+struct FocalLengthCalibrationResult {
+  // Optimized focal lengths per camera.
+  std::unordered_map<camera_t, double> focal_lengths;
+  // Squared calibration error per image pair.
+  std::unordered_map<image_pair_t, double> calibration_errors_sq;
+  // Whether the Ceres solver succeeded.
```
ahojnnes:
```suggestion
  // Whether the calibration succeeded.
```

src/colmap/estimators/view_graph_calibration.cc
```diff
@@ -0,0 +1,458 @@
...
+  Eigen::Matrix3d F;
+  camera_t camera_id1;
+  camera_t camera_id2;
+};
+
+// Result of focal length calibration.
+struct FocalLengthCalibrationResult {
+  // Optimized focal lengths per camera.
+  std::unordered_map<camera_t, double> focal_lengths;
+  // Squared calibration error per image pair.
```
ahojnnes: Specify the unit of error?

src/colmap/estimators/view_graph_calibration.h
```diff
@@ -0,0 +1,93 @@
...
+  double relpose_min_inlier_ratio = 0.25;
+
+  ViewGraphCalibrationOptions() {
+    solver_options.num_threads = -1;
+    solver_options.max_num_iterations = 100;
+    solver_options.function_tolerance = 1e-5;
+  }
+
+  // Create loss function for given options.
+  std::unique_ptr<ceres::LossFunction> CreateLossFunction() const {
```
ahojnnes: Needs to be inlined or maybe better move to .cc file.

src/colmap/estimators/view_graph_calibration.h
```diff
@@ -0,0 +1,93 @@
...
+  }
+};
+
+// Calibrate the view graph by estimating focal lengths from fundamental
+// matrices. This function operates directly on the database, reading both
+// UNCALIBRATED and CALIBRATED two-view geometries along with their associated
+// cameras. It optimizes focal lengths and updates the camera intrinsics in the
+// database. Image pairs with low calibration error have their essential
+// matrices computed and relative poses re-estimated, then are upgraded to
+// CALIBRATED. Pairs with high calibration error are tagged as DEGENERATE_VGC.
```
ahojnnes: I suggest not introducing a new two view geometry config value. Do we really need this? I think for the tests we can work around this and just identify the pairs that flipped from non-degenerate to degenerate after calibration?

ahojnnes (approved):

src/colmap/estimators/view_graph_calibration.cc
```diff
@@ -0,0 +1,458 @@
...
+    auto error_it = calib_result.calibration_errors_sq.find(pair_id);
+    if (error_it == calib_result.calibration_errors_sq.end()) continue;
+
+    if (error_it->second > max_calibration_error_sq) {
+      invalid_counter++;
+      tvg.config = TwoViewGeometry::DEGENERATE_VGC;
+      database->UpdateTwoViewGeometry(image_id1, image_id2, tvg);
+    } else {
+      const Camera& c1 = cameras.at(image_to_camera.at(image_id1));
+      const Camera& c2 = cameras.at(image_to_camera.at(image_id2));
```
ahojnnes: camera1 camera2

src/colmap/estimators/view_graph_calibration.cc
```diff
@@ -0,0 +1,458 @@
...
+  FocalLengthCalibrationResult result;
+
+  if (inputs.empty()) {
+    result.success = true;
+    return result;
+  }
+
+  // Initialize focal lengths from cameras.
+  std::unordered_map<camera_t, double> focals;
+  std::unordered_map<camera_t, double> original_focals;
```
ahojnnes: reserve?

src/colmap/estimators/view_graph_calibration.cc
```diff
@@ -0,0 +1,458 @@
...
+    const std::unordered_map<camera_t, Camera>& cameras) {
+  FocalLengthCalibrationResult result;
+
+  if (inputs.empty()) {
+    result.success = true;
+    return result;
+  }
+
+  // Initialize focal lengths from cameras.
+  std::unordered_map<camera_t, double> focals;
```
ahojnnes: Use a single unordered_map for both focals with a struct or std::pair?

src/colmap/estimators/view_graph_calibration.cc
```diff
@@ -0,0 +1,458 @@
...
+          loss_function.get(),
+          &(focals[input.camera_id1]));
+    } else {
+      problem.AddResidualBlock(
+          FetzerFocalLengthCostFunctor::Create(
+              input.F,
+              cameras.at(input.camera_id1).PrincipalPoint(),
+              cameras.at(input.camera_id2).PrincipalPoint()),
+          loss_function.get(),
+          &(focals[input.camera_id1]),
```
ahojnnes: nit
```suggestion
          &focals[input.camera_id1],
```

src/colmap/estimators/view_graph_calibration.cc
```diff
@@ -0,0 +1,458 @@
...
+    }
+  }
+
+  // Parameterize cameras (fix those with prior, set lower bound).
+  size_t num_cameras = 0;
+  for (const auto& [camera_id, camera] : cameras) {
+    if (!problem.HasParameterBlock(&(focals[camera_id]))) continue;
+
+    num_cameras++;
+    problem.SetParameterLowerBound(&(focals[camera_id]), 0, 1e-3);
```
ahojnnes: Extract 1-e3 as a constexpr?

ahojnnes: Surprised we are not setting an upper bound.

src/colmap/estimators/view_graph_calibration.cc
```diff
@@ -0,0 +1,458 @@
...
+    if (camera.has_prior_focal_length) {
+      problem.SetParameterBlockConstant(&(focals[camera_id]));
+      num_cameras--;
```
ahojnnes: Maybe just increment in the else branch instead of incr/decr?

src/colmap/estimators/view_graph_calibration.cc
ahojnnes: Also wondering whether it makes sense to bound the parameter using options.min_focal_length_ratio/options.max_focal_length_ratio?

src/colmap/estimators/view_graph_calibration.cc
```diff
@@ -0,0 +1,458 @@
...
+
+  // Read cameras and build image_id -> camera_id mapping.
+  std::unordered_map<camera_t, Camera> cameras;
+  for (const Camera& cam : database->ReadAllCameras()) {
+    cameras[cam.camera_id] = cam;
+  }
+  std::unordered_map<image_t, camera_t> image_to_camera;
+  for (const Image& img : database->ReadAllImages()) {
+    image_to_camera[img.ImageId()] = img.CameraId();
+  }
```
ahojnnes: nit: camera, image

src/colmap/estimators/view_graph_calibration.cc
```diff
@@ -0,0 +1,458 @@
...
+
+    if (tvg.config == TwoViewGeometry::CALIBRATED) {
+      camera_counter[camera_id1].first++;
+      camera_counter[camera_id2].first++;
+      camera_counter[camera_id1].second++;
+      camera_counter[camera_id2].second++;
+    } else if (tvg.config == TwoViewGeometry::UNCALIBRATED) {
+      camera_counter[camera_id1].first++;
+      camera_counter[camera_id2].first++;
+    }
```
ahojnnes, line 91:
```suggestion
    auto& [total1, calibrated1] = camera_counter[camera_id1];
    auto& [total2, calibrated2] = camera_counter[camera_id2];
    ++total1;
    ++total2;
    if (tvg.config == TwoViewGeometry::CALIBRATED) {
      ++calibrated1;
      ++calibrated2;
    }
```

src/colmap/estimators/view_graph_calibration.cc
```diff
@@ -0,0 +1,458 @@
...
+  }
+
+  // Camera is valid if majority (>50%) of its pairs are calibrated.
+  std::unordered_map<camera_t, bool> camera_validity;
```
ahojnnes, line 95: reserve

src/colmap/estimators/view_graph_calibration.cc
```diff
@@ -0,0 +1,458 @@
...
+  for (auto& [camera_id, counter] : camera_counter) {
+    camera_validity[camera_id] = counter.second * 1. / counter.first > 0.5;
```
ahojnnes: extract as constexpr's or better expose as parameter in the ViewGraphCalibOptions.

src/colmap/estimators/view_graph_calibration.cc
```diff
@@ -0,0 +1,458 @@
...
+  // Convert UNCALIBRATED pairs to CALIBRATED if both cameras are valid.
+  for (auto& [pair_id, tvg] : pairs) {
+    if (tvg.config != TwoViewGeometry::UNCALIBRATED) continue;
+
+    const auto [image_id1, image_id2] = PairIdToImagePair(pair_id);
+    const camera_t camera_id1 = image_to_camera.at(image_id1);
+    const camera_t camera_id2 = image_to_camera.at(image_id2);
+
+    if (camera_validity[camera_id1] && camera_validity[camera_id2]) {
+      tvg.config = TwoViewGeometry::CALIBRATED;
```
ahojnnes, line 115: This is a bit dangerous, if there is no essential matrix estimated for the pair, as subsequent steps may assume presence of E in the tvg struct.

ahojnnes: ... or I guess this is resolved by re-estimating the relative poses below?

src/colmap/estimators/view_graph_calibration.cc
```diff
@@ -0,0 +1,458 @@
...
+    const std::unordered_map<image_t, camera_t>& image_to_camera,
+    Database* database) {
+  LOG(INFO) << "Re-estimating relative poses for " << pairs.size() << " pairs";
+
+  TwoViewGeometryOptions two_view_options;
+  two_view_options.compute_relative_pose = true;
+  two_view_options.ransac_options.max_error = options.relpose_max_error;
+  two_view_options.min_num_inliers = options.relpose_min_num_inliers;
+  two_view_options.min_inlier_ratio = options.relpose_min_inlier_ratio;
+  if (options.random_seed >= 0) {
```
ahojnnes: No if condition needed?

src/colmap/estimators/view_graph_calibration.cc
```diff
@@ -0,0 +1,458 @@
...
+    two_view_options.ransac_options.random_seed = options.random_seed;
+  }
+
+  // Pre-read all keypoints and matches from database (SQLite is not
+  // thread-safe).
+  std::unordered_map<image_t, std::vector<Eigen::Vector2d>> image_points;
```
ahojnnes: reserve?

src/colmap/estimators/view_graph_calibration.cc
```diff
@@ -0,0 +1,458 @@
...
+    std::vector<Eigen::Vector2d> points(keypoints.size());
+    for (size_t j = 0; j < keypoints.size(); ++j) {
+      points[j] = Eigen::Vector2d(keypoints[j].x, keypoints[j].y);
+    }
+    image_points[image_id] = std::move(points);
+  }
+
+  for (size_t i = 0; i < pairs.size(); ++i) {
+    const auto [image_id1, image_id2] = PairIdToImagePair(pairs[i].first);
+    pair_matches[i] = database->ReadMatches(image_id1, image_id2);
```
ahojnnes: Loading all matches into memory is unnecessary. How about loading them from the threads? A simple mutex should work.

src/colmap/estimators/view_graph_calibration.cc
```diff
@@ -0,0 +1,458 @@
...
+bool CalibrateViewGraph(const ViewGraphCalibrationOptions& options,
+                        Database* database) {
+  THROW_CHECK_NOTNULL(database);
+
+  // Read cameras and build image_id -> camera_id mapping.
+  std::unordered_map<camera_t, Camera> cameras;
+  for (const Camera& cam : database->ReadAllCameras()) {
+    cameras[cam.camera_id] = cam;
+  }
+  std::unordered_map<image_t, camera_t> image_to_camera;
```
ahojnnes: Could instead store directly the camera pointer to avoid the extra hash map lookup.

src/colmap/estimators/view_graph_calibration_test.cc
```diff
@@ -0,0 +1,231 @@
...
+  std::unordered_map<camera_t, double> gt_focals;
+  for (const auto& [camera_id, camera] : reconstruction.Cameras()) {
+    gt_focals[camera_id] = camera.MeanFocalLength();
+  }
+
+  // Change pairs to UNCALIBRATED so F matrices are used directly.
+  for (const auto& [pair_id, tvg] : database->ReadTwoViewGeometries()) {
+    const auto [image_id1, image_id2] = PairIdToImagePair(pair_id);
+    TwoViewGeometry uncalib_tvg = tvg;
+    uncalib_tvg.config = TwoViewGeometry::UNCALIBRATED;
```
ahojnnes: We should maybe instead fix this in the dataset synthesis. If camera_has_prior_focal_length==false, we set uncalibrated?

src/colmap/estimators/view_graph_calibration_test.cc
```diff
@@ -0,0 +1,231 @@
...
+    Camera camera = database->ReadCamera(camera_id);
+    const double noise = RandomUniformReal(-50.0, 50.0);
+    for (const size_t idx : camera.FocalLengthIdxs()) {
+      camera.params[idx] += noise;
+    }
+    database->UpdateCamera(camera);
+  }
+
+  ViewGraphCalibrationOptions calib_options;
+  calib_options.random_seed = 42;
```
ahojnnes: Should be unnecessary if using SetPRNG seed?

src/colmap/estimators/view_graph_calibration_test.cc
```diff
@@ -0,0 +1,231 @@
...
+  options.camera_has_prior_focal_length = true;
+
+  Reconstruction reconstruction;
+  SynthesizeDataset(options, &reconstruction, database.get());
+
+  // Store original focal lengths (which have priors).
+  std::unordered_map<camera_t, double> original_focals;
+  for (const auto& [camera_id, camera] : reconstruction.Cameras()) {
+    original_focals[camera_id] = camera.MeanFocalLength();
+    EXPECT_TRUE(camera.has_prior_focal_length);
```
ahojnnes: Would remove this, as this is testing SynthesizeDataset rather than view graph calib.

src/colmap/estimators/view_graph_calibration_test.cc
```diff
@@ -0,0 +1,231 @@
...
+  size_t degenerate_count = 0;
+  for (const auto& [pid, geom] : database->ReadTwoViewGeometries()) {
+    if (geom.config == TwoViewGeometry::CALIBRATED) {
+      calibrated_count++;
+    } else if (geom.config == TwoViewGeometry::DEGENERATE_VGC) {
+      degenerate_count++;
+    }
+  }
+  // At least some pairs should be processed.
+  EXPECT_GT(calibrated_count + degenerate_count, 0);
```
ahojnnes: I am not sure I understand this logic. Should we instead expect both calibrated and degenerate to individually be non-zero?

src/colmap/estimators/view_graph_calibration_test.cc
```diff
@@ -0,0 +1,231 @@
...
+  Reconstruction reconstruction;
+  SynthesizeDataset(options, &reconstruction, database.get());
+
+  // Store ground truth relative poses.
+  std::unordered_map<image_pair_t, Rigid3d> gt_poses;
+  for (const auto& [pair_id, tvg] : database->ReadTwoViewGeometries()) {
+    const auto [image_id1, image_id2] = PairIdToImagePair(pair_id);
+    const Image& image1 = reconstruction.Image(image_id1);
+    const Image& image2 = reconstruction.Image(image_id2);
+    gt_poses[pair_id] = image2.CamFromWorld() * Inverse(image1.CamFromWorld());
```
ahojnnes, line 199: Should we perturb the input geometries by more than the rotation/translation thresholds below to make sure the function does not cheat by copying through the values?

src/colmap/estimators/view_graph_calibration.cc
B1ueber2y: Lets for now extract 1e-3 as a constexpr and keep it as it is, to keep the logic equivalent, as bounding the parameter directly may end up not reverting those diverged focals (not sure if there are any). Maybe this can be resolved by checking whether the focal lengths are exactly at the boundary after optimization, but this needs some studies and goes beyond the scope of this PR.

src/colmap/estimators/view_graph_calibration.cc
B1ueber2y: Added a line to recalculate E from F, in case relative pose re-estimation is skipped.

src/colmap/estimators/view_graph_calibration_test.cc
B1ueber2y: Indeed. This test was a bit meaningless. I fixed it now.

src/colmap/estimators/view_graph_calibration.h
B1ueber2y: Indeed. This new config value is removed now.

PR #3952, opened 2026-01-05
Migrate glomap exe into colmap.

B1ueber2y:
In preparation for pulling out the view graph calibrator. Logic is equivalent.

### DIFF
--- a/src/colmap/controllers/option_manager.cc
+++ b/src/colmap/controllers/option_manager.cc
@@ -728,6 +728,7 @@ void OptionManager::AddGlobalMapperOptions() {
       "POINTS_AND_CAMERAS}");
 
   // Bundle adjustment options.
+  // TODO: Consolidate these with existing BA options (e.g., Mapper.ba_*).
   AddAndRegisterDefaultOption("GlobalMapper.BundleAdjustment.use_gpu",
                               &global_mapper->mapper.bundle_adjustment.use_gpu);
   AddAndRegisterDefaultOption(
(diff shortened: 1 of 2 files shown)


src/colmap/exe/sfm.cc
```diff
@@ -351,6 +358,62 @@ int RunMapper(int argc, char** argv) {
...
+                               reconstruction_manager);
+  global_mapper.Run();
+
+  if (reconstruction_manager->Size() == 0) {
+    LOG(ERROR) << "Failed to create sparse model";
+    return EXIT_FAILURE;
+  }
+
+  // Write reconstructions with specified format.
+  for (size_t i = 0; i < reconstruction_manager->Size(); ++i) {
```
ahojnnes: Suggest to just use ReconstructionManager::Write here and not support multiple output formats. If somebody really wants txt, they can run the model_converter afterwards.

src/colmap/controllers/option_manager.cc
```diff
@@ -636,6 +642,160 @@ void OptionManager::AddMapperOptions() {
...
+      &global_mapper->mapper.bundle_adjustment.optimize_rotations);
+  AddAndRegisterDefaultOption(
+      "GlobalMapper.BundleAdjustment.optimize_translation",
+      &global_mapper->mapper.bundle_adjustment.optimize_translation);
+  AddAndRegisterDefaultOption(
+      "GlobalMapper.BundleAdjustment.optimize_intrinsics",
+      &global_mapper->mapper.bundle_adjustment.optimize_intrinsics);
+  AddAndRegisterDefaultOption(
+      "GlobalMapper.BundleAdjustment.optimize_principal_point",
+      &global_mapper->mapper.bundle_adjustment.optimize_principal_point);
```
ahojnnes, line 751: Can we name these consistently with existing BA options?

B1ueber2y: This is non-trivial since the options have slightly different semantic meanings. Added a TODO here and left for future work.

PR #3959, opened 2026-01-06
Add Ceres autodiff support for CamFromImg via implicit function theorem.

B1ueber2y:
Exciting new feature. Did not expect this to work this well. All tests on grid pixels for every camera model works. Each pixel is added a random noise from -10 to 10 on x and y, and there is also a check to make sure that at least 80% of the test cases are not skipped due to bad initialization.


src/colmap/sensor/models.h
```diff
@@ -712,16 +756,98 @@ bool BaseCameraModel<CameraModel>::IterativeUndistortion(const double* params,
...
+bool BaseCameraModel<CameraModel>::IterativeUndistortionWithJacobian(
+    const T* extra_params, T* u, T* v) {
+  // Extract scalar values
+  constexpr size_t N = CameraModel::num_extra_params;
+  double extra_params_scalar[N];
+  for (size_t i = 0; i < N; ++i) {
+    extra_params_scalar[i] = GetScalarValue(extra_params[i]);
+  }
+  double u_scalar = GetScalarValue(*u);
+  double v_scalar = GetScalarValue(*v);
```
ahojnnes: const

src/colmap/sensor/models.h
```diff
@@ -712,16 +756,98 @@ bool BaseCameraModel<CameraModel>::IterativeUndistortion(const double* params,
...
   *u = x(0);
   *v = x(1);
+  if (J_out) *J_out = J;
 
   return false;
 }
 
+template <typename CameraModel>
+template <typename T>
+bool BaseCameraModel<CameraModel>::IterativeUndistortionWithJacobian(
```
ahojnnes: It feels like we could just have one IterativeUndistortion function and add the templating to the one above?

src/colmap/sensor/models.h
```diff
@@ -664,9 +709,8 @@ T BaseCameraModel<CameraModel>::CamFromImgThreshold(const T* params,
 }
 
 template <typename CameraModel>
-bool BaseCameraModel<CameraModel>::IterativeUndistortion(const double* params,
-                                                         double* u,
-                                                         double* v) {
+bool BaseCameraModel<CameraModel>::IterativeUndistortion(
+    const double* params, double* u, double* v, Eigen::Matrix2d* J_out) {
```
ahojnnes: Should we call the "extra_params" here as well?

src/colmap/sensor/models_test.cc
```diff
@@ -29,11 +29,87 @@
...
+    T u, v;
+    if (!CameraModel::CamFromImg(params_T, xy[0], xy[1], &u, &v)) {
+      return false;
+    }
+    residuals[0] = u - T(target_u_);
+    residuals[1] = v - T(target_v_);
+    return true;
+  }
+
+  const double* params_;
```
ahojnnes, line 64: extra_params?

src/colmap/sensor/models_test.cc
```diff
@@ -29,11 +29,87 @@
...
+  double target_u, target_v;
+  if (!CameraModel::CamFromImg(params.data(), x0, y0, &target_u, &target_v)) {
+    return false;  // Skip if undistortion fails
+  }
+
+  // Add random noise to pixel in range [-10, 10]
+  double xy[2] = {x0 + RandomUniformReal(-10.0, 10.0),
+                  y0 + RandomUniformReal(-10.0, 10.0)};
+
+  // Skip if noisy pixel is out of valid range
```
ahojnnes, line 86: Could clip to valid range?

ahojnnes (commented):
Great idea and implementation. Didn't double check the math and relying on passing tests :-) Wondering how fast/slow this is, given that each residual evaluation now runs an iterative undistortion?

B1ueber2y:
> Great idea and implementation. Didn't double check the math and relying on passing tests :-) Wondering how fast/slow this is, given that each residual evaluation now runs an iterative undistortion?

The jacobians was already calculated in the process of IterativeUndistortion so there should not be any runtime degradation for double type. For ceres::Jet there is one overhead of chaining Jacobians, which is only run once after IterativeUndistortion.

src/colmap/sensor/models.h
B1ueber2y: In this current implementation, there is only one jacobian chaining via implicit function theorem after IterativeUndistortion, assuming that the convergence is reached. If we directly use template in IterativeUndistortion there would be jacobian propagation in each iteration, which would be much slower and potentially unstable.

B1ueber2y: I suggest to keep the two methods as it is. I can make IterativeUndistortion private though.

ahojnnes:
> > Great idea and implementation. Didn't double check the math and relying on passing tests :-) Wondering how fast/slow this is, given that each residual evaluation now runs an iterative undistortion?
> 
> The jacobians was already calculated in the process of IterativeUndistortion so there should not be any runtime degradation for double type. For ceres::Jet there is one overhead of chaining Jacobians, which is only run once after IterativeUndistortion.

Yes, I am not suggesting it is slower than before, as we never used CamFromImg in a cost function so far. I was just curious how fast or slow this is in generaly given the iterative loop inside the residual evaluation.

src/colmap/sensor/models.h
B1ueber2y: Renamed to IterativeUndistortion and IterativeUndistortionScalar, and moved the latter to private.

src/colmap/sensor/models_test.cc
B1ueber2y: Here this is actually params instead of extra_params.

src/colmap/sensor/models_test.cc
B1ueber2y: The clipping is tricky for Fisheye models, for which not all pixels are valid for undistortion.

PR #3960, opened 2026-01-06
Add relative poses to DatabaseCache and update the signature of global mapper. 

B1ueber2y:
Breaking change: Now the static methods of DatabaseCache, including Create, CreateFromCache, and Load, take a DatabaseCache::Options object rather than scattered arguments.

Replace database_ member with database_cache_ throughout the global mapper logic except for GlobalPipeline, which requires view graph calibrator to live outside the pipeline logic.

### DIFF
--- a/src/pycolmap/scene/database_cache.cc
+++ b/src/pycolmap/scene/database_cache.cc
@@ -1,5 +1,6 @@
 #include "colmap/scene/database_cache.h"
 
+#include "pycolmap/helpers.h"
 #include "pycolmap/pybind11_extension.h"
 
 #include <pybind11/eigen.h>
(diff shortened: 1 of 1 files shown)


src/colmap/scene/database_cache.cc
```diff
@@ -267,15 +268,47 @@ void DatabaseCache::Load(const Database& database,
...
+  }
 }
 
 std::shared_ptr<DatabaseCache> DatabaseCache::Create(
     const Database& database,
     const size_t min_num_matches,
     const bool ignore_watermarks,
-    const std::unordered_set<std::string>& image_names) {
+    const std::unordered_set<std::string>& image_names,
+    const bool load_relative_pose) {
```
ahojnnes: Maybe it's now time to introduce options `class DatabaseCache { struct Options { ... }; };`

src/colmap/scene/database_cache.h
```diff
@@ -62,16 +63,23 @@ class DatabaseCache {
   //                              frame is included, all other images in the
   //                              same frame will also be included. All images
   //                              are used if empty.
+  // @param load_relative_pose    Whether to load/compute relative poses
+  //                              (cam2_from_cam1) for image pairs. If the pose
+  //                              is not stored in the database, it will be
+  //                              decomposed from the essential/fundamental
+  //                              matrix.
```
ahojnnes: I cannot find this part?

ahojnnes (approved):
LGTM

src/colmap/scene/database_cache.h
B1ueber2y: Sorry this is legacy from the development. Let me remove it.

src/colmap/scene/database_cache.cc
B1ueber2y: Nice idea. This is done. To align the pybind interfaces with C++, we would have to introduce breaking changes there. Updated the PR description.

B1ueber2y:
The regression issues actually happened here. If every camera is calibrated, the view graph calibration will directly return true without re-estimating relative poses, leaving all cam2_from_cam1 empty. Will send in a fix.

PR #3963, opened 2026-01-07
Add test for reading/writing options from ini

ahojnnes:
As part of this new test, I found that not all options were registered correctly and could be parsed from the ini. Now registered all options by default.


src/colmap/controllers/base_option_manager.h
```diff
@@ -179,19 +171,6 @@ void BaseOptionManager::AddDefaultOption(const std::string& name,
         boost::program_options::value<T>(option)->default_value(*option),
         help_text.c_str());
   }
```
B1ueber2y, line 172: I m confused. Should we do RegisterOption also here?

ahojnnes: It's here. The diff presentation is just confusing :-)

PR #3967, opened 2026-01-08
Add unit tests for incremental triangulator

ahojnnes:


src/colmap/sfm/incremental_triangulator_test.cc
```diff
@@ -48,5 +93,262 @@ TEST(IncrementalTriangulator, Print) {
...
+
+  DeleteOneObservationFromEachTrack(reconstruction);
+
+  IncrementalTriangulator triangulator(cache->CorrespondenceGraph(),
+                                       reconstruction);
+
+  triangulator.CompleteImage(IncrementalTriangulator::Options(),
+                             reconstruction.RegImageIds().at(0));
+  EXPECT_EQ(reconstruction.NumPoints3D(), synthetic_options.num_points3D);
+  EXPECT_EQ(
```
B1ueber2y, line 208: Relevant to the other comment, is there a risk that this gets affected by the cheirality check when synthesizing recnostruction? Same for all the tests below.

src/colmap/sfm/incremental_triangulator_test.cc
```diff
@@ -48,5 +93,262 @@ TEST(IncrementalTriangulator, Print) {
...
+  IncrementalTriangulator triangulator(cache->CorrespondenceGraph(),
+                                       reconstruction);
+  size_t total_tris = 0;
+  for (const image_t image_id : reconstruction.RegImageIds()) {
+    total_tris += triangulator.TriangulateImage(
+        IncrementalTriangulator::Options(), image_id);
+  }
+
+  EXPECT_EQ(reconstruction.NumPoints3D(), synthetic_options.num_points3D);
+  EXPECT_GT(total_tris, synthetic_options.num_points3D);
```
B1ueber2y: Why not do EXPECT_EQ here similar to the other tests?

src/colmap/sfm/incremental_triangulator_test.cc
ahojnnes: No, the synthesis makes sure points are infront of the cameras.

src/colmap/sfm/incremental_triangulator_test.cc
ahojnnes: Good catch. Vibe coding mistake :-)

src/colmap/sfm/incremental_triangulator_test.cc
B1ueber2y: From my understanding, in the synthesis the observation was not added if it is behind the camera: https://github.com/colmap/colmap/blob/main/src/colmap/scene/synthetic.cc#L383, which can potentially affect the counting in this PR. If we intentionally require that this never happens, it should be a throw rather than continue there?

ahojnnes: I turned it into a THROW_CHECK.

PR #3968, opened 2026-01-08
Add option to synthesize sparse view graph

ahojnnes:
New tooling for writing more realistic tests.

### DIFF
--- a/src/colmap/scene/synthetic.cc
+++ b/src/colmap/scene/synthetic.cc
@@ -57,9 +57,10 @@ void AddOutlierMatches(double inlier_ratio,
 
 std::vector<image_pair_t> ExtractExhaustiveImagePairs(
     const Reconstruction& reconstruction) {
+  const size_t num_exhaustive_pairs =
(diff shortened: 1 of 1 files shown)


src/colmap/scene/synthetic.cc
```diff
@@ -174,6 +207,99 @@ void SynthesizeChainedMatches(double inlier_match_ratio,
...
+  if (sparsity == 0.0) {
+    SynthesizeExhaustiveMatches(inlier_match_ratio, reconstruction, database);
+    return;
+  }
+
+  if (sparsity == 1.0) {
+    return;
+  }
+
+  std::map<image_pair_t, TwoViewGeometry> all_two_view_geometries =
```
B1ueber2y: Nit: we actually do not need exhaustive two-view geometry in this function? we can first do the edge sampling and then build only on the active edges?

ahojnnes: Right, thanks and fixed.

PR #3969, opened 2026-01-08
Bugfix in pose decomposition interface. Update pipeline tests.

B1ueber2y:
Another bug fix for https://github.com/colmap/colmap/pull/3960. These new tests should have been added in https://github.com/colmap/colmap/pull/3966. 

Updated all sfm pipeline tests to always take database without cam2_from_cam1 available, which is the common case from the results of colmap feature matching. Also updated the AutomaticReconstructorController test to cover all three modes (INCREMENTAL, HIERARCHICAL, GLOBAL).

### DIFF
--- a/src/colmap/scene/synthetic.cc
+++ b/src/colmap/scene/synthetic.cc
@@ -386,10 +386,8 @@ void SynthesizeDataset(const SyntheticDatasetOptions& options,
           Point2D point2D;
           const std::optional<Eigen::Vector2d> proj_point2D =
               camera.ImgFromCam(cam_from_world * point3D.xyz);
-          if (!proj_point2D.has_value()) {
-            continue;  // Point is behind the camera.
-          }
-          point2D.xy = proj_point2D.value();
+          THROW_CHECK(proj_point2D.has_value());
+          point2D.xy = *proj_point2D;
           if (point2D.xy(0) >= 0 && point2D.xy(1) >= 0 &&
               point2D.xy(0) <= camera.width && point2D.xy(1) <= camera.height) {
             point2D.point3D_id = point3D_id;
--- a/src/colmap/sfm/incremental_triangulator_test.cc
+++ b/src/colmap/sfm/incremental_triangulator_test.cc
@@ -29,15 +29,60 @@
 
 #include "colmap/sfm/incremental_triangulator.h"
 
+#include "colmap/scene/database_cache.h"
+#include "colmap/scene/database_sqlite.h"
+#include "colmap/scene/synthetic.h"
+
+#include <gmock/gmock.h>
 #include <gtest/gtest.h>
 
 namespace colmap {
 namespace {
 
+void DeleteAllPoints3D(Reconstruction& reconstruction) {
+  std::vector<point3D_t> point3D_ids_to_delete;
+  for (const auto point3D_id : reconstruction.Point3DIds()) {
+    point3D_ids_to_delete.push_back(point3D_id);
+  }
+  for (const auto point3D_id : point3D_ids_to_delete) {
+    reconstruction.DeletePoint3D(point3D_id);
+  }
+  EXPECT_EQ(reconstruction.NumPoints3D(), 0);
+}
+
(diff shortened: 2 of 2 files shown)


B1ueber2y:
The automatic reconstructor pipeline integration test with the SynthesizeImages is a bit shaky for the global pipeline.

src/colmap/controllers/incremental_pipeline_test.cc
```diff
@@ -51,6 +51,7 @@ TEST(IncrementalPipeline, WithoutNoise) {
   synthetic_dataset_options.num_frames_per_rig = 7;
   synthetic_dataset_options.num_points3D = 50;
   synthetic_dataset_options.camera_has_prior_focal_length = false;
+  synthetic_dataset_options.two_view_geometry_has_relative_pose = false;
```
ahojnnes: Why are we disabling this for all the tests?

src/colmap/controllers/global_pipeline_test.cc
```diff
@@ -53,6 +53,7 @@ TEST(GlobalPipeline, Nominal) {
   synthetic_dataset_options.num_frames_per_rig = 7;
   synthetic_dataset_options.num_points3D = 50;
   synthetic_dataset_options.camera_has_prior_focal_length = false;
+  synthetic_dataset_options.two_view_geometry_has_relative_pose = false;
```
ahojnnes: Should we have one test where it's true but the inputs are bad to test whether the re-estimation does the right thing?

B1ueber2y:
> The automatic reconstructor pipeline integration test with the SynthesizeImages is a bit shaky for the global pipeline.

Disabled the GLOBAL mode for now in the automatic reconstructor test and added a TODO, to first unblock the fix from this PR to be merged.

src/colmap/controllers/incremental_pipeline_test.cc
ahojnnes: Before it was always present?

ahojnnes: If it is the common case, then I suggest to default it to false in the synthetic options and set it to true in the tests that require it. I assume this is mostly all the two-view geometry tests?

B1ueber2y: Done: https://github.com/colmap/colmap/pull/3969/commits/45e11a095e08d512a574774fe325a8c35cbcc82f. Even the two view geometry tests do not need it to be true, as it uses a custom function to synthesize test data. The only few places that need cam2_from_cam1 are guided geometric verifier and some tests for glomap modules, including global mapper, rotation averaging and gravity refinement.

src/colmap/controllers/global_pipeline_test.cc
B1ueber2y: Thanks! This is added: https://github.com/colmap/colmap/pull/3969/commits/9b5e1f67c868aa1b20226d5a8ecdfaf26e43f215. However, since we plan to move view graph calibration outside global pipeline soon, this test will be removed in the near future. This was also already covered in view_graph_calibration_test.cc.

src/colmap/estimators/two_view_geometry_test.cc
```diff
@@ -889,5 +889,47 @@ TEST(EstimateMultipleTwoViewGeometries, MultipleGeometries) {
...
+  SyntheticDatasetOptions synthetic_dataset_options;
+  synthetic_dataset_options.num_rigs = 2;
+  synthetic_dataset_options.num_cameras_per_rig = 1;
+  synthetic_dataset_options.num_frames_per_rig = 1;
+  synthetic_dataset_options.num_points3D = 50;
+  synthetic_dataset_options.camera_has_prior_focal_length = true;
+  SynthesizeDataset(synthetic_dataset_options, &reconstruction, database.get());
+
+  const image_t image_id1 = 1;
+  const image_t image_id2 = 2;
```
ahojnnes: nit: kImageId1, kImageId2

ahojnnes (approved):

PR #3954, opened 2026-01-05
Replace ImagePair with RelativePoseData and rename ViewGraph to PoseGraph. 

B1ueber2y:
Dependent on https://github.com/colmap/colmap/pull/3951.

With the view graph calibration update (upstream PR), we can finally remove the need of config, E, F, H, matches in ImagePair struct. The whole pipeline only needs cam2_from_cam1 and inlier_matches, which is put into a new minimal struct RelativePoseData. The ViewGraph therefore is renamed to PoseGraph to avoid concept conflicts with ViewGraphCalibration. 

Next step would be to add cam2_from_cam1 into DatabaseCache and move PoseDecomposition into DatabaseCache::Load. In this way, we will be able to finally rely on DatabaseCache rather than Database in global mapper. We will also be able to move pose_graph.h/cc into colmap/scene.

PR #3971, opened 2026-01-09
Revert valid field from Edge struct to PoseGraph-managed state.

B1ueber2y:
Reverting this small change from https://github.com/colmap/colmap/pull/3954 from the PR discussion there. 

Thinking more about it, managing valid field from PoseGraph has several benefits:
1) valid is a attribute specific to PoseGraph rather than the relative pose edge. One relative pose data can be valid in one graph but invalid in another, depending on the connectivity.
2) Managing valid in PoseGraph as private member prevents the call site to update the flag in the edge directly, and thus improves safety.

The overhead is negligible in our use case.

### DIFF
--- a/src/glomap/scene/pose_graph.h
+++ b/src/glomap/scene/pose_graph.h
@@ -28,9 +28,6 @@ class PoseGraph {
     // Inlier feature matches between the two images.
     colmap::FeatureMatches inlier_matches;
 
-    // Whether this edge is valid for reconstruction.
-    bool valid = true;
-
     // Invert the geometry to match swapped image order.
     void Invert() {
       cam2_from_cam1 = colmap::Inverse(cam2_from_cam1);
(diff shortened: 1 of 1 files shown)


ahojnnes:
I am not sure I understand the motivation but I am probably lacking context. In particular:

1. "One relative pose data can be valid in one graph but invalid in another, depending on the connectivity." -- There is no way to share edges in different graphs though and they would be duplicated.
2. "prevents the call site to update the flag in the edge directly, and thus improves safety" -- Which additional safety do we achieve with this? There is no associated logic to also update some other state when invalidating an edge? The only thing SetValid/SetInvalid does is change the boolean flag?

I would even go as far as saying that we can remove the `SetValid`, `SetInvalid`, `IsValid` methods, because they don't really do anything other than reading/writing the flag?

PR #3975, opened 2026-01-10
Store two view geometries in correspondence graph

ahojnnes:
* Stores two view geometries in correspondences graph.
* Removes "relative poses" from database cache.
* Various other naming consistency changes and smaller perf optimizations.
* The only breaking change is that "CorrespondenceGraph::AddMatches" is now "CorrespondenceGraph::AddTwoViewGeometry". The other changes are backwards compatible with deprecations in the pycolmap interface.

### DIFF
--- a/src/colmap/scene/correspondence_graph.h
+++ b/src/colmap/scene/correspondence_graph.h
@@ -81,11 +81,11 @@ class CorrespondenceGraph {
   // Get the number of correspondences per image.
   inline point2D_t NumCorrespondencesForImage(image_t image_id) const;
 
-  // Get the number of correspondences between a pair of images.
+  // Get the number of matches between a pair of images.
   inline point2D_t NumMatchesBetweenImages(image_t image_id1,
                                            image_t image_id2) const;
 
-  // Get the number of correspondences between all images.
+  // Get the number of matches between all images.
   std::unordered_map<image_pair_t, point2D_t> NumMatchesBetweenAllImages()
       const;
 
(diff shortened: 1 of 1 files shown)


src/colmap/scene/correspondence_graph.h
```diff
@@ -162,8 +170,10 @@ class CorrespondenceGraph {
   };
 
   struct ImagePair {
-    // The number of correspondences between pairs of images.
-    point2D_t num_correspondences = 0;
+    // The number of matches between pairs of images.
+    point2D_t num_matches = 0;
```
B1ueber2y, line 174: With this name change, does it make sense to improve the comment here to explicitly say that this is generally number of inlier matches, to avoid confusion?

src/pycolmap/scene/correspondence_graph.cc
```diff
@@ -118,4 +114,13 @@ void BindCorrespondenceGraph(py::module& m) {
              return CorrespondenceGraph(self);
            })
       .def("__repr__", &CreateRepresentation<CorrespondenceGraph>);
+  DefDeprecation(PyCorrespondenceGraph,
```
B1ueber2y, line 117: Do we want to mention these DefDeprecation in the PR description as well, in case it is forgotten at the release?

src/colmap/scene/correspondence_graph.h
ahojnnes: Done

src/pycolmap/scene/correspondence_graph.cc
ahojnnes: Done

PR #3981, opened 2026-01-10
Do not store inlier matches in pose graph

ahojnnes:


src/colmap/scene/correspondence_graph.h
```diff
@@ -89,7 +89,7 @@ class CorrespondenceGraph {
   std::unordered_map<image_pair_t, point2D_t> NumMatchesBetweenAllImages()
       const;
 
-  // Two-view geometry with inlier matches. Inverted if necessary.
+  // Two-view geometry without matches. Inverted if necessary.
```
B1ueber2y: I m slightly worried that this will cause confusion and would personally slightly prefer the previous logic to also extract matches, for better readability. I assume this step is not the bottleneck. Or, alternative, how about a load_inlier_matches option and default to false?

ahojnnes: Done.

src/colmap/scene/correspondence_graph.h
```diff
@@ -140,6 +136,10 @@ class CorrespondenceGraph {
                                    image_t image_id2,
                                    FeatureMatches& matches) const;
 
+  // Two-view geometry with optional inlier matches. Inverted if necessary.
+  struct TwoViewGeometry ExtractTwoViewGeometry(
+      image_t image_id1, image_t image_id2, bool extract_inlier_matches) const;
```
B1ueber2y, line 141: Does it make sense to have a default value in both C++ and pybind? Otherwise this would be considered a breaking change.

PR #3983, opened 2026-01-10
Setup priors when loading database, filter watermarks when creating database cache from another cache

ahojnnes:
* Simplifies the loading and setup of pose priors.
* Filters watermarks now that we have access to the two-view geometry in the cache.

### DIFF
--- a/src/colmap/controllers/incremental_pipeline.cc
+++ b/src/colmap/controllers/incremental_pipeline.cc
@@ -234,9 +234,7 @@ IncrementalPipeline::IncrementalPipeline(
       CreateDatabaseCacheOptions(*options_, *reconstruction_manager_));
   timer.PrintMinutes();
 
-  RegisterCallback(INITIAL_IMAGE_PAIR_REG_CALLBACK);
-  RegisterCallback(NEXT_IMAGE_REG_CALLBACK);
-  RegisterCallback(LAST_IMAGE_REG_CALLBACK);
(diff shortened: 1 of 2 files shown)


src/colmap/controllers/incremental_pipeline.cc
```diff
@@ -264,18 +256,14 @@ IncrementalPipeline::IncrementalPipeline(
     }
   }
 
-  DatabaseCache::Options cache_options;
-  cache_options.min_num_matches =
+  DatabaseCache::Options database_cache_options;
+  database_cache_options.min_num_matches =
```
B1ueber2y: Then we need to propagate ignore_watermarks here?

src/colmap/scene/database_cache.h
```diff
@@ -60,6 +60,9 @@ class DatabaseCache {
     // that if one image of a frame is included, all other images in the same
     // frame will also be included. All images are used if empty.
     std::unordered_set<std::string> image_names;
+
+    // Whether to convert pose priors to ENU coordinate system.
+    bool convert_pose_priors_to_enu = false;
```
B1ueber2y, line 65: Would need pybind for this.

src/colmap/controllers/incremental_pipeline.cc
ahojnnes: Good catch.

PR #3985, opened 2026-01-11
Misc view graph calibration and global mapper improvements

ahojnnes:


src/glomap/scene/pose_graph.h
```diff
@@ -46,7 +46,7 @@ class PoseGraph {
   inline void Clear();
 
   // Load edges from the database cache.
```
ahojnnes: Update.

src/colmap/scene/image.h
```diff
@@ -117,6 +117,7 @@ class Image {
   inline struct Point2D& Point2D(point2D_t point2D_idx);
   inline const std::vector<struct Point2D>& Points2D() const;
   inline std::vector<struct Point2D>& Points2D();
+  std::vector<Eigen::Vector2d> Points2DVector() const;
```
ahojnnes: Test

ahojnnes: Search for other instances where this can be used?

src/colmap/estimators/view_graph_calibration.cc
```diff
@@ -360,101 +357,108 @@ bool CalibrateViewGraph(const ViewGraphCalibrationOptions& options,
...
+  }
+
+  LOG(INFO) << "Building calibrated graph";
+
+  CorrespondenceGraph calibrated_graph;
+  for (const auto& [image_id, image] : reconstruction.Images()) {
+    calibrated_graph.AddImage(image_id, image.NumPoints2D());
+  }
+
+  // TODO(jsch): This will only add CALIBRATED/UNCALIBRATED pairs to the
```
ahojnnes: Fix

src/colmap/estimators/view_graph_calibration.h
```diff
@@ -81,6 +82,7 @@ struct ViewGraphCalibrationOptions {
   std::unique_ptr<ceres::LossFunction> CreateLossFunction() const;
 };
 
+// TODO(jsch): Update.
```
ahojnnes: Fix

src/colmap/estimators/view_graph_calibration.h
```diff
@@ -89,6 +91,8 @@ struct ViewGraphCalibrationOptions {
 // matrices computed and relative poses re-estimated, then are upgraded to
 // CALIBRATED. Pairs with high calibration error are tagged as DEGENERATE.
 bool CalibrateViewGraph(const ViewGraphCalibrationOptions& options,
```
B1ueber2y, line 91: To be honest, I think this change is going backwards. I dont think one should need a reconstruction and correspondence graph in this function. And from the top level, I dont think the view graph calibration belongs to the global pipeline. 

I understand that the challenge of using database cache for view graph calibration comes from the fact that we dont have all raw matches stored in the cache. But this is exactly something we should do outside the global pipeline, in the sense that we dont want the pipeline to rely on the raw matches.

So i think its better to keep the previous interface (only on database) and make it a stand-alone step in between feature_matching and global_mapper when the user calls substeps from the CLI. For automatic reconstructor we can just add it so it wont be an issue there.

src/colmap/estimators/view_graph_calibration.cc
B1ueber2y: This only becomes an issue in the new logic, where the view graph calibration does not operate on the database anymore.

src/glomap/scene/pose_graph.h
```diff
@@ -45,8 +45,8 @@ class PoseGraph {
   inline bool Empty() const;
   inline void Clear();
 
-  // Load edges from the database cache.
-  void Load(const colmap::DatabaseCache& cache);
+  // Load edges from the correspondence graph.
+  void Load(const colmap::CorrespondenceGraph& corr_graph);
```
B1ueber2y, line 49: Nice improvement!

src/glomap/sfm/global_mapper.h
```diff
@@ -79,16 +79,20 @@ class GlobalMapper {
              std::unordered_map<frame_t, int>& cluster_ids);
 
   // Run rotation averaging to estimate global rotations.
-  bool RotationAveraging(const RotationEstimatorOptions& options);
+  bool RotationAveraging(const RotationEstimatorOptions& options,
```
B1ueber2y: Should we mention in the comments here how would pose_graph change in this function? (filtered by relative rotations)

src/colmap/estimators/view_graph_calibration.h
B1ueber2y: Also, this would make the logic update of view graph calibrator CLI (currently broken in this branch: https://github.com/colmap/colmap/blob/5c6ccd71afd2e0a00b7e8a73a4745c86d215720f/src/colmap/exe/sfm.cc#L794) quite complicated?

src/colmap/estimators/view_graph_calibration.cc
ahojnnes: Relatively easy to fix.

src/colmap/estimators/view_graph_calibration.h
ahojnnes: I am neither super convinced with the existing implementation nor the new proposal here. Here are some thoughts that led to the new proposal:
1. As per original design principle, the database was meant to be "immutable" in the sense that processes should only add but not remove or alter the data. Reconstruction results should go into the reconstruction object. This has been true until a few days ago when we started applying view graph calibration directly on the database.
1. Running view graph calibration repeatedly on the same database is not an idempotent operation and the calibration parameters would slowly drift away.
1. View graph calibration is a global estimation/reconstruction problem, so conceptually it fits into the global pipeline (imo).
1. My understanding is that we want a separate view graph calibration tool for experimentation and evaluation of the estimated camera calibration quality. The idea would be to build this tool and instead of writing out all the two view geometries, we simply write the estimated cameras in the reconstruction, which can then be inspected for the calibration output?

B1ueber2y: Re 2. It is idempotent from my understanding, since no view graph calibration will be done if all cameras become calibrated, which should be set in the first pass of view graph calibration. So the second pass would be just a no-op.

Re 3. I dont agree with this. Since in my opinion view graph calibration should be considered a step after two view geometry estimation at feature matching. There is nothing preventing view graph calibration to be used ahead of incremental pipeline and hierarchical pipeline, so this is not something specific to global pipeline. 

With these being said, I think view graph calibration should operate on the database, similar to how feature_matching does.

B1ueber2y: I can see the point of 1. But since we are only using UpdateCamera and UpdateTwoViewGeometry which were already exposed in the database, I dont think this is a significant issue.

ahojnnes: > Re 2. It is idempotent from my understanding, since no view graph calibration will be done if all cameras become calibrated, which should be set in the first pass of view graph calibration. So the second pass would be just a no-op.

I think the issue is if not all cameras become calibrated, then a different subset of cameras is cross-validated to be calibrated and the E/F conversion etc will be different. I don't think it is really idempotent.

> Re 3. I dont agree with this. Since in my opinion view graph calibration should be considered a step after two view geometry estimation at feature matching. There is nothing preventing view graph calibration to be used ahead of incremental pipeline and hierarchical pipeline, so this is not something specific to global pipeline.

I am not arguing against running view graph calibration also for the hierarchical / incremental pipeline. I suggest also running it in-process there. Correct me if I am wrong but it is very fast and should have negligible runtime compared to all the other stages in any of the mapping pipelines?

> With these being said, I think view graph calibration should operate on the database, similar to how feature_matching does.

Why do we need to persist the outputs of view graph calibration in the database instead of computing it on-the-fly? If the only purpose is debugging and experimentation, then there are other means to achieve that, no?

B1ueber2y: > I am not arguing against running view graph calibration also for the hierarchical / incremental pipeline. I suggest also running it in-process there. Correct me if I am wrong but it is very fast and should have negligible runtime compared to all the other stages in any of the mapping pipelines?
> Why do we need to persist the outputs of view graph calibration in the database instead of computing it on-the-fly? If the only purpose is debugging and experimentation, then there are other means to achieve that, no?

The thing is that, there is a step to re-estimate two view geometry in view graph calibration, and to achieve this we need raw matches, which can only come from the database unless we add them to the database cache. This would make database dependent again for all pipelines if we want to put view graph calibration inside. Now that in this PR the global pipeline still has both database and database cache as private members, and we would need to introduce the database back for incremental and hierarchical pipeline if we want to add this step. 

By design, the final step of the feature matching is to run two-view geometry estimation for geometric verification. So i think this two view re-estimation naturally fits in there.

And yes, I can confirm that the view graph calibration is very fast and has negligible runtime.

B1ueber2y: > I think the issue is if not all cameras become calibrated, then a different subset of cameras is cross-validated to be calibrated and the E/F conversion etc will be different. I don't think it is really idempotent.

Yes probably, but can this really drift? My intuition is that it will most likely get better? My experience on running ETH3D after one round of view graph calibration all cameras became calibrated, probably because that all cameras are shared so there are a lot of redundant constraints.

ahojnnes: Yeah, running it as a final step in the matcher after verification (even though the name is not very appropriate then anymore) probably makes the most sense. 

I can undo the database change in view graph calibration in this PR then and we can call it from there?

B1ueber2y: Sounds good!

ahojnnes: Done

src/glomap/sfm/global_mapper.h
```diff
@@ -67,6 +67,7 @@ struct GlobalMapperOptions {
 class GlobalMapper {
  public:
   explicit GlobalMapper(
+      const GlobalMapperOptions& options,
```
B1ueber2y: This is inconsistent with the design of IncrementalMapper. 

Actually, not directly relevant to this PR, the IncrementalMapper does not have such interface Solve(). Instead, the core logic is put in the IncrementalPipelineController. Not sure if we want to do the same thing.

ahojnnes: I don't have a strong opinion. It just felt a bit unnecessary and redundant to pass around the options?

B1ueber2y: I also dont have a strong preference here. Just thinking that it would be nice to align the two interfaces (in either way).

ahojnnes: The incremental mapper was designed this as it allows a user to modify options via the GUI during the processing. This is not so relevant for the global mapper?

B1ueber2y: I would revert these changes. These options in submethods can be super useful for prototyping after we pybind each step?

ahojnnes: Not on my machine for the rest of the day. Feel free to push to the branch if you want to do anything that needs this in the meanwhile.

B1ueber2y: Done.

ahojnnes: Thank you!

Issue #3588, opened 2025-08-20
colmap pose_prior_mapper abnormal exit when using rig


tangxianbo:
**Describe the bug**

windows 11,  colmap 3.12.4
picture was taken with GoPro and contains GPS data
colmap pose_prior_mapper abnormal exit when using rig

<img width="2214" height="690" alt="Image" src="https://github.com/user-attachments/assets/4e5e62f7-8b3a-4de1-8107-74d3e194d2d2" />

**To Reproduce**

```
colmap feature_extractor --database_path ./database.db --image_path  ./images --ImageReader.single_camera_per_folder 1

colmap rig_configurator --database_path ./database.db --rig_config_path ./rig_config.json 

colmap exhaustive_matcher --database_path ./database.db 

mkdir sparse
colmap pose_prior_mapper --database_path ./database.db --image_path  ./images --output_path ./sparse --overwrite_priors_covariance 1
```

If rig_configurator is not called, pose_prior_mapper runs normally

pictures: https://drive.google.com/file/d/1DwSydYYZK9wYISH_27c7AW0dVH8F4xYw/view?usp=sharing

ipekoztas:
I have the same issue while running the same pipeline (but using mapper instead of pose_prior_mapper)

DuVogel87:
Any solution yet? @tangxianbo @ipekoztas

tangxianbo:
Unfortunately, no @DuVogel87

ahojnnes:
Thanks for sharing the data. I looked at it and this is expected -- even though we should improve the error message and not crash the process.

The problem is that colmap's incremental mapper does not support unknown rig extrinsics and it requires a two-step reconstruction process, see: https://colmap.github.io/rigs.html#unknown-rig-sensor-poses

Alternatively, you can be adventurous and try our new global mapper, which can directly estimate also the rig extrinsics. But it does not support pose priors yet.

ahojnnes:
See: https://github.com/colmap/colmap/pull/3988

PR #3988, opened 2026-01-11
Do not crash but return error on unknown sensor_from_rig poses in incremental mapper

ahojnnes:
If not all sensor_from_rig poses are known, the incremental mapper currently crashes without any error message.

Fixes: https://github.com/colmap/colmap/issues/3588


src/colmap/controllers/incremental_pipeline.h
```diff
@@ -228,7 +234,9 @@ class IncrementalPipeline : public BaseController {
     return database_cache_;
   }
 
-  void Reconstruct(IncrementalMapper& mapper,
+  // Returns true, if the main loop should call Reconstruct() again with relaxed
```
B1ueber2y: The boolean value return confused me a lot, since we would generally expect true to be "success"? How about returning Status instead?

ahojnnes: I created a new "STOP" and "CONTINUE" status.

Issue #3924, opened 2026-01-01
Rotation averaging terminates with exception


leonngl:
Hi,

thanks for this project, really cool work!

**Describe the bug**
After recent changes to the rotation averager, I am having troubles running it via the command line. It crashes with 
"terminate called after throwing an instance of 'std::domain_error'
  what():  Camera model does not exist"

**To Reproduce**
I have uploaded a small dataset to reproduce the error [here](https://github.com/leonngl/colmap/releases/tag/debug). It is a subset of a neat dataset from [here](https://projects.pascalsommer.ch/gravity-sfm/).

The error occurs when I invoke the rotation averager e.g. with
`colmap rotation_averager --relpose_path relposes.txt --output_path rotations_out.txt`

**Expected behavior**
I think the rotation averager should work fine for this dataset. Using the older commit 28ad83a2005877435cf6e56f74429744e3808501, it produces a seemingly valid output file. The output rotations when additionally passing a gravity-file are identical up to small numeric errors.

**Environment:**
 - OS: Ubuntu 24.04.3
 - COLMAP Version 3.14, commit 6aea89fb3629bee12df751f66ab1611212200c7c (but after brief testing, I think the issue is also in the newest commit)

**Additional Notes**
I think this issue occurs when building the reconstruction object and adding the cameras with invalid model-id in RunRotationAverager.
Apart from this, I think there is a slightly trickier bug in "SolveRotationAveraging" that occurs when using gravity-priors. I believe the issue is a removed call to "KeepLargestConnectedComponent" on the gravity_view_graph. In case it helps, I uploaded two small fixes to [this fork](https://github.com/leonngl/colmap/tree/rotation-averaging-debugging), that solved the issues for me. This is probably not the most efficient/elegant way to solve this, though.

Thanks in advance and happy new year!

B1ueber2y:
Thank you for reporting! This should be fixed in the linked PR. Can you double check if it works for all your local test cases? @leonngl

leonngl:
Cool, thanks a lot for looking into this so quickly!
In my tests the rotation_averager runs without issues now and produces the same output as in the older commit.

PR #3990, opened 2026-01-11
Use database in rotation averaging CLI. Add a python example for legacy format.

B1ueber2y:
The python example can be used to convert legacy txt format to a database. Tested through with the data from https://github.com/colmap/colmap/issues/3924.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -65,7 +65,7 @@ test-command = "pwsh -File {project}/python/ci/test-colmap-windows.ps1"
 # annotations that was fixed in version 2.3.0, but numpy stopped shipping
 # wheels for Python 3.10 in 2.3.0. This causes mypy to fail.
 [[tool.cibuildwheel.overrides]]
-select = "cp310-*"
+select = "cp310-{macosx,manylinux}*"
 test-command = """\
     python -c "import pycolmap; print(pycolmap.__version__)" && \
     cd {project} && \
--- a/src/pycolmap/scene/camera.cc
+++ b/src/pycolmap/scene/camera.cc
@@ -36,18 +36,28 @@ void BindCamera(py::module& m) {
 
   py::classh<Camera> PyCamera(m, "Camera");
   PyCamera.def(py::init<>())
-      .def_static("create",
+      .def_static("create_from_model_id",
                   &Camera::CreateFromModelId,
                   "camera_id"_a,
                   "model"_a,
                   "focal_length"_a,
                   "width"_a,
                   "height"_a)
+      .def_static("create_from_model_name",
+                  &Camera::CreateFromModelName,
+                  "camera_id"_a,
+                  "model_name"_a,
+                  "focal_length"_a,
+                  "width"_a,
+                  "height"_a,
+                  "Create camera from model name string.")
       .def_readwrite(
           "camera_id", &Camera::camera_id, "Unique identifier of the camera.")
       .def_property_readonly(
           "sensor_id", &Camera::SensorId, "Unique identifier of the sensor.")
       .def_readwrite("model", &Camera::model_id, "Camera model.")
+      .def_property_readonly(
+          "model_name", &Camera::ModelName, "Camera model name as string.")
       .def_readwrite("width", &Camera::width, "Width of camera sensor.")
       .def_readwrite("height", &Camera::height, "Height of camera sensor.")
       .def("mean_focal_length", &Camera::MeanFocalLength)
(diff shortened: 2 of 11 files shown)


src/colmap/exe/sfm.cc
```diff
@@ -638,114 +634,48 @@ void RunPointTriangulatorImpl(
...
 
-  if (!gravity_path.empty() && !ExistsFile(gravity_path)) {
-    LOG(ERROR) << "`gravity_path` is not a file";
+  if (!ExistsDir(output_path)) {
+    LOG(ERROR) << "`output_path` is not a directory";
     return EXIT_FAILURE;
   }
 
-  glomap::RotationEstimatorOptions rotation_averager_options;
-  rotation_averager_options.skip_initialization = true;
```
B1ueber2y, line 671: Previously, skip_initialization is always set to true here. This does not seem correct? Instead I changed to setting skip_initialization to true only when gravity priors are available.

src/colmap/exe/sfm.cc
```diff
@@ -638,114 +634,48 @@ void RunPointTriangulatorImpl(
...
-      }
-    }
-    pose_graph.InvalidatePairsOutsideActiveImageIds(active_image_ids);
+  auto database = Database::Open(*options.database_path);
+  auto reconstruction = std::make_shared<Reconstruction>();
 
-    glomap::RunGravityRefinement(
-        *options.gravity_refiner, pose_graph, reconstruction, pose_priors);
-  }
+  RotationAveragingController controller(
```
ahojnnes: nit: RotationAveragingPipeline?

python/examples/legacy_rotation_averaging_io.py
```diff
@@ -0,0 +1,293 @@
...
+    colmap rotation_averager --database_path database.db --output_path output/
+
+Usage:
+    python legacy_rotation_averaging_io.py \
+        --relpose_path relative_poses.txt \
+        --database_path database.db \
+        [--gravity_path gravity_priors.txt]
+"""
+
+from __future__ import annotations
```
ahojnnes: Given that we only support Python 3.10+, this shouldn't be necessary?

python/examples/legacy_rotation_averaging_io.py
```diff
@@ -0,0 +1,293 @@
...
+        for _, camera in reconstruction.cameras.items():
+            db.write_camera(camera, use_camera_id=True)
+
+        for _, rig in reconstruction.rigs.items():
+            db.write_rig(rig, use_rig_id=True)
+
+        for _, frame in reconstruction.frames.items():
+            db.write_frame(frame, use_frame_id=True)
+
+        for _, image in reconstruction.images.items():
```
ahojnnes: .values()

python/examples/legacy_rotation_averaging_io.py
```diff
@@ -0,0 +1,293 @@
...
+        for rel_pose in relative_poses:
+            id1 = image_name_to_id[rel_pose.image_name1]
+            id2 = image_name_to_id[rel_pose.image_name2]
+
+            two_view_geom = pycolmap.TwoViewGeometry()
+            two_view_geom.config = (
+                pycolmap.TwoViewGeometryConfiguration.CALIBRATED
+            )
+            two_view_geom.cam2_from_cam1 = rel_pose.cam2_from_cam1
+            two_view_geom.inlier_matches = np.array([[0, 0]], dtype=np.uint32)
```
ahojnnes: Why not empty?

B1ueber2y: Two view geometry with empty inliers are skipped while reading. Fixed with https://github.com/colmap/colmap/pull/3990/commits/0f933fef4f1bdb404de8a947de71755b9e1f2c57

B1ueber2y:
The new workflow was successfully tested with the data provided from https://github.com/colmap/colmap/issues/3924.

python/examples/legacy_rotation_averaging_io.py
ahojnnes: Another option could be to load based on config value? For example, we would still want to load the pair, if only one of E, F, H exist (e.g., when compute_relative_pose was not set during feature matching)?

B1ueber2y: If E, F, H exists there should be inlier matches already present there, which would get rows > 0 and therefore the pair gets loaded? If we want to also add a OR branch for config, should WATERMARK, DEGENERATE be loaded? Afraid that this would change some behaviors.

ahojnnes: Not necessarily, if somebody manually adds them to the database. This is exactly the issue you are trying to address in this PR? The safe thing is probably:
```
"SELECT * FROM two_view_geometries WHERE rows > 0 OR F IS NOT NULL OR E IS NOT NULL OR H IS NOT NULL OR qvec IS NOT NULL OR tvec IS NOT NULL;"
```

B1ueber2y: Sure. Updated.

ahojnnes:
Thank you, great work.

PR #3992, opened 2026-01-12
Reduce incremental mapper's minimum size for small models

ahojnnes:
I think this new hyper-parameter is a more sensible choice.

### DIFF
--- a/src/colmap/controllers/incremental_pipeline.cc
+++ b/src/colmap/controllers/incremental_pipeline.cc
@@ -540,7 +540,7 @@ IncrementalPipeline::Status IncrementalPipeline::ReconstructSubModel(
           const size_t kMinNumInitialRegTrials = 30;
           if (reg_trial >= kMinNumInitialRegTrials &&
               reconstruction->NumRegImages() <
-                  static_cast<int>(options_->min_model_size)) {
+                  static_cast<size_t>(options_->min_model_size)) {
             break;
           }
         }
@@ -686,7 +686,7 @@ IncrementalPipeline::Status IncrementalPipeline::Reconstruct(
 
         // Always keep the first reconstruction, independent of size.
         if ((options_->multiple_models && reconstruction_manager_->Size() > 1 &&
-             num_reg_images < static_cast<int>(options_->min_model_size)) ||
+             num_reg_images < static_cast<size_t>(options_->min_model_size)) ||
             num_reg_images == 0) {
           LOG(WARNING) << "Discarding reconstruction due to insufficient size";
           mapper.EndReconstruction(/*discard=*/true);


B1ueber2y:
This has a slight risk of affecting the benchmarking, when many small models are saved and we reach max_num_models more quickly?

On the other hand, min_model_size is calculated with respect to number of images, but tested against number of registered frames. Is this a bug?

ahojnnes:
> This has a slight risk of affecting the benchmarking, when many small models are saved and we reach max_num_models more quickly?

Not sure if this is an issue. There can only be 2 models with at least half the number of images. So, unless one configured max_num_models=1, this shouldn't cause an issue?

> On the other hand, min_model_size is calculated with respect to number of images, but tested against number of registered frames. Is this a bug?

This seems to be a bug. I fixed it.

B1ueber2y:
> Not sure if this is an issue. There can only be 2 models with at least half the number of images. So, unless one configured max_num_models=1, this shouldn't cause an issue?

From my understanding, in the mode with multiple models, the second model does not exclude the images from the first model. 

> This seems to be a bug. I fixed it.

This seems to be the case not only in Python but also in C++, and in two places.

ahojnnes:
> From my understanding, in the mode with multiple models, the second model does not exclude the images from the first model.

There can be overlap. I am still convinced that if we reconstruct more than ~50% of the images, then we should keep the model. There is very little chance that when starting from another image pair, we would reconstruct more.

> This seems to be the case not only in Python but also in C++, and in two places.

Yes, and it should be fixed in both.

B1ueber2y:
> Yes, and it should be fixed in both.

This seems to be missed: https://github.com/colmap/colmap/blob/390077fdf8e0e5299cfd1e1710532d6071622e10/src/colmap/controllers/incremental_pipeline.cc#L528

B1ueber2y (approved):

src/colmap/controllers/incremental_pipeline.cc
```diff
@@ -489,12 +489,13 @@ IncrementalPipeline::Status IncrementalPipeline::ReconstructSubModel(
...
-          // If initial pair fails to continue for some time,
+          // If initial model fails to continue for some time,
           // abort and try different initial pair.
           const size_t kMinNumInitialRegTrials = 30;
+          const size_t min_model_size = std::min<size_t>(
+              0.5 * database_cache_->NumImages(), options_->min_model_size);
           if (reg_trial >= kMinNumInitialRegTrials &&
-              reconstruction->NumRegFrames() <
-                  static_cast<size_t>(options_->min_model_size)) {
+              reconstruction->NumRegImages() < min_model_size) {
```
B1ueber2y: but then we also need to update the python pipeline as well.

ahojnnes:
OK, I went the extra mile to make sure there is no regression with this change. A: master, B: this branch
```
I20260112 14:02:13.463241 20457612 compare.py:main:58] Results A - B:

botanical_garden  -0.01  -0.00  -0.00  -0.00      0     0    0       0
boulders          -0.00  -0.00  -0.00  -0.00      0     0    0       0
bridge             0.00   0.00   0.00   0.00      0     0    0       0
courtyard         -0.02  -0.87  -0.45  -0.23      0     0    0       0
delivery_area     -0.00  -0.00  -0.00  -0.00      0     0    0       0
door               0.00   0.00   0.00  -0.00      0     0    0       0
electro           -0.00  -0.00  -0.00  -0.00      0     0    0       0
exhibition_hall   -0.20   0.08  -0.06   0.22      0     0    0       0
facade            -0.01  -0.01  -0.00  -0.00      0     0    0       0
kicker            -1.07  -0.49  -2.04  -3.77      0     0    0       0
lecture_room      -0.00  -0.00  -0.00  -0.00      0     0    0       0
living_room       -0.00  -0.00  -0.00  -0.00      0     0    0       0
lounge           -23.87 -28.62 -32.39 -32.86     -5     0   -1      -4
meadow             0.00   0.00   0.00   0.00      0     0    0       0
observatory       -1.42  -1.28  -0.35  -0.18      0     0    0       0
office             0.00   0.00   0.00   0.00      0     0    0       0
old_computer      -0.00  -0.00  -0.00  -0.00      0     0    0       0
pipes              4.23   3.16   0.68   0.34      0     0    0       0
playground        -0.00  -0.00  -0.00  -0.00      0     0    0       0
relief            -0.00  -0.00  -0.00  -0.00      0     0    0       0
relief_2          -0.00  -0.00  -0.00  -0.00      0     0    0       0
statue            -0.00  -0.00  -0.00  -0.00      0     0    0       0
terrace           -0.00  -0.00  -0.00  -0.00      0     0    0       0
terrace_2          0.00   0.00   0.00   0.00      0     0    0       0
terrains           1.11   1.15   0.24   0.12      0     0    0       0
overall           -0.05  -0.05  -0.12  -0.13     -5     0   -1      -4
average           -0.85  -1.08  -1.38  -1.45      0     0    0       0
```

src/colmap/controllers/incremental_pipeline.cc
ahojnnes: Done

PR #3997, opened 2026-01-12
Add missing Python bindings

ahojnnes:
Asked Claude to add any missing method/fields in existing class/struct bindings. Needed to manually fix a few things but otherwise decent job.

### DIFF
--- a/src/pycolmap/scene/reconstruction.cc
+++ b/src/pycolmap/scene/reconstruction.cc
@@ -258,9 +258,6 @@ void BindReconstruction(py::module& m) {
            "Import from PLY format. Note that these import functions are"
            "only intended for visualization of data and usable for "
            "reconstruction.")
-      .def("convert_to_PLY",
-           &Reconstruction::ConvertToPLY,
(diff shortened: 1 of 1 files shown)


src/pycolmap/sensor/bitmap.cc
```diff
@@ -137,5 +137,68 @@ void BindBitmap(pybind11::module& m) {
...
+      .def(
+          "exif_altitude",
+          [](const Bitmap& self) -> py::typing::Optional<double> {
+            double altitude;
+            if (self.ExifAltitude(&altitude)) {
+              return py::cast(altitude);
+            }
+            return py::none();
+          },
+          "Extract EXIF altitude. Returns None if not available.")
```
sarlinpe, line 202: It would be nice to instead change the interface to return `optional<float>`, this would simplify the bindings.

sarlinpe (approved):
Nice!

src/pycolmap/sensor/bitmap.cc
ahojnnes: Good idea. Will follow up to this PR.

PR #3998, opened 2026-01-12
Switch to Python 3.11 in the PR build. Disable mypy for Python 3.10. 

B1ueber2y:
To avoid type annotation bug in numpy < 2.3.0.

### DIFF
--- a/python/examples/custom_incremental_pipeline.py
+++ b/python/examples/custom_incremental_pipeline.py
@@ -225,7 +225,7 @@ def reconstruct_sub_model(
                 kMinNumInitialRegTrials = 30
                 if (
                     reg_trial >= kMinNumInitialRegTrials
-                    and reconstruction.num_reg_frames() < options.min_model_size
+                    and reconstruction.num_reg_images() < options.min_model_size
                 ):
                     break
             if reg_next_success:
@@ -344,15 +344,13 @@ def reconstruct(
             reconstruction_manager.delete(reconstruction_idx)
             return IncrementalPipelineStatus.CONTINUE
         elif status == IncrementalPipelineStatus.SUCCESS:
+            num_reg_images = reconstruction.num_reg_images()
             total_num_reg_images = mapper.num_total_reg_images()
-            min_model_size = min(
-                0.8 * database_cache.num_images(), options.min_model_size
-            )
             if (
                 options.multiple_models
                 and reconstruction_manager.size() > 1
-                and reconstruction.num_reg_frames() < min_model_size
-            ) or reconstruction.num_reg_frames() == 0:
+                and num_reg_images < options.min_model_size
+            ) or num_reg_images == 0:
                 logging.info(
                     "Discarding reconstruction due to insufficient size"
                 )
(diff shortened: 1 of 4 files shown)


python/generate_stubs_test.py
```diff
@@ -0,0 +1,61 @@
...
+
+def test_two_view_geometry_inlier_matches() -> None:
+    """TwoViewGeometry.inlier_matches should accept uint32 array with 2 columns."""
+    matches: NDArray[np.uint32] = np.array([[0, 1], [2, 3], [4, 5]], dtype=np.uint32)
+    two_view_geom = pycolmap.TwoViewGeometry()
+    two_view_geom.inlier_matches = matches
+    assert two_view_geom.inlier_matches is not None
+
+
+if __name__ == "__main__":
```
ahojnnes: This should not be needed if you invoke with pytest.

B1ueber2y: Thanks!

B1ueber2y:
Managed to reproduce the errors with the tests from CI (seems to fail in both ubuntu 24.04 and macos-15 arm64): https://github.com/colmap/colmap/actions/runs/20923717516/job/60115792676. 

Strange enough, I cannot reproduce this issue locally.

python/generate_stubs.sh
```diff
@@ -26,5 +26,11 @@ perl -i -pe's/(__hash__:? .*= None)$/\1  # type: ignore/g' $FILES
 # pybind issue: dictionary keys should not be cast to the more generic types.
 perl -i -pe's/Mapping\[typing.Supports(Int|Float)/Mapping\[\L\1/g' $FILES
 
+# Eigen column vectors (Nx1) should also accept 1D arrays (N,).
+perl -i -pe's/tuple\[Literal\[(\d+)\], Literal\[1\]\]/tuple[Literal[\1]] | tuple[Literal[\1], Literal[1]]/g' $FILES
+
+# Eigen dynamic dimensions use Never but should be int.
+perl -i -pe's/tuple\[Never,/tuple[int,/g' $FILES
+
```
sarlinpe: I think that we don't need any of this.

python/generate_stubs_test.py
```diff
@@ -0,0 +1,41 @@
+"""Tests for generated pycolmap stubs.
```
sarlinpe: Can we avoid committing this file?

B1ueber2y: Done.

ahojnnes:
I am confused. So why did this not trigger before as we have 3.10 running in CI? What's the concrete issue? If we don't test 3.10 in CI anymore, I expect we'll frequently break master because changes slip through that will be incompatible with 3.10?

B1ueber2y:
> I am confused. So why did this not trigger before as we have 3.10 running in CI? What's the concrete issue? If we don't test 3.10 in CI anymore, I expect we'll frequently break master because changes slip through that will be incompatible with 3.10?

We did not hit the type annotation bug before. Alternatively, what about enabling both 3.10 and 3.11 in the PR, but skip mypy for 3.10? This will make the CI run a bit longer.

ahojnnes:
Let's see how it goes. In the first breakage in master, let's reenable 3.10 in PR checks.

PR #4000, opened 2026-01-12
Do not override test commands for Python 3.10 on windows.

B1ueber2y:
Bugfix for https://github.com/colmap/colmap/pull/3998.


B1ueber2y:
We may decide whether to include py310 in PRs in the future. I m fine with either. Worst case we can come back to fix the issue once the py310 throws from the main branch CI.

ahojnnes:
Yeah, let's remove 3.10 and see when it breaks. YOLO.

PR #4003, opened 2026-01-13
Add test for reconstruction_clustering and sort cluster id by cluster size

lpanaf:
- Add an additional sorting step to the end of reconstruciton_clustering so that clusters are sorted by their sizes.
- Cluster with too few frames are marked with id -1
- Add tests for reconstruction_clustering.

### DIFF
--- a/src/colmap/scene/reconstruction_clustering.cc
+++ b/src/colmap/scene/reconstruction_clustering.cc
@@ -116,8 +116,7 @@ std::unordered_map<frame_t, int> EstablishStrongClusters(
 std::unordered_map<frame_t, int> ClusterReconstructionFrames(
     const ReconstructionClusteringOptions& options,
     Reconstruction& reconstruction) {
-  // Cluster with fewer than 2 frames is not meaningful.
-  THROW_CHECK_GE(options.min_num_reg_frames, 2);
+  options.Check();
 
   // Step 1: Compute covisibility counts between all frame pairs.
   // For each 3D point, increment the count for every pair of frames that sees
--- a/src/colmap/scene/reconstruction_clustering.h
+++ b/src/colmap/scene/reconstruction_clustering.h
@@ -30,6 +30,7 @@
 #pragma once
 
 #include "colmap/scene/reconstruction.h"
+#include "colmap/util/logging.h"
 #include "colmap/util/types.h"
 
 #include <unordered_map>
@@ -49,6 +50,12 @@ struct ReconstructionClusteringOptions {
   // Minimum number of registered frames required for a cluster to be kept.
   // Clusters with fewer frames will be discarded.
   int min_num_reg_frames = 3;
+
+  void Check() const {
+    THROW_CHECK_GE(min_covisibility_count, 1);
+    THROW_CHECK_GT(min_edge_weight_threshold, 0.0);
+    THROW_CHECK_GE(min_num_reg_frames, 2);
+  }
 };
 
 // Clusters frames based on 3D point covisibility and removes weakly connected


src/glomap/processors/reconstruction_pruning_test.cc
```diff
@@ -0,0 +1,355 @@
...
+
+      // Mark observations from other clusters for deletion, but keep some
+      // based on keep_ratio.
+      for (const auto& [cluster_id, observations] : cluster_observations) {
+        if (cluster_id != chosen_cluster) {
+          for (size_t i = 0; i < observations.size(); ++i) {
+            // Use a deterministic pseudo-random decision based on point3D_id
+            // and observation index to decide whether to keep this observation.
+            const double hash_value =
+                static_cast<double>((point3D_id * 31 + i * 17) % 1000) / 1000.0;
```
ahojnnes: Did an LLM come up with this? Does it make sense to use RandomUniformReal here instead?

src/glomap/processors/reconstruction_pruning_test.cc
```diff
@@ -0,0 +1,355 @@
...
+  // All frames should still be registered before pruning.
+  EXPECT_EQ(reconstruction.NumRegFrames(), initial_num_reg_frames);
+
+  const auto cluster_ids = PruneWeaklyConnectedFrames(reconstruction);
+
+  // After pruning, only the well-connected frames should remain (5 frames).
+  // The disconnected frames should be de-registered.
+  const size_t expected_remaining = initial_num_reg_frames - num_to_disconnect;
+  EXPECT_EQ(reconstruction.NumRegFrames(), expected_remaining);
+  EXPECT_EQ(cluster_ids.size(), expected_remaining);
```
ahojnnes: Ideally we also check that these are all in the same cluster?

src/glomap/processors/reconstruction_pruning_test.cc
```diff
@@ -0,0 +1,355 @@
...
+
+  // Verify frames from the largest cluster are assigned cluster ID 0.
+  for (size_t i = 0; i < frame_to_keep; ++i) {
+    EXPECT_EQ(cluster_ids.at(all_frame_ids[i]), 0);
+  }
+}
+
+TEST(PruneWeaklyConnectedFrames, MultipleDisconnectedClusters) {
+  // Create a reconstruction with 12 frames, all initially well-connected.
+  int max_cluster_size = 25;
```
ahojnnes: nit: const int kMaxClusterSize, etc.

src/glomap/processors/reconstruction_pruning_test.cc
```diff
@@ -0,0 +1,355 @@
...
+  EXPECT_EQ(initial_num_reg_frames, max_cluster_size + other_cluster_size * 2);
+
+  // Get all frame IDs and sort them.
+  std::vector<colmap::frame_t> all_frame_ids;
+  for (const auto& [frame_id, frame] : reconstruction.Frames()) {
+    if (frame.HasPose()) {
+      all_frame_ids.push_back(frame_id);
+    }
+  }
+  std::sort(all_frame_ids.begin(), all_frame_ids.end());
```
ahojnnes: Extract as ExtractSortedFrameIds helper and reuse?

src/glomap/processors/reconstruction_pruning_test.cc
lpanaf: Thanks! I thought this suggestion was some common practice that I am not aware of :)

lpanaf:
> Thanks, great to have tests for this.

Thanks! I incoporated the comments and changed the test cases a bit.

src/colmap/scene/reconstruction_clustering.cc
```diff
@@ -120,12 +122,36 @@ std::unordered_map<frame_t, int> EstablishStrongClusters(
...
 
-  // Count clusters with at least kMinClusterSize frames.
-  constexpr int kMinClusterSize = 2;
+  // Count cluster sizes.
   std::unordered_map<int, int> cluster_sizes;
   for (const auto& [node, cluster_id] : cluster_ids) {
     cluster_sizes[cluster_id]++;
   }
+
+  // Phase 4: Reassign cluster IDs so they are sorted by size (largest first).
```
B1ueber2y: The logic here can be largely simplified by combining the sorting logic into Phase 3, such that there is no Phase 4 needed. For example, we can store a ``std::unordered_map<int, std::vector<frame_t>> root_to_nodes`` instead in the iteration of Phase 3?

src/colmap/scene/reconstruction_clustering.cc
```diff
@@ -120,12 +122,36 @@ std::unordered_map<frame_t, int> EstablishStrongClusters(
...
+    old_to_new_cluster_id[cluster_size_vec[i].first] = static_cast<int>(i);
+  }
+
+  // Apply the new cluster IDs.
+  for (auto& [node, cluster_id] : cluster_ids) {
+    cluster_id = old_to_new_cluster_id[cluster_id];
+  }
+
+  // Count clusters with at least kMinClusterSize frames.
+  constexpr int kMinClusterSize = 2;
```
B1ueber2y: Is this a bug from resolving merging conflicts? I thought that this parameter has been moved to the option class in the other PR.

src/colmap/scene/reconstruction_clustering.h
```diff
@@ -75,13 +75,16 @@ struct ReconstructionClusteringOptions {
 //      absolute deviation (MAD).
 //   3. Cluster frames using union-find: first merge strongly connected frames,
 //      then iteratively merge clusters connected by multiple weaker edges.
+//   4. Assign cluster IDs sorted by cluster size in descending order (i.e.,
```
B1ueber2y: Lets be more specific on "cluster size", to say "number of frames"?

src/colmap/scene/reconstruction_clustering.h
```diff
@@ -75,13 +75,16 @@ struct ReconstructionClusteringOptions {
...
+//      cluster ID 0 is the largest cluster).
 //
 // Args:
 //   options: Configuration options for clustering.
 //   reconstruction: The reconstruction containing frames and 3D points.
 //
 // Returns:
-//   Map from frame_id to cluster_id for all registered frames.
+//   Map from frame_id to cluster_id for all registered frames. Cluster IDs are
+//   sorted by cluster size (largest cluster has ID 0).
```
B1ueber2y: same here. Be more specific on cluster size?

src/colmap/scene/reconstruction_clustering.cc
B1ueber2y: We can potentially use the new Compress() and Parents() there to further improve the readability.

src/colmap/scene/reconstruction_clustering.cc
B1ueber2y: Looking at the main branch, seems that the new option min_num_reg_frames was not respected there, either.

lpanaf: > Is this a bug from resolving merging conflicts? I thought that this parameter has been moved to the option class in the other PR.

This one was not a bug, but was for logging only. However, for the min_num_reg_frames, I am not sure how exactly it should be used. Currently, it is only used in the controller, and for this function, every images would be assigned a cluster. Potentially, it can be changed that only large enough clusters are kept.

B1ueber2y: Maybe i m missing something, but why not just use min_num_reg_frames instead of this constexpr in this function?

lpanaf: Here, it just showing number of clusters which has at least 2 frames. I can change it that it shows the specified number of frames.

src/colmap/scene/reconstruction_clustering_test.cc
```diff
@@ -0,0 +1,389 @@
...
+#include "colmap/scene/reconstruction_clustering.h"
+
+#include "colmap/math/random.h"
+#include "colmap/scene/reconstruction.h"
+#include "colmap/scene/synthetic.h"
+
+#include <gmock/gmock.h>
+#include <gtest/gtest.h>
+
+namespace colmap {
```
B1ueber2y, line 39: Can we cover a rig case in the test?

src/colmap/scene/reconstruction_clustering.cc
B1ueber2y: Looking at it again, how about we just filtering by min_num_reg_frames in this EstablishStrongClusters function. This can even be combined into phase 3 after sorting.

src/colmap/scene/reconstruction_clustering.cc
```diff
@@ -109,59 +107,46 @@ std::unordered_map<frame_t, int> EstablishStrongClusters(
...
+  // First ensure all nodes are in the union-find structure (isolated nodes
+  // may not have been added during Phase 1/2 if they had no edges).
   for (const frame_t node : nodes) {
-    frame_t root = uf.Find(node);
-    auto it = root_to_cluster.find(root);
-    if (it == root_to_cluster.end()) {
-      root_to_cluster[root] = next_cluster_id++;
-    }
-    cluster_ids[node] = root_to_cluster[root];
+    uf.Find(node);
```
B1ueber2y: Unnecessary legacy loop?

src/colmap/scene/reconstruction_clustering_test.cc
```diff
@@ -385,5 +416,216 @@ TEST(ClusterReconstructionFrames, MultipleDisjointClusters) {
   EXPECT_EQ(result_clusters[2], expected_clusters[2]);
 }
 
+// Tests with non-trivial rigs (multiple cameras per frame).
+
+TEST(ClusterReconstructionFrames, RigOneMajorConnectedComponent) {
+  // Create a reconstruction with 10 frames from a rig with 3 cameras each.
+  // Test that clustering correctly handles multi-camera rigs.
+  // Note: With multi-camera rigs, covisibility between frames is stronger
```
B1ueber2y, line 427: Not sure if we want to fix this. We left a TODO in the logic for overcounting: https://github.com/colmap/colmap/pull/3886#discussion_r2650182612. Should we just avoid overcounting? Or was this intended?

lpanaf: I never test the behavior of it on real data with rigs, so I cannot tell which way is better. I think we can keep it as it is now, and address that todo later when improving the module itself.

src/colmap/scene/reconstruction_clustering.cc
lpanaf: actually no, if some of the images have no connected edges, they are never added to the structure

B1ueber2y: Yes. But these islands are not relevant anymore, since we have min_num_reg_frames checking? I assume it should be >=2?

lpanaf: In this function, it still requires each image to be assigned one cluster id.

B1ueber2y: I think we should set the cluster of those filtered out images to -1 (or some other constant representing invalid clusters), since min_num_reg_images is part of the options that are being passed into this module.

lpanaf: Okay, I like the idea of -1. Done.

B1ueber2y: Realizing that, probably still need to do this loop if min_num_reg_images <= 1?

src/colmap/scene/reconstruction_clustering.cc
```diff
@@ -107,35 +110,51 @@ std::unordered_map<frame_t, int> EstablishStrongClusters(
...
```
B1ueber2y: This Find() is only needed when min_num_reg_frames <= 1?

lpanaf: I think it is better for every registered frame to get assigned one cluster id. In this case, it still needs to touch every node at least once.

B1ueber2y: Sorry maybe i m missing something. What s the benefit?

lpanaf: just for the sake of consistency, no other reason.

B1ueber2y: I dont understand what consistency it benefits, as these island nodes are in the end assigned to -1. From my understanding this would be purely wasting memory in the UnionFind structs.

lpanaf: This assignment can be use for determining which cluster a camera is assigned. If this is considered as a waste of memory, then I would remove the return for other -1 clusters. Otherwise it is strange that some disconnected points are assigned -1 as cluster while others do not.

B1ueber2y: > This assignment can be use for determining which cluster a camera is assigned.

I am really lost here. This is an inlined function and there is no sense of camera assignment in this module?

> Otherwise it is strange that some disconnected points are assigned -1 as cluster while others do not.

You could check with FindIfExists and assign the missing nodes as -1?

lpanaf: Sorry, I mean a frame. I can do either way: I can assign the missing nodes as -1 in this function, or check whether a frame is assigned any number at all. But I think it would be better if it is consistent

src/colmap/scene/reconstruction_clustering.cc
```diff
@@ -61,34 +63,50 @@ std::unordered_map<frame_t, int> EstablishStrongClusters(
     }
   }
 
-  // Assign sequential cluster IDs based on union-find roots.
-  std::unordered_map<frame_t, int> root_to_cluster;
-  int next_cluster_id = 0;
-
-  for (const frame_t node : nodes) {
-    frame_t root = uf.Find(node);
```
B1ueber2y, line 69: Again, you would need to call this when min_num_reg_frames <= 1.

src/colmap/scene/reconstruction_clustering.cc
```diff
@@ -61,34 +63,50 @@ std::unordered_map<frame_t, int> EstablishStrongClusters(
...
+      // Clusters smaller than min_num_reg_frames are discarded.
+      for (const frame_t node : cluster_nodes) {
+        cluster_ids[node] = -1;
+      }
     }
   }
 
+  // Ensure all nodes are assigned a cluster ID.
+  for (const auto node : nodes) {
+    if (cluster_ids.find(node) == cluster_ids.end()) cluster_ids[node] = -1;
```
B1ueber2y: nit: here the cluster_ids grow in each iteration of the loop. uf.FindIfExists(node).has_value() is much better.

src/colmap/scene/reconstruction_clustering.cc
lpanaf: added to the end

ahojnnes:
@lpanaf Should we merge this or what's the status here?

lpanaf:
I think it is ready to merge :)

src/colmap/controllers/reconstruction_clustering.cc
```diff
@@ -122,19 +124,20 @@ void ReconstructionClustererController::Run() {
...
-        LOG(INFO) << "Skipping cluster " << comp << " with only "
-                  << cluster_reconstruction.NumRegFrames()
-                  << " registered frames";
-      }
+      THROW_CHECK_GE(
+          cluster_reconstruction.NumRegFrames(),
+          static_cast<size_t>(
+              options_.min_num_reg_frames));  // Should always be true
+      size_t idx = reconstruction_manager_->Add();
+      *reconstruction_manager_->Get(idx) = cluster_reconstruction;
```
ahojnnes:
```suggestion
      *reconstruction_manager_->Get(idx) = std::move(cluster_reconstruction);
```

lpanaf: Thanks for the review!

src/colmap/scene/reconstruction_clustering.cc
lpanaf: I added an assert to not allow this case :)

src/colmap/scene/reconstruction_clustering.cc
```diff
@@ -98,11 +116,19 @@ std::unordered_map<frame_t, int> EstablishStrongClusters(
 std::unordered_map<frame_t, int> ClusterReconstructionFrames(
     const ReconstructionClusteringOptions& options,
     Reconstruction& reconstruction) {
+  // Cluster with fewer than 2 frames is not meaningful.
+  THROW_CHECK_GE(options.min_num_reg_frames, 2);
```
B1ueber2y: If we want to be consistent with what is done in other places of colmap, we can add a Check() in the option class.

lpanaf: Thanks! Done.

PR #4004, opened 2026-01-14
Switch paths from std::string to std::filesystem::path (continued)

sarlinpe:


src/colmap/controllers/image_reader.cc
```diff
@@ -52,16 +52,18 @@ ImageReader::ImageReader(const ImageReaderOptions& options, Database* database)
 
   // Get a list of all files in the image path, sorted by image name.
   if (options_.image_names.empty()) {
-    options_.image_names = GetRecursiveFileList(options_.image_path);
```
sarlinpe, line 55: `image_names` is supposed to have relative normalized paths, the previous usage was incorrect.

src/colmap/util/file.h
```diff
@@ -183,16 +178,4 @@ void OverwriteDownloadCacheDir(std::filesystem::path path);
 // Throws runtime exception if download is not supported.
 std::filesystem::path MaybeDownloadAndCacheFile(const std::string& uri);
 
-////////////////////////////////////////////////////////////////////////////////
-// Implementation
-////////////////////////////////////////////////////////////////////////////////
-
-template <typename... T>
-std::string JoinPaths(T const&... paths) {
```
B1ueber2y, line 191: Need to remove the definition as well?

ahojnnes (approved):

PR #4007, opened 2026-01-14
Switch to perform clustering on all registered frames

lpanaf:
Instead of clustering only the largest connected component, performs it over all frames. This would be helpful if the generated reconstruction contains several disconnected components.

### DIFF
--- a/src/colmap/scene/reconstruction_clustering.cc
+++ b/src/colmap/scene/reconstruction_clustering.cc
@@ -172,7 +172,7 @@ std::unordered_map<frame_t, int> ClusterReconstructionFrames(
   }
 
   // Filter edges to keep only reliable connections.
-  std::unordered_map<frame_pair_t, int> edge_weights;
+  std::unordered_map<image_pair_t, int> edge_weights;
   for (const auto& [pair_id, count] : frame_covisibility_count) {
     if (count < options.min_covisibility_count) continue;
     edge_weights[pair_id] = count;


src/colmap/scene/reconstruction_clustering.cc
```diff
@@ -205,8 +177,8 @@ std::unordered_map<frame_t, int> PruneWeaklyConnectedFrames(
   const double threshold = std::max(median - mad, kMinEdgeWeightThreshold);
   LOG(INFO) << "Threshold for Strong Clustering: " << threshold;
 
-  // Step 5: Cluster frames based on covisibility weights.
```
B1ueber2y, line 228: nit: lets drop the number count (1,2,3,4) in the steps? I realized that this is inconsistent with the comments in the header.

src/glomap/processors/reconstruction_pruning.h
```diff
@@ -16,21 +16,17 @@ namespace glomap {
...
+//   3. Cluster frames using union-find: first merge strongly connected frames,
 //      then iteratively merge clusters connected by two weaker (0.75 *
 //      threshold) edges.
 //
 // Args:
 //   reconstruction: The reconstruction containing frames and 3D points.
-//      Frames outside the largest connected component will be de-registered.
 //
 // Returns:
-//   Map from frame_id to cluster_id for frames in the largest component.
```
B1ueber2y: As discussed earlier this week, can we make an option class for this function?

B1ueber2y (commented):
Thanks. Does it make sense to add a test with a multi-CC input?

src/glomap/processors/reconstruction_pruning.h
lpanaf: I will add it in a seperate PR. I will move the functions to be a seperate module as well.

lpanaf:
> Thanks. Does it make sense to add a test with a multi-CC input?

I will add it to the PR with the test functions.

ahojnnes (approved):
Thanks, LGTM.

PR #4008, opened 2026-01-14
Make reconstruction clustering logic a stand-alone module

lpanaf:
Remove the reconstruction from the default global pipeline and move it to a standalone module

### DIFF
--- a/src/colmap/controllers/reconstruction_clustering.cc
+++ b/src/colmap/controllers/reconstruction_clustering.cc
@@ -129,7 +129,7 @@ void ReconstructionClustererController::Run() {
       if (cluster_reconstruction.NumRegFrames() >=
           static_cast<size_t>(options_.min_num_reg_frames)) {
         size_t idx = reconstruction_manager_->Add();
-        *reconstruction_manager_->Get(idx) = std::move(cluster_reconstruction);
+        *reconstruction_manager_->Get(idx) = cluster_reconstruction;
       } else {
         LOG(INFO) << "Skipping cluster " << comp << " with only "
                   << cluster_reconstruction.NumRegFrames()


src/colmap/controllers/reconstruction_pruning.h
```diff
@@ -0,0 +1,63 @@
...
+
+#include "colmap/scene/reconstruction.h"
+#include "colmap/util/base_controller.h"
+
+#include <filesystem>
+#include <string>
+
+namespace colmap {
+
+struct ReconstructionPruningOptions {
```
B1ueber2y: Can we make these parameters into options?

lpanaf: Oh yes. Done :)

src/colmap/controllers/global_pipeline.cc
```diff
@@ -95,33 +94,13 @@ void GlobalPipeline::Run() {
...
-      if (!options_.image_path.empty()) {
-        output_reconstruction.ExtractColorsForAllImages(options_.image_path);
-      }
-    }
-    LOG(INFO) << "Exported " << max_cluster_id + 1 << " reconstructions";
+  // Output the reconstruction.
+  Reconstruction& output_reconstruction =
+      *reconstruction_manager_->Get(reconstruction_manager_->Add());
+  output_reconstruction = *reconstruction;
+  if (!options_.image_path.empty()) {
```
B1ueber2y, line 101: I dont think the color extraction belongs to this module. Can we drop it?

src/colmap/controllers/reconstruction_pruning.h
```diff
@@ -0,0 +1,68 @@
...
+  // Path to images for extracting colors (optional).
+  std::string image_path;
+
+  // Output path for pruned reconstruction(s).
+  std::filesystem::path output_path;
+
+  // Pruning algorithm options.
+  glomap::ReconstructionPruningOptions pruning;
```
B1ueber2y: Now that the whole module is already moved out of GlobalMapper, I think we can just inline the logic into a function in the empty namespace here. This already becomes a separate CLI step that is irrelevant to global_mapper, which simplifies a lot.

src/colmap/exe/colmap.cc
```diff
@@ -127,6 +127,8 @@ int main(int argc, char** argv) {
   commands.emplace_back("pose_prior_mapper", &colmap::RunPosePriorMapper);
   commands.emplace_back("poisson_mesher", &colmap::RunPoissonMesher);
   commands.emplace_back("project_generator", &colmap::RunProjectGenerator);
+  commands.emplace_back("reconstruction_pruning",
```
B1ueber2y: Maybe ``reconstruction_pruner``? To align the naming with other modules. Same for the name of the controller class.

src/colmap/controllers/reconstruction_pruning.cc
```diff
@@ -0,0 +1,95 @@
...
+    if (!options_.image_path.empty()) {
+      LOG(INFO) << "Extracting colors ...";
+      reconstruction_->ExtractColorsForAllImages(options_.image_path);
+    }
+    reconstruction_->Write(options_.output_path);
+  } else {
+    // Split by cluster and output multiple reconstructions
+    for (int comp = 0; comp <= max_cluster_id; comp++) {
+      Reconstruction cluster_reconstruction =
+          glomap::SubReconstructionByClusterId(
```
B1ueber2y: We can also inline this function to this file as well. And then the colmap_io.h/cc can finally be removed : )

src/colmap/controllers/global_pipeline.cc
lpanaf: I assume you mean the color extraction from the pruning module. Changed there.

src/colmap/exe/sfm.cc
```diff
@@ -739,4 +742,38 @@ int RunViewGraphCalibrator(int argc, char** argv) {
   return EXIT_SUCCESS;
 }
 
+int RunReconstructionPruning(int argc, char** argv) {
+  std::filesystem::path input_path;
+  std::filesystem::path output_path;
+
+  OptionManager options;
+  options.AddRequiredOption("input_path", &input_path);
+  options.AddRequiredOption("output_path", &output_path);
```
B1ueber2y, line 749: Add to option manager and thereby expose these parameters to CLI?

lpanaf: Sounds good. Done :)

lpanaf:
@B1ueber2y and I had a dicsussion on change the module from `reconstruction_pruner` to `reconstruction_clustering", and agree to keep the current `reconstruction_pruner`. Do you have any other suggestions on this, @ahojnnes ?

B1ueber2y:
> @B1ueber2y and I had a dicsussion on change the module from `reconstruction_pruner` to `reconstruction_clustering", and agree to keep the current `reconstruction_pruner`. Do you have any other suggestions on this, @ahojnnes ?

Actually, thinking again about it, we should probably update the name to reconstruction_clusterer and also rename the controller file, since there is a reconstruction_pruning module in colmap/scene serving for a different purpose? Also, if we already rename the option class we should probably align the CLI.

lpanaf:
> > @B1ueber2y and I had a dicsussion on change the module from `reconstruction_pruner` to `reconstruction_clustering", and agree to keep the current `reconstruction_pruner`. Do you have any other suggestions on this, @ahojnnes ?
> 
> Actually, thinking again about it, we should probably update the name to reconstruction_clusterer and also rename the controller file, since there is a reconstruction_pruning module in colmap/scene serving for a different purpose? Also, if we already rename the option class we should probably align the CLI.

sure, changed.

src/colmap/controllers/CMakeLists.txt
```diff
@@ -44,6 +44,7 @@ COLMAP_ADD_LIBRARY(
         image_reader.h image_reader.cc
         incremental_pipeline.h incremental_pipeline.cc
         option_manager.h option_manager.cc
+        reconstruction_clusterer.h reconstruction_clusterer.cc
```
B1ueber2y: nit: maybe reconstruction_clustering.h/cc, using the same name as in colmap/scene.

ahojnnes: +1

src/colmap/exe/sfm.cc
```diff
@@ -739,4 +740,45 @@ int RunViewGraphCalibrator(int argc, char** argv) {
...
+  auto reconstruction = std::make_shared<Reconstruction>();
+  reconstruction->Read(input_path);
+
+  auto reconstruction_manager = std::make_shared<ReconstructionManager>();
+
+  ReconstructionClusteringOptions clustering_options =
+      *options.reconstruction_clusterer;
+
+  ReconstructionClustererController controller(
+      clustering_options, reconstruction, reconstruction_manager);
```
ahojnnes: nit: no need for a copy
```suggestion
  ReconstructionClustererController controller(
      *options.reconstruction_clusterer, reconstruction, reconstruction_manager);
```

lpanaf:
> Thank you, I think that's a great change. I suggest we have at least some minimal tests on the controller side. We don't need to test the underlying algo in detail (as we have another PR already that does that), but at least check that the input reconstruction is split into the expected number of sub-reconstructions with the expected frames, etc.?

I added some tests to the controller itself :)

PR #4014, opened 2026-01-15
Add EQUIRECTANGULAR camera model for 360° panoramic images

zalo:
# Watch a Youtube Video of it working here: https://youtu.be/P1p5VzBWg1M

## Summary
- Add support for equirectangular (360° panoramic) camera model
- Enables Structure-from-Motion reconstruction with 360° cameras like Insta360, Ricoh Theta, etc.
- Skip positive depth / cheirality checks for spherical cameras which can see in all directions

## Changes
1. **New EQUIRECTANGULAR camera model** (`src/colmap/sensor/models.h`)
   - Model ID 10, no parameters (uses image dimensions for projection)
   - `CamFromImg`: Maps pixel coordinates to 3D bearing vectors on unit sphere
   - `ImgFromCam`: Projects 3D points to equirectangular image coordinates
   - Full 360° field of view support

2. **Skip depth checks for spherical cameras** (`src/colmap/estimators/triangulation.cc`)
   - Traditional cheirality constraint (z>0) doesn't apply to 360° cameras
   - Add `IsSphericalCamera()` helper to detect equirectangular cameras
   - Skip positive depth checks in two-view and multi-view triangulation

3. **Skip depth filtering for spherical cameras** (`src/colmap/sfm/observation_manager.cc`)
   - `FilterObservationsWithNegativeDepth()` now skips equirectangular cameras
   - Allows 3D points "behind" camera center to remain valid

4. **Bug fix: Duplicate correspondence crash** (`src/colmap/sfm/observation_manager.cc`)
   - Fix crash where `num_tri_corrs > num_total_corrs` during track completion
   - Convert assertion to warning with counter capping

## Test plan
- [x] Build passes
- [x] Tested with real 360° panoramic dataset (330 images)
- [x] Successfully reconstructed point cloud with GLOMAP pipeline
- [x] Visualization confirms 360° coverage

🤖 Generated with [Claude Code](https://claude.ai/claude-code)


ahojnnes:
Thanks for the attempt. This will be a deep rabbit hole. Triangulation will not work with the proposed changes, because normalized camera coordinates using CamFromImg() are 2d points with z=1. We decided against supporting equirectangular images directly but instead tile the images and then model the camera as a camera rig. See panorama_sfm.py. Did you look at this?

zalo:
Ah yes; I actually wrote a pipeline earlier by hand that updated panorama_sfm to use two 180° fisheye cameras (a lot faster to FFmpeg equirect video to images -> unwarp to 2x fisheye in parallel -> standard rig process (glomap added)); I can PR that too, if desired. That one is battle tested.

But the promise of direct equirect into both Colmap + Glomap is too tantalizing to ignore… why let Metashape hog all the fun? 

Perhaps triangulation can be solved by taking the convex hull of the normalized bearing vectors?  I’ll noodle on the purpose of the triangulation routine a bit…

sarlinpe:
> Ah yes; I actually wrote a pipeline earlier by hand that updated panorama_sfm to use two 180° fisheye cameras

This could be interesting to have, happy to have a look.

> But the promise of direct equirect into both Colmap + Glomap is too tantalizing to ignore… why let Metashape hog all the fun?

I don't understand this argument. Feature detection and matching do not work well on equirectangular images, there is really no benefit to it now that COLMAP supports multi-camera rigs.

zalo:
My motivation for the fisheye rig support was speed; it seems to improve roughly linearly with the number of virtual cameras in the rig.  Can’t do better than 1. 😄  Native equirect saves on a bunch of preprocessing steps too.  And the new gen of Gaussian splatters (like Lichtfeld) can take in Equirect via GUT, which I bet will train faster.

Likewise; rig support was shaky in glomap for a while (I need to do more benchmarks now that it’s merged…)

Feature matching seems to work alright from my tests; doesn’t do great around the seams or towards the poles, but better than I’d feared.   Maybe some rotation independent or equirect aware matcher can get full performance.  I think the existing one is good enough for 90% of cases now though.

ichsan2895:
Hello, I think @zalo is right

Recently, 360 videos into 3DGS is more popular. Instead of creates cubemap/planar of equirectangular images, it can directly train the equirectangular images with Lichtfeld-Studio/others that has GUT feature.

Only agisoft metashape that can handle equirectangular poses. Hopefully colmap can handle it too.

ahojnnes:
> And the new gen of Gaussian splatters (like Lichtfeld) can take in Equirect via GUT, which I bet will train faster.

This can be easily supported by just rewriting the resulting colmap reconstruction and "converting" the virtual camera rig poses to the camera poses for the spherical image.

zalo:
@ahojnnes @sarlinpe 
I'll be back to finish cleaning up the equirect branches here, but for posterity, here are my previous explorations doing equirect reconstruction with rigs and two fisheye views:

https://github.com/user-attachments/assets/8bbe5248-b391-4efb-b279-411deacd981e

Modified Script for ffmpeg video -> .jpgs -> fisheye -> reconstruction: [panorama_sfm_fisheye.py](https://github.com/user-attachments/files/24698649/panorama_sfm_fisheye.py)
Rerun visualizer for .txt serialized colmap scenes: [colmap_txt_visualizer.py](https://github.com/user-attachments/files/24698650/colmap_txt_visualizer.py)

A colmap mode that can do this automatically might be worth it (if I can't make direct to equirect work), but it'll be about half the speed, which will be sad.

zalo:
Here is a video of this PR estimating incremental mapping directly on equirectangular footage: https://youtu.be/P1p5VzBWg1M

Let me know how else I can help!   Datasets, Visualizations, Tests, etc.

src/colmap/geometry/triangulation.h
```diff
@@ -71,6 +71,25 @@ bool TriangulateMidPoint(const Rigid3d& cam2_from_cam1,
...
+// For equirectangular and other spherical camera models.
+//
+// @param cam_from_world1   Pose of the first camera.
+// @param cam_from_world2   Pose of the second camera.
+// @param bearing1          Unit bearing vector in first camera frame.
+// @param bearing2          Unit bearing vector in second camera frame.
+// @param point3D           Triangulated 3D point in world coordinates.
+//
+// @return                  Whether triangulation was successful.
+bool TriangulatePointBearing(const Rigid3d& cam_from_world1,
```
ahojnnes, line 87: This is the exact same as TriangulateMidPoint above just with different inputs?

ahojnnes: I see, I misunderstood the last message in that I thought it was produced from the changes in this PR.

chpatrick:
Does this have quantifiable benefits compared to reprojected pinhole camera rigs? For example the poles are massively distorted by equirectangular projection so I would expect the feature extractor to work very poorly there directly on equirectangular. I think the best solution would be some quality of life feature that automatically reprojects equirectangular to rigs so that you don't need to do it with an external tool.

zalo:
It’s entirely a speed thing; I’d have to check against the latest Colmap commits, but my recollection is that speed is linear with the number of cameras in the rig: 4 rectilinear cameras is half as fast as 2 fisheye cameras, which is half as fast again as 1 equirectangular camera.

zalo:
Actually, https://github.com/colmap/colmap/pull/4433 might just be better; I'll close this one and add some review comments to that one.

zalo:
It seems like #4433 's author abandoned that PR, so I'm putting my latest improvements in: #4441

PR #4016, opened 2026-01-16
Use colmap::BundleAdjustmentOptions in global pipeline and drop custom option class.

B1ueber2y:
Enabled by https://github.com/colmap/colmap/pull/4012. 

* Add a min_track_length option in colmap BA. 
* Drop custom option class and directly store colmap::BundleAdjustmentOptions.
* Align the new names in option manager. 
* Drop the confusing "optimize_rotations" option as the default in global mapper is to run BA for two times (one with rotation fixed and one with full BA). Add an option to skip the fixed-rotation stage in bundle adjustment in the global pipeline, and default it to false.

We therefore can drop the custom BA wrapper in glomap and directly use the colmap one in the global mapper. 

Logic is kept equivalent to before.

### DIFF
--- a/src/colmap/controllers/bundle_adjustment_test.cc
+++ b/src/colmap/controllers/bundle_adjustment_test.cc
@@ -31,6 +31,7 @@
 
 #include "colmap/controllers/option_manager.h"
 #include "colmap/math/random.h"
+#include "colmap/scene/reconstruction_matchers.h"
 #include "colmap/scene/synthetic.h"
 
 #include <gtest/gtest.h>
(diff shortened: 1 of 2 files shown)


src/colmap/controllers/option_manager.cc
```diff
@@ -703,32 +709,37 @@ void OptionManager::AddGlobalMapperOptions() {
...
   AddDefaultOption("GlobalMapper.ba_gpu_index",
                    &global_mapper->mapper.bundle_adjustment.gpu_index);
-  AddDefaultOption("GlobalMapper.ba_optimize_rig_poses",
-                   &global_mapper->mapper.bundle_adjustment.optimize_rig_poses);
-  AddDefaultOption("GlobalMapper.ba_optimize_rotations",
-                   &global_mapper->mapper.bundle_adjustment.optimize_rotations);
   AddDefaultOption(
-      "GlobalMapper.ba_optimize_translation",
-      &global_mapper->mapper.bundle_adjustment.optimize_translation);
+      "GlobalMapper.ba_refine_focal_length",
```
B1ueber2y, line 717: It is still a bit tricky to consolidate with colmap in the option manager, so I for now leave it like this.

lpanaf: Initialy, it allows the extra option to skip the stage of optimizing everything. Do you have more context on why such option is removed? For global_mapper, the rotation results can already be good, and only the translation needs to be optimized if speed is really a concern.

src/glomap/sfm/global_mapper.cc
```diff
@@ -381,20 +414,20 @@ bool GlobalMapper::IterativeRetriangulateAndRefine(
...
   // Set up bundle adjustment options for colmap's incremental mapper.
-  colmap::BundleAdjustmentOptions colmap_ba_options;
-  colmap_ba_options.solver_options.num_threads =
+  colmap::BundleAdjustmentOptions custom_ba_options;
+  custom_ba_options.solver_options.num_threads =
       ba_options.solver_options.num_threads;
-  colmap_ba_options.solver_options.max_num_iterations = 50;
-  colmap_ba_options.solver_options.max_linear_solver_iterations = 100;
-  colmap_ba_options.print_summary = false;
+  custom_ba_options.solver_options.max_num_iterations = 50;
```
lpanaf, line 423: Is it still possible to control the number of iterations in BA?

src/colmap/controllers/option_manager.cc
B1ueber2y: The intention is to align the bundle adjustment CLI with the one for incremental pipeline, which is a follow up of https://github.com/colmap/colmap/pull/3868#issuecomment-3695020017. 

> Initialy, it allows the extra option to skip the stage of optimizing everything.

Not sure I understand this. Which option are you referring to? Do you mean to skip the full optimization in BA and only do the fixed-rotation BA? If so, I can add such an option but I dont think it was there before.

src/glomap/sfm/global_mapper.cc
B1ueber2y: This is for retriangulation which is never controllable.

src/colmap/controllers/option_manager.cc
lpanaf: Yes, initially, the original design was that if this skip_rotation_optimization is set to false, then only the first stage BA is performed, and the second stage is skipped. It can be used for faster mapper

src/glomap/sfm/global_mapper.cc
lpanaf: Okay, I see. Got confused initially :)

src/colmap/controllers/option_manager.cc
B1ueber2y: Okay. This option was already dropped in https://github.com/colmap/colmap/pull/3868 as I did not know skipping full optimization is a reasonable option. I just brought back this support: https://github.com/colmap/colmap/pull/4016/commits/a2f6aa17a302e116f160655d78690301a0dff9fa. PTAL.

src/glomap/sfm/global_mapper.h
```diff
@@ -31,7 +33,24 @@ struct GlobalMapperOptions {
   // Options for each component
   RotationEstimatorOptions rotation_averaging;
   GlobalPositionerOptions global_positioning;
-  BundleAdjusterOptions bundle_adjustment;
+  colmap::BundleAdjustmentOptions bundle_adjustment = [] {
+    colmap::BundleAdjustmentOptions opts;
```
ahojnnes: nit
```suggestion
    colmap::BundleAdjustmentOptions options;
```

PR #4017, opened 2026-01-16
Analytical Jacobians for reprojection error cost function with SimpleRadial cameras

ahojnnes:
Makes BA ~10% faster on south building with ~100 cameras and synthetic tests. More camera models and cost functions to be added later.

### DIFF
--- a/src/colmap/estimators/cost_functions.h
+++ b/src/colmap/estimators/cost_functions.h
@@ -648,6 +648,7 @@ struct Point3DAlignmentCostFunctor
 template <template <typename> class CostFunctor, typename... Args>
 ceres::CostFunction* CreateCameraCostFunction(
     const CameraModelId camera_model_id, Args&&... args) {
+  // NOLINTBEGIN(bugprone-macro-parentheses)
   switch (camera_model_id) {
 #define CAMERA_MODEL_CASE(CameraModel)                                        \
   case CameraModel::model_id:                                                 \
@@ -665,6 +666,7 @@ ceres::CostFunction* CreateCameraCostFunction(
 
 #undef CAMERA_MODEL_CASE
   }
+  // NOLINTEND(bugprone-macro-parentheses)
 }
 
 // Compute polynomial coefficients from cross-products of SVD-derived vectors
--- a/src/colmap/sensor/models_test.cc
+++ b/src/colmap/sensor/models_test.cc
@@ -244,7 +244,9 @@ void TestModel(const std::vector<double>& params) {
       params, params[pp_idxs.at(0)], params[pp_idxs.at(1)]);
 
   if constexpr (CameraModel::has_img_from_cam_with_jac) {
+    // NOLINTNEXTLINE(clang-analyzer-security.FloatLoopCounter)
     for (double u = -0.5; u <= 0.5; u += 0.1) {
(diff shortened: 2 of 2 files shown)


src/colmap/estimators/cost_functions.h
```diff
@@ -43,6 +43,192 @@
...
+    J_out[9] = 2.0 * (-qw_px + qz_py - 2.0 * qy_pz);
+    J_out[10] = 2.0 * (qx_px + qy_py);
+    J_out[11] = 2.0 * (-qy_px + qx_py);
+  }
+
+  return pt_out;
+}
+
+// Eigen quaternion to 3x3 rotation matrix.
+inline Eigen::Matrix3d QuaternionToScaledRotation(const double* q) {
```
B1ueber2y: I might be wrong, but I think this function can be replaced with EigenQuaternionMap<double>(q).toRotationMatrix().

src/colmap/estimators/bundle_adjustment.cc
```diff
@@ -781,6 +781,8 @@ class DefaultBundleAdjuster : public BundleAdjuster {
 
     ceres::Solve(solver_options, problem_.get(), &summary);
 
+    LOG(INFO) << summary.FullReport();
```
B1ueber2y: This is already present in PrintSolverSummary.

src/colmap/estimators/cost_functions.h
```diff
@@ -495,6 +679,14 @@ struct Point3DAlignmentCostFunctor
 template <template <typename> class CostFunctor, typename... Args>
 ceres::CostFunction* CreateCameraCostFunction(
     const CameraModelId camera_model_id, Args&&... args) {
+  if constexpr (std::is_same<
```
B1ueber2y: Until we make the jacobian support into BaseCameraModel, we can use type traits here to detect whether ImgFromCamWithJac exists.

ahojnnes: Yes, that was the idea. This was just the initial hacky solution to get some first numbers.

ahojnnes:
@B1ueber2y If have some hacky benchmark: https://github.com/colmap/colmap/pull/4022

src/colmap/estimators/cost_functions_test.cc
```diff
@@ -74,6 +74,46 @@ TEST(ReprojErrorCostFunctor, Nominal) {
   EXPECT_EQ(residuals[1], 2);
 }
 
+TEST(AnalyticalReprojErrorCostFunction, GradientChecker) {
+  for (const double x : {0, 100, 200, 300}) {
```
B1ueber2y: In this test, the jacobian at each loop is actually the same since the 3D point fixed to (2,1,4). Wondering if it makes sense to initialize the 3D point by using CamFromImg plus some random noise.

B1ueber2y (approved):

B1ueber2y:
Reminder not to forget to update the PR title to specify SimpleRadial : )

src/colmap/estimators/cost_functions_test.cc
ahojnnes: Good catch. Fixed.

PR #4019, opened 2026-01-16
Add nominal tests on reconstruction accuracy for bundle adjustment. 

B1ueber2y:
Realizing that we are missing important nominal tests on the default bundle adjustment estimator in terms of reconstruction accuracy. Until now the test only covers number of DoFs, whether the variables are constant, and accuracy test on PosePriorBundleAdjustment. 

Also improved the previous test on bundle adjustment controller by using the reconstruction matcher. The previous test only requires the residuals to go down.

### DIFF
--- a/src/colmap/controllers/bundle_adjustment_test.cc
+++ b/src/colmap/controllers/bundle_adjustment_test.cc
@@ -31,6 +31,7 @@
 
 #include "colmap/controllers/option_manager.h"
 #include "colmap/math/random.h"
+#include "colmap/scene/reconstruction_matchers.h"
 #include "colmap/scene/synthetic.h"
 
 #include <gtest/gtest.h>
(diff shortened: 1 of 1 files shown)


ahojnnes (approved):

ichsan2895:
Sorry noobs question. How to use this feature?

It is interesting if we compare the default bundle-adjustment (Sparse-schur CuDSS & Dense-schur CUDA) versus bundle-adjustment with caspar

B1ueber2y:
> Sorry noobs question. How to use this feature?
> 
> It is interesting if we compare the default bundle-adjustment (Sparse-schur CuDSS & Dense-schur CUDA) versus bundle-adjustment with caspar

This is just an improve test to ensure that the bundle adjustment is working and converges to a good reconstruction, so cannot be used for comparison.

PR #3920, opened 2025-12-31
Drop experimental modes in global positioning. Add module signature and improve naming. 

B1ueber2y:
Also added a reference for the BATA cost.

PR #4021, opened 2026-01-17
Specify custom linear BA solver ordering

ahojnnes:
For a full south-building problem, roughly halves the runtime of the preprocessor from previously ~0.4s to ~0.2s, thus leading to overall ~10% speedup of such problems.
```
Time (in seconds):
Preprocessor                         0.189728

  Residual only evaluation           0.039847 (7)
  Jacobian & residual evaluation     0.167393 (8)
  Linear solver                      1.446735 (7)
Minimizer                            1.740785

Postprocessor                        0.003945
Total                                1.934458
```

### DIFF
--- a/src/colmap/estimators/bundle_adjustment.h
+++ b/src/colmap/estimators/bundle_adjustment.h
@@ -191,6 +191,7 @@ struct BundleAdjustmentOptions {
 
   // Whether to use custom parameter block ordering for Schur-based solvers.
   // Disable for deterministic behavior when using a fixed random seed.
+  // See: https://groups.google.com/g/ceres-solver/c/2ITKHCi3-oU/m/OlxDBAq6BYAJ
   bool use_parameter_block_ordering = true;
 
   // Ceres-Solver options.
--- a/src/glomap/estimators/global_positioning.h
+++ b/src/glomap/estimators/global_positioning.h
@@ -61,6 +61,7 @@ struct GlobalPositionerOptions {
 
   // Whether to use custom parameter block ordering for Schur-based solvers.
   // Disable for deterministic behavior when using a fixed random seed.
+  // See: https://groups.google.com/g/ceres-solver/c/2ITKHCi3-oU/m/OlxDBAq6BYAJ
   bool use_parameter_block_ordering = true;
 
   // The options for the solver


src/colmap/estimators/bundle_adjustment.cc
```diff
@@ -620,11 +583,58 @@ void FixGaugeWithTwoCamsFromWorld(
   }
 }
 
+void ParameterizeCameras(const BundleAdjustmentOptions& options,
```
ahojnnes, line 586: Just moved down.

B1ueber2y:
Thanks! We might need to disable parameter block ordering depending on the random seed if we hit again the non-deterministic issue. Similar to the other PR, would this benefit from a benchmark with our synthetic tooling?

src/colmap/estimators/bundle_adjustment.cc
```diff
@@ -41,6 +41,11 @@
 namespace colmap {
 namespace {
 
+constexpr int kPointsEliminationGroup = 0;
+constexpr int kFramesEliminationGroup = 1;
+constexpr int kCamerasEliminationGroup = 2;
```
B1ueber2y: I think here the ordering is actually wrong (should be the inverse). Ceres eliminates the last group rather than the first.

B1ueber2y: Also, the Ceres internal implementation only eliminates one group and does not support cascaded elimination, so the ordering of the other three groups only contributes to improving the linear system structure. So I am not sure if the current ``kEliminationGroup`` here is the best naming.

src/glomap/sfm/global_mapper.cc
```diff
@@ -381,11 +382,8 @@ bool GlobalMapper::IterativeRetriangulateAndRefine(
   }
 
   // Set up bundle adjustment options for colmap's incremental mapper.
-  colmap::BundleAdjustmentOptions colmap_ba_options;
-  colmap_ba_options.solver_options.num_threads =
-      ba_options.solver_options.num_threads;
-  colmap_ba_options.solver_options.max_num_iterations = 50;
-  colmap_ba_options.solver_options.max_linear_solver_iterations = 100;
+  colmap::BundleAdjustmentOptions colmap_ba_options =
```
ahojnnes: I am not sure if the previous custom parameters were intentional? It seems like we were not correctly passing through options (e.g., optimize_rigs, etc.) here?

B1ueber2y: This is for the bundle adjustment in the retriangulation, which used the incremental pipeline.

src/colmap/estimators/bundle_adjustment.cc
ahojnnes: I think the ordering should be right but maybe I am misreading the documentation. The part that makes me confident is that the automatic ordering leads to the same ordering in the full report.

I will rename the parameters, agreed on this being strictly not correct.

src/glomap/sfm/global_mapper.cc
B1ueber2y: Oh yes I get what you are saying. Yes some of them might not be well propagated here to the retriangulation. Should be solved here: https://github.com/colmap/colmap/pull/4016.

src/colmap/estimators/bundle_adjustment.cc
B1ueber2y: > I think the ordering should be right but maybe I am misreading the documentation. The part that makes me confident is that the automatic ordering leads to the same ordering in the full report.
> 
> I will rename the parameters, agreed on this being strictly not correct.

Indeed. I got confused. I am also using the same ordering : )

src/colmap/estimators/bundle_adjustment.cc
```diff
@@ -41,6 +41,14 @@
 namespace colmap {
 namespace {
 
+// Note that Ceres only supports two elimination groups, so this effectively
```
lpanaf: In this page: http://ceres-solver.org/nnls_solving.html, it specifies that  when linear_solver_ordering_type = AMD (can use this to search for the item), all the numbers are used for elimination

src/pycolmap/estimators/bundle_adjustment.cc
```diff
@@ -198,6 +198,11 @@ void BindBundleAdjuster(py::module& m) {
               "Whether to automatically select solver type based on "
               "problem size. When False, uses the linear_solver_type "
               "and preconditioner_type from solver_options directly.")
+          .def_readwrite("use_parameter_block_ordering",
+                         &BAOpts::use_parameter_block_ordering,
+                         "Whether to use custom parameter block ordering for "
```
lpanaf, line 203: Why would it be deterministic when this is disabled and non-deterministic otherwise? I thought both case would be non-deterministic?

ahojnnes: I added a link to the explanation in the BundleAdjustmentOptions.

ahojnnes:
@B1ueber2y @lpanaf Fixing the reordering issue on the Ceres side, so this maybe becomes less of a headache in the future: https://ceres-solver-review.googlesource.com/c/ceres-solver/+/27660

ahojnnes:
After some more benchmarking, I am not convinced we should enable this and maybe even disable in the global positioning. Depending on the specific problem configuration (num_rigs/cams/frames/points), automatic ordering more often than not performs better than manual ordering.

lpanaf:
> After some more benchmarking, I am not convinced we should enable this and maybe even disable in the global positioning. Depending on the specific problem configuration (num_rigs/cams/frames/points), automatic ordering more often than not performs better than manual ordering.

Thanks for the experiments! This behavior sounds plausible for bundle adjustment but is a bit of unexpected for global positioning. Ther is a large number of scale parameters, and the elimination can potentially takes quite some time to discover that they can be eliminated? But maybe the automatic detector is very powerful on the other hand.

ahojnnes:
> Thanks for the experiments! This behavior sounds plausible for bundle adjustment but is a bit of unexpected for global positioning. Ther is a large number of scale parameters, and the elimination can potentially takes quite some time to discover that they can be eliminated? But maybe the automatic detector is very powerful on the other hand.

I have not looked at the implementation, but I would assume this is a "simple" clustering of the dependency graph of parameters, where all the instances of a parameter are independent of each other. Then, just find the largest cluster of independent parameters and eliminate it first?

B1ueber2y:
Some comments from my side: I am not surprised by the fact that manual ordering does not help in the default bundle adjustment problem, as I got the same conclusion two years ago with this branch: https://github.com/B1ueber2y/colmap/tree/features/ordering_schur. 

I did more experiments last year for the custom bundle adjustment problem I had (with not only points). The key for the elimination to help in my practice is to have single-node continous-memory for each parameter block to be eliminated (e.g., in the colmap context the 3D point is indeed a single node, but the camera pose has two nodes (rotation and translation)). The reason for this is likely that ceres internally uses the maximum independent set algorithm to detect the first elimination group, which is a "simple" algorithm (via BFS) to detect leaf nodes which is the case for each point parameter block in colmap. This process happens before the optimization so is not the bottleneck in each loop. 

At some point I tried to use a single 7-dimensional parameter rather than the separate rotation / translation vectors in Rigid3d but did not get much success in improving the runtime, probably because it is already quite dense after point elimination so the structure does not matter anymore. 

I agree that doing more benchmark on global positioning can help decide whether it is really useful, which probably can be done in the same way once the bundle adjustment runtime benchmark gets stabilized. Things can also be easier after https://github.com/colmap/colmap/pull/3920 gets merged.

ahojnnes:
@B1ueber2y I tried this now again that we have a single pose parameter block. It doesn't make a difference. Abandoning this PR.

Issue #4011, opened 2026-01-14
image_undistorter: Unexpectedly large file size increase in output images


deniaqgoog-droid:
<h3>Description</h3>
<p>I am experiencing a significant increase in file size when using the <code>image_undistorter</code> tool.</p>

<h3>Observed Behavior</h3>
<p>Input distorted images are approximately <strong>1MB</strong> in size (JPEG). After running the undistorter with <code>--min_scale 1</code> and <code>--max_scale 1</code>, the resulting undistorted images are <strong>5MB to 6MB</strong> each.</p>

<p>I suspect this is because the tool defaults to a "Superb" JPEG quality (100) or high-density resampling that inflates the footprint. For datasets with thousands of frames (e.g., for Gaussian Splatting), this creates a massive storage bottleneck.</p>

<h3>Command Used</h3>
<pre><code>colmap image_undistorter \
    --image_path ./images \
    --input_path ./sparse/0 \
    --output_path ./undistorted \
    --min_scale 1 \
    --max_scale 1</code></pre>

<h3>Feature Request</h3>
<p>Would it be possible to expose a <code>--jpeg_quality</code> or <code>--compression</code> flag in the CLI for the <code>image_undistorter</code>?</p>

<h3>Contribution / Implementation Question</h3>
<p>I am willing to implement a solution and submit a PR for this. Based on my initial look at the source, it seems the default quality is hardcoded in <code>src/colmap/sensor/bitmap.cc</code> (using <code>JPEG_QUALITY_SUPERB</code>).</p>

<p>Could you confirm if the following files are the most relevant for adding this flag to the CLI and the undistortion controller?</p>
<ul>
  <li><code>src/colmap/exe/image.cc</code> (to add the CLI argument)</li>
  <li><code>src/colmap/controllers/image_undistorter.cc</code> (to pass the option)</li>
  <li><code>src/colmap/sensor/bitmap.cc</code> (to apply the quality setting during save)</li>
</ul>

<h3>Desktop Info</h3>
<ul>
  <li><strong>OS:</strong> Ubuntu 24.04</li>
  <li><strong>COLMAP Version:</strong> 3.13.0 (Commit 62b82454)</li>
</ul>

ahojnnes:
That's a good idea. There could be a jpeg_image_quality setting in the undistorter settings and if the image extension is jpeg, we set `bitmap.SetMetaData("Compression", "jpeg:" + std::tostring(option.jpeg_quality));`?

[...]

PR #4027, opened 2026-01-18
Add option to specify jpeg quality during undistortion

ahojnnes:
Also some other code quality improvements. Addresses: https://github.com/colmap/colmap/issues/4011

### DIFF
--- a/src/colmap/controllers/feature_extraction.cc
+++ b/src/colmap/controllers/feature_extraction.cc
@@ -343,7 +343,7 @@ class FeatureExtractorController : public Thread {
       if (ExistsFile(reader_options_.camera_mask_path)) {
         camera_mask = std::make_shared<Bitmap>();
         if (!camera_mask->Read(reader_options_.camera_mask_path,
-                               /*as_rgb*/ false)) {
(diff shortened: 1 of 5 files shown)


src/colmap/image/undistortion.cc
```diff
@@ -41,6 +41,15 @@
 namespace colmap {
 namespace {
 
+void MaybeSetJpegQuality(const std::filesystem::path& path,
+                         Bitmap& bitmap,
+                         int jpeg_quality) {
+  if ((HasFileExtension(path, ".jpg") || HasFileExtension(path, ".jpeg")) &&
+      jpeg_quality > 0) {
+    bitmap.SetMetaData("Compression", "jpeg:" + std::to_string(jpeg_quality));
```
sarlinpe, line 49: Wouldn't it be nicer to have this as an option to `Bitmap::Write`? It's easy to mistype these strings and cannot be caught at compile time.

ahojnnes: Thanks, found a middle ground by exposing an explicit SetJpegQuality method instead.

PR #4032, opened 2026-01-20
Clear model outputs in all estimators and RANSACs

ahojnnes:
* Both P3P and P4PF were missing calls to clear(). In practice, this was only an issue in P4PF, as poselib clears the solutions in P3P.
* Added further protection against such bugs in the future by clearing the sampled models explicitly also in the RANSACs before calling Estimate().


B1ueber2y (approved):
Does it make sense to also fix this in the poselib side? Ideally the clear() should be called at each solver of poselib.

ahojnnes:
Makes sense, I am admittedly too busy with other things right now.

PR #4033, opened 2026-01-20
Move undistortion controllers to controllers folder

ahojnnes:
Just moving code, no functional change (except for minor cosmetics like const and include cleanup).


B1ueber2y (approved):
Did not read in details as it's hard to diff. This is a half-blind approval : )

PR #4039, opened 2026-01-21
Add fisheye (equidistant) camera model

lpanaf:
Add fisheye (equidistant) camera model, which is essentially the OpenCV fisheye camera model without distortion.

### DIFF
--- a/src/colmap/sensor/models.h
+++ b/src/colmap/sensor/models.h
@@ -535,12 +535,8 @@ struct DivisionCameraModel : public BaseCameraModel<DivisionCameraModel> {
 //
 struct SimpleFisheyeCameraModel
     : public BaseFisheyeCameraModel<SimpleFisheyeCameraModel> {
-  CAMERA_MODEL_DEFINITIONS(CameraModelId::kSimpleFisheye,
-                           "SIMPLE_FISHEYE",
-                           1,
-                           2,
-                           0,
-                           false)
+  CAMERA_MODEL_DEFINITIONS(
+      CameraModelId::kSimpleFisheye, "SIMPLE_FISHEYE", 1, 2, 0, false)
   FISHEYE_CAMERA_MODEL_DEFINITIONS
 };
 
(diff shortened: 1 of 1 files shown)


B1ueber2y:
In the COLMAP naming context, would it make sense to call these two models ``SimpleFisheyeCameraModel`` and ``FisheyeCameraModel``?

lpanaf:
> In the COLMAP naming context, would it make sense to call these two models `SimpleFisheyeCameraModel` and `FisheyeCameraModel`?

This one I am not sure. It is standard, but I don't know whether fisheye camera model can potentially refer to other models. Maybe @ahojnnes can comment on this?

src/colmap/sensor/models.h
```diff
@@ -2149,6 +2188,169 @@ void DivisionCameraModel::Distortion(
...
+  return true;
+}
+
+template <typename T>
+void EquiangularModel::Distortion(
+    const T* extra_params, const T& u, const T& v, T* du, T* dv) {
+  // No distortion
+  *du = T(0);
+  *dv = T(0);
+}
```
ahojnnes: I think we can also just not define this method and it should work?

src/colmap/sensor/models.h
```diff
@@ -517,6 +523,39 @@ struct DivisionCameraModel : public BaseCameraModel<DivisionCameraModel> {
...
+//
+// Uses the equiangular (theta = r) fisheye projection without distortion
+// parameters. Suitable for fish-eye cameras where distortion can be ignored
+// or has been pre-corrected. This model has a single focal length.
+//
+// Parameter list is expected in the following order:
+//
+//    f, cx, cy
+//
+struct SimpleEquiangularModel
```
ahojnnes: Maybe "Equidistant" instead of "Equiangular"?

ahojnnes (commented):
Thank you!

src/colmap/sensor/models.h
ahojnnes: Or maybe even better as @B1ueber2y suggests.

lpanaf: Sounds good, thanks for the suggestions!

src/colmap/sensor/models.h
```diff
@@ -517,6 +523,43 @@ struct DivisionCameraModel : public BaseCameraModel<DivisionCameraModel> {
...
+// or has been pre-corrected. This model has a single focal length.
+//
+// Parameter list is expected in the following order:
+//
+//    f, cx, cy
+//
+struct SimpleFisheyeCameraModel
+    : public BaseFisheyeCameraModel<SimpleFisheyeCameraModel> {
+  CAMERA_MODEL_DEFINITIONS(CameraModelId::kSimpleFisheyeCamera,
+                           "SIMPLE_FISHEYE_CAMERA",
```
ahojnnes: nit: Let's name them just "SIMPLE_FISHEYE" and "FISHEYE" in the enum/str representation.

B1ueber2y: Need to update pybind as well.

PR #4041, opened 2026-01-22
Store Rigid3/Sim3 params as single vector

ahojnnes:
* Makes it easier to manage parameter vectors in Ceres optimization problems.
* Fully backwards compatible on the Python side, breaking change for C++ users as in need for changing from rotation/translation to rotation()/translation().

### DIFF
--- a/src/pycolmap/geometry/eigen.cc
+++ b/src/pycolmap/geometry/eigen.cc
@@ -16,22 +16,13 @@ void BindEigenGeometry(py::module& m) {
   using Rotation3dWrapper = pycolmap::Rotation3dWrapper;
   py::classh_ext<Rotation3dWrapper> PyRotation3d(m, "Rotation3d");
   PyRotation3d.def(py::init<>())
-      .def(py::init([](const Eigen::Vector4d& xyzw) {
-             Rotation3dWrapper w;
-             w.map().coeffs() = xyzw;
-             return w;
-           }),
+      .def(py::init<const Eigen::Vector4d&>(),
            "xyzw"_a,
            "Quaternion in [x,y,z,w] format.")
-      .def(py::init([](const Eigen::Matrix3d& matrix) {
-             return Rotation3dWrapper(Eigen::Quaterniond(matrix));
-           }),
+      .def(py::init<const Eigen::Matrix3d&>(),
            "matrix"_a,
            "3x3 rotation matrix.")
-      .def(py::init([](const Eigen::Vector3d& vec) {
-             return Rotation3dWrapper(Eigen::Quaterniond(
-                 Eigen::AngleAxis<double>(vec.norm(), vec.normalized())));
-           }),
+      .def(py::init<const Eigen::Vector3d&>(),
            "axis_angle"_a,
            "Axis-angle 3D vector.")
       .def_static(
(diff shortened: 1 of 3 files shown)


src/pycolmap/geometry/rigid3.cc
```diff
@@ -23,8 +23,22 @@ void BindRigid3(py::module& m) {
       .def(py::init(&Rigid3d::FromMatrix),
            "matrix"_a,
            "3x4 transformation matrix.")
-      .def_readwrite("rotation", &Rigid3d::rotation)
-      .def_readwrite("translation", &Rigid3d::translation)
+      .def_property(
```
B1ueber2y, line 33: We would like to expose params as well, to be able to work with pyceres with the switch of cost functions in the future.

src/pycolmap/geometry/rigid3.cc
```diff
@@ -23,8 +23,22 @@ void BindRigid3(py::module& m) {
...
+          "rotation",
+          [](const Rigid3d& self) -> Eigen::Quaterniond {
```
B1ueber2y: I am wondering if it makes sense to add a rigid3d_test in the pycolmap side, to check if we can do in-place parameter setting with auto-binded Eigen::Map.

B1ueber2y: I can confirm that this is working the same way as before in my quick test. So probably no need for the test.

src/pycolmap/geometry/rigid3.cc
```diff
@@ -23,7 +23,10 @@ void BindRigid3(py::module& m) {
       .def(py::init(&Rigid3d::FromMatrix),
            "matrix"_a,
            "3x4 transformation matrix.")
-      .def_property("params", &Rigid3d::params)
+      .def_property(
+          "params",
+          [](const Rigid3d& self) -> Eigen::Vector7d { return self.params; },
```
B1ueber2y: This wont work because this returns by copy rather than reference.

src/pycolmap/geometry/sim3.cc
```diff
@@ -25,25 +25,39 @@ void BindSim3(py::module& m) {
...
+          "params",
+          [](Sim3d& self) -> Eigen::Vector8d& { return self.params; },
+          [](Sim3d& self, const Eigen::Vector8d& params) {
+            self.params = params;
+          })
       .def_property(
           "scale",
-          [](Sim3d& self) {
-            return py::array({}, {}, &self.scale, py::cast(self));
+          [](Sim3d& self) { return self.params(7); },
```
B1ueber2y: This might prevent in-place optimization on top of scale with pyceres.

B1ueber2y: Okay alternatively, we could update this cost function: https://github.com/colmap/colmap/blob/main/src/colmap/estimators/cost_functions.h#L614 to use continuous parameter, as the cost function is not usable anymore after this update. But then it might make sense to store log_scale instead.

B1ueber2y:
Should we also add the product manifold alias for SE(3) and ST(3) in this PR?

ahojnnes:
> Should we also add the product manifold alias for SE(3) and ST(3) in this PR?

Let's do this separately. The PR is already big enough and is self-contained as it stands now.

src/pycolmap/geometry/sim3.cc
ahojnnes: The cost function is updated in my other PR, where this becomes a single parameter. Then, you just have take the log before the optimization and exp afterwards.

PR #4042, opened 2026-01-22
Abstract general bundle adjuster and solver summary.

B1ueber2y:
In preparation for supporting different backends.

The breakage includes:
1) changes on the BundleAdjustmentOptions struct, in both C++ and pycolmap.
2) The return of BundleAdjuster now becomes ``std::shared_ptr<BundleAdjusterSummary>`` instead of the ``ceres::SolverSummary`` before. 
3) The ceres-related properties in bundle adjustment CLI becomes ``BundleAdjustmentCeres.xxx``, and in global pipeline becomes ``GlobalMapper.ba_ceres_xxx``.

### DIFF
--- a/benchmark/runtime/bundle_adjustment.cc
+++ b/benchmark/runtime/bundle_adjustment.cc
@@ -118,7 +118,7 @@ BENCHMARK_DEFINE_F(BM_BundleAdjustment, Solve)(benchmark::State& state) {
     state.PauseTiming();
     num_iterations += GetCeresSummary(summary.get()).num_successful_steps;
     if (summary->termination_type ==
-        BATerminationType::NO_CONVERGENCE) {
+        BundleAdjustmentTerminationType::NO_CONVERGENCE) {
       state.SkipWithError("Bundle adjustment did not converge");
     }
     state.ResumeTiming();
--- a/src/colmap/estimators/bundle_adjustment.cc
+++ b/src/colmap/estimators/bundle_adjustment.cc
@@ -34,15 +34,16 @@
 namespace colmap {
 
 bool BundleAdjustmentSummary::IsSolutionUsable() const {
-  return termination_type == BATerminationType::CONVERGENCE ||
-         termination_type == BATerminationType::NO_CONVERGENCE ||
-         termination_type == BATerminationType::USER_SUCCESS;
+  return termination_type == BundleAdjustmentTerminationType::CONVERGENCE ||
+         termination_type == BundleAdjustmentTerminationType::NO_CONVERGENCE ||
+         termination_type == BundleAdjustmentTerminationType::USER_SUCCESS;
 }
 
 std::string BundleAdjustmentSummary::BriefReport() const {
-  return "termination: " +
-         std::string(BATerminationTypeToString(termination_type)) +
-         ", num_residuals: " + std::to_string(num_residuals);
+  return "Bundle adjustment report: termination=" +
+         std::string(
+             BundleAdjustmentTerminationTypeToString(termination_type)) +
+         ", num_residuals=" + std::to_string(num_residuals);
 }
 
 ////////////////////////////////////////////////////////////////////////////////
--- a/src/colmap/estimators/bundle_adjustment.h
+++ b/src/colmap/estimators/bundle_adjustment.h
@@ -48,7 +48,7 @@ MAKE_ENUM_CLASS_OVERLOAD_STREAM(
     BundleAdjustmentGauge, -1, UNSPECIFIED, TWO_CAMS_FROM_WORLD, THREE_POINTS);
 
 // Termination type for bundle adjustment, independent of solver backend.
-MAKE_ENUM_CLASS_OVERLOAD_STREAM(BATerminationType,
+MAKE_ENUM_CLASS_OVERLOAD_STREAM(BundleAdjustmentTerminationType,
                                 0,
                                 CONVERGENCE,
                                 NO_CONVERGENCE,
@@ -61,7 +61,8 @@ MAKE_ENUM_CLASS_OVERLOAD_STREAM(BundleAdjustmentBackend, 0, CERES);
 
 // Summary of bundle adjustment results, independent of solver backend.
 struct BundleAdjustmentSummary {
-  BATerminationType termination_type = BATerminationType::FAILURE;
+  BundleAdjustmentTerminationType termination_type =
+      BundleAdjustmentTerminationType::FAILURE;
   // Number of residuals connected to at least one variable parameter block.
   // Excludes residuals where all connected parameters are constant.
   int num_residuals = 0;
--- a/src/colmap/estimators/bundle_adjustment_ceres.cc
+++ b/src/colmap/estimators/bundle_adjustment_ceres.cc
@@ -42,22 +42,22 @@ namespace colmap {
 
 namespace {
 
-BATerminationType CeresTerminationTypeToTerminationType(
+BundleAdjustmentTerminationType CeresTerminationTypeToTerminationType(
     ceres::TerminationType ceres_type) {
   switch (ceres_type) {
     case ceres::CONVERGENCE:
-      return BATerminationType::CONVERGENCE;
+      return BundleAdjustmentTerminationType::CONVERGENCE;
     case ceres::NO_CONVERGENCE:
-      return BATerminationType::NO_CONVERGENCE;
+      return BundleAdjustmentTerminationType::NO_CONVERGENCE;
     case ceres::FAILURE:
-      return BATerminationType::FAILURE;
+      return BundleAdjustmentTerminationType::FAILURE;
     case ceres::USER_SUCCESS:
-      return BATerminationType::USER_SUCCESS;
+      return BundleAdjustmentTerminationType::USER_SUCCESS;
     case ceres::USER_FAILURE:
-      return BATerminationType::USER_FAILURE;
+      return BundleAdjustmentTerminationType::USER_FAILURE;
   }
   LOG(FATAL_THROW) << "Unknown Ceres termination type: " << ceres_type;
-  return BATerminationType::FAILURE;
+  return BundleAdjustmentTerminationType::FAILURE;
 }
 
 std::unique_ptr<ceres::LossFunction> CreateLossFunction(
--- a/src/colmap/estimators/bundle_adjustment_test.cc
+++ b/src/colmap/estimators/bundle_adjustment_test.cc
@@ -79,19 +79,19 @@ TEST(PosePriorBundleAdjustmentOptions, Copy) {
 TEST(BundleAdjustmentSummary, IsSolutionUsable) {
   BundleAdjustmentSummary summary;
 
-  summary.termination_type = BATerminationType::CONVERGENCE;
+  summary.termination_type = BundleAdjustmentTerminationType::CONVERGENCE;
   EXPECT_TRUE(summary.IsSolutionUsable());
 
-  summary.termination_type = BATerminationType::NO_CONVERGENCE;
+  summary.termination_type = BundleAdjustmentTerminationType::NO_CONVERGENCE;
   EXPECT_TRUE(summary.IsSolutionUsable());
 
-  summary.termination_type = BATerminationType::USER_SUCCESS;
+  summary.termination_type = BundleAdjustmentTerminationType::USER_SUCCESS;
   EXPECT_TRUE(summary.IsSolutionUsable());
 
-  summary.termination_type = BATerminationType::FAILURE;
+  summary.termination_type = BundleAdjustmentTerminationType::FAILURE;
   EXPECT_FALSE(summary.IsSolutionUsable());
 
-  summary.termination_type = BATerminationType::USER_FAILURE;
+  summary.termination_type = BundleAdjustmentTerminationType::USER_FAILURE;
   EXPECT_FALSE(summary.IsSolutionUsable());
 }
 
--- a/src/pycolmap/estimators/bundle_adjustment.cc
+++ b/src/pycolmap/estimators/bundle_adjustment.cc
@@ -52,14 +52,16 @@ class PyCeresBundleAdjuster : public CeresBundleAdjuster,
 void BindBundleAdjuster(py::module& m) {
   IsPyceresAvailable();  // Try to import pyceres to populate the docstrings.
 
-  auto PyBATerminationType =
-      py::enum_<BATerminationType>(m, "BATerminationType")
-          .value("CONVERGENCE", BATerminationType::CONVERGENCE)
-          .value("NO_CONVERGENCE", BATerminationType::NO_CONVERGENCE)
-          .value("FAILURE", BATerminationType::FAILURE)
-          .value("USER_SUCCESS", BATerminationType::USER_SUCCESS)
-          .value("USER_FAILURE", BATerminationType::USER_FAILURE);
-  AddStringToEnumConstructor(PyBATerminationType);
+  auto PyBundleAdjustmentTerminationType =
+      py::enum_<BundleAdjustmentTerminationType>(
+          m, "BundleAdjustmentTerminationType")
+          .value("CONVERGENCE", BundleAdjustmentTerminationType::CONVERGENCE)
+          .value("NO_CONVERGENCE",
+                 BundleAdjustmentTerminationType::NO_CONVERGENCE)
+          .value("FAILURE", BundleAdjustmentTerminationType::FAILURE)
+          .value("USER_SUCCESS", BundleAdjustmentTerminationType::USER_SUCCESS)
+          .value("USER_FAILURE", BundleAdjustmentTerminationType::USER_FAILURE);
+  AddStringToEnumConstructor(PyBundleAdjustmentTerminationType);
 
   using BASummary = BundleAdjustmentSummary;
   auto PyBundleAdjustmentSummary =
--- a/src/colmap/estimators/bundle_adjustment_ceres_test.cc
+++ b/src/colmap/estimators/bundle_adjustment_ceres_test.cc
@@ -150,7 +150,8 @@ TEST(DefaultBundleAdjuster, Nominal) {
   std::unique_ptr<BundleAdjuster> bundle_adjuster =
       CreateDefaultCeresBundleAdjuster(options, config, reconstruction);
   const auto summary = bundle_adjuster->Solve();
-  ASSERT_NE(summary->termination_type, BATerminationType::FAILURE);
+  ASSERT_NE(summary->termination_type,
+            BundleAdjustmentTerminationType::FAILURE);
 
   EXPECT_THAT(gt_reconstruction,
               ReconstructionNear(reconstruction,
@@ -189,7 +190,8 @@ TEST(DefaultBundleAdjuster, NominalMultiCameraRig) {
   std::unique_ptr<BundleAdjuster> bundle_adjuster =
       CreateDefaultCeresBundleAdjuster(options, config, reconstruction);
   const auto summary = bundle_adjuster->Solve();
-  ASSERT_NE(summary->termination_type, BATerminationType::FAILURE);
+  ASSERT_NE(summary->termination_type,
+            BundleAdjustmentTerminationType::FAILURE);
 
   EXPECT_THAT(gt_reconstruction,
               ReconstructionNear(reconstruction,
@@ -221,7 +223,8 @@ TEST(DefaultBundleAdjuster, TwoView) {
   std::unique_ptr<BundleAdjuster> bundle_adjuster =
       CreateDefaultCeresBundleAdjuster(options, config, reconstruction);
   const auto summary = bundle_adjuster->Solve();
-  ASSERT_NE(summary->termination_type, BATerminationType::FAILURE);
+  ASSERT_NE(summary->termination_type,
+            BundleAdjustmentTerminationType::FAILURE);
 
   EXPECT_EQ(config.NumResiduals(reconstruction),
             GetCeresProblem(*bundle_adjuster).NumResiduals());
@@ -270,7 +273,8 @@ TEST(DefaultBundleAdjuster, TwoViewRig) {
   std::unique_ptr<BundleAdjuster> bundle_adjuster =
       CreateDefaultCeresBundleAdjuster(options, config, reconstruction);
   const auto summary = bundle_adjuster->Solve();
-  ASSERT_NE(summary->termination_type, BATerminationType::FAILURE);
+  ASSERT_NE(summary->termination_type,
+            BundleAdjustmentTerminationType::FAILURE);
 
   EXPECT_EQ(config.NumResiduals(reconstruction),
             GetCeresProblem(*bundle_adjuster).NumResiduals());
@@ -324,7 +328,8 @@ TEST(DefaultBundleAdjuster, ManyViewRig) {
   std::unique_ptr<BundleAdjuster> bundle_adjuster =
       CreateDefaultCeresBundleAdjuster(options, config, reconstruction);
   const auto summary = bundle_adjuster->Solve();
-  ASSERT_NE(summary->termination_type, BATerminationType::FAILURE);
+  ASSERT_NE(summary->termination_type,
+            BundleAdjustmentTerminationType::FAILURE);
 
   EXPECT_EQ(config.NumResiduals(reconstruction),
             GetCeresProblem(*bundle_adjuster).NumResiduals());
@@ -380,7 +385,8 @@ TEST(DefaultBundleAdjuster, ManyViewRigConstantSensorFromRig) {
   std::unique_ptr<BundleAdjuster> bundle_adjuster =
       CreateDefaultCeresBundleAdjuster(options, config, reconstruction);
   const auto summary = bundle_adjuster->Solve();
-  ASSERT_NE(summary->termination_type, BATerminationType::FAILURE);
+  ASSERT_NE(summary->termination_type,
+            BundleAdjustmentTerminationType::FAILURE);
 
   EXPECT_EQ(config.NumResiduals(reconstruction),
             GetCeresProblem(*bundle_adjuster).NumResiduals());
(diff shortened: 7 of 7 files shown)


B1ueber2y:
Sorry that this PR gets a bit huge. The main idea is to do another level of abstraction on bundle adjustment to be able to add new backends. The logic is kept equivalent.

src/colmap/controllers/option_manager.cc
```diff
@@ -466,21 +456,44 @@ void OptionManager::AddBundleAdjustmentOptions() {
                    &bundle_adjustment->constant_rig_from_world_rotation);
   AddDefaultOption("BundleAdjustment.min_track_length",
                    &bundle_adjustment->min_track_length);
-  AddDefaultOption("BundleAdjustment.use_gpu", &bundle_adjustment->use_gpu);
-  AddDefaultOption("BundleAdjustment.gpu_index", &bundle_adjustment->gpu_index);
+
+  // Ceres-specific options
+  AddDefaultOption(
+      "BundleAdjustment.max_num_iterations",
```
B1ueber2y: Ideally the CLI should be ``BundleAdjustment.ceres.max_num_iterations``. I did not update it only for the reason of backwards compatibility.

src/colmap/estimators/bundle_adjustment.h
```diff
@@ -123,13 +142,11 @@ class BundleAdjustmentConfig {
   std::unordered_set<frame_t> constant_rig_from_world_poses_;
 };
 
+// Solver-agnostic bundle adjustment options.
```
ahojnnes: * Should we have some common options here like max_num_iterations?
* Loss function type and scale?

... and then equivalent to the FeatureExtractionOptions a "solver_backend" enum?

B1ueber2y: If our target is caspar, I am not sure how max_num_iterations is controlled there. And from my understanding a new loss function and a different scale would result in a new kernel in caspar, which it is still unclear whether we gonna support it. 

The backend in``FeatureExtractionOptions`` allows dynamic ``FeatureExtractor::Create``, while in our case CeresBundleAdjuster is still a virtual class with multiple extended classes (e.g., default, PosePriorBA). So I did not see how a backend parameter is useful there.

src/colmap/estimators/CMakeLists.txt
```diff
@@ -36,6 +36,7 @@ COLMAP_ADD_LIBRARY(
         affine_transform.h affine_transform.cc
         alignment.h alignment.cc
         bundle_adjustment.h bundle_adjustment.cc
+        ceres_bundle_adjustment.h ceres_bundle_adjustment.cc
```
ahojnnes: nit: suggest "bundle_adjustment_ceres.h/.cc" so the files are grouped together in the explorer (equivalent to database and database_sqlite).

src/colmap/sfm/incremental_mapper.cc
```diff
@@ -858,7 +861,8 @@ bool IncrementalMapper::RegisterNextStructureLessImage(const Options& options,
                                   abs_pose_refinement_config,
                                   *reconstruction_);
   const auto abs_pose_summary = abs_pose_refinement->Solve();
-  if (abs_pose_summary.termination_type == ceres::FAILURE) {
+  if (abs_pose_summary->termination_type ==
+      BundleAdjustmentTerminationType::FAILURE) {
```
ahojnnes: nit: it's a bit of a mouthful, maybe just `BATerminationType` ?

src/pycolmap/estimators/bundle_adjustment.cc
```diff
@@ -106,31 +151,82 @@ void BindBundleAdjuster(py::module& m) {
                              &BACfg::ConstantRigFromWorldPoses);
   MakeDataclass(PyBundleAdjustmentConfig);
 
-  using BAOpts = BundleAdjustmentOptions;
+  // Ceres-specific bundle adjustment options
+  using CeresBAOpts = CeresBundleAdjustmentOptions;
   auto PyBALossFunctionType =
```
ahojnnes: nit: PyCeresLossFunctionType ?

src/colmap/sfm/incremental_mapper.cc
```diff
@@ -1164,6 +1171,7 @@ void IncrementalMapper::IterativeLocalRefinement(
     const BundleAdjustmentOptions& ba_options,
     const IncrementalTriangulator::Options& tri_options,
     const image_t image_id) {
+  // NOLINTNEXTLINE(performance-unnecessary-copy-initialization)
```
ahojnnes: Why do we need this now?

python/examples/custom_bundle_adjustment.py
```diff
@@ -45,11 +46,11 @@ def adjust_global_bundle(
 
     # Use stricter convergence criteria for first registered images
     if len(reg_frame_ids) < 10:  # kMinNumRegImagesForFastBA = 10
-        ba_options_tmp.solver_options.function_tolerance /= 10
-        ba_options_tmp.solver_options.gradient_tolerance /= 10
-        ba_options_tmp.solver_options.parameter_tolerance /= 10
-        ba_options_tmp.solver_options.max_num_iterations *= 2
-        ba_options_tmp.solver_options.max_linear_solver_iterations = 200
+        ba_options_tmp.ceres.solver_options.function_tolerance /= 10
```
ahojnnes: nit: custom_ba_options?

src/colmap/controllers/option_manager.cc
ahojnnes: We'll eventually need to break it I think, once we add caspar as another backend, we'll likely have a "backend" enum or so and switch between the two.

src/colmap/estimators/bundle_adjustment.cc
```diff
@@ -288,993 +268,45 @@ const BundleAdjustmentConfig& BundleAdjuster::Config() const { return config_; }
...
-#endif  // CERES_VERSION_MAJOR
-  }
-
-  std::string solver_error;
-  THROW_CHECK(custom_solver_options.IsValid(&solver_error)) << solver_error;
-  return custom_solver_options;
+BundleAdjustmentOptions::BundleAdjustmentOptions()
+    : ceres(std::make_shared<CeresBundleAdjustmentOptions>()) {}
+
+BundleAdjustmentOptions::BundleAdjustmentOptions(
```
ahojnnes: I am not a fan of this, because it will be very hard to keep in sync with the options. How about we create a default copy constructor / assignment operator, make it private to prevent anybody from using it, and then create a public Clone() method.  The clone method uses the default copy operation and then patches this default copy to also create a copy of the shared_ptr?

src/colmap/estimators/bundle_adjustment.h
```diff
@@ -252,19 +202,4 @@ class BundleAdjuster {
...
-    BundleAdjustmentOptions options,
-    BundleAdjustmentConfig config,
-    Reconstruction& reconstruction);
-
-std::unique_ptr<BundleAdjuster> CreatePosePriorBundleAdjuster(
-    BundleAdjustmentOptions options,
-    PosePriorBundleAdjustmentOptions prior_options,
-    BundleAdjustmentConfig config,
-    std::vector<PosePrior> pose_priors,
-    Reconstruction& reconstruction);
```
ahojnnes, line 265: I think the constructors should stay here. Once we have Caspar as another backend, we will add a switch/case to this factory functions.

src/colmap/estimators/bundle_adjustment_test.cc
```diff
@@ -29,79 +29,10 @@
```
ahojnnes, line 29: We should probably leave some generic "interface tests" in this file which are not specific to a specific implementation and should pass with all the backends.

src/colmap/controllers/option_manager.cc
B1ueber2y: Should we break it here then?

src/colmap/estimators/bundle_adjustment_test.cc
B1ueber2y: Done.

src/colmap/controllers/option_manager.cc
B1ueber2y: I just broke it: https://github.com/colmap/colmap/pull/4042/commits/4eddc71ed1814ed590f311b640869a3019eb4faf

benchmark/runtime/bundle_adjustment.cc
```diff
@@ -51,6 +51,12 @@ void GenerateArguments(benchmark::internal::Benchmark* b) {
   }
 }
 
+inline ceres::Solver::Summary& GetCeresSummary(
+    const std::shared_ptr<BundleAdjustmentSummary>& summary) {
```
ahojnnes:
```suggestion
    const BundleAdjustmentSummary* summary) {
```

src/colmap/estimators/bundle_adjustment.cc
```diff
@@ -288,999 +268,42 @@ const BundleAdjustmentConfig& BundleAdjuster::Config() const { return config_; }
...
-    if (num_images <= max_num_images_direct_dense_solver) {
-      custom_solver_options.linear_solver_type = ceres::DENSE_SCHUR;
-    } else if (has_sparse &&
-               num_images <= max_num_images_direct_sparse_solver) {
-      custom_solver_options.linear_solver_type = ceres::SPARSE_SCHUR;
-    } else {  // Indirect sparse (preconditioned CG) solver.
-      custom_solver_options.linear_solver_type = ceres::ITERATIVE_SCHUR;
-      custom_solver_options.preconditioner_type = ceres::SCHUR_JACOBI;
-    }
+BundleAdjustmentOptions BundleAdjustmentOptions::Clone() const {
```
ahojnnes: Thanks for the change, but there was a misunderstanding. What I meant to suggest was the following, which does away with manually copying all the fields (except for the shared_ptr) and reduces the risk of forgetting to update it appropriately in the future. By making the copy operators private, the class can only be copied/cloned through the Clone() method.
```
struct TestOptions {
  TestOptions() = default;

  int field1 = 1;
  std::string field2 = "test";
  std::shared_ptr<std::string> field3 = std::make_shared<std::string>("foobar");

  TestOptions Clone() const {
    TestOptions options = *this;
    options.field3 = std::make_shared<std::string>(*field3);
    return options;
  }

 private:
  TestOptions(const TestOptions& other) = default;
  TestOptions& operator=(const TestOptions& other) = default;
};

TEST(TestOptions, Clone) {
  TestOptions options;
  options.field1 = 42;
  options.field2 = "test_42";
  *options.field3 += "test_42";
  TestOptions options_clone = options.Clone();
  EXPECT_EQ(options.field1, options_clone.field1);
  EXPECT_EQ(options.field2, options_clone.field2);
  EXPECT_NE(options.field3.get(), options_clone.field3.get());
  EXPECT_EQ(*options.field3, *options_clone.field3);
  *options.field3 += "test_42";
  EXPECT_NE(*options.field3, *options_clone.field3);
}

```

src/colmap/estimators/bundle_adjustment.h
```diff
@@ -252,19 +214,12 @@ class BundleAdjuster {
   BundleAdjustmentConfig config_;
 };
 
+// Factory function to create bundle adjusters.
+// Currently uses Ceres as the backend, but can be extended to support
+// other backends (e.g., Caspar) in the future.
 std::unique_ptr<BundleAdjuster> CreateDefaultBundleAdjuster(
```
ahojnnes, line 226: For consistency, I suggest to have the same factory for the prior bundle adjuster. The only implementation is Ceres for now (and probably we won't implement it in Caspar soon either).

The factory can also always fall back to Ceres as the default backend if no other backend is available (e.g., in case compiled without CUDA).

src/glomap/sfm/global_mapper.cc
```diff
@@ -13,6 +13,34 @@
 #include <algorithm>
 
 namespace glomap {
+
+GlobalMapperOptions GlobalMapperOptions::Clone() const {
```
ahojnnes: Same suggestion about copy operators + Clone as above.

src/pycolmap/estimators/bundle_adjustment.cc
```diff
@@ -36,6 +52,33 @@ class PyBundleAdjuster : public BundleAdjuster,
...
+  using BASummary = BundleAdjustmentSummary;
+  auto PyBundleAdjustmentSummary =
+      py::classh<BASummary>(m, "BundleAdjustmentSummary")
+          .def(py::init<>())
+          .def_readwrite("termination_type", &BASummary::termination_type)
+          .def_readwrite("num_residuals", &BASummary::num_residuals)
+          .def("is_solution_usable", &BASummary::IsSolutionUsable);
+  MakeDataclass(PyBundleAdjustmentSummary);
+
+  using CeresBASummary = CeresBundleAdjustmentSummary;
```
ahojnnes, line 76: Should we change all structs/classes to use the shorter "BA" instead of "BundleAdjustment"?
```suggestion
  using CeresBASummary = CeresBASummary;
```

src/colmap/controllers/global_pipeline.cc
```diff
@@ -39,11 +39,26 @@
 
 namespace colmap {
 
+GlobalPipelineOptions GlobalPipelineOptions::Clone() const {
```
ahojnnes: Same comment about copy / clone.

src/colmap/estimators/bundle_adjustment.h
```diff
@@ -39,13 +37,36 @@
...
+MAKE_ENUM_CLASS_OVERLOAD_STREAM(BATerminationType,
+                                -1,
+                                CONVERGENCE,
+                                NO_CONVERGENCE,
+                                FAILURE,
+                                USER_SUCCESS,
+                                USER_FAILURE);
+
+// Backend for bundle adjustment solver.
+MAKE_ENUM_CLASS_OVERLOAD_STREAM(BundleAdjustmentBackend, -1, CERES);
```
ahojnnes: Do we start from -1 or from 0? Same for other enums?

Maybe we can also let the default (-1) be "UNDEFINED" or "AUTO" and it picks the fastest backend automatically?

src/colmap/estimators/bundle_adjustment.h
```diff
@@ -39,13 +37,36 @@
...
+
+// Summary of bundle adjustment results, independent of solver backend.
+struct BundleAdjustmentSummary {
+  BATerminationType termination_type = BATerminationType::FAILURE;
+  int num_residuals = 0;
```
ahojnnes, line 68: Maybe a short comment about whether this is "residuals_reduced" or "residuals"?

src/colmap/estimators/bundle_adjustment.h
```diff
@@ -162,87 +184,27 @@ struct BundleAdjustmentOptions {
...
-
-struct PosePriorBundleAdjustmentOptions {
-  // Fallback if no prior position covariance is provided.
-  double prior_position_fallback_stddev = 1.0;
-
-  // Loss function for prior position loss.
-  BundleAdjustmentOptions::LossFunctionType prior_position_loss_function_type =
-      BundleAdjustmentOptions::LossFunctionType::TRIVIAL;
+  // Solver backend to use for bundle adjustment.
+  BundleAdjustmentBackend backend = BundleAdjustmentBackend::CERES;
```
ahojnnes, line 201: Make the default "AUTO" as suggested above?

src/colmap/estimators/bundle_adjustment_test.cc
```diff
@@ -139,7 +108,11 @@ TEST(BundleAdjustmentConfig, NumResiduals) {
   EXPECT_EQ(config.NumResiduals(reconstruction), 792);
 }
 
-TEST(DefaultBundleAdjuster, Nominal) {
+// Parameterized test for generic BundleAdjuster interface across backends.
+class BundleAdjusterBackendTest
+    : public ::testing::TestWithParam<BundleAdjustmentBackend> {};
```
ahojnnes, line 134:
```suggestion
    : public ::testing::TestWithParam<BundleAdjustmentBackend> {}
```

src/colmap/estimators/bundle_adjustment_test.cc
```diff
@@ -217,1311 +160,9 @@ TEST(DefaultBundleAdjuster, NominalMultiCameraRig) {
...
-  EXPECT_THAT(gt_reconstruction,
-              ReconstructionNear(reconstruction,
-                                 /*max_rotation_error_deg=*/0.1,
-                                 /*max_proj_center_error=*/0.1,
-                                 /*max_scale_error=*/std::nullopt,
-                                 /*num_obs_tolerance=*/0.02));
-}
+INSTANTIATE_TEST_SUITE_P(BundleAdjusterBackends,
+                         BundleAdjusterBackendTest,
+                         ::testing::Values(BundleAdjustmentBackend::CERES));
```
ahojnnes, line 186: Nice, thanks.

src/colmap/estimators/bundle_adjustment.cc
B1ueber2y: This class is used inside GlobalMapperOptions, which is used in GlobalPipelineOptions. I don't think we want to use friend class here? The copying won't work in these classes with this update, so I decided to just delete the copying for these three classes for the consistency.

src/colmap/estimators/bundle_adjustment.h
B1ueber2y: Good catch. I think we generally only start from -1 when we have UNDEFINED or INVALID. I make it start from zero instead. Lets add AUTO only when we have multiple backends.

src/colmap/estimators/bundle_adjustment.h
B1ueber2y: Done.

src/pycolmap/estimators/bundle_adjustment.cc
B1ueber2y: We have BundleAdjustOptions and BundleAdjustmentConfig binded to pycolmap. Changing all structs will incur unnecessary breaking changes. I would keep it like it is for now.

src/colmap/estimators/bundle_adjustment.h
B1ueber2y: The num_residuals was used in the colmap incremental mapper to determine the number of adjusted observations. Looking at it again, I think it should actually use the reduced num_residuals there. I fixed it here: https://github.com/colmap/colmap/pull/4042/commits/dcab4707fe17767477ab7ab6f8a39ed0714c0ab5

B1ueber2y:
> Thanks for this massive effort. I tried to look at everything but it's hard to tell if something got messed up during the copying of the code. Maybe run a benchmark on ETH3D and check for same results? Possibly even deterministically (num_threads=1, random_seed=42)?

I have tested this on two scenes of ETH3D and the results are completely equivalent.

src/colmap/estimators/bundle_adjustment.cc
```diff
@@ -288,998 +268,85 @@ const BundleAdjustmentConfig& BundleAdjuster::Config() const { return config_; }
...
-    LOG(INFO) << summary.FullReport();
+  switch (options.backend) {
+    case BundleAdjustmentBackend::CERES:
+      return CreatePosePriorCeresBundleAdjuster(options,
+                                                prior_options,
+                                                config,
+                                                std::move(pose_priors),
+                                                reconstruction);
+    default:
+      LOG(FATAL) << "Unknown bundle adjustment backend: "
```
ahojnnes: LOG(FATAL_THROW), same above

src/colmap/estimators/bundle_adjustment_ceres.cc
```diff
@@ -0,0 +1,1103 @@
...
+    case ceres::NO_CONVERGENCE:
+      return BATerminationType::NO_CONVERGENCE;
+    case ceres::FAILURE:
+      return BATerminationType::FAILURE;
+    case ceres::USER_SUCCESS:
+      return BATerminationType::USER_SUCCESS;
+    case ceres::USER_FAILURE:
+      return BATerminationType::USER_FAILURE;
+  }
+  LOG(FATAL) << "Unknown Ceres termination type: " << ceres_type;
```
ahojnnes: FATAL_THROW

src/colmap/estimators/bundle_adjustment_ceres.cc
```diff
@@ -0,0 +1,1103 @@
...
+}
+
+std::unique_ptr<ceres::LossFunction> CreateLossFunction(
+    CeresBundleAdjustmentOptions::LossFunctionType loss_function_type,
+    double loss_function_scale) {
+  switch (loss_function_type) {
+    case CeresBundleAdjustmentOptions::LossFunctionType::TRIVIAL:
+      return std::make_unique<ceres::TrivialLoss>();
+      break;
```
ahojnnes: break; not needed due to return;

src/colmap/sfm/incremental_mapper.cc
```diff
@@ -858,7 +861,7 @@ bool IncrementalMapper::RegisterNextStructureLessImage(const Options& options,
                                   abs_pose_refinement_config,
                                   *reconstruction_);
   const auto abs_pose_summary = abs_pose_refinement->Solve();
-  if (abs_pose_summary.termination_type == ceres::FAILURE) {
+  if (abs_pose_summary->termination_type == BATerminationType::FAILURE) {
```
ahojnnes: Should this be IsSolutionUsable instead?

src/colmap/sfm/incremental_mapper.cc
```diff
@@ -1130,7 +1133,8 @@ bool IncrementalMapper::AdjustGlobalBundle(
 
   // Optimize the redundant 3D points with all other parameters fixed.
   if (!is_small_reconstruction && options.ba_global_ignore_redundant_points3D) {
-    if (bundle_adjuster->Solve().termination_type == ceres::FAILURE) {
+    if (bundle_adjuster->Solve()->termination_type ==
```
ahojnnes: IsSolutionUsable?

src/colmap/sfm/incremental_mapper.cc
```diff
@@ -1154,7 +1158,8 @@ bool IncrementalMapper::AdjustGlobalBundle(
         custom_ba_options, ba_config, *reconstruction_);
   }
 
-  return bundle_adjuster->Solve().termination_type != ceres::FAILURE;
+  return bundle_adjuster->Solve()->termination_type !=
```
ahojnnes: IsSolutionUsable?

src/colmap/estimators/bundle_adjustment.cc
ahojnnes: How about this? https://github.com/B1ueber2y/colmap/pull/1

python/examples/custom_bundle_adjustment.py
```diff
@@ -5,18 +5,17 @@
...
 
 
 def solve_bundle_adjustment(
     reconstruction: pycolmap.Reconstruction,
     ba_options: pycolmap.BundleAdjustmentOptions,
     ba_config: pycolmap.BundleAdjustmentConfig,
-) -> pyceres.SolverSummary:
-    bundle_adjuster = pycolmap.create_default_bundle_adjuster(
+) -> pycolmap.CeresBundleAdjustmentSummary:
+    bundle_adjuster = pycolmap.create_default_ceres_bundle_adjuster(
```
ahojnnes: Shuld this be create_default_bundle_adjuster?

python/examples/custom_bundle_adjustment.py
```diff
@@ -80,9 +82,11 @@ def adjust_global_bundle(
     ba_config.fix_gauge(pycolmap.BundleAdjustmentGauge.THREE_POINTS)
 
     # Run bundle adjustment
-    summary = solve_bundle_adjustment(reconstruction, ba_options_tmp, ba_config)
+    summary = solve_bundle_adjustment(
+        reconstruction, custom_ba_options, ba_config
+    )
     logging.info("Global Bundle Adjustment")
-    logging.info(summary.BriefReport())
+    logging.info(summary.ceres_summary.BriefReport())
```
ahojnnes: Would it make sense to implement the BriefReport() for the abstract summary for each backend?

B1ueber2y: Done.

src/colmap/estimators/bundle_adjustment.cc
```diff
@@ -29,39 +29,21 @@
...
+
+bool BundleAdjustmentSummary::IsSolutionUsable() const {
+  return termination_type == BATerminationType::CONVERGENCE ||
+         termination_type == BATerminationType::NO_CONVERGENCE ||
+         termination_type == BATerminationType::USER_SUCCESS;
 }
 
-}  // namespace
+std::string BundleAdjustmentSummary::BriefReport() const {
+  return "termination: " +
```
ahojnnes:
```suggestion
  return "Bundle Adjustment Report - termination:" +
```

src/colmap/estimators/bundle_adjustment.h
```diff
@@ -39,13 +38,40 @@
...
+                                NO_CONVERGENCE,
+                                FAILURE,
+                                USER_SUCCESS,
+                                USER_FAILURE);
+
+// Backend for bundle adjustment solver.
+MAKE_ENUM_CLASS_OVERLOAD_STREAM(BundleAdjustmentBackend, 0, CERES);
+
+// Summary of bundle adjustment results, independent of solver backend.
+struct BundleAdjustmentSummary {
```
ahojnnes, line 63: Not super urgent directly in this PR, but I feel the following fields will be universal and useful for any backend: iterations, initial cost, final cost.

src/colmap/estimators/bundle_adjustment.h
```diff
@@ -39,13 +38,40 @@
...
 namespace colmap {
 
+struct CeresBundleAdjustmentOptions;
+struct CeresPosePriorBundleAdjustmentOptions;
+
 MAKE_ENUM_CLASS_OVERLOAD_STREAM(
     BundleAdjustmentGauge, -1, UNSPECIFIED, TWO_CAMS_FROM_WORLD, THREE_POINTS);
 
+// Termination type for bundle adjustment, independent of solver backend.
+MAKE_ENUM_CLASS_OVERLOAD_STREAM(BATerminationType,
```
ahojnnes: nit: Sorry for randomization on the naming. I originally proposed to change all the structs to use "BA" instead of "BundleAdjustment". Now this is the only struct with "BA". Even though it's long, I suggest to call it `BundleAdjustmentTerminationType` to be consistent?

src/colmap/estimators/bundle_adjustment.h
B1ueber2y: I had a look into the caspar implementation. Unfortunately, for now, the only value that is returned from Caspar::GraphSolver::Solve is the final cost, as a float value. So before they support more information from the interface, we cannot get iterations and initial cost for now. 

cc: @tordnat @emil-martens-skydio

ahojnnes: OK, let's leave it as is for now. Thanks.

emil-martens-skydio: You are right. Caspar yields very little feedback. This is definitely something I'll add in the future, but I cannot give an ETA. I'll discuss internally if I can allocate some resources to implement this.

PR #4052, opened 2026-01-25
Add test for rig verification and fixes

ahojnnes:
* Adds basic integration test for rig verification.
* Fixes issue where we only overwrote inlier matches and not the two view geometry due to recent renaming of the database methods.
* Fixes issue where we would redundant matches for same-frame verification.

### DIFF
--- a/.github/workflows/build-ubuntu.yml
+++ b/.github/workflows/build-ubuntu.yml
@@ -235,7 +235,7 @@ jobs:
         if: ${{ matrix.config.cmakeBuildType != 'ClangTidy' }}
         run: |
           if [ "${{ matrix.config.cudaEnabled }}" == "true" ]; then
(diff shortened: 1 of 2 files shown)


src/colmap/controllers/feature_matching_test.cc
```diff
@@ -332,6 +332,45 @@ TEST(CreateGeometricVerifier, Nominal) {
   EXPECT_GE(database->ReadTwoViewGeometries().size(), 3);
 }
 
+TEST(CreateGeometricVerifier, RigVerification) {
```
B1ueber2y: I am confused here. It does not seem that this is testing the rig verification step, as all two view geometries are good here.

ahojnnes: I am not sure I understand the comment. I made the test a bit stricter to ensure that rig verification actually ran by checking that the right two-view geometry configurations are in the database.

PR #4054, opened 2026-01-26
Modularize cost functions

ahojnnes:
* Decompose cost functions into individual files grouped by topic.
* Move to cost_functions sub-folder to keep estimators folder manageable.


src/colmap/estimators/cost_functions/calibration.h
```diff
@@ -0,0 +1,201 @@
+// Copyright (c), ETH Zurich and UNC Chapel Hill.
```
B1ueber2y, line 1: This might not be only focal length, if we want to include distortion in the future.

src/colmap/estimators/cost_functions/alignment.h
```diff
@@ -0,0 +1,71 @@
+// Copyright (c), ETH Zurich and UNC Chapel Hill.
```
B1ueber2y, line 1: How about just alignment rather than point_alignment?

src/colmap/estimators/cost_functions/calibration.h
ahojnnes: Renamed to calibration.

src/colmap/estimators/cost_functions/alignment.h
ahojnnes: Done

ahojnnes:
Let's keep it as cost_functions for now, we can rename it later or introduce another level in the hierarchy with a ceres and caspar sub-folder.

PR #4075, opened 2026-02-02
Update to latest vcpkg

ahojnnes:


src/colmap/controllers/incremental_pipeline_test.cc
```diff
@@ -681,32 +661,12 @@ TEST(IncrementalPipeline, PriorBasedSfMWithRandomSeedStability) {
...
-  {
-    auto reconstruction_manager0 =
-        run_mapper(/*num_threads=*/1, /*random_seed=*/kRandomSeed);
-    auto reconstruction_manager1 =
-        run_mapper(/*num_threads=*/1, /*random_seed=*/kRandomSeed);
-    EXPECT_THAT(*reconstruction_manager0->Get(0),
-                ReconstructionEq(*reconstruction_manager1->Get(0)));
-  }
-
-  // Multi-threaded execution.
```
B1ueber2y, line 694: If we disable the multithreading test in the incremental pipeline, we might also want to do the same for global pipeline.

PR #4080, opened 2026-02-05
Associate feature descriptors with type and propagate through pipeline and storage

ahojnnes:
* Backwards compatible for old databases
* Breaking change on the pycolmap interface
* Fails visual index and matching if descriptor types are inconsistent

### DIFF
--- a/src/colmap/exe/vocab_tree.cc
+++ b/src/colmap/exe/vocab_tree.cc
@@ -58,7 +58,8 @@ FeatureDescriptorsFloat LoadRandomDatabaseDescriptors(
 
   std::vector<size_t> image_idxs;
   size_t num_descriptors = 0;
-  if (max_num_images < 0 || max_num_images >= images.size()) {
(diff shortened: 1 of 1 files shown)


src/colmap/feature/types.h
```diff
@@ -86,10 +90,43 @@ struct FeatureKeypoint {
...
+
+// Returns the descriptor dimension for the given feature extractor type.
+inline constexpr int FeatureDescriptorDim(FeatureExtractorType type) {
+  switch (type) {
+    case FeatureExtractorType::SIFT:
+      return 128;
+    case FeatureExtractorType::UNDEFINED:
+      return 0;
+  }
+  return 0;
```
sarlinpe: Or rather raise an exception?

ahojnnes: It's a constexpr function. Cannot get this to work across platforms. clang-tidy should already warn us when one of the enum's is not covered in the switch-case. I'll leave it as is.

PR #4086, opened 2026-02-07
Update project dependencies

ahojnnes:
## Summary
- Update cibuildwheel from v3.1.2 to v3.3.1
- Update ccache from 4.8.2 to 4.12.2
- Update Jimver/cuda-toolkit from v0.2.29 to v0.2.30
- Update sphinx from 7.4.7 to 9.1.0
- Update sphinx-rtd-theme from 2.0.0 to 3.1.0
- Update sphinx-toolbox from 3.8.0 to 4.1.2
- Update enlighten from 1.13.0 to 1.14.1

## Test plan
- [x] CI passes for Ubuntu builds (ccache update)
- [x] CI passes for Windows builds (cuda-toolkit update)
- [x] CI passes for pycolmap wheel builds (cibuildwheel update)
- [x] Documentation builds successfully (sphinx updates)

PR #4092, opened 2026-02-10
Fix doc ordering for sphinx==0.9.1

sarlinpe:
https://github.com/colmap/colmap/pull/4086 broke the ordering of the API documentation, as reflected at https://colmap.github.io/pycolmap/pycolmap.html (we want class members to be ordered by source, not alphabetically, such that properties are grouped together). This PR fixes it. I haven't figured out how to use the newer, non-legacy API, so leaving this for later.


ahojnnes:
Thanks, good catch. Didn't notice this.

PR #4096, opened 2026-02-12
Extract loading of random descriptors and sample randomly across images

ahojnnes:
* Extracts the loading function and adds tests
* Changes the logic by not randomly picking images but randomly picking descriptors across all images.


ahojnnes:
> This should be considered a breaking change. 

Strictly yes, in practice I don't think anybody is training their own vocabulary tree?

PR #4098, opened 2026-02-13
Add LightGlue ONNX feature matching

ahojnnes:


ahojnnes:
> Thank you! One thing that concerns me: if I understand correctly, in the current design, SIFT + ALIKED_LIGHTGLUE or ALIKED + SIFT_LIGHTGLUE will just run without a warning or error message?

No, it should crash but not in an understandable way. Will fix it. Thanks for the catch.

Phil26AT:
Wow, many thanks to you @ahojnnes for implementing this!

ahojnnes:
> Wow, many thanks to you @ahojnnes for implementing this!

Thanks for building such an impressive model 🙂

Issue #3237, opened 2025-03-27
pycolmap: How to constrain translation during camera pose estimation?


winterclincke:
When localizing cameras using pycolmap.absolute_pose_estimation, there are cases where the camera’s position is known approximately or even exactly (e.g. CCTV setups), but its orientation and intrinsics are unknown. In such scenarios, it would be useful to constrain the translation during pose estimation — either by fixing it completely or by incorporating it as a prior during RANSAC and refinement.

Currently, it seems that pycolmap doesn’t provide a direct way to enforce such constraints. I’m wondering if there's a recommended way to integrate this kind of spatial prior. Would it be feasible to pass an initial pose or a fixed translation into the estimation or refinement steps?

The use case is camera localization in a stadium, where camera positions are known, but intrinsics and orientation vary per image. We’d like to leverage this positional information to improve stability and accuracy in pose estimation.

Thanks in advance for any suggestions or pointers!

ahojnnes:
This is currently not possible out of the box. The pose_prior_mapper can take advantage of position priors and may be helpful for your use case, but it is not directly usable from pycolmap and you would have to wrap the commandline executable using e.g. subprocess.call.

ahojnnes:
Closing: pose_prior_mapper suggested as workaround via subprocess.

PR #4107, opened 2026-02-17
Add camera-center prior to absolute pose refinement

winterclincke:
Adds an optional camera-center (world position) prior during absolute pose non-linear refinement (`RefineAbsolutePose` / PnP refinement).

- New refinement option to enable a camera-center prior (position in world coordinates)
- Adds a Ceres residual using the existing position prior cost functor
- Exposed through pycolmap
- Unit test added

Refs #3237

### DIFF
--- a/src/colmap/controllers/feature_matching_utils.cc
+++ b/src/colmap/controllers/feature_matching_utils.cc
@@ -166,8 +166,8 @@ namespace {
 
 class VerifierWorker : public Thread {
  public:
-  typedef FeatureMatcherData Input;
-  typedef FeatureMatcherData Output;
+  using Input = FeatureMatcherData;
+  using Output = FeatureMatcherData;
 
   VerifierWorker(const TwoViewGeometryOptions& options,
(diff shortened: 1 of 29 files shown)


src/colmap/estimators/pose.cc
```diff
@@ -190,40 +192,54 @@ bool RefineAbsolutePose(const AbsolutePoseRefinementOptions& options,
...
-                                   focal_length_idxs.end());
-      }
 
-      if (!options.refine_extra_params) {
-        const span<const size_t> extra_params_idxs = camera->ExtraParamsIdxs();
-        camera_params_const.insert(camera_params_const.end(),
-                                   extra_params_idxs.begin(),
-                                   extra_params_idxs.end());
-      }
+  if (options.use_position_prior && options.position_prior_weight > 0.0) {
```
B1ueber2y: Instead of ScaledLoss with a weight, which strictly requires manual tuning, I would pass a covariance matrix as in the position prior based BA.

Issue #4109, opened 2026-02-18
libonnxruntime.so.1: cannot open shared object file


bashtavenko:
**Describe the bug**
colmap CLI does not open after fresh install

`colmap: error while loading shared libraries: libonnxruntime.so.1: cannot open shared object file: No such file or directory`

I guess an ONNX shared lib is missing and I don't see the apt get in the docker or in the doc

**To Reproduce**

Today (02/18/2026)

```
git pull
rm -rf build && cd build
cmake .. -GNinja -DCMAKE_CUDA_ARCHITECTURES=89
ninja
sudo ninja install
colmap gui 
```

**Environment:**
 Ubuntu 24.04

jchai01:
Same error, I had to run it like this:

```
export LD_LIBRARY_PATH=/home/<YOUR_USERNAME>/repos/colmap/build/_deps/onnxruntime-src/lib:$LD_LIBRARY_PATH
colmap gui
```

ahojnnes:
Thank you. Could you provide me with the console output of the cmake + ninja install commands?

jchai01:
I am using Ubuntu 22.

`cmake .. -GNinja -DCMAKE_CUDA_ARCHITECTURES=native`:

```
-- The C compiler identification is GNU 10.5.0
-- The CXX compiler identification is GNU 10.5.0
-- Detecting C compiler ABI info
-- Detecting C compiler ABI info - done
-- Check for working C compiler: /usr/bin/gcc-10 - skipped
-- Detecting C compile features
-- Detecting C compile features - done
-- Detecting CXX compiler ABI info
-- Detecting CXX compiler ABI info - done
-- Check for working CXX compiler: /usr/bin/g++-10 - skipped
-- Detecting CXX compile features
-- Detecting CXX compile features - done
-- Enabling LSD support
-- Found OpenMP_C: -fopenmp (found version "4.5")
-- Found OpenMP_CXX: -fopenmp (found version "4.5")
-- Found OpenMP: TRUE (found version "4.5") found components: C CXX
-- Performing Test CMAKE_HAVE_LIBC_PTHREAD
-- Performing Test CMAKE_HAVE_LIBC_PTHREAD - Success
-- Found Threads: TRUE
-- Found ZLIB: /usr/lib/x86_64-linux-gnu/libz.so (found version "1.2.11")
-- Found ZLIB: /usr/lib/x86_64-linux-gnu/libz.so (found suitable version "1.2.11", minimum required is "1.2.11")
[...]

PR #4114, opened 2026-02-19
Correctly set the rpath in CMake

sarlinpe:
On macOS, the ONNX runtime is dynamically linked as rpath and correctly copied into $CMAKE_INSTALL_PREFIX/lib, but the executable fails to find it:
```bash
$ colmap
dyld[11547]: Library not loaded: @rpath/libonnxruntime.1.24.1.dylib
  Referenced from: <A09CACEF-7AC0-306B-A18D-81290DCBB49E> $HOME/.local/bin/colmap
  Reason: no LC_RPATH's found

$ otool -L $(which colmap)
$HOME/.local/bin/colmap:
        ...
        @rpath/libonnxruntime.1.24.1.dylib (compatibility version 0.0.0, current version 1.24.1)
        ...
```
Setting CMAKE_INSTALL_RPATH seems to fix it.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -168,6 +168,17 @@ if(IS_MACOS)
     add_link_options(LINKER:-no_warn_duplicate_libraries)
 endif()
 
+# Correctly set RPATH to look up dependencies.
+if(NOT IS_MSVC)
+    if (IS_MACOS)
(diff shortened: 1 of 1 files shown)


ahojnnes:
Will this also fix this issue https://github.com/colmap/colmap/issues/4109 ?

PR #4120, opened 2026-02-20
Fix install destination for onnxruntime

ahojnnes:
Attempt to fix: https://github.com/colmap/colmap/issues/4109

### DIFF
--- a/cmake/FindDependencies.cmake
+++ b/cmake/FindDependencies.cmake
@@ -329,20 +329,19 @@ if(ONNX_ENABLED)
         set(onnxruntime_LIBRARY_DIR_HINTS ${onnxruntime_BINARY_DIR}/lib CACHE PATH "ONNX Runtime library directory hints")
         find_package(onnxruntime ${COLMAP_FIND_TYPE})
 
-        install(DIRECTORY "${onnxruntime_BINARY_DIR}/include/"
-                DESTINATION "${CMAKE_INSTALL_INCLUDEDIR}")
+        install(DIRECTORY "${onnxruntime_BINARY_DIR}/include/" TYPE INCLUDE)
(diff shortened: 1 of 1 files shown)


sarlinpe:
This is equivalent to https://github.com/colmap/colmap/pull/4115 (and I think that using `TYPE LIB` is preferred by CMake). This should fix the initial issue but causes problems for any external project since we export `share/onnxruntime/cmake/onnxruntimeTargets.cmake` packaged by ONNX and this one still points to `lib64`.

ahojnnes:
@sarlinpe I managed to fix it, I believe. I am now fixing up the cmake config instead of moving the files to lib64.

cmake/FindDependencies.cmake
```diff
@@ -298,10 +298,10 @@ if(ONNX_ENABLED)
             file(MAKE_DIRECTORY ${ONNX_INCLUDE_DIR})
             file(COPY ${onnxruntime_SOURCE_DIR}/include/ DESTINATION ${ONNX_INCLUDE_DIR}/)
         endif()
-        set(onnxruntime_LIB_DIR ${onnxruntime_BINARY_DIR}/${onnxruntime_LIB_DIR_NAME})
+        set(onnxruntime_LIB_DIR ${onnxruntime_BINARY_DIR}/lib)
```
sarlinpe, line 301: Should we be using `CMAKE_INSTALL_LIBDIR` instead of hardcoding `lib`? This is also used by `install(... TYPE LIB)`.

ahojnnes: This path is references the downloaded onnxruntime binaries and not the install destination. I'll keep all of them as hardcoded paths for consistency.

PR #4121, opened 2026-02-21
Fix missing runtime library path causing COLMAP binaries to fail without LD_LIBRARY_PATH

bashtavenko:
Fixes a runtime linking issue affecting installed COLMAP binaries on Linux where shared libraries are installed into `lib` or `lib64`.

It works on Ubuntu 24.04 - open `colmap gui` without setting `LD_LIBRARY_PATH`

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -59,6 +59,15 @@ option(FETCH_FAISS "Whether to consume faiss using FetchContent or find_package"
 option(FETCH_ONNX "Whether to consume ONNX using FetchContent or find_package" ON)
 option(ALL_SOURCE_TARGET "Whether to create a target for all source files (for Visual Studio / XCode development)" OFF)
 
+# COLMAP installs executables to bin and shared libraries to
+# lib or lib64. Some dependencies (e.g. ONNXRuntime) fails at runtime
+# unless manually set LD_LIBRARY_PATH.
+if(UNIX AND NOT APPLE)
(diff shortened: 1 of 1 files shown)


CMakeLists.txt
```diff
@@ -59,6 +59,15 @@ option(FETCH_FAISS "Whether to consume faiss using FetchContent or find_package"
 option(FETCH_ONNX "Whether to consume ONNX using FetchContent or find_package" ON)
 option(ALL_SOURCE_TARGET "Whether to create a target for all source files (for Visual Studio / XCode development)" OFF)
 
+# COLMAP installs executables to bin and shared libraries to
+# lib or lib64. Some dependencies (e.g. ONNXRuntime) fails at runtime
+# unless manually set LD_LIBRARY_PATH.
+if(UNIX AND NOT APPLE)
+    set(CMAKE_INSTALL_RPATH "$ORIGIN/../lib;$ORIGIN/../lib64")
```
sarlinpe, line 66: See similar https://github.com/colmap/colmap/pull/4114 for macOS, which already sets CMAKE_INSTALL_RPATH. We can simply update it to include both lib and lib64.

ahojnnes: I am still inclined to instead just consistently install onnx also under lib instead of lib64?

ahojnnes:
Thank you for the fix. I think this shouldn't be needed anymore with this most recent commit: https://github.com/colmap/colmap/pull/4120

Please let us know if the issue still persists on your end. Thanks.

PR #4122, opened 2026-02-21
Read image orientation from EXIF and extract upright features

sarlinpe:
Tested on http://vision.maths.lth.se/calledataset/de_guerre/de_guerre.zip

SIFT:
Metric | HEAD -- non-upright | HEAD -- upright, model 1/2 ❌ | this PR -- upright, single model ✅
-- | -- | -- | --
Registered Images | 35 | 26 | 35
Points | 44,836 | 26,845 | 38,328
Observations | 236,754 | 154,588 | 225,184
Mean Track Length | 5.280444 | 5.758540 | 5.875183
Obs per Image | 6764.400000 | 5945.692308 | 6433.828571
Reprojection Error | 0.558117px | 0.376756px | 0.391964px

ALIKED:
Metric | HEAD -- model 1/2 ❌ | HEAD -- model 2/2 ❌ | this PR -- single model ✅
-- | -- | -- | --
Registered Images | 26 | 9 | 35
Points | 5,468 | 3,536 | 6,816
Observations | 41,332 | 14,256 | 60,786
Mean Track Length | 7.558888 | 4.031674 | 8.918134
Obs per Image | 1589.692308 | 1584.000000 | 1736.742857
Reprojection Error | 0.595125px | 0.271122px | 0.545645px

</body></html>

### DIFF
--- a/src/colmap/geometry/pose_prior.cc
+++ b/src/colmap/geometry/pose_prior.cc
@@ -29,6 +29,7 @@
 
 #include "colmap/geometry/pose_prior.h"
 
+#include "colmap/math/math.h"
 #include "colmap/util/logging.h"
 
 namespace colmap {


sarlinpe:
LightGlue does seem to struggle with the rotation. In a follow-up PR, we can rotate its keypoints too.

Metric | rotated ALIKED (brute force) | rotated ALIKED (LightGlue)
-- | -- | --
Images / Frames | 35 | 35
Registered Images | 35 | 35
Points | 6,816 | 8,992
Observations | 60,786 | 63,135
Mean Track Length | 8.918134 | 7.021241
Obs per Image | 1736.742857 | 1803.857143
Mean Reprojection Error | 0.545645px | 0.659852px

src/colmap/geometry/pose_prior.cc
```diff
@@ -80,4 +81,31 @@ std::ostream& operator<<(std::ostream& stream, const PosePrior& prior) {
...
+    case 1:  // Normal
+      return Eigen::Vector3d(0, 1, 0);
+    case 3:  // Rotate 180
+      return Eigen::Vector3d(0, -1, 0);
+    case 6:  // Rotate 90 CW
+      return Eigen::Vector3d(1, 0, 0);
+    case 8:  // Rotate 270 CW
+      return Eigen::Vector3d(-1, 0, 0);
+    default:
+      LOG(FATAL) << "Unknown EXIF orientation: " << orientation;
```
B1ueber2y, line 95: The program will throw for these corner cases. Would it make sense to return std::optional<Eigen::Vector3d> for this function and skip if the orientation belongs to the other cases (2, 4, 5, 7), potentially with a warning message?

B1ueber2y (commented):
Thank you! However, the population of gravity from exif orientation is a bit dangerous and may break the global pipeline, as in the current codebase we enable use_gravity by default (https://github.com/colmap/colmap/blob/main/src/colmap/estimators/rotation_averaging.h#L55, or should we disable it?) in the global sfm pipeline. Can we avoid touching the pose prior fields? For example, maybe we can call bitmap.ExifOrientation() directly in the feature matching?

sarlinpe:
Good point, though there are some reasons for doing it this way: 1) we need to store this information in the database to reuse it later during matching, 2) rotating images based on the gravity direction is also useful when the gravity is obtained differently (eg, GeoCalib), 3) IIRC the rotation averaging paper did show that using the upright direction as proxy gravity brings improvements (otherwise we should disable it by default).

src/colmap/controllers/feature_extraction.cc
```diff
@@ -42,13 +42,14 @@
 namespace colmap {
 namespace {
 
-void ScaleKeypoints(const Bitmap& bitmap,
+void ScaleKeypoints(int bitmap_width,
+                    int bitmap_height,
                     const Camera& camera,
```
ahojnnes, line 47: In this case also camera_width and camera_height?

src/colmap/controllers/feature_extraction.cc
```diff
@@ -191,11 +192,29 @@ class FeatureExtractorThread : public Thread {
         auto& image_data = input_job.Data();
 
         if (image_data.status == ImageReader::Status::SUCCESS) {
+          const int orig_width = image_data.bitmap.Width();
+          const int orig_height = image_data.bitmap.Height();
+          int rot90 =
```
ahojnnes, line 197: const?

src/colmap/geometry/pose_prior.cc
```diff
@@ -29,6 +29,7 @@
 
 #include "colmap/geometry/pose_prior.h"
 
+#include "colmap/math/math.h"
```
ahojnnes, line 32: New include really needed here?

src/colmap/geometry/pose_prior.cc
```diff
@@ -80,4 +81,31 @@ std::ostream& operator<<(std::ostream& stream, const PosePrior& prior) {
...
+    default:
+      LOG(FATAL) << "Unknown EXIF orientation: " << orientation;
+  }
+}
+
+int ComputeRot90FromGravity(const Eigen::Vector3d& gravity) {
+  // Calculate the angle of gravity in image space, then find number of 90 deg
+  // CCW rotations needed to make the image upright (where gravity is at pi/2).
+  const double angle = std::atan2(gravity.y(), gravity.x());
+  const double halfpi = M_PI / 2.0;
```
ahojnnes, line 103: nit: constexpr double kHalfPi

src/colmap/sensor/bitmap_test.cc
```diff
@@ -355,6 +355,39 @@ TEST(Bitmap, RescaleGrey) {
   EXPECT_EQ(bitmap2.Channels(), 1);
 }
```
ahojnnes, line 357: Suggest adding a short test also for ExifOrientation.

ahojnnes (approved):
Thank you, great improvement!

Regarding the concern from Shaohui, how about the following options/approach:
(1) Disable gravity-aligned rotation averaging by default - it seems to have some convergence issues anyway right now, at least when I last talked to @lpanaf ? My intuition is it will work not well with Exif orientation gravities.
(2) Store a gravity uncertainty and set it to a large value / infinity for the ExifOrientation gravities. Only let gravity aligned rotation averaging run if gravities have gravity uncertainty < X degrees.

ahojnnes:
I didn't see the auto complete...

src/colmap/geometry/pose_prior.cc
sarlinpe: On Windows, M_PI is defined in this header.

PR #4126, opened 2026-02-22
Follow-up fixes on EXIF auto-rotation

sarlinpe:
As requested in https://github.com/colmap/colmap/pull/4122. I will not add a gravity uncertainty until the need arises for other components besides the rotation averaging.

### DIFF
--- a/doc/cli.rst
+++ b/doc/cli.rst
@@ -124,6 +124,9 @@ The available commands can be listed using the command::
           exhaustive_matcher
           feature_extractor
           feature_importer
+          geometric_verifier
+          global_mapper
+          guided_geometric_verifier
           hierarchical_mapper
           image_deleter
           image_filterer
@@ -148,11 +151,14 @@ The available commands can be listed using the command::
           pose_prior_mapper
           poisson_mesher
           project_generator
+          reconstruction_clusterer
           rig_configurator
+          rotation_averager
           sequential_matcher
           spatial_matcher
           stereo_fusion
           transitive_matcher
+          view_graph_calibrator
           vocab_tree_builder
           vocab_tree_matcher
           vocab_tree_retriever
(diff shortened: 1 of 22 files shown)


src/colmap/controllers/image_reader.cc
```diff
@@ -313,7 +313,10 @@ ImageReader::Status ImageReader::Next(Rig* rig,
 
     const std::optional<int> orientation = bitmap->ExifOrientation();
     if (orientation.has_value()) {
-      pose_prior->gravity = GravityFromExifOrientation(orientation.value());
+      const auto gravity = GravityFromExifOrientation(orientation.value());
+      if (gravity.has_value()) {
+        pose_prior->gravity = gravity.value();
+      }
```
ahojnnes, line 319: I guess we can just skip the inner if condition here, since we anyway return an optional now?

src/colmap/geometry/pose_prior.h
```diff
@@ -77,9 +77,9 @@ struct PosePrior {
 
 std::ostream& operator<<(std::ostream& stream, const PosePrior& prior);
 
-// Extract gravity vector from EXIF orientation. Returns NaN vector if not an
-// upright orientation.
-Eigen::Vector3d GravityFromExifOrientation(int orientation);
+// Extract gravity vector from EXIF orientation. Returns std::nullopt if not an
```
ahojnnes, line 81: Needs `<optional>` include now.

src/colmap/controllers/image_reader.cc
sarlinpe: `PosePrior::gravity` is not an optional, we simply treat it as invalid if any of the values is NaN.

ahojnnes: Oh, right, misread.

PR #4131, opened 2026-02-23
Add Python bindings for feature matcher

ahojnnes:


src/pycolmap/feature/matching.cc
```diff
@@ -73,5 +170,63 @@ void BindFeatureMatching(py::module& m) {
...
+          "Match features between two images. Returns FeatureMatches.")
+      .def(
+          "match_guided",
+          [](FeatureMatcher& self,
+             double max_error,
+             const FeatureMatcher::Image& image1,
+             const FeatureMatcher::Image& image2,
+             TwoViewGeometry& two_view_geometry) {
+            self.MatchGuided(max_error, image1, image2, &two_view_geometry);
+          },
```
sarlinpe, line 255: We don't need the lambda here, no?
```suggestion
          &FeatureMatcher::MatchGuided,
```

sarlinpe (approved):

PR #4132, opened 2026-02-24
Rotate LightGlue inputs based on gravity

sarlinpe:
ALIKED Rotated:
Metric | Brute-force | LightGlue (Unrotated) | LightGlue (Rotated)
-- | -- | -- | --
Images | 35 | 35 | 35
Registered Images | 35 | 35 | 35
Points | 6,816 | 8,992 | 6,555
Observations | 60,786 | 63,135 | 65,621
Mean Track Length | 8.918134 | 7.021241 | 10.010831
Obs per Image | 1736.742857 | 1803.857143 | 1874.885714
Mean Reproj. Error | 0.545645px | 0.659852px | 0.670427px

PR #4137, opened 2026-02-24
Add missing include

sarlinpe:
Fixes an error at HEAD following a race condition between https://github.com/colmap/colmap/pull/4132 and https://github.com/colmap/colmap/pull/4135.


ahojnnes:
I already fixed it in the feature matcher bindings PR because it broke the CI there already.

PR #4140, opened 2026-02-25
Parse unregistered config options with warning instead of error

StonerLing:
Enable allow_unregistered when parsing config files, unregistered options are preserved and logged as warnings instead of errors.

This is necessary because many custom options are defined within COLMAP’s controllers and are not pre-registered in the `OptionManager`. Without enabling allow_unregistered, these controller-specific parameters would cause parsing failures.

### DIFF
--- a/src/colmap/controllers/base_option_manager.cc
+++ b/src/colmap/controllers/base_option_manager.cc
@@ -221,7 +221,7 @@ bool BaseOptionManager::Read(const std::filesystem::path& path,
     std::ifstream file(path);
     THROW_CHECK_FILE_OPEN(file, path);
 
-    config::parsed_options parsed_options =
+    const config::parsed_options parsed_options =
         config::parse_config_file(file, *desc_, allow_unregistered);
     config::store(parsed_options, vmap);
 


src/colmap/controllers/base_option_manager.cc
```diff
@@ -219,7 +220,19 @@ bool BaseOptionManager::Read(const std::filesystem::path& path) {
   try {
     std::ifstream file(path);
     THROW_CHECK_FILE_OPEN(file, path);
-    config::store(config::parse_config_file(file, *desc_), vmap);
+
+    config::parsed_options parsed_options =
```
ahojnnes: nit: const

src/colmap/controllers/base_option_manager.cc
```diff
@@ -219,7 +220,19 @@ bool BaseOptionManager::Read(const std::filesystem::path& path) {
...
+        config::parse_config_file(file, *desc_, allow_unregistered);
+    config::store(parsed_options, vmap);
+
+    if (allow_unregistered) {
+      for (const auto& option : parsed_options.options) {
+        if (option.unregistered) {
+          LOG(WARNING) << "Unrecognized config option: " << option.string_key;
```
ahojnnes: nit:
```suggestion
          LOG(WARNING) << "Unrecognized option key: " << option.string_key;
```

PR #4141, opened 2026-02-25
Revamp logging configuration and fix UI log sink formatting

StonerLing:
- Refactor logging options in BaseOptionManager: introduced Logging.log_target/path/level/severity/color.
- Fix text copy logic in src/colmap/ui/log_widget.cc GlogSink to correctly place message after severity prefix.
- Fix scripts/format/c++.sh: avoid "Argument list too long".

### DIFF
--- a/src/colmap/controllers/base_option_manager.cc
+++ b/src/colmap/controllers/base_option_manager.cc
@@ -263,7 +263,8 @@ bool BaseOptionManager::Parse(const int argc, char** argv) {
   return true;
 }
 
-bool BaseOptionManager::Read(const std::filesystem::path& path) {
+bool BaseOptionManager::Read(const std::filesystem::path& path,
+                             bool allow_unregistered) {
   config::variables_map vmap;
 
   if (!ExistsFile(path)) {
@@ -274,7 +275,19 @@ bool BaseOptionManager::Read(const std::filesystem::path& path) {
   try {
     std::ifstream file(path);
     THROW_CHECK_FILE_OPEN(file, path);
-    config::store(config::parse_config_file(file, *desc_), vmap);
+
+    const config::parsed_options parsed_options =
+        config::parse_config_file(file, *desc_, allow_unregistered);
+    config::store(parsed_options, vmap);
+
+    if (allow_unregistered) {
+      for (const auto& option : parsed_options.options) {
+        if (option.unregistered) {
+          LOG(WARNING) << "Unrecognized option key: " << option.string_key;
+        }
+      }
+    }
+
     vmap.notify();
     ApplyEnumConversions();
   } catch (std::exception& e) {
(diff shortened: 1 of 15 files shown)


src/colmap/controllers/base_option_manager.cc
```diff
@@ -155,6 +164,32 @@ void BaseOptionManager::ApplyEnumConversions() {
   }
 }
 
+void BaseOptionManager::ApplyLogFlags() {
+  FLAGS_logtostderr = false;
+  FLAGS_logtostdout = false;
+  FLAGS_alsologtostderr = false;
```
ahojnnes, line 179: Should we set this to true by default in ResetImpl() ? Probably set to false in the GUI or not sure what the log widget would do in this case otherwise?

ahojnnes (approved):
Thank you, LGTM. One small suggestion.

src/colmap/controllers/base_option_manager.cc
StonerLing: I handle the logging reset in the latest commit.

src/colmap/controllers/base_option_manager.cc
```diff
@@ -94,14 +103,21 @@ void BaseOptionManager::AddImageOptions() {
...
 
 void BaseOptionManager::ResetOptions(const bool reset_paths) {
   ResetOptionsImpl(reset_paths);
 }
 
-void BaseOptionManager::ResetImpl() {
-  FLAGS_logtostderr = true;
+void BaseOptionManager::ResetImpl(bool reset_logging) {
+  if (reset_logging) {
+    log_target_ = "stderr_and_file";
```
StonerLing, line 113: If we want to keep the previous behavior, this should be "stderr", but that would be inconsistent with the constructor default.

src/colmap/controllers/base_option_manager.cc
```diff
@@ -72,8 +72,17 @@ void BaseOptionManager::AddLogOptions() {
   }
   added_log_options_ = true;
 
-  AddDefaultOption("log_to_stderr", &FLAGS_logtostderr);
-  AddDefaultOption("log_level", &FLAGS_v);
+  AddDefaultOption("Logging.log_target",
```
ahojnnes: This breaks backwards compatibility. I suggest to remove the "Logging." prefix?

src/colmap/controllers/base_option_manager.cc
ahojnnes: I think it's good to log to stderr and file by default. That's a useful change.

src/colmap/controllers/base_option_manager.cc
```diff
@@ -94,14 +103,21 @@ void BaseOptionManager::AddImageOptions() {
...
 
-void BaseOptionManager::ResetImpl() {
-  FLAGS_logtostderr = true;
+void BaseOptionManager::ResetImpl(bool reset_logging) {
+  if (reset_logging) {
+    log_target_ = "stderr_and_file";
+    FLAGS_log_dir = "";
+    FLAGS_v = 0;
+    FLAGS_minloglevel = 0;
+    FLAGS_colorlogtostderr = 0;
```
ahojnnes: Is there a drawback to setting this to true by default?

src/colmap/controllers/base_option_manager.cc
StonerLing: Done

src/colmap/controllers/base_option_manager.cc
StonerLing: I don’t see a strong drawback to setting it to true by default; the previous false was mainly inherited from glog’s default, default to true now.

PR #4142, opened 2026-02-25
Parallelize structure-less registration on different images with ThreadPool

B1ueber2y:
Structure-less registration becomes the bottleneck in large-scale datasets. This PR attempts to speed it up with multi-threading.


ahojnnes:
My gut reaction is that this is a bit complicated and I'd rather have us have a parallelized implementation of RANSAC.

B1ueber2y:
Closing due to https://github.com/colmap/colmap/pull/4169.

PR #4010, opened 2026-01-14
Move view graph calibration from the global pipeline into feature matching pipeline. 

B1ueber2y:
As a result, we can rely on database cache rather than database throughout the global pipeline, and thereby align its interface with the incremental pipeline. Also, we should be able to test its effect on the incremental pipeline. 

Edit: However, regarding the incremental pipeline, guided matching appears to be quite crucial in its performance. Before heading to benchmarking the incremental pipeline with VGC, we would need to first resolve https://github.com/colmap/colmap/blob/main/src/colmap/controllers/automatic_reconstruction.cc#L106-L108. Actually, with VGC being moved to the matching stage, we can actually do a guided matching after VGC.

PR #3984, opened 2026-01-10
Decompose missing relative poses in upon loading database cache

ahojnnes:
Pending view graph calibration to operate on database cache.

PR #4143, opened 2026-02-25
Make view graph calibration a standalone module

ahojnnes:


src/colmap/estimators/view_graph_calibration_test.cc
```diff
@@ -74,6 +74,7 @@ TEST(CalibrateViewGraph, Nominal) {
     for (const size_t idx : camera.FocalLengthIdxs()) {
       camera.params[idx] += noise;
     }
+    camera.has_prior_focal_length = false;
```
ahojnnes, line 77: This should be the only behavioral change. Can this have caused the issue with the incremental pipeline? @B1ueber2y

src/colmap/controllers/global_pipeline.cc
```diff
@@ -52,19 +79,10 @@ GlobalPipeline::GlobalPipeline(
...
-    vgc_options.random_seed = options_.random_seed;
-    vgc_options.solver_options.num_threads = options_.num_threads;
-    if (!CalibrateViewGraph(vgc_options, database_.get())) {
-      LOG(ERROR) << "View graph calibration failed";
-      return;
-    }
-    LOG(INFO) << "View graph calibration done in " << run_timer.ElapsedSeconds()
-              << " seconds";
+  const bool has_insufficient_prior_focal_lengths =
+      HasInsufficientPriorFocalLengths(*database_);
```
B1ueber2y: Should this work on database_cache instead? We dont need database anymore in the global pipeline with this update, as in https://github.com/colmap/colmap/pull/4010.

src/colmap/estimators/view_graph_calibration_test.cc
B1ueber2y: Its possible. The focal length from vgc might not be as good as that from pnpf. The new guided matching would still be crucial I guess.

B1ueber2y: On the other hand, i think this variable is already false in the test so this line of change is not necessary.

src/colmap/controllers/global_pipeline.cc
ahojnnes: Done.

ahojnnes: There will also be this follow-up: https://github.com/colmap/colmap/pull/3984

PR #4149, opened 2026-02-27
relax assertion for the global bundle adjustment

lpanaf:
It should be possible to skip the global_bundle_adjustment


B1ueber2y:
What is the use case of this? With this value set to zero, retriangulation and track completion/merging still runs without global refinement, which sounds to me more of an "accidentally doesn't crash" situation than a "meaningfully supported workflow." If we really want to skip the global ba maybe its better to add a skip_global_ba option for it?

lpanaf:
> What is the use case of this? With this value set to zero, retriangulation and track completion/merging still runs without global refinement, which sounds to me more of an "accidentally doesn't crash" situation than a "meaningfully supported workflow." If we really want to skip the global ba maybe its better to add a skip_global_ba option for it?

When camera poses are not good, performing bundle adjustment actually removes a lot of useful points.

PR #4151, opened 2026-02-27
Load keypoints for all images to database cache for triangulation.

B1ueber2y:
For triangulation with known poses, the keypoints for images with zero valid two-view geometry can still have value as they may benefit structure-less registration at localizing new images.

### DIFF
--- a/src/colmap/scene/database_cache.cc
+++ b/src/colmap/scene/database_cache.cc
@@ -232,11 +232,18 @@ void DatabaseCache::Load(const Database& database, const Options& options) {
       images_.emplace(image_id, std::move(image));
     }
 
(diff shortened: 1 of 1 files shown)


ahojnnes:
I am not sure I understand how these are useful without correspondences. Do you do some on-demand matching to establish correspondences?

B1ueber2y:
There may not be matches for this triangulation, but can be useful to register a new image on top of the triangulated reconstruction, where there are useful matches (but not in this database). The current triangulation pipeline results in all images registered, but some of them may have zero keypoints in the reconstruction because of this filtering step.

B1ueber2y:
The Gemini summary is misleading there. This was not an attempt to improve the triangulation itself, but to load keypoints for images that were filtered from the start and ended up without 3D point tracks.

PR #4152, opened 2026-02-27
Add separate trial counter for structure-less registration.

B1ueber2y:
The structure-less registration does not have the mechanism of max trial counter and deprioritizing frames after failing.

### DIFF
--- a/src/colmap/sfm/incremental_mapper.cc
+++ b/src/colmap/sfm/incremental_mapper.cc
@@ -138,8 +138,8 @@ std::vector<image_t> IncrementalMapper::FindNextImages(const Options& options,
       *obs_manager_,
       filtered_frames_,
       /*num_reg_trials=*/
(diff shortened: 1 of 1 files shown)


ahojnnes (commented):
LGTM. So, this essentially means that if an image has failed already with normal image registration, we would never try it again with structureless anymore?

B1ueber2y:
> LGTM. So, this essentially means that if an image has failed already with normal image registration, we would never try it again with structureless anymore?

Without this fix, we would never try it again with structureless anymore if the image has failed three times for normal registration. On the other hand, if the image did not reach three times for normal registration, we will keep trying structureless every time until all images fail, which can be a serious performance issue. What makes it worse is that it will try always in the same order because there is no priority. 

The update in this PR should fix all these situations.

ahojnnes:
Great, thanks for the fix. Do you see a boost in recall in practice with this chance?

PR #4153, opened 2026-02-27
Fix the usage of filter_frames_ in FindNextImages.

B1ueber2y:
Bugfix.


ahojnnes:
See: https://github.com/colmap/colmap/issues/4155

PR #4154, opened 2026-02-27
Read inputs in parallel in patch match

ahojnnes:
Loading the images and especially resizing them becomes a bottleneck in the current single-threaded setup.


B1ueber2y (approved):
Would it make sense to update here? https://github.com/colmap/colmap/blob/main/src/colmap/controllers/automatic_reconstruction.cc#L106

PR #4159, opened 2026-02-28
Fix missing colmap namespace in file macro.

B1ueber2y:
Already addressed in https://github.com/colmap/colmap/pull/3770, but somehow regressed in https://github.com/colmap/colmap/pull/4058


ahojnnes:
The Gemini comments are valid. I guess we could do `const auto& path_val = path;` and then use `path_val`?

ahojnnes (approved):

PR #4160, opened 2026-02-28
Fix file THROW_CHECK_* macros from double evaluation of path expressions

ahojnnes:
If `path` was an expression, it would lead to unintuitive double evaluation, which this PR prevents.

### DIFF
--- a/src/colmap/util/file.h
+++ b/src/colmap/util/file.h
@@ -69,7 +69,7 @@
 #define THROW_CHECK_HAS_FILE_EXTENSION(path, ext)                              \
   {                                                                            \
     const auto& path_val = (path);                                             \
-    const auto& ext_val = ext;                                                 \
+    const auto& ext_val = (ext);                                               \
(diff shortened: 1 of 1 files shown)


src/colmap/util/file.h
```diff
@@ -39,28 +39,41 @@
...
-#define THROW_CHECK_HAS_FILE_EXTENSION(path, ext)                        \
-  THROW_CHECK(colmap::HasFileExtension(path, ext))                       \
-      << "Path " << (path) << " does not match file extension " << (ext) \
-      << "."
+      << ". Is the path a directory or does the parent dir not exist?";
+
+#define THROW_CHECK_HAS_FILE_EXTENSION(path, ext)                              \
+  {                                                                            \
+    const auto& path_val = (path);                                             \
+    const auto& ext_val = ext;                                                 \
```
B1ueber2y: Also do (ext) here?

PR #4167, opened 2026-03-02
Improve ONNX Runtime error handling and CUDA compatibility warnings

StonerLing:
- Emit a CMake warning for potential CUDA / ONNX Runtime GPU binary mismatches (ONNX Runtime ≥ 1.22 built with CUDA ≥ 12)
- Catch and log ONNX Runtime exception messages

### DIFF
--- a/src/colmap/feature/onnx_utils.cc
+++ b/src/colmap/feature/onnx_utils.cc
@@ -45,10 +45,9 @@ namespace colmap {
 #ifdef COLMAP_ONNX_ENABLED
 
 namespace {
-// NOLINTNEXTLINE(performance-unnecessary-value-param)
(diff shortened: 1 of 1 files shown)


src/colmap/feature/onnx_utils.cc
```diff
@@ -43,6 +44,27 @@ namespace colmap {
 
 #ifdef COLMAP_ONNX_ENABLED
 
+namespace {
+// NOLINTNEXTLINE(performance-unnecessary-value-param)
```
ahojnnes: We can probably do away with the eptr argument altogether and just use std::current_exception() in the function instead of passing it?

StonerLing: Good point

PR #4169, opened 2026-03-02
Add multi-threading support for (LO)RANSAC loop

ahojnnes:
Parallel LORANSAC Benchmark — South Building Dataset (Serial (1 thread) vs Parallel (4 threads))
```
  ┌────────────────────┬───────┬────────────┬──────────────┬─────────┬────────────┬───────┬──────────────┬────────────────┐
  │     Estimator      │ Calls │ Serial Avg │ Parallel Avg │ Speedup │ Avg Trials │ Avg n │ Serial Total │ Parallel Total │
  ├────────────────────┼───────┼────────────┼──────────────┼─────────┼────────────┼───────┼──────────────┼────────────────┤
  │ StructureLess_GR6P │   126 │  159.78 ms │     47.98 ms │   3.33x │        101 │  6684 │      20.133s │         6.046s │
  ├────────────────────┼───────┼────────────┼──────────────┼─────────┼────────────┼───────┼──────────────┼────────────────┤
  │ P3P                │   126 │    2.80 ms │      1.07 ms │   2.63x │        102 │  1593 │       0.353s │         0.134s │
  ├────────────────────┼───────┼────────────┼──────────────┼─────────┼────────────┼───────┼──────────────┼────────────────┤
  │ Essential          │     7 │    2.40 ms │      1.45 ms │   1.66x │         31 │  2486 │       0.017s │         0.029s │
  ├────────────────────┼───────┼────────────┼──────────────┼─────────┼────────────┼───────┼──────────────┼────────────────┤
  │ Homography         │     7 │    1.93 ms │      1.28 ms │   1.51x │        228 │  2486 │       0.014s │         0.026s │
  ├────────────────────┼───────┼────────────┼──────────────┼─────────┼────────────┼───────┼──────────────┼────────────────┤
  │ Fundamental        │     7 │    0.86 ms │      1.09 ms │   0.79x │         31 │  2486 │       0.006s │         0.022s │
  └────────────────────┴───────┴────────────┴──────────────┴─────────┴────────────┴───────┴──────────────┴────────────────┘
```

Overhead of New LORANSAC at num_threads=1 vs Original
```  
  ┌───────────────────────────┬──────────┬──────────┬───────────┐
  │          Metric           │ Original │ New (1T) │ Overhead  │
  ├───────────────────────────┼──────────┼──────────┼───────────┤
  │ Total P3P RANSAC time     │ 0.3615s  │ 0.3639s  │ +0.7%     │
  ├───────────────────────────┼──────────┼──────────┼───────────┤
  │ Avg time per call         │ 2.869 ms │ 2.888 ms │ +0.019 ms │
  ├───────────────────────────┼──────────┼──────────┼───────────┤
  │ Median per-call diff      │ —        │ —        │ -0.001 ms │
  ├───────────────────────────┼──────────┼──────────┼───────────┤
  │ Calls where new is faster │ —        │ —        │ 64 / 126  │
  ├───────────────────────────┼──────────┼──────────┼───────────┤
  │ Calls where new is slower │ —        │ —        │ 61 / 126  │
  └───────────────────────────┴──────────┴──────────┴───────────┘
```

### DIFF
--- a/src/pycolmap/optim/bindings.cc
+++ b/src/pycolmap/optim/bindings.cc
@@ -27,6 +27,7 @@ void BindOptim(py::module& m) {
           .def_readwrite("min_num_trials", &RANSACOptions::min_num_trials)
           .def_readwrite("max_num_trials", &RANSACOptions::max_num_trials)
           .def_readwrite("random_seed", &RANSACOptions::random_seed)
+          .def_readwrite("num_threads", &RANSACOptions::num_threads)
           .def("check", &RANSACOptions::Check);
   MakeDataclass(PyRANSACOptions);
 }


B1ueber2y:
Amazing. Thanks for the exciting update! One thing we have to be careful though: we need to have a mechanism to avoid cascaded spawning. I assume this might happen to VGC and feature matching?

ahojnnes:
> Amazing. Thanks for the exciting update! One thing we have to be careful though: we need to have a mechanism to avoid cascaded spawning. I assume this might happen to VGC and feature matching?

The options default to a single thread, so we need to explicitly enable multi-threading for specific RANSAC calls. So, I don't see a risk currently, but maybe I misunderstand your concern.

B1ueber2y:
> > Amazing. Thanks for the exciting update! One thing we have to be careful though: we need to have a mechanism to avoid cascaded spawning. I assume this might happen to VGC and feature matching?
> 
> The options default to a single thread, so we need to explicitly enable multi-threading for specific RANSAC calls. So, I don't see a risk currently, but maybe I misunderstand your concern.

Thanks for the clarification. This makes sense to me. Maybe we should do explicit Check() on the ransac options to make sure to use single thread for ransac in these modules where there is already multi-processing in the top level of ransac.

PR #4173, opened 2026-03-03
Improve test coverage for coordinate frame estimators

ahojnnes:


src/colmap/estimators/coordinate_frame_test.cc
```diff
@@ -175,5 +217,52 @@ TEST(CoordinateFrame, AlignToENUPlane) {
...
+  }
+
+  // Apply a non-unit scale to simulate a scaled model
+  const double model_scale = 2.0;
+  Sim3d pre_scale(
+      model_scale, Eigen::Quaterniond::Identity(), Eigen::Vector3d::Zero());
+  reconstruction.Transform(pre_scale);
+
+  // Record scaled distances for comparison
+  std::vector<double> scaled_dists;
```
lpanaf: This scaled_dists does not seem to be used?

ahojnnes: Thanks, good catch.

PR #4176, opened 2026-03-03
Improve reconstruction test coverage

ahojnnes:


src/colmap/scene/reconstruction_test.cc
```diff
@@ -1098,5 +1100,237 @@ TEST(Reconstruction, IsValid) {
...
+  reconstruction.DeRegisterFrame(1);
+  EXPECT_EQ(reconstruction.NumRegFrames(), 2);
+  EXPECT_EQ(reconstruction.NumRegImages(), 2);
+  EXPECT_TRUE(reconstruction.ExistsFrame(1));
+  EXPECT_FALSE(reconstruction.Frame(1).HasPose());
+
+  // De-registering an already de-registered frame is a no-op (with warning)
+  reconstruction.DeRegisterFrame(1);
+  EXPECT_EQ(reconstruction.NumRegFrames(), 2);
+}
```
lpanaf, line 1127: Maybe add the check that the 3d point is removed?

src/colmap/scene/reconstruction_test.cc
```diff
@@ -1098,5 +1100,237 @@ TEST(Reconstruction, IsValid) {
...
+  // De-register one frame to create an unregistered frame
+  const auto reg_frames_before = reconstruction.NumRegFrames();
+  const auto frame_ids = reconstruction.RegFrameIds();
+  reconstruction.DeRegisterFrame(frame_ids[0]);
+  EXPECT_EQ(reconstruction.NumRegFrames(), reg_frames_before - 1);
+
+  // TearDown should remove the unregistered frame and its images
+  const auto num_frames_before = reconstruction.NumFrames();
+  reconstruction.TearDown();
+  EXPECT_LT(reconstruction.NumFrames(), num_frames_before);
```
lpanaf, line 1147: maybe can also check that reconstruction.NumFrames() == reconstruction.NumRegFrames()

src/colmap/scene/reconstruction_test.cc
```diff
@@ -1098,5 +1100,237 @@ TEST(Reconstruction, IsValid) {
...
+    loaded.Read(test_dir);
+
+    EXPECT_THAT(loaded, ReconstructionEq(reconstruction));
+    ExpectValidPtrs(loaded);
+  }
+
+  // Write text and verify Read auto-detects text format
+  {
+    const auto test_dir = CreateTestDir();
+    reconstruction.WriteText(test_dir);
```
lpanaf, line 1317: It seems that they would write to the same folder, then the read operation would choose to read the binary file.

ahojnnes: The code already uses separate CreateTestDir() calls for each block.

PR #4177, opened 2026-03-03
Add tests for MVS model

ahojnnes:


src/colmap/mvs/model_test.cc
```diff
@@ -0,0 +1,192 @@
...
+  const float T2[3] = {1, 0, 0};
+  model.images.emplace_back("img0.jpg", 100, 100, K, R, T1);
+  model.images.emplace_back("img1.jpg", 100, 100, K, R, T2);
+
+  model.points.emplace_back(Model::Point{0.5f, 0.0f, 10.0f, {0, 1}});
+
+  const std::vector<std::map<int, float>> angles =
+      model.ComputeTriangulationAngles(50);
+  EXPECT_EQ(angles.size(), 2);
+  EXPECT_GT(angles[0].at(1), 0.0f);
```
lpanaf: maybe can replace with the actual value: arctan(0.5 / 10) * 2?

src/colmap/mvs/model_test.cc
lpanaf, line 1: Is it that PMVs would not be supported, or that the test for the relevant functions would be added later?

PR #4184, opened 2026-03-03
Add claude settings template for easy setup of new machine

ahojnnes:
Reduce permission noise by allowlisting mostly benign commands without having to specify `--dangerously-skip-permissions`. Just a template that everyone can adjust and modify into settings.local.json to become effective.


B1ueber2y:
Personally I think this may be personal preference varying from developers and should not be added the repo. We also do not want to limit to claude (although I am a heavy user myself), given the fast pace of agentic coding developments. 

Setting up of new machine would be quite straightforward and if really needed we can provide a download link for this script. Are there any other use cases in your mind?

ahojnnes:
No, I am just copying this file on every new machine I allocate. It becomes cumbersome.

PR #4186, opened 2026-03-04
Add more log options in GUI

StonerLing:
- Extends the Log Options dialog
- Adds color rendering in LogWidget


src/colmap/ui/log_widget.cc
```diff
@@ -73,7 +73,7 @@ class GlogSink : public google::LogSink {
       default:
         break;
     }
-    log_widget_->Append(text);
+    log_widget_->Append(severity, text);
```
ahojnnes: Not your change, but we could avoid the copy here by moving the text and changing the interface to
```
void LogWidget::Append(google::LogSeverity severity, std::string text)
```
as well as moving the text into the text_queue_.

StonerLing: Done

PR #4205, opened 2026-03-08
Homogenize undistorter options

ahojnnes:


src/colmap/controllers/undistorters.h
```diff
@@ -83,7 +87,18 @@ class COLMAPUndistorter : public BaseController {
 // Undistort images and prepare data for CMVS/PMVS.
 class PMVSUndistorter : public BaseController {
  public:
-  PMVSUndistorter(const UndistortCameraOptions& camera_options,
+  struct Options {
```
B1ueber2y, line 90: These two fields seem to be relevant to all undistortion option classes. Wondering if it makes sense to abstract a base to consolidate.

ahojnnes: I don't think it's worth deduplicating these fields vs. the loss in discoverability of the parameters. It's easier for humans (and AI) to just have all the parameters declared next to each other.

PR #4206, opened 2026-03-08
Add pose graph synthesizer support and apply to relevant tests.

B1ueber2y:
Until now we had to synthesize database with features and a full reconstruction with 3D points, which is not relevant to initializing a pose graph. This PR adds a stand-alone pose graph synthesizer and applies it to the tests of rotation averaging and gravity refinement.

### DIFF
--- a/src/colmap/scene/synthetic.cc
+++ b/src/colmap/scene/synthetic.cc
@@ -42,6 +42,14 @@
 namespace colmap {
 namespace {
 
+Eigen::Vector3d RandomPerpendicularAxis(const Eigen::Vector3d& v) {
+  Eigen::Vector3d axis;
+  do {
+    axis = v.cross(Eigen::Vector3d::Random());
+  } while (axis.squaredNorm() < 1e-12);
+  return axis.normalized();
+}
+
 void AddOutlierMatches(double inlier_ratio,
                        int num_points2D1,
                        int num_points2D2,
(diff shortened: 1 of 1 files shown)


src/colmap/scene/synthetic.h
```diff
@@ -130,4 +136,30 @@ void SynthesizeImages(const SyntheticImageOptions& options,
                       const Reconstruction& reconstruction,
                       const std::filesystem::path& image_path);
 
+using SyntheticPoseGraphOptions = BaseSyntheticOptions;
+
+struct SyntheticPoseGraphData {
```
ahojnnes, line 141: Why decompose the existing SynthesizeDataset into a separate pose graph synthesizer and not instead always synthesize the pose graph or make it optional?

The SynthesizeDataset could return a SyntheticDataset struct to make it more easily manageable?

src/colmap/scene/synthetic.h
```diff
@@ -30,23 +30,20 @@
...
+#include "colmap/scene/pose_graph.h"
 #include "colmap/scene/reconstruction.h"
 #include "colmap/sensor/models.h"
 
 #include <filesystem>
 
 namespace colmap {
 
-struct SyntheticDatasetOptions {
+struct BaseSyntheticOptions {
```
ahojnnes, line 43: I'd avoid inheritance for options structs.

src/colmap/scene/synthetic.cc
```diff
@@ -840,4 +876,187 @@ void SynthesizeImages(const SyntheticImageOptions& options,
...
+  int total_num_images = 0;
+
+  for (int rig_idx = 0; rig_idx < options.num_rigs; ++rig_idx) {
+    const auto [rig_id, camera_sensor_ids] =
+        SynthesizeCamerasAndRig(options,
+                                rig_idx,
+                                /*has_prior_focal_length=*/false,
+                                &data.reconstruction);
+
+    for (int frame_idx = 0; frame_idx < options.num_frames_per_rig;
```
ahojnnes, line 905: This is mostly redundant to the SynthesizeDataset?

B1ueber2y:
> Thank you, this is useful functionality to have. However, I suggest integrating this into the existing SynthesizeDataset function. We can just always synthesize it or add a boolean flag, if we want to avoid the overhead. I like the idea of bundling the Data into a SyntheticDataset struct instead of the current input/output arguments.

Good point. This will however incur changes on the interfaces and therefore lead to updates on a large number of files. Closing this and will reopen another PR with the new design.

PR #4207, opened 2026-03-08
Add support for importing surface mesh in viewer

ahojnnes:
Added simple mesh viewing functionality with optional wire-frame mode.

Poisson meshing results on south-building:

mesh

Wire-frame mode:
mesh-wireframe


src/colmap/ui/main_window.cc
```diff
@@ -1073,7 +1118,7 @@ void MainWindow::ExportAll() {
           .toUtf8()
           .constData();
 
-  // Selection canceled?
+  // Selection cancelled?
```
B1ueber2y: nit: the comment style is not consistent in this file, some are outside the if and some are inside.

PR #4208, opened 2026-03-08
Add support for synthesizing pose graph and apply to relevant tests.

B1ueber2y:
Rework on https://github.com/colmap/colmap/pull/4206

### DIFF
--- a/benchmark/reconstruction/evaluation/utils_test.py
+++ b/benchmark/reconstruction/evaluation/utils_test.py
@@ -362,7 +362,9 @@ def create_test_reconstruction():
     synthetic_dataset_options.num_cameras_per_rig = 1
     synthetic_dataset_options.num_frames_per_rig = 5
     synthetic_dataset_options.num_points3D = 0
-    return pycolmap.synthesize_dataset(synthetic_dataset_options)
+    gt_dataset = pycolmap.SyntheticDataset()
+    pycolmap.synthesize_dataset(synthetic_dataset_options, gt_dataset)
+    return gt_dataset.reconstruction
 
 
 class TestComputeAbsErrors:


src/colmap/scene/synthetic.h
```diff
@@ -95,26 +97,59 @@ struct SyntheticDatasetOptions {
...
+  Reconstruction reconstruction;
+  PoseGraph pose_graph;
+  std::vector<PosePrior> pose_priors;
+};
+
+// Synthesize a full dataset including reconstruction, pose graph, and pose
+// priors. Optionally populates the database. Appends to existing data.
+void SynthesizeDataset(const SyntheticDatasetOptions& options,
+                       SyntheticDataset* dataset,
+                       Database* database = nullptr);
```
ahojnnes: Should we add this to the dataset and control it via a bool in the options?

src/colmap/scene/synthetic.h
```diff
@@ -95,26 +97,59 @@ struct SyntheticDatasetOptions {
...
+struct SyntheticDataset {
+  Reconstruction reconstruction;
+  PoseGraph pose_graph;
+  std::vector<PosePrior> pose_priors;
+};
+
+// Synthesize a full dataset including reconstruction, pose graph, and pose
+// priors. Optionally populates the database. Appends to existing data.
+void SynthesizeDataset(const SyntheticDatasetOptions& options,
+                       SyntheticDataset* dataset,
```
ahojnnes: Return the dataset instead of output argument?

src/colmap/scene/synthetic.h
```diff
@@ -95,26 +97,59 @@ struct SyntheticDatasetOptions {
...
+                       Database* database = nullptr);
+// Convenience overload that only populates a reconstruction and optionally the
+// database, without producing a pose graph or pose priors.
 void SynthesizeDataset(const SyntheticDatasetOptions& options,
```
ahojnnes, line 120: Should we remove this in a follow-up and control all the optional outputs via options?

src/colmap/scene/synthetic.cc
```diff
@@ -716,9 +778,59 @@ void SynthesizeNoise(const SyntheticNoiseOptions& options,
...
-    for (auto& pose_prior : database->ReadAllPosePriors()) {
+  reconstruction->UpdatePoint3DErrors();
+}
+
+void SynthesizePoseGraphNoise(const PoseGraphNoiseOptions& options,
+                              PoseGraph* pose_graph) {
+  THROW_CHECK_GE(options.rel_rotation_noise_deg, 0.);
+  THROW_CHECK_GE(options.rel_translation_noise_deg, 0.);
+
+  if (options.rel_rotation_noise_deg > 0.0 ||
```
ahojnnes: nit: return if both are zero instead to reduce indentation?

src/colmap/scene/synthetic.cc
```diff
@@ -716,9 +778,59 @@ void SynthesizeNoise(const SyntheticNoiseOptions& options,
...
+        const double angle = std::clamp(
+            RandomGaussian<double>(0, options.rel_rotation_noise_deg),
+            -180.0,
+            180.0);
+        edge.cam2_from_cam1.rotation() *= Eigen::Quaterniond(Eigen::AngleAxisd(
+            DegToRad(angle), Eigen::Vector3d::Random().normalized()));
+      }
+      if (options.rel_translation_noise_deg > 0.0) {
+        const Eigen::Vector3d t = edge.cam2_from_cam1.translation();
+        if (t.norm() > 1e-10) {
```
ahojnnes: nit: squaredNorm

src/colmap/scene/synthetic.h
```diff
@@ -95,26 +97,59 @@ struct SyntheticDatasetOptions {
...
+// Add Gaussian noise to frame poses, 2D/3D points in the reconstruction.
+// Optionally updates keypoints in the database.
+void SynthesizeReconstructionNoise(const ReconstructionNoiseOptions& options,
+                                   Reconstruction* reconstruction,
+                                   Database* database = nullptr);
+// Add noise to relative poses in pose graph edges.
+void SynthesizePoseGraphNoise(const PoseGraphNoiseOptions& options,
+                              PoseGraph* pose_graph);
+// Add noise to position/gravity pose priors. At least one of pose_priors or
+// database must be non-null. If pose_priors is null, priors are read from the
```
ahojnnes, line 161: The automatic loading of pose_priors feels a bit magical. Suggest to just always require the pose_priors?

src/colmap/scene/synthetic.h
```diff
@@ -95,26 +97,59 @@ struct SyntheticDatasetOptions {
...
-void SynthesizeNoise(const SyntheticNoiseOptions& options,
-                     Reconstruction* reconstruction,
-                     Database* database = nullptr);
+// Add Gaussian noise to frame poses, 2D/3D points in the reconstruction.
+// Optionally updates keypoints in the database.
+void SynthesizeReconstructionNoise(const ReconstructionNoiseOptions& options,
+                                   Reconstruction* reconstruction,
+                                   Database* database = nullptr);
+// Add noise to relative poses in pose graph edges.
+void SynthesizePoseGraphNoise(const PoseGraphNoiseOptions& options,
```
ahojnnes, line 155: For consistency, should this also update the two-view geometries in the database?

src/colmap/scene/synthetic.h
B1ueber2y: I originally also went for the returning. However, returning the dataset will break the support of synthesizing twice, which was covered by a test case in synthetic_test.cc. I thus switched back to using output argument.

src/colmap/scene/synthetic.h
B1ueber2y: Sure we can remove it. There are >30 call sites for this that will require updates though. I kept it for this PR to minimize the diff.

ahojnnes: Yes, definitely not in this PR. :-)

src/colmap/scene/synthetic.h
B1ueber2y: With the legacy interface (in the other comment) on synthesizing reconstruction and database, always requiring pose priors will result in reading pose priors out of the database and updating twice, which is not ideal. I can change it to required argument in a follow-up PR along with dropping the legacy interface.

B1ueber2y:
> Thanks for the changes, looks mostly good apart from some minor comments. Overall, I am still a bit torn on decomposing the noise synthesis into different functions instead of just keeping the existing SynthesizeDatasetNoise. For example, synthesizing keypoint noise doesn't really fit into the "ReconstructionNoise"?

Similar to the other comments, keeping those separate functions is nice with the existing legacy interface. I also dont see much drawback just keeping them, as it only improves flexibility.

src/colmap/scene/synthetic.h
ahojnnes: I think this complication goes away if we stay with single SynthesizeDataset and SynthesizeDatasetNoise functions instead of bifurcating into different functions and redundant overloads? It looks unnecessarily complicated now for relatively little benefit.

B1ueber2y: I agree with your point. However, there was no concept of SyntheticDatabase in the legacy interface of synthesizing reconstruction and database directly. So resolving this complication will also happen after we drop the legacy interface in the follow-up PR.

src/colmap/scene/synthetic.h
B1ueber2y: One thing about this though: if the database lives in the dataset, it will remain opened throughout the life of the ``SyntheticDataset``, which is not ideal. Do we want to revert this change? @ahojnnes

Issue #239, opened 2017-10-16
"dense_stereo" is not being compiled


biesseck:
I've compiled COLMAP source code (downloaded from https://github.com/colmap/colmap/archive/master.zip) on Ubuntu but **dense_stereo** program is not being compiled into **colmap/build/src/exe** folder. The compilation process works without error untill the end, **dense_fuser** and **dense_mesher** are generated normally but **dense_stereo** not. I found **colmap/src/exe/dense_stereo.cc** file but **make** does not knows **dense_stereo** argument. Here are the steps I used:

```
wget -c https://github.com/colmap/colmap/archive/master.zip
unzip master.zip
mv colmap-master colmap
mkdir colmap/build
cd colmap/build
cmake ..
make
```

CMAKE OUTPUT:
```
-- The C compiler identification is GNU 4.8.4
-- The CXX compiler identification is GNU 4.8.4
-- Check for working C compiler: /usr/bin/cc
-- Check for working C compiler: /usr/bin/cc -- works
-- Detecting C compiler ABI info
-- Detecting C compiler ABI info - done
-- Check for working CXX compiler: /usr/bin/c++
-- Check for working CXX compiler: /usr/bin/c++ -- works
-- Detecting CXX compiler ABI info
-- Detecting CXX compiler ABI info - done
-- Performing Test HAS_SSE_EXTENSION
-- Performing Test HAS_SSE_EXTENSION - Success
-- Performing Test HAS_SSE2_EXTENSION
-- Performing Test HAS_SSE2_EXTENSION - Success
-- Performing Test HAS_SSE3_EXTENSION
-- Performing Test HAS_SSE3_EXTENSION - Success
-- Performing Test HAS_SSE41_EXTENSION
-- Performing Test HAS_SSE41_EXTENSION - Success
-- Performing Test HAS_SSE42_EXTENSION
-- Performing Test HAS_SSE42_EXTENSION - Success
-- Performing Test HAS_AVX_EXTENSION
-- Performing Test HAS_AVX_EXTENSION - Success
-- Try OpenMP C flag = [-fopenmp]
-- Performing Test OpenMP_FLAG_DETECTED
-- Performing Test OpenMP_FLAG_DETECTED - Success
-- Try OpenMP CXX flag = [-fopenmp]
-- Performing Test OpenMP_FLAG_DETECTED
-- Performing Test OpenMP_FLAG_DETECTED - Success
-- Boost version: 1.54.0
-- Found the following Boost libraries:
--   program_options
[...]

PR #4217, opened 2026-03-09
Improved robustness to solver failures

ahojnnes:
Fixes: https://github.com/colmap/glomap/issues/239 and generally improves robustness to solver failures.

It's a risky change, as I am not certain under what circumstances Ceres returns non-usable solutions and if this unnecessarily fails the reconstruction process, if a single solver run fails due to numerical issues.

### DIFF
--- a/src/pycolmap/estimators/bundle_adjustment.cc
+++ b/src/pycolmap/estimators/bundle_adjustment.cc
@@ -83,7 +83,6 @@ void BindBundleAdjuster(py::module& m) {
           .def("is_solution_usable", &BASummary::IsSolutionUsable)
           .def("is_unrecoverable_failure", &BASummary::IsUnrecoverableFailure)
           .def("brief_report", &BASummary::BriefReport);
-  MakeDataclass(PyBundleAdjustmentSummary);
 
   using CeresBASummary = CeresBundleAdjustmentSummary;
   auto PyCeresBundleAdjustmentSummary =


src/colmap/controllers/bundle_adjustment.cc
```diff
@@ -100,7 +100,19 @@ void BundleAdjustmentController::Run() {
...
       CreateDefaultBundleAdjuster(ba_options, ba_config, *reconstruction_);
-  bundle_adjuster->Solve();
+  const auto summary = bundle_adjuster->Solve();
+  if (!summary->IsSolutionUsable()) {
+    if (summary->IsUnrecoverableFailure()) {
+      LOG(ERROR) << "Bundle adjustment failed due to unrecoverable error: "
+                 << summary->message;
+    } else {
+      LOG(WARNING)
+          << "Bundle adjustment solver failed with a recoverable error: "
```
B1ueber2y: nit: can we align these two logs. One is with "solver" but the other is without.

src/colmap/controllers/incremental_pipeline.h
```diff
@@ -217,6 +217,7 @@ class IncrementalPipeline : public BaseController {
     NO_INITIAL_PAIR,
     BAD_INITIAL_PAIR,
     UNKNOWN_SENSOR_FROM_RIG,
+    SOLVER_FAILURE,
```
B1ueber2y: I think this naming is highly misleading. Its actually UNRECOVERABLE_SOLVER_FAILURE. not sure how to make it more concise.

src/colmap/sfm/incremental_mapper.h
```diff
@@ -262,7 +263,8 @@ class IncrementalMapper {
                           const BundleAdjustmentOptions& ba_options);
 
   // Perform multiple rounds of local bundle adjustment.
-  void IterativeLocalRefinement(
+  // Returns false if the bundle adjustment solver fails.
```
B1ueber2y: This comment was wrong. It only returns false when the bundle adjustment solver fails with an unrecoverable failure.

src/colmap/sfm/incremental_mapper.h
```diff
@@ -271,7 +273,8 @@ class IncrementalMapper {
       image_t image_id);
 
   // Perform multiple rounds of global bundle adjustment.
-  void IterativeGlobalRefinement(
+  // Returns false if the bundle adjustment solver fails.
```
B1ueber2y: Same here. The comment was wrong.

src/colmap/sfm/incremental_mapper.cc
```diff
@@ -1145,8 +1157,17 @@ bool IncrementalMapper::AdjustGlobalBundle(
 
   // Optimize the redundant 3D points with all other parameters fixed.
   if (!is_small_reconstruction && options.ba_global_ignore_redundant_points3D) {
-    if (!bundle_adjuster->Solve()->IsSolutionUsable()) {
-      return false;
+    const auto redundant_summary = bundle_adjuster->Solve();
+    if (!redundant_summary->IsSolutionUsable()) {
```
B1ueber2y, line 1161: There is a logic change here for the return value of ``AdjustGlobalBundle`` for recoverable failures. Before it would return false but now it will return true. However, this return value was not used anywhere before so this change does not break any logic. Maybe worth documenting in the header though.

DuVogel87:
Hi @ahojnnes, thanks for pointing me from glomap issue #239 to this PR.

I may be misunderstanding: my main problem is `CUDSS_STATUS_ALLOC_FAILED` in global positioning (then my pipeline falls back to CPU).  
Does PR #4217 address the OOM itself (memory usage / avoiding fallback), or mainly improve failure detection/propagation?

If it is only error handling, could you suggest the recommended settings/workaround for issue #239 on a 24GB RTX 3090?

ahojnnes:
> Hi @ahojnnes, thanks for pointing me from glomap issue #239 to this PR.
> 
> I may be misunderstanding: my main problem is `CUDSS_STATUS_ALLOC_FAILED` in global positioning (then my pipeline falls back to CPU). Does PR #4217 address the OOM itself (memory usage / avoiding fallback), or mainly improve failure detection/propagation?
> 
> If it is only error handling, could you suggest the recommended settings/workaround for issue #239 on a 24GB RTX 3090?

@DuVogel87 It only addresses the failure handling in that it doesn't silently continue. It doesn't actually switch to CPU in case of insufficient memory. This is something we could do as a follow-up.

ahojnnes:
@DuVogel87 I added a fallback mechanism for the GPU use case. I appreciate, if you can test if this works for you.

ahojnnes:
Abandoning this in favor of https://github.com/colmap/colmap/pull/4230 and https://github.com/colmap/colmap/pull/4231

PR #4225, opened 2026-03-11
Fix 180-degree flipped solutions in gravity-aligned rotation averaging

ahojnnes:
Enable MST initialization for the gravity case to properly initialize yaw angles, preventing convergence to 180-degree flipped local minima. Preserve already-known sensor_from_rig values during MST initialization to avoid corrupting calibrated rig extrinsics.

Tested this with 1000 different random seeds. The existing implementation suffers from ~5/1000 flips. The new one doesn't produce any flips.

### DIFF
--- a/src/colmap/estimators/rotation_averaging.cc
+++ b/src/colmap/estimators/rotation_averaging.cc
@@ -514,11 +514,16 @@ bool InitializeRigRotationsFromImages(
   std::vector<double> weights;
   for (auto& [camera_id, rig_id_and_samples] : cam_from_rig_samples) {
     auto& [rig_id, samples] = rig_id_and_samples;
-    // Skip if sensor_from_rig is already fully known.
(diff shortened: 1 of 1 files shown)


src/colmap/estimators/rotation_averaging.cc
```diff
@@ -397,8 +397,8 @@ bool RotationEstimator::SolveRotationAveraging(
     const std::vector<PosePrior>& pose_priors,
     const std::unordered_set<image_t>& active_image_ids,
     Reconstruction& reconstruction) {
-  // Initialize rotations from maximum spanning tree if no gravity priors.
-  if (!options_.skip_initialization && !UseGravity(options_, pose_priors)) {
+  // Initialize rotations from maximum spanning tree.
```
lpanaf: maybe add a remark that gravity aligned case would have a flip without the intialization

Issue #4234, opened 2026-03-12
ALIKED feature extraction in Docker consumes unbounded RAM and swap until killed


eriksoonn:
When running `automatic_reconstructor` with `--feature aliked` inside the official Docker-based build of COLMAP `3.14.dev0` (`7b88a520b95aa5369ba5dc8bbdc3de1285293aad`), feature extraction appears to run on CPU instead of GPU and host memory usage grows continuously until the process is killed.

On my system, RAM usage grows until all 64 GB are consumed, and swap usage also keeps increasing until the full 256 GB swap file is exhausted. The process is then killed during feature extraction.

I am not sure whether this is an expected CPU fallback with very high memory use, or a real memory leak / unbounded memory growth bug. The reason I suspect a leak is that swap usage keeps increasing during feature extraction instead of stabilizing.

## To Reproduce

1. Build and run COLMAP in Docker from commit `7b88a520b95aa5369ba5dc8bbdc3de1285293aad`.
2. Run the following command:

   ```bash
   colmap automatic_reconstructor \
     --image_path image_path \
     --workspace_path workspace_path \
     --data_type video \
     --quality high \
     --single_camera 1 \
     --sparse 1 \
     --dense 0 \
     --feature aliked
   ```

3. Observe resource usage during feature extraction.
4. RAM usage keeps growing, swap usage also keeps growing, and eventually the process is killed.

## Expected behavior

I would expect one of the following:

- ALIKED feature extraction uses the available GPUs, or
- if CPU fallback is intended, memory usage should remain bounded and not grow until all RAM and swap are exhausted.

At minimum, it would be helpful to know whether ALIKED extraction is currently expected to run on CPU in Docker and whether this level of memory growth is expected.

## Screenshots

I am attaching a screenshot showing that RAM and swap become fully occupied during feature extraction.

## Environment

- OS: Docker / Linux
- COLMAP Version: `3.14.dev0`
- COLMAP Commit: `7b88a520b95aa5369ba5dc8bbdc3de1285293aad`
- Input type: video frames
- CPU: AMD Ryzen 9 7900
- RAM: 64 GB
- GPU: 2x RTX 3090
- Swap: 256 GB

## Additional context

The input consists of 300 frames at 3840x2160 resolution.

The process is killed during feature extraction, for example:

```text
I20260312 22:34:02.988494   318 feature_extraction.cc:280] Processed file [90/300]
[...]

PR #4235, opened 2026-03-13
Prevent nested threading for ONNX

ahojnnes:
Partial fix for: https://github.com/colmap/colmap/issues/4234

- Fix unbounded memory growth during ALIKED (and other ONNX-based) GPU feature
  extraction caused by each extractor thread inheriting the full
  `hardware_concurrency()` thread count for its internal ONNX session, leading
  to massive per-thread memory arenas that never shrink.
- Refactor GPU vs CPU extraction dispatch to be feature-type-aware instead of
  relying on SIFT-specific options (`domain_size_pooling`,
  `estimate_affine_shape`) for all extractor types.
- Prevent InterOp threads and explicitly set sequential ONNX execution mode.


DELL-888:
Hii🖐️

Issue #4095, opened 2026-02-11
Add CMake option to build COLMAP without MVS (SfM-only builds)


shuta-ochiai:
# Summary
I would like to request a CMake option (e.g., MVS_ENABLED=OFF) to build COLMAP without the Multi-View Stereo (MVS) module, enabling lightweight SfM-only builds.

# Background
Currently, COLMAP always builds and links the MVS library (colmap_mvs), even when only sparse reconstruction (SfM) is needed. The top-level CMakeLists.txt provides many feature flags (e.g., GUI_ENABLED, CGAL_ENABLED), but there is no option to exclude MVS . As a result:

- The colmap_controllers and colmap_exe targets always link colmap_mvs .
- Build times and binary sizes include MVS components even when unused.

# Proposed Solution
Add a new CMake option MVS_ENABLED (default ON for backward compatibility). When OFF:

- Exclude src/colmap/mvs/ from the build.
- Skip MVS-related targets (colmap_mvs, colmap_mvs_cuda if CUDA enabled).
- Omit MVS commands from the CLI and MVS bindings from PyCOLMAP.

# Expected Benefits
- Faster builds for SfM-only workflows.
- Smaller binaries and Docker images.
- Clearer separation between SfM and MVS pipelines.

PR #4252, opened 2026-03-16
Add MVS_ENABLED option to build without MVS module.

B1ueber2y:
Fixes https://github.com/colmap/colmap/issues/4095

This is actually a feature I personally also have wanted to have for a while, as this can be quite useful to reduce building overhead when colmap is used as a third party. The build with MVS_ENABLED=OFF was tested successfully locally **but currently is not protected in the CI pipeline**.


B1ueber2y:
These Gemini reviews seem to flag two pre-existing minor issues in the codebase. Fixed in a separate PR.

B1ueber2y:
Ideally, we might need to add a new matrix with MVS_ENABLED=OFF, as there is no good place in the existing workflows to set this option without dropping test coverage of other crucial features.

PR #4269, opened 2026-03-19
Upgrade to pybind11 3.X native_enum

ahojnnes:


python/CMakeLists.txt
```diff
@@ -53,7 +53,7 @@ else()
 endif()
 find_package(Python REQUIRED COMPONENTS Interpreter ${DEV_MODULE} REQUIRED)
 
-find_package(pybind11 2.13.0 REQUIRED)
+find_package(pybind11 3.0.0 REQUIRED)
```
B1ueber2y: Should we use 3.0.1 here to align with the pycolmap setup?

src/pycolmap/estimators/bundle_adjustment.cc
```diff
@@ -63,16 +64,14 @@ class PyCeresBundleAdjuster : public CeresBundleAdjuster,
...
-          m, "BundleAdjustmentTerminationType")
-          .value("CONVERGENCE", BundleAdjustmentTerminationType::CONVERGENCE)
-          .value("NO_CONVERGENCE",
-                 BundleAdjustmentTerminationType::NO_CONVERGENCE)
-          .value("FAILURE", BundleAdjustmentTerminationType::FAILURE)
-          .value("USER_SUCCESS", BundleAdjustmentTerminationType::USER_SUCCESS)
-          .value("USER_FAILURE", BundleAdjustmentTerminationType::USER_FAILURE);
-  AddStringToEnumConstructor(PyBundleAdjustmentTerminationType);
+  py::native_enum<BundleAdjustmentTerminationType>(
+      m, "BundleAdjustmentTerminationType", "enum.IntEnum")
```
sarlinpe, line 68: What's the rationale for using IntEnum? this enables arithmetic operations on enums, but this isn't relevant for any of our enums, right? We likely don't want to support `NO_CONVERGENCE + 1 == FAILURE`. Sadly we can't use StrEnum since it requires Python 3.11.

ahojnnes: Yes, arithmetic operations are not needed but then you can implicitly convert integers to enums, e.g.: `BundleAdjustmentTerminationType(1)`, which I thought would be useful?

ahojnnes: I guess, this prevents `BundleAdjustmentTerminationType("CONVERGENCE")` from working now though...

sarlinpe: I think that init from string is more useful than init from integer. The current setup also supports implicit conversion from string to enum, is this supported by native_enum? do we need to add `py::implicitly_convertible<std::string, T>();` to make it work?

ahojnnes: This now works:
```
>>> import pycolmap
>>> pycolmap.BundleAdjustmentTerminationType("CONVERGENCE")
<BundleAdjustmentTerminationType.CONVERGENCE: 0>
>>> pycolmap.BundleAdjustmentTerminationType(0)
<BundleAdjustmentTerminationType.CONVERGENCE: 0>
```

PR #4272, opened 2026-03-20
Update pybind11 to 3.0.2

ahojnnes:


src/pycolmap/estimators/alignment.cc
```diff
@@ -143,8 +144,11 @@ void BindAlignmentEstimator(py::module& m) {
...
-        &AlignReconstructionToOrigRigScales,
-        "orig_rigs"_a,
-        "reconstruction"_a);
+  m.def(
+      "align_reconstruction_to_orig_rig_scales",
+      [](const RigMap& orig_rigs, Reconstruction& reconstruction) {
+        return AlignReconstructionToOrigRigScales(orig_rigs, &reconstruction);
+      },
+      "orig_rigs"_a,
+      "reconstruction"_a);
```
sarlinpe: We might instead want to edit the stubs to remove all occurrences of `SupportsIndex`, since it anyway looks confusing in the API docs.

ahojnnes: You mean at the stubgen layer or another perl hack?

sarlinpe: With a perl hack, until we send a PR to pybind11-stubgen to add this new feature. `SupportsIndex -> int` and `int | int -> int` should do, no?

ahojnnes: I found a different way to fix the issue in this specific PR. The opaque types were actually not correctly included/defined in all files. We can follow up with the type annotation change you suggest separately, as it seems unrelated to the issue here.

src/pycolmap/scene/database_cache.cc
```diff
@@ -75,10 +76,18 @@ void BindDatabaseCache(py::module& m) {
            py::overload_cast<image_t>(&DatabaseCache::Image),
            py::return_value_policy::reference_internal,
            "image_id"_a)
-      .def_property_readonly("rigs", &DatabaseCache::Rigs)
-      .def_property_readonly("cameras", &DatabaseCache::Cameras)
-      .def_property_readonly("frames", &DatabaseCache::Frames)
-      .def_property_readonly("images", &DatabaseCache::Images)
+      .def_property_readonly("rigs",
+                             &DatabaseCache::Rigs,
+                             py::return_value_policy::reference_internal)
```
sarlinpe, line 81: I believe that reference_internal should already be the default for def_property_readonly and def_property https://github.com/pybind/pybind11/blob/cd538ed1184d41cf685dd4da81c1eca80d24f353/include/pybind11/pybind11.h#L2407

PR #4279, opened 2026-03-21
Support incremental CorrespondenceGraph and ObservationManager construction

B1ueber2y:
* Allow CorrespondenceGraph to be queried without calling Finalize(), enabling streaming/online use where images and matches arrive incrementally. Finalize() remains available as a memory optimization. Track num_observations incrementally in AddTwoViewGeometry instead of deferring to Finalize().
* Add ObservationManager::AddImage() so that new images can be registered and triangulated without rebuilding the entire   ObservationManager. Includes visibility propagation for already-triangulated points.

All changes are additive and have no effects on any existing logics.

### DIFF
--- a/src/colmap/sfm/observation_manager.cc
+++ b/src/colmap/sfm/observation_manager.cc
@@ -71,7 +71,7 @@ ObservationManager::ObservationManager(
   // Add image stats.
   image_stats_.reserve(reconstruction_.NumImages());
   for (const auto& [image_id, image] : reconstruction_.Images()) {
-    image_stats_.emplace(image_id, InitImageStat(image_id));
+    image_stats_.emplace(image_id, InitImageStat(image_id, image));
   }
 
   // If an existing model was loaded from disk and there were already images
@@ -105,8 +105,8 @@ void ObservationManager::AddImage(const image_t image_id) {
         << "Image " << image_id
         << " must be added to the CorrespondenceGraph first";
   }
-  image_stats_.emplace(image_id, InitImageStat(image_id));
   const Image& image = reconstruction_.Image(image_id);
+  image_stats_.emplace(image_id, InitImageStat(image_id, image));
 
   if (correspondence_graph_) {
     // Add image pair stats for all pairs involving the new image.
@@ -151,8 +151,7 @@ void ObservationManager::AddImage(const image_t image_id) {
 }
 
 ObservationManager::ImageStat ObservationManager::InitImageStat(
-    const image_t image_id) const {
-  const Image& image = reconstruction_.Image(image_id);
+    const image_t image_id, const Image& image) const {
   const Camera& camera = *image.CameraPtr();
   ImageStat image_stat;
   image_stat.point3D_visibility_pyramid = VisibilityPyramid(
--- a/src/colmap/sfm/observation_manager.h
+++ b/src/colmap/sfm/observation_manager.h
@@ -71,7 +71,8 @@ class ObservationManager {
   // Add image stats for streaming/online SfM, so that the image can be
   // registered and triangulated without rebuilding the ObservationManager.
   // The image must already be added to the Reconstruction and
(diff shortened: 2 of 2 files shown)


src/colmap/scene/correspondence_graph_test.cc
```diff
@@ -392,5 +392,76 @@ TEST(CorrespondenceGraph, UpdateTwoViewGeometrySwapped) {
               Rigid3dNear(Inverse(cam1_from_cam0), 1e-6, 1e-6));
 }
 
+TEST(CorrespondenceGraph, WithoutFinalize) {
```
ahojnnes: Alternatively, we could use a parameterized gtest with a boolean "finalize" to make sure all tests work with and without finalize?

src/colmap/sfm/observation_manager.cc
```diff
@@ -108,6 +108,75 @@ ObservationManager::ObservationManager(
   }
 }
 
+void ObservationManager::AddImage(const image_t image_id) {
+  THROW_CHECK(reconstruction_.ExistsImage(image_id))
+      << "Image " << image_id << " must be added to the Reconstruction first";
+  THROW_CHECK(image_stats_.find(image_id) == image_stats_.end())
```
ahojnnes, line 99: nit: move this as the first check?

src/colmap/sfm/observation_manager.cc
```diff
@@ -108,6 +108,75 @@ ObservationManager::ObservationManager(
...
+    }
+
+    // Propagate visibility from already-triangulated points.
+    // In the batch pipeline, the constructor handles this by iterating
+    // all registered images and propagating triangulation visibility to
+    // their correspondences. Since this image was not present during
+    // construction, it missed that propagation. We catch up here by
+    // scanning the new image for correspondences to points that are
+    // already triangulated in other images.
+    for (point2D_t point2D_idx = 0; point2D_idx < image.NumPoints2D();
```
ahojnnes, line 136: Would it make sense to reuse the code in the constructor and here with a private helper method? Same for the other blocks above?

src/colmap/scene/correspondence_graph_test.cc
B1ueber2y: Thanks. There are only two tests concerning operations after "Finalize". I added the parameterized on these two and dropped the "WithoutFinalize" test.

src/colmap/sfm/observation_manager.cc
B1ueber2y: Thanks. This is done now.

src/colmap/sfm/observation_manager.cc
```diff
@@ -108,6 +95,80 @@ ObservationManager::ObservationManager(
...
+  THROW_CHECK(image_stats_.find(image_id) == image_stats_.end())
+      << "Image " << image_id << " already exists in the ObservationManager";
+  THROW_CHECK(reconstruction_.ExistsImage(image_id))
+      << "Image " << image_id << " must be added to the Reconstruction first";
+  if (correspondence_graph_) {
+    THROW_CHECK(correspondence_graph_->ExistsImage(image_id))
+        << "Image " << image_id
+        << " must be added to the CorrespondenceGraph first";
+  }
+  image_stats_.emplace(image_id, InitImageStat(image_id));
```
ahojnnes: InitImageStat can just directly take an `Image&` to avoid the redunandant lookup here and above in the constructor?

src/colmap/sfm/observation_manager.h
```diff
@@ -68,6 +68,12 @@ class ObservationManager {
   inline const std::unordered_map<image_pair_t, ImagePairStat>& ImagePairs()
       const;
 
+  // Add image stats for streaming/online SfM, so that the image can be
```
ahojnnes, line 71: Maybe a brief note about quadratic complexity?

PR #4283, opened 2026-03-22
Handle CHOLMOD includes not being in a subdirectory

### LINKED ISSUE #4266: Suitesparse subdirectory in include
Hi, in [FindCHOLMOD.cmake](https://github.com/colmap/colmap/blob/d927f7e518fc20afa33390712c4cc20d85b730b8/cmake/FindCHOLMOD.cmake#L60) it always looks in a `/suitesparse` subdirectory for the headers, but on NixOS for example we don't have these in a subdirectory. Would it be possible to make this more flexible?

chpatrick:
Fixes #4266


ahojnnes:
Thanks!

PR #4309, opened 2026-03-30
Parallelize color extraction for all images

ahojnnes:
Replaces: https://github.com/colmap/colmap/pull/4300

PR #4310, opened 2026-03-30
Fix thread oversubscription in hierarchical mapper.

B1ueber2y:


B1ueber2y:
Issues found when reviewing https://github.com/colmap/colmap/pull/4309

PR #4312, opened 2026-03-31
Format also staged files

ahojnnes:
Previous logic only formatted committed files, which is inconvenient when working on a branch to submit local changes without going through an intermediate commit.


ahojnnes:
OK, agreed. I keep formatting staged files though, as the intention for those is to be committed and it solves most of the inconvenience of having to first commit to then format.

PR #4314, opened 2026-03-31
Add new camera specs

ahojnnes:
With the help of Claude.


ahojnnes:
1. Some from OpenMVG, which seems like a copy from COLMAP with some new ones added manually later (based on git history). Some from a Claude web-crawl (iPhone, Canon).
2. They are all derived from the raw sensor specs ignoring distortion.
3. Yes, had the same thought already. Not sure about json/yaml and how to package it into the binary (maybe through Q_RESOURCE). Let's separate this change from the PR here.

PR #4318, opened 2026-04-01
Add FinalizeDataIds() guard to prevent Frame mutation after reconstruction insertion

ahojnnes:
- Add `data_ids_finalized_` flag to `Frame`, set by `Reconstruction::AddFrame`, that prevents `AddDataId`/`ClearDataIds` from being called on frames already in a reconstruction
- Remove mutable `DataIds()` accessor to prevent bypassing the guard
- Custom copy constructor does not propagate finalized state, so copies of frames from a reconstruction can be freely modified
- Restructure `CreateExpandedReconstruction` in rotation_averaging.cc to build frames fully before adding to reconstruction
- Expose `finalize_data_ids`/`final_data_ids` to Python bindings

### DIFF
--- a/src/colmap/scene/frame.h
+++ b/src/colmap/scene/frame.h
@@ -33,8 +33,6 @@
 #include "colmap/sensor/rig.h"
 #include "colmap/util/types.h"
 
-#include <cstdint>
-#include <memory>
 #include <optional>
 #include <set>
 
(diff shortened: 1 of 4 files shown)


src/colmap/scene/frame.h
```diff
@@ -45,19 +45,30 @@ namespace colmap {
...
   inline size_t NumDataIds() const;
 
   // Check whether the data is associated with the frame.
   inline bool HasDataId(data_t data_id) const;
 
+  // Finalize the data ids, preventing further modifications. This is called
+  // when the frame is added to a reconstruction to ensure consistency of
+  // cached counters like num_reg_images_.
+  inline void FinalizeDataIds();
+  inline bool FinalDataIds() const;
```
B1ueber2y: Does it make sense to use "is" or "has" here to improve the readability?

PR #4323, opened 2026-04-03
Decouple reg_stats from ObservationManager in RegisterFrameEvent/DeRegisterFrameEvent

ahojnnes:
## Summary

Fixes two bugs in `IncrementalMapper` that caused `num_visible_correspondences` (a `uint32_t`) to be incorrect, visible as negative values in the log output (`"Image sees -109 / 289 correspondences"`).

### Bug 1: Double de-registration in `FilterFrames`

`IncrementalMapper::FilterFrames()` called `ObservationManager::FilterFrames()` which internally de-registered each filtered frame (decrementing `num_visible_correspondences` for all correspondence partners), then called `DeRegisterFrameEvent()` which called `DeRegisterFrame()` again — double-decrementing the unsigned counter and causing underflow.

### Bug 2: Double initialization in `BeginReconstruction`

The `ObservationManager` constructor already initializes `num_visible_correspondences` for pre-existing registered frames. Then `BeginReconstruction` called `RegisterFrameEvent()` → `RegisterFrame()` for those same frames, doubling the count.

### Fix

- `RegisterFrameEvent` / `DeRegisterFrameEvent` now only manage `reg_stats_` (the `IncrementalMapper`'s own bookkeeping). They no longer call `obs_manager_->RegisterFrame()` / `DeRegisterFrame()`.
- Callers explicitly call `obs_manager_->RegisterFrame()` or `obs_manager_->DeRegisterFrame()` where needed.
- `BeginReconstruction` only calls `RegisterFrameEvent` for pre-existing frames (reg_stats only), relying on the `ObservationManager` constructor for observation bookkeeping. No double-counting.
- Replace `ObservationManager::FilterFrames()` with `FindFramesToFilter()` that only identifies frames without de-registering them. Callers handle de-registration explicitly.

### DIFF
--- a/src/colmap/sfm/incremental_mapper_test.cc
+++ b/src/colmap/sfm/incremental_mapper_test.cc
@@ -482,8 +482,10 @@ TEST_F(IncrementalMapperLargeDatasetTest,
             num_reg_images_before - num_images_in_frame);
   EXPECT_EQ(mapper_->FilteredFrames().count(target_frame_id), 1);
 
-  const std::unordered_set<image_t> filtered_image_ids = {
-      target_frame.ImageIds().begin(), target_frame.ImageIds().end()};
+  std::unordered_set<image_t> filtered_image_ids;
+  for (const data_t& data_id : target_frame.ImageIds()) {
+    filtered_image_ids.insert(data_id.id);
+  }
 
   // After filtering, verify exact num_visible_correspondences for every image.
   // DeRegisterFrame decrements the correspondence partners' counters, so each


src/pycolmap/sfm/observation_manager.cc
```diff
@@ -93,16 +93,19 @@ void BindObservationManager(py::module& m) {
...
-           "Filter frames without observations or bogus camera parameters."
-           "Return the identifiers of the filtered frames.")
+           "min_num_observations"_a,
+           "Find frames that should be filtered due to having no observations "
+           "or bogus camera parameters, without de-registering them. Pass them "
+           "to DeRegisterFrame to reset their pose."
+           "Return the identifiers of the frames to filter.")
       .def("register_frame",
-           &ObservationManager::DeRegisterFrame,
+           &ObservationManager::RegisterFrame,
```
ahojnnes, line 106: Faulty binding.

B1ueber2y (approved):

PR #4337, opened 2026-04-10
Add advancing front surface reconstruction meshing

ahojnnes:
## Summary

Implement the Advancing Front Surface Reconstruction (AFSR) algorithm based on Cohen-Steiner & Da (The Visual Computer, 2004), with visibility-based filtering following the [cvg/pcdmeshing](https://github.com/cvg/pcdmeshing) reference implementation.

### Key differences from the pcdmeshing reference

- **Integrated into COLMAP pipeline**: CLI command (`colmap advancing_front_mesher`), Python bindings (`pycolmap.advancing_front_meshing()`), automatic reconstruction pipeline, and option manager — not a standalone tool
- **Visibility from COLMAP workspace**: Loads visibility directly from `fused.ply.vis` and sparse reconstruction, rather than requiring separate endpoint/observation PLY files
- **Consistent visibility threshold**: Both pre-filtering and post-filtering modes use `visibility_filtering_max_intersections` to only remove faces exceeding N ray intersections. The reference post-filtering removes any face hit by a single ray
- **Block-wise visibility filtering**: Rays are assigned to blocks via segment-AABB intersection tests, so visibility filtering works correctly across block boundaries. The reference disables visibility in block-wise mode entirely
- **Per-thread local accumulation**: Visibility counters and face sets use per-thread local maps merged after the parallel region, avoiding the `#pragma omp critical` bottleneck in the reference
- **CGAL 6.x compatibility**: Uses `CGAL::AABB_traits_3` (non-deprecated), `CGAL::do_intersect` instead of `CGAL::assign`, and `CGAL::Euler::remove_face` for correct halfedge connectivity
- **EPIC float32 kernel**: Same float32 approach as the reference for memory efficiency, but using `CGAL::Filtered_kernel_adaptor` for exact predicates
- **File decomposition**: Existing `meshing.h/cc` split into `poisson_meshing`, `delaunay_meshing`, `advancing_front_meshing` with separate test files

Sample result:
advancing front


src/colmap/mvs/advancing_front_meshing.cc
```diff
@@ -0,0 +1,899 @@
...
+  for (const size_t ray_idx : ray_indices) {
+    block_rays.push_back(all_rays[ray_idx]);
+  }
+  return block_rays;
+}
+
+colmap::PlyMesh CropMeshToRegion(const colmap::PlyMesh& mesh,
+                                 const Eigen::AlignedBox3d& crop_box) {
+  colmap::PlyMesh cropped;
+  cropped.vertices = mesh.vertices;
```
B1ueber2y: Does it make sense to clean the vertices in this function?

src/colmap/mvs/advancing_front_meshing.cc
```diff
@@ -0,0 +1,899 @@
...
+        block_rays = GatherBlockRays(all_rays, block_ray_indices[block_idx]);
+      }
+
+      const auto block_mesh =
+          ReconstructBlock(block_points, block_rays, options);
+
+      // Crop to the core region plus half the overlap as seam margin.
+      int bx, by, bz;
+      grid.BlockCoords(block_idx, bx, by, bz);
+      const double margin = grid.overlap * 0.5;
```
B1ueber2y, line 785: If I understand correctly, there might be duplicated triangles for this block method. However, I do not have a good solution to avoid it.

B1ueber2y (approved):
Thanks!

src/colmap/mvs/advancing_front_meshing.cc
ahojnnes: Done

src/colmap/mvs/advancing_front_meshing.cc
ahojnnes: I think it's fine to have some duplicated triangles in the margin.

PR #4341, opened 2026-04-13
Using IndexIVFScalarQuantizer instead of IndexIVFFlat to accelerate exhaustive_matcher

baohongmin:
Performance Optimization: Achieved 1.25x speed improvement by replacing IndexIVFFlat with IndexIVFScalarQuantizer, utilizing uint8 quantization for direct vector similarity calculations while preserving accuracy.

### DIFF
--- a/src/colmap/feature/index.cc
+++ b/src/colmap/feature/index.cc
@@ -34,8 +34,8 @@
 #include <faiss/IndexFlat.h>
 #include <faiss/IndexIVFFlat.h>
 #include <faiss/IndexIVFPQ.h>
+#include <faiss/IndexIVFScalarQuantizer.h>
 #include <faiss/IndexPQ.h>
-#include <faiss/IndexScalarQuantizer.h>
 #include <omp.h>
 
 namespace colmap {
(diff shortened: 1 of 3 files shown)


ahojnnes:
Thank you. What's the impact of this change on matching accuracy? I assume this is not for free and comes with some trade-off?

baohongmin:
> Thank you. What's the impact of this change on matching accuracy? I assume this is not for free and comes with some trade-off?

1.Since SIFT-extracted vectors are natively represented in uint8 format, there is no accuracy impact from quantization. However, FAISS cannot directly build indices from uint8 data. Consequently, Colmap first converts SIFT uint8 vectors to floating-point format before index building.
2. IndexIVFScalarQuantizer stores the original SIFT uint8 vectors, while "QT_8bit_direct" enables direct uint8 computation for L2 distance calculations, resulting in superior performance.

src/thirdparty/CMakeLists.txt
```diff
@@ -55,11 +55,12 @@ endif()
...
     include(FetchContent)
     FetchContent_Declare(faiss
-        URL https://github.com/ahojnnes/faiss/archive/36b77353dc435383e0c23a709e7997a29d049041.zip
-        URL_HASH SHA256=aaf90b3dd3e353ff0bc3391581788cda7c510749023d1937707fc99c38ef2587
+        URL https://github.com/facebookresearch/faiss/archive/refs/tags/v1.14.1.zip
+        URL_HASH SHA256=4b1ae7e7a0a46385b4084f0e3945623a15fcf99d793bf44d82aae8e24f11e5f5
         ${_fetch_content_declare_args}
     )
     message(STATUS "Configuring faiss...")
+    set(FAISS_OPT_LEVEL dd)
```
ahojnnes: Should we revert this change or what does "dd" do? I could not find a reference to this in the faiss repo.

src/colmap/feature/index.cc
```diff
@@ -35,6 +35,7 @@
 #include <faiss/IndexIVFFlat.h>
```
ahojnnes, line 35: We can probably now remove this include?

ahojnnes (approved):
Thanks for clarifying, makes sense, and I now understand.

src/colmap/feature/index.cc
baohongmin: yes, we can remove it.

src/thirdparty/CMakeLists.txt
baohongmin: https://github.com/facebookresearch/faiss/blob/main/faiss/docs/simd_dynamic_dispatch_migration.md
https://github.com/facebookresearch/faiss/blob/main/faiss/utils/simd_levels.cpp#L59
The libfaiss.a library compiled with dynamic dispatch (dd) automatically selects and executes the highest SIMD instruction set supported by the target machine, such as AVX512.

ahojnnes:
I pushed some changes to your branch. Since we now also support ALIKED features, which are natively float32 descriptors, I fallback to the old implementation for all but SIFT descriptors.

baohongmin:
> I pushed some changes to your branch. Since we now also support ALIKED features, which are natively float32 descriptors, I fallback to the old implementation for all but SIFT descriptors.

It seems that there is no IndexIVFScalarQuantizer. h, so the compilation may have failed.

PR #4342, opened 2026-04-13
Fix inconsistent doc strings

ahojnnes:
Asked Claude to make a pass through all doc strings and fix inconsistencies/mistakes/stale information.


B1ueber2y (approved):
Thanks!

PR #4345, opened 2026-04-15
Bind ReprojectionErrorType and additional point filter methods in pycolmap

lpanaf:
Expose ObservationManager::FilterPoints3DWithLargeReprojectionError, which supports PIXEL, NORMALIZED, and ANGULAR reprojection error metrics, along with FilterPoints3DWithSmallTriangulationAngle. Also bind the ReprojectionErrorType enum so the angular-error mode is reachable from Python.


ahojnnes (approved):
LGTM.

PR #4347, opened 2026-04-16
Change std::filesystem::relative to lexically_relative for NormalizePath.

laxnpander:
Currently, NormalizePath will use std::filesystem::relative to compute the relative location of the images to the image_path in the image_reader. However, as std::filesystem::relative is defined as using weakly_canonical, it will resolve symlinks [(ref)](https://en.cppreference.com/w/cpp/filesystem/relative.html). 

This leads to counterintuitive issues when processing folders which contain images as symlinks. An example:

/path/to/real/image.png
/path/to/symlinks/image.png

image_path: /path/to/symlinks

Now NormalizePath will resolve to e.g. ../real/image.png instead of symlinks/image.png. This in turn can cause images to be duplicated during feature extraction as the image names are suddenly "new".

### DIFF
--- a/src/colmap/util/file.cc
+++ b/src/colmap/util/file.cc
@@ -169,7 +169,7 @@ std::string NormalizePath(const std::filesystem::path& path) {
 
 std::string GetNormalizedRelativePath(const std::filesystem::path& full_path,
                                       const std::filesystem::path& base_path) {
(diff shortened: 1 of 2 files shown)


ahojnnes (approved):
Thanks, I added a test to cover the issue.

PR #4351, opened 2026-04-17
Add comprehensive smoke tests for pycolmap Python bindings

ahojnnes:
## Summary
- Add 812 smoke tests covering all pycolmap binding modules (util, geometry, optim, sensor, feature, scene, image, estimators, retrieval, sfm, pipeline, mvs)
- Each test file lives next to its corresponding `.cc` binding file in `src/pycolmap/` for easy association
- Tests verify fields can be read/written, methods can be called, and enums have expected values — no deep functional testing
- Cross-cutting tests for the MakeDataclass protocol (summary, repr, todict, pickle, etc.)
- Updates `pyproject.toml`: adds `src/pycolmap` as test path, enables `importlib` import mode, adds mypy checking

## Test plan
- [x] `pytest .` passes (866 tests total: 812 new + 54 existing)
- [x] `mypy --package pycolmap` passes
- [x] `mypy src/pycolmap` passes (65 source files, no errors)

### DIFF
--- a/src/pycolmap/estimators/alignment_test.py
+++ b/src/pycolmap/estimators/alignment_test.py
@@ -1,3 +1,5 @@
+import pytest
+
 import pycolmap
 
 
@@ -24,20 +26,17 @@ def test_image_alignment_error_proj_center_error_readwrite():
     assert error.proj_center_error == 0.01
 
 
-def test_align_reconstructions_via_reprojections_is_callable():
-    assert callable(pycolmap.align_reconstructions_via_reprojections)
-
-
-def test_align_reconstructions_via_proj_centers_is_callable():
-    assert callable(pycolmap.align_reconstructions_via_proj_centers)
-
-
-def test_align_reconstructions_via_points_is_callable():
-    assert callable(pycolmap.align_reconstructions_via_points)
-
-
-def test_compare_reconstructions_is_callable():
-    assert callable(pycolmap.compare_reconstructions)
+@pytest.mark.parametrize(
+    "name",
+    [
+        "align_reconstructions_via_reprojections",
+        "align_reconstructions_via_proj_centers",
+        "align_reconstructions_via_points",
+        "compare_reconstructions",
+    ],
+)
+def test_public_api_callable(name):
+    assert callable(getattr(pycolmap, name))
 
 
 def test_compare_reconstructions_with_synthetic(synthetic_reconstruction):
--- a/src/pycolmap/estimators/alignment_test.py
+++ b/src/pycolmap/estimators/alignment_test.py
@@ -1,0 +1,49 @@
+import pycolmap
+
+
+def test_image_alignment_error_default_init():
+    error = pycolmap.ImageAlignmentError()
+    assert error is not None
+
+
+def test_image_alignment_error_image_name_readwrite():
+    error = pycolmap.ImageAlignmentError()
+    error.image_name = "test_image.jpg"
+    assert error.image_name == "test_image.jpg"
+
+
+def test_image_alignment_error_rotation_error_deg_readwrite():
+    error = pycolmap.ImageAlignmentError()
+    error.rotation_error_deg = 1.5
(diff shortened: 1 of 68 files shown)


src/pycolmap/estimators/ceres_bindings.cc
```diff
@@ -164,7 +164,12 @@ void BindCeresSolver(py::module& m) {
   py::classh<Options> PyOptions(m, "SolverOptions", py::module_local());
   PyOptions.def(py::init<>())
       .def(py::init<const Options&>())
-      .def("IsValid", &Options::IsValid)
+      .def("IsValid",
```
ahojnnes, line 167: @B1ueber2y @sarlinpe Should we sync this to pyceres? The previous interface didn't make sense, because the error string couldn't be retrieved.

src/pycolmap/estimators/alignment_test.py
```diff
@@ -0,0 +1,49 @@
...
+
+
+def test_image_alignment_error_proj_center_error_readwrite():
+    error = pycolmap.ImageAlignmentError()
+    error.proj_center_error = 0.01
+    assert error.proj_center_error == 0.01
+
+
+def test_align_reconstructions_via_reprojections_is_callable():
+    assert callable(pycolmap.align_reconstructions_via_reprojections)
```
B1ueber2y: The per-file assert callable(...) tests only catch symbol removal, not signature or behavior. Does it make sense to consolidate them into one parametrized presence test over a name list, and keep the colocated files for tests that actually invoke the function? 

For example, claude gives me something like this:
```
PUBLIC_CALLABLES = [
    "estimate_fundamental_matrix",
    "estimate_essential_matrix",
    "estimate_homography_matrix",
]

@pytest.mark.parametrize("name", PUBLIC_CALLABLES)
def test_public_api_callable(name):
    assert callable(getattr(pycolmap, name))
```

src/pycolmap/estimators/bundle_adjustment_test.py
```diff
@@ -0,0 +1,237 @@
+import pycolmap
+
+
+def test_bundle_adjustment_termination_type_convergence():
+    assert pycolmap.BundleAdjustmentTerminationType.CONVERGENCE is not None
```
B1ueber2y: ``assert Enum.VALUE is not None`` can't fail in the sense that 1) enum members are never None 2) a missing member raises AttributeError before the assert. So this is actually equivalent to hasattr, and does not pin the integer values.

How about something like this as a single test:
```
assert {m.name: int(m) for m in pycolmap.BundleAdjustmentTerminationType} == {
    "CONVERGENCE": 0, "NO_CONVERGENCE": 1, "FAILURE": 2,
    "USER_SUCCESS": 3, "USER_FAILURE": 4,
}
```

src/pycolmap/estimators/ceres_bindings.cc
B1ueber2y: Done: https://github.com/cvg/pyceres/pull/89.

B1ueber2y: Actually, given that we are already here, should this be ``is_valid`` instead to be consistent with the pycolmap style?

sarlinpe: All functions in PyCeres use Camel case, so we shouldn't rename it, for consistency.

src/pycolmap/estimators/bundle_adjustment_test.py
```diff
@@ -0,0 +1,220 @@
+import pycolmap
+
+
+def test_bundle_adjustment_termination_type_enum():
+    assert {
+        m.name: int(m) for m in pycolmap.BundleAdjustmentTerminationType
```
B1ueber2y: Sorry. It should be ``xxxType.__members__.items()``

B1ueber2y (approved):
Thanks!

PR #4360, opened 2026-04-24
Add live progress display for reconstruction benchmark evaluation

ahojnnes:
## Summary
- Adds an opt-in live progress display (powered by `rich`) to `benchmark/reconstruction/evaluate.py` showing an overall scene completion bar plus a per-scene row with the current phase (setup, extraction, matching, reconstruction, evaluation).
- Workers publish phase transitions to a `multiprocessing.Manager` dict consumed by a monitor thread in the parent process.
- Controlled via `--progress` / `--no-progress`; auto-enabled when stdout is a TTY.


lpanaf (approved):
Thanks!

PR #4018, opened 2026-01-16
GPU Accelerated Bundle Adjustment with Caspar

tordnat:
This PR includes my attempt of integrating [Caspar](https://github.com/symforce-org/symforce/tree/main/symforce/experimental/caspar), a CUDA accelerator for symbolic programming, into COLMAP as an alternative to Ceres Solver. Testing Caspar on the BAL datasets results in a 5-20x speed improvement, as well as improved memory efficiency (Caspar results to be published in upcoming paper). I have implemented factors for the simple radial camera model. Caspar uses damped LM, which should work without gauge fixing. However, I have also implemented logic for gauge fixing. The CasparBundleAdjuster inherits from the base bundle adjustment class, and should be an easy drop-in replacement. 

image

How to build and run:
1. Build latest version of [Symforce](https://github.com/symforce-org/symforce)
2. Generate and build the Caspar library by running: `python generate.py` in the colmap estimator directory
3. Enable Caspar support with the CMake flag -DCASPAR_ENABLED=ON

What works and what doesn't:
I have tested the implementation on synthetic data both in isolation and in COLMAP's unit testing framework. This seems to pass, and work as expected. However, when running on datasets in COLMAP, the reconstruction seems to degrade and eventually fail. I have spent a lot of time trying to isolate the reason for this behavior, but the bug has thus far not revealed itself (hence all the remaining debugging and logging in the code). 

I'm hoping to get some feedback on my initial attempt, and perhaps gain some insights in what might be the missing piece in making this implementation work. 

Cheers!

PR #4374, opened 2026-05-05
Update CI cuda architecture from 50 to 75.

B1ueber2y:
Per discussion from https://github.com/colmap/colmap/pull/4018.

However, I believe this would mean that we may not catch issues that break compilation on older arches (50, 60, 70) and may only bump into it at the release multi-arch build runs.


ahojnnes:
Since this would only catch issues at compile-time, I am not sure there is a big risk. If this helps unblocking the Caspar PR, I am fine with going ahead with this.

B1ueber2y:
I already patched it in the caspar PR. The docker CI does not need to be updated as we did not enable caspar there. I think merging directly with the caspar changes would be fine enough. Closing for now.

PR #4037, opened 2026-01-20
Performance improvement on generalized absolute pose.

B1ueber2y:
main (on the test case from https://github.com/colmap/colmap/issues/4023):
```
Running EstimateGeneralizedAbsolutePose with max_trials=1000
[RANSAC] 1000 trials, 2277 models, 58 inliers
  Estimation: 1.173 ms (11.5875%)
  Residuals:  8.005 ms (79.0773%)
  Scoring:    0.945 ms (9.33518%)
  Total:      10.123 ms
EstimateGeneralizedAbsolutePose RANSAC time: 0.010301 s, num_trials: 1000
```

After precomputing rig_from_world matrix:
```
Running EstimateGeneralizedAbsolutePose with max_trials=1000
[RANSAC] 1000 trials, 2277 models, 58 inliers
  Estimation: 1.14 ms (13.3834%)
  Residuals:  6.451 ms (75.7337%)
  Scoring:    0.927 ms (10.8828%)
  Total:      8.518 ms
EstimateGeneralizedAbsolutePose RANSAC time: 0.008678 s, num_trials: 1000
```

After precomputing cam_from_rig matrix:
```
Running EstimateGeneralizedAbsolutePose with max_trials=1000
[RANSAC] 1000 trials, 2279 models, 58 inliers
  Estimation: 1.196 ms (15.7431%)
  Residuals:  5.373 ms (70.7253%)
  Scoring:    1.028 ms (13.5317%)
  Total:      7.597 ms
EstimateGeneralizedAbsolutePose RANSAC time: 0.007778 s, num_trials: 1000
```

Will follow up another update on generalized relative pose. 

Edit: In the generalized relative pose case poselib::gen_relpose_6pt is the bottleneck, so there is not much to optimize in the colmap side. 

Edit: after switching to eigen expressions:
```
Running EstimateGeneralizedAbsolutePose with max_trials=1000
[RANSAC] 1000 trials, 2279 models, 58 inliers
  Estimation: 1.198 ms (18.0503%)
  Residuals:  4.455 ms (67.1237%)
  Scoring:    0.984 ms (14.826%)
  Total:      6.637 ms
EstimateGeneralizedAbsolutePose RANSAC time: 0.006816 s, num_trials: 1000
```

PR #4029, opened 2026-01-19
Single pose parameter block and angle axis for BA cost functions

ahojnnes:
* Leads to ~10-15% speedup compared to the main branch.
* Breaks compatibility of cost functions, but there errors should be obvious on the user side, where Ceres will complain that the parameter block sizes don't match.

PR #4375, opened 2026-05-05
Fix: Missing rig pose manifold in RefineGeneralizedAbsolutePose

nushakrishnan:
Hi

I was facing an error with `pycolmap.estimate_and_refine_generalized_absolute_pose` when installing pycolmap via pip. 

Error:
```
*** Aborted at 1777983489 (unix time) try "date -d @1777983489" if you are using GNU date ***
PC: @     0x73ded64220cb (/local/home/akrishnan/lamar_env_fix/lib/python3.10/site-packages/pycolmap/_core.cpython-310-x86_64-linux-gnu.so+0x10220ca)
Segmentation fault
```

The breaking PR was #4041. I was able to validate this by installing pycolmap from source on the commit before (#4037). The script below should work on PR #4037.

Test data: [Rig_pose_data](https://drive.google.com/file/d/125Y-Pap9FhQUVCayLVfU6iTWXQSHJaS7/view?usp=sharing)

Testing script:

```python
import pickle, time
import numpy as np
import pycolmap
from tqdm import tqdm

PKL = '/local/home/akrishnan/Downloads/rig_pose_debug.pkl'

with open(PKL, 'rb') as f:
    all_data = pickle.load(f)

print(f'pycolmap {pycolmap.__version__} from {pycolmap.__file__}')
print(f'{len(all_data)} entries')

n_ok = 0
n_none = 0
t0 = time.time()
for i, d in enumerate(tqdm(all_data, desc='rig PnP')):
    p2d_per_cam = d['p2d']
    p3d_per_cam = d['p3d']
    camera_dicts = d['camera_dicts']
    qvecs = d['qvecs'] 
    tvecs = d['tvecs']
    thresh = float(d['thresh'])

    p2d_flat = np.concatenate(p2d_per_cam, axis=0).astype(np.float64)
    p3d_flat = np.concatenate(p3d_per_cam, axis=0).astype(np.float64)
    camera_idxs = []
    for ci, p in enumerate(p2d_per_cam):
        camera_idxs.extend([ci] * len(p))

    cams_from_rig = []
    for q_wxyz, t in zip(qvecs, tvecs):
        q_xyzw = np.array([q_wxyz[1], q_wxyz[2], q_wxyz[3], q_wxyz[0]], dtype=np.float64)
        cams_from_rig.append(pycolmap.Rigid3d(
            pycolmap.Rotation3d(q_xyzw),
            np.asarray(t, dtype=np.float64),
        ))

    # Build cameras
    cameras = [
        pycolmap.Camera(
            model=cd['model'],
            width=int(cd['width']),
            height=int(cd['height']),
            params=list(cd['params']),
        )
        for cd in camera_dicts
    ]

    opts = pycolmap.RANSACOptions()
    opts.max_error = thresh

    ret = pycolmap.estimate_and_refine_generalized_absolute_pose(
        p2d_flat, p3d_flat, camera_idxs, cams_from_rig, cameras,
        estimation_options=opts, return_covariance=True,
    )
    if ret is None:
        n_none += 1
    else:
        n_ok += 1

dt = time.time() - t0
print(f'\ndone in {dt:.1f}s — succeeded: {n_ok}/{len(all_data)}, '
      f'returned None: {n_none}/{len(all_data)}')
```

### DIFF
--- a/src/colmap/estimators/generalized_pose.cc
+++ b/src/colmap/estimators/generalized_pose.cc
@@ -409,10 +409,12 @@ bool RefineGeneralizedAbsolutePose(const AbsolutePoseRefinementOptions& options,
     }
   }
 
-  SetManifold(&problem,
-              rig_from_world->params.data(),
-              CreateProductManifold(CreateEigenQuaternionManifold(),
-                                    CreateEuclideanManifold<3>()));
+  if (problem.NumResiduals() > 0) {
+    SetManifold(&problem,
+                rig_from_world->params.data(),
+                CreateProductManifold(CreateEigenQuaternionManifold(),
+                                      CreateEuclideanManifold<3>()));
+  }
 
   ceres::Solver::Options solver_options;
   solver_options.gradient_tolerance = options.gradient_tolerance;
--- a/src/colmap/estimators/generalized_pose.cc
+++ b/src/colmap/estimators/generalized_pose.cc
@@ -407,9 +407,6 @@ bool RefineGeneralizedAbsolutePose(const AbsolutePoseRefinementOptions& options,
         }
       }
     }
-  }
-
-  if (problem.NumResiduals() > 0) {
     SetManifold(&problem,
                 rig_from_world->params.data(),
                 CreateProductManifold(CreateEigenQuaternionManifold(),
--- a/src/colmap/estimators/generalized_pose.cc
+++ b/src/colmap/estimators/generalized_pose.cc
@@ -409,6 +409,11 @@ bool RefineGeneralizedAbsolutePose(const AbsolutePoseRefinementOptions& options,
     }
   }
 
+  SetManifold(&problem,
+              rig_from_world->params.data(),
+              CreateProductManifold(CreateEigenQuaternionManifold(),
+                                    CreateEuclideanManifold<3>()));
+
   ceres::Solver::Options solver_options;
   solver_options.gradient_tolerance = options.gradient_tolerance;
   solver_options.max_num_iterations = options.max_num_iterations;


src/colmap/estimators/generalized_pose.cc
```diff
@@ -409,6 +409,13 @@ bool RefineGeneralizedAbsolutePose(const AbsolutePoseRefinementOptions& options,
     }
   }
 
+  if (problem.NumResiduals() > 0) {
```
ahojnnes:
```suggestion

```

ahojnnes (approved):
Thank you for this fix. LGTM.

B1ueber2y:
Thanks! I dont see how this was introduced in https://github.com/colmap/colmap/pull/4041 though. Was this always buggy as we were missing the EigenQuaternionManifold parameterization even before the single-parameter block change? Also curious why this would give segfault.

ahojnnes:
> Thanks! I dont see how this was introduced in #4041 though. Was this always buggy as we were missing the EigenQuaternionManifold parameterization even before the single-parameter block change? Also curious why this would give segfault.

The only explanation for segfault I have is that this produces an invalid Quaternion and then some Eigen operation produces a crash? In either case, it's a genuine bug that we should fix.

nushakrishnan:
While this is closed, just a follow-up. I dug into it a bit and I think that -

[Here](https://github.com/colmap/colmap/blob/754b0993dc010bcdaaa5bc2b87381822cbe8612f/src/colmap/estimators/generalized_pose.cc#L355), the tangent space size is 3. Whereas it is missing [here](https://github.com/colmap/colmap/blob/12dbe8a5cb41f167343a4136c950aa535636cce8/src/colmap/estimators/generalized_pose.cc#L424-L436). The tangent space size is returned as 7. 

In addition, PR #4029 dropped the SetQuaternionManifold without updating to the new one.

ahojnnes:
Thanks Anusha, I think the covariance extraction explains it, as Ceres would write a 7x7 matrix but we only pass enough memory for a 6x6 matrix, which likely leads to the observed segmentation fault. Thanks for digging more.

PR #4377, opened 2026-05-05
Fix: Apply refine_sensor_from_rig to all stages of the global mapper

TannerGilbert:
## Changes                                                                                                                                                    
                                                                                                                                                              
**Fix: `refine_sensor_from_rig=False` now honored by the global mapper**                                                                                      
Previously, the flag was respected only in incremental BA. In the global SfM , both rotation averaging and global positioning silently overwrote per-sensor `cam_from_rig` data — RA reset translations, GP NaN-poisoned them for re-estimation — even when callers had explicitly opted out of refinement.

This PR plumbs the flag through both stages so a pre-calibrated rig stays exactly as provided when `refine_sensor_from_rig=False`:                                                                                                      
- `RotationEstimatorOptions::refine_sensor_from_rig` (new, default `true`) → when `false`, only `rig_from_world` is updated; `sensor_from_rig` is left untouched.                                                                                                                                             
- `GlobalPositionerOptions::refine_sensor_from_rig` (new, default `true`) → when `false`, the `cam_in_rig` direction term uses the provided translation (zeroing only on NaN) and `ConvertBackResults` skips the sensor-from-rig writeback.                                                                                                                                  
- pycolmap bindings exposed for both.                                                                                                                         
                                                                                                                                                              
**Tests**                                              
                                                                                                                                                              
- `global_positioning_test`: new case verifies sensor_from_rig is preserved end-to-end through global positioning when the flag is `false`.                                                                                             
- `rotation_averaging_test`: matching case for rotation averaging — exact equality of both rotation (quaternion coeffs) and translation against a pre-RA snapshot.                                                                                                                                            
                                                                                                                                                              
## Discussion / follow-up                              
                                                                                                                                                              
`refine_sensor_from_rig` now exists as three independent fields on `BundleAdjustmentOptions`, `RotationEstimatorOptions` and `GlobalPositionerOptions`. Should this be a single field instead?

### DIFF
--- a/src/colmap/estimators/global_positioning.cc
+++ b/src/colmap/estimators/global_positioning.cc
@@ -246,7 +246,8 @@ void GlobalPositioner::AddPoint3DToProblem(point3D_t point3D_id,
         // re-estimated, which requires refine_sensor_from_rig=true.
         THROW_CHECK(options_.refine_sensor_from_rig)
             << "sensor_from_rig has NaN translation but "
-               "refine_sensor_from_rig=false";
+               "refine_sensor_from_rig=false (image_id="
+            << observation.image_id << ")";
         const sensor_t sensor_id = image.CameraPtr()->SensorId();
         if (cams_in_rig_.find(sensor_id) == cams_in_rig_.end()) {
           // Will be initialized to random values in ParameterizeVariables().
--- a/src/colmap/estimators/global_positioning.cc
+++ b/src/colmap/estimators/global_positioning.cc
@@ -227,10 +227,16 @@ void GlobalPositioner::AddPoint3DToProblem(point3D_t point3D_id,
       Rig& rig = reconstruction.Rig(rig_id);
       Rigid3d& cam_from_rig = rig.SensorFromRig(image.CameraPtr()->SensorId());
 
-      if (!cam_from_rig.translation().hasNaN()) {
+      const bool keep_rig_fixed =
+          !cam_from_rig.translation().hasNaN() ||
+          !options_.refine_sensor_from_rig;
+      if (keep_rig_fixed) {
+        Eigen::Vector3d cam_in_rig = cam_from_rig.translation();
+        if (cam_in_rig.hasNaN()) {
+          cam_in_rig.setZero();
+        }
         const Eigen::Vector3d cam_from_rig_dir =
-            image.CamFromWorld().rotation().inverse() *
-            cam_from_rig.translation();
+            image.CamFromWorld().rotation().inverse() * cam_in_rig;
 
         ceres::CostFunction* cost_function =
             RigBATAPairwiseDirectionConstantRigCostFunctor::Create(
@@ -426,17 +432,19 @@ void GlobalPositioner::ConvertBackResults(Reconstruction& reconstruction) {
     rig_from_world.translation() = rig_from_world.rotation() * -center;
   }
 
-  // Convert optimized cam_in_rig back to sensor_from_rig translations.
-  for (const auto& [sensor_id, center] : cams_in_rig_) {
-    // Find the rig containing this sensor.
-    for (const auto& [rig_id, rig] : reconstruction.Rigs()) {
-      if (!rig.HasSensor(sensor_id)) {
-        continue;
+  if (options_.refine_sensor_from_rig) {
+    // Convert optimized cam_in_rig back to sensor_from_rig translations.
+    for (const auto& [sensor_id, center] : cams_in_rig_) {
+      // Find the rig containing this sensor.
+      for (const auto& [rig_id, rig] : reconstruction.Rigs()) {
+        if (!rig.HasSensor(sensor_id)) {
+          continue;
+        }
+        Rigid3d& sensor_from_rig =
+            reconstruction.Rig(rig_id).SensorFromRig(sensor_id);
+        sensor_from_rig.translation() = sensor_from_rig.rotation() * -center;
+        break;
       }
-      Rigid3d& sensor_from_rig =
-          reconstruction.Rig(rig_id).SensorFromRig(sensor_id);
-      sensor_from_rig.translation() = sensor_from_rig.rotation() * -center;
-      break;
     }
   }
 }
--- a/src/colmap/estimators/global_positioning.cc
+++ b/src/colmap/estimators/global_positioning.cc
@@ -227,16 +227,10 @@ void GlobalPositioner::AddPoint3DToProblem(point3D_t point3D_id,
       Rig& rig = reconstruction.Rig(rig_id);
       Rigid3d& cam_from_rig = rig.SensorFromRig(image.CameraPtr()->SensorId());
 
-      const bool keep_rig_fixed =
-          !cam_from_rig.translation().hasNaN() ||
-          !options_.refine_sensor_from_rig;
-      if (keep_rig_fixed) {
-        Eigen::Vector3d cam_in_rig = cam_from_rig.translation();
-        if (cam_in_rig.hasNaN()) {
-          cam_in_rig.setZero();
-        }
+      if (!cam_from_rig.translation().hasNaN()) {
         const Eigen::Vector3d cam_from_rig_dir =
-            image.CamFromWorld().rotation().inverse() * cam_in_rig;
+            image.CamFromWorld().rotation().inverse() *
+            cam_from_rig.translation();
 
         ceres::CostFunction* cost_function =
             RigBATAPairwiseDirectionConstantRigCostFunctor::Create(
@@ -248,8 +242,11 @@ void GlobalPositioner::AddPoint3DToProblem(point3D_t point3D_id,
                                    frame_centers_[image.FrameId()].data(),
                                    &scale);
       } else {
-        // If the cam_from_rig contains nan values, it needs to be re-estimated.
-        // Initialize cams_in_rig_ if not already done.
+        // NaN translation means the sensor's cam_from_rig must be
+        // re-estimated, which requires refine_sensor_from_rig=true.
+        THROW_CHECK(options_.refine_sensor_from_rig)
+            << "sensor_from_rig has NaN translation but "
+               "refine_sensor_from_rig=false";
         const sensor_t sensor_id = image.CameraPtr()->SensorId();
         if (cams_in_rig_.find(sensor_id) == cams_in_rig_.end()) {
           // Will be initialized to random values in ParameterizeVariables().
@@ -432,19 +429,16 @@ void GlobalPositioner::ConvertBackResults(Reconstruction& reconstruction) {
     rig_from_world.translation() = rig_from_world.rotation() * -center;
   }
 
-  if (options_.refine_sensor_from_rig) {
-    // Convert optimized cam_in_rig back to sensor_from_rig translations.
-    for (const auto& [sensor_id, center] : cams_in_rig_) {
-      // Find the rig containing this sensor.
-      for (const auto& [rig_id, rig] : reconstruction.Rigs()) {
-        if (!rig.HasSensor(sensor_id)) {
-          continue;
-        }
-        Rigid3d& sensor_from_rig =
-            reconstruction.Rig(rig_id).SensorFromRig(sensor_id);
-        sensor_from_rig.translation() = sensor_from_rig.rotation() * -center;
-        break;
+  for (const auto& [sensor_id, center] : cams_in_rig_) {
+    // Find the rig containing this sensor.
+    for (const auto& [rig_id, rig] : reconstruction.Rigs()) {
+      if (!rig.HasSensor(sensor_id)) {
+        continue;
       }
+      Rigid3d& sensor_from_rig =
+          reconstruction.Rig(rig_id).SensorFromRig(sensor_id);
+      sensor_from_rig.translation() = sensor_from_rig.rotation() * -center;
+      break;
     }
   }
 }
--- a/src/colmap/estimators/global_positioning_test.cc
+++ b/src/colmap/estimators/global_positioning_test.cc
@@ -196,10 +196,10 @@ TEST(GlobalPositioning, RefineSensorFromRigFalsePreservesRig) {
   for (const auto& [rig_id, rig] : reconstruction.Rigs()) {
     for (const auto& [sensor_id, sensor_from_rig_after] : rig.NonRefSensors()) {
       ASSERT_TRUE(sensor_from_rig_after.has_value())
-          << "rig=" << rig_id << " sensor=" << sensor_id.id;
+          << "rig_id=" << rig_id << ", sensor_id=" << sensor_id.id;
       const auto& sensor_from_rig_before = snapshot.at({rig_id, sensor_id});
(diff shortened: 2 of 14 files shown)


src/colmap/estimators/global_positioning.cc
```diff
@@ -227,10 +227,16 @@ void GlobalPositioner::AddPoint3DToProblem(point3D_t point3D_id,
       Rig& rig = reconstruction.Rig(rig_id);
       Rigid3d& cam_from_rig = rig.SensorFromRig(image.CameraPtr()->SensorId());
 
-      if (!cam_from_rig.translation().hasNaN()) {
+      const bool keep_rig_fixed =
```
B1ueber2y: This logic is not right imo. Should instead do THROW_CHECK(!cam_from_rig.translation().hasNaN()) if the refine_sensor_from_rig is set to false.

src/colmap/estimators/global_positioning.cc
```diff
@@ -426,17 +432,19 @@ void GlobalPositioner::ConvertBackResults(Reconstruction& reconstruction) {
     rig_from_world.translation() = rig_from_world.rotation() * -center;
   }
 
-  // Convert optimized cam_in_rig back to sensor_from_rig translations.
-  for (const auto& [sensor_id, center] : cams_in_rig_) {
-    // Find the rig containing this sensor.
-    for (const auto& [rig_id, rig] : reconstruction.Rigs()) {
-      if (!rig.HasSensor(sensor_id)) {
-        continue;
+  if (options_.refine_sensor_from_rig) {
```
B1ueber2y: Related to the other comments, once we defend against the NaN value when refine_cam_from_rig is false, this change in ConvertBackResults becomes unnecessary, as cams_in_rig_ should not be populated in the GP logic.

B1ueber2y (commented):
Thanks for the PR! This is an important fix to have. 

A key thing that is missing is to propagate these changes to the global mapper. We should have a centralized option ``refine_cam_from_rig`` and propagate it to the three suboptions in the existing ``InitializeOptions`` method. This will also mean 1) do not set the value of ba_refine_cam_from_rig in the construction. 2) dropping the ba_refine_cam_from_rig interface in option manager and instead use ``refine_cam_from_rig``.

B1ueber2y:
To make the design even more consistent we can add some new interfaces like done here: https://github.com/colmap/colmap/blob/main/src/colmap/controllers/incremental_pipeline.h#L196-L199

TannerGilbert:
Tried to address the changes and make the global mapper consistent with the incremental pipeline based on the code-snippet you linked above

src/colmap/estimators/global_positioning.cc
```diff
@@ -242,8 +242,11 @@ void GlobalPositioner::AddPoint3DToProblem(point3D_t point3D_id,
                                    frame_centers_[image.FrameId()].data(),
                                    &scale);
       } else {
-        // If the cam_from_rig contains nan values, it needs to be re-estimated.
-        // Initialize cams_in_rig_ if not already done.
+        // NaN translation means the sensor's cam_from_rig must be
+        // re-estimated, which requires refine_sensor_from_rig=true.
+        THROW_CHECK(options_.refine_sensor_from_rig)
+            << "sensor_from_rig has NaN translation but "
```
ahojnnes, line 248: nit: Here it could be helpful to log out the image_id for debugging.

src/colmap/estimators/global_positioning_test.cc
```diff
@@ -137,5 +140,75 @@ TEST(GlobalPositioning, MultiCameraRig) {
...
+  for (const auto& [frame_id, _] : reconstruction.Frames()) {
+    Frame& frame = reconstruction.Frame(frame_id);
+    frame.SetRigFromWorld(
+        Rigid3d(frame.RigFromWorld().rotation(), Eigen::Vector3d::Zero()));
+  }
+
+  // Snapshot the rig BEFORE GP.
+  std::map<std::pair<rig_t, sensor_t>, Rigid3d> snapshot;
+  for (const auto& [rig_id, rig] : reconstruction.Rigs()) {
+    for (const auto& [sensor_id, sfr] : rig.NonRefSensors()) {
```
ahojnnes:
```suggestion
    for (const auto& [sensor_id, sensor_from_rig] : rig.NonRefSensors()) {
```

src/colmap/estimators/global_positioning_test.cc
```diff
@@ -137,5 +140,75 @@ TEST(GlobalPositioning, MultiCameraRig) {
...
+  options.use_gpu = false;
+  options.random_seed = 42;
+  options.solver_options.minimizer_progress_to_stdout = false;
+  options.refine_sensor_from_rig = false;
+
+  ASSERT_TRUE(RunGlobalPositioning(options, pose_graph, reconstruction));
+
+  // Every sensor_from_rig must match the snapshot exactly.
+  for (const auto& [rig_id, rig] : reconstruction.Rigs()) {
+    for (const auto& [sensor_id, sfr_after] : rig.NonRefSensors()) {
```
ahojnnes:
```suggestion
    for (const auto& [sensor_id, sensor_from_rig_after] : rig.NonRefSensors()) {
```

src/colmap/estimators/global_positioning_test.cc
```diff
@@ -137,5 +140,75 @@ TEST(GlobalPositioning, MultiCameraRig) {
...
+      ASSERT_TRUE(sfr_after.has_value())
+          << "rig=" << rig_id << " sensor=" << sensor_id.id;
+      const auto& sfr_before = snapshot.at({rig_id, sensor_id});
```
ahojnnes:
```suggestion
      const auto& sensor_from_rig_before = snapshot.at({rig_id, sensor_id});
```

src/colmap/estimators/global_positioning_test.cc
```diff
@@ -137,5 +140,75 @@ TEST(GlobalPositioning, MultiCameraRig) {
...
+      EXPECT_EQ(sfr_after->rotation().coeffs(), sfr_before.rotation().coeffs())
```
ahojnnes: `EXPECT_THAT(sensor_from_rig_after, Rigid3dEq(sensor_from_rig_before));`

src/colmap/estimators/rotation_averaging.cc
```diff
@@ -521,25 +525,27 @@ bool InitializeRigRotationsFromImages(
...
+    for (auto& [camera_id, rig_id_and_samples] : cam_from_rig_samples) {
+      auto& [rig_id, samples] = rig_id_and_samples;
+      const auto sensor_id = sensor_t(SensorType::CAMERA, camera_id);
+      const auto existing =
+          reconstruction.Rig(rig_id).MaybeSensorFromRig(sensor_id);
+      // If sensor_from_rig rotation is already known (translation is valid),
+      // preserve the known rotation instead of overwriting with MST-derived
+      // approximation. Reset translation to NaN so downstream code (global
+      // positioning) still estimates it using the variable-rig formulation.
+      if (existing.has_value() && !existing->translation().hasNaN()) {
```
ahojnnes, line 536: I think I asked this before @B1ueber2y and I am still not sure I understand the logic here. If the translation is known, why would we want to re-estimate it? We should just assume it as known, no?

src/colmap/estimators/rotation_averaging.cc
```diff
@@ -521,25 +525,27 @@ bool InitializeRigRotationsFromImages(
...
-          sensor_id, Rigid3d(existing->rotation(), kUnknownTranslation));
-      continue;
+          sensor_id, Rigid3d(cam_from_rig, kUnknownTranslation));
     }
-    weights.resize(samples.size(), 1.0);
-    const Eigen::Quaterniond cam_from_rig =
-        AverageQuaternions(samples, weights);
-    reconstruction.Rig(rig_id).SetSensorFromRig(
-        sensor_id, Rigid3d(cam_from_rig, kUnknownTranslation));
   }
```
ahojnnes, line 556: Does it make sense to check that no rotations are unknown/NaN in the case where we don't refine? Otherwise, we'll have NaNs?

src/colmap/estimators/rotation_averaging_impl.cc
```diff
@@ -113,21 +113,23 @@ size_t RotationAveragingProblem::AllocateParameters(
...
-      if (camera_id_to_param_idx_.find(camera_id) ==
-          camera_id_to_param_idx_.end()) {
-        // Mark for later allocation (actual index set below).
-        camera_id_to_param_idx_[camera_id] = -1;
-        if (cam_from_rig.has_value()) {
-          cam_from_rig_rotations[camera_id] =
-              Eigen::AngleAxisd(cam_from_rig->rotation());
+  if (options_.refine_sensor_from_rig) {
+    for (const auto& [camera_id, rig_id] : camera_id_to_rig_id_) {
+      sensor_t sensor_id(SensorType::CAMERA, camera_id);
```
ahojnnes: const

src/colmap/estimators/rotation_averaging_impl.cc
```diff
@@ -113,21 +113,23 @@ size_t RotationAveragingProblem::AllocateParameters(
...
+      if (reconstruction.Rig(rig_id).IsRefSensor(sensor_id)) continue;
+
+      auto cam_from_rig =
```
ahojnnes: const

ahojnnes (commented):
Thanks for the fix. I am not deep into the weeds of the logic anymore. Left a few comments and questions.

src/colmap/estimators/rotation_averaging.cc
B1ueber2y: This indeed seems buggy and needs a fix. I did not remember having this, and this seems to be something brought in by https://github.com/colmap/colmap/pull/4225. Maybe there was a motivation, or was it from AI?

B1ueber2y: Okay looking more closely, this bug seems to be always there inherited from the original glomap codebase, and https://github.com/colmap/colmap/pull/4225 was actually a partial fix that fixes the rotation part. The translation part still requires another fix.

src/colmap/sfm/global_mapper.h
```diff
@@ -27,12 +27,14 @@ struct GlobalMapperOptions {
   // If not specified, all point colors will be black.
   std::filesystem::path image_path;
 
+  // When false, treat each non-ref sensor's cam_from_rig as a pre-calibrated
+  bool refine_sensor_from_rig = true;
+
   // Options for each component
   RotationEstimatorOptions rotation_averaging;
```
B1ueber2y, line 34: Thank you for the changes. However, now that I am concerned that these options are accessible but do not correspond to the value used in the pipeline. Does it make sense to further align with incremental pipeline to break them to fields, e.g., ra_max_num_l1_iterations, etc. 

Also, with the current design, we cannot properly bind out these new const getter methods to pybind because of the name collapse.

src/colmap/estimators/rotation_averaging.cc
TannerGilbert: I now changed this so known translations aren't set to `NaN` anymore and added a corresponding test-case. Please have a look if I understood your concerns correctly.

src/colmap/estimators/rotation_averaging.cc
TannerGilbert: Added a `else` condition with `THROW_CHECK` similar to what was added for the translation

src/colmap/sfm/global_mapper.h
TannerGilbert: Thanks for the review. I tried to address all other comments but I'm not to sure about how to address this one. Is your concern that a user would set one of the individual options to `false` for example but then use the `global_mapper` and there the option is set to `true` and would overwrite the local one?

Regarding the binding should we address this in a separate PR?

src/colmap/estimators/rotation_averaging.cc
```diff
@@ -521,25 +525,34 @@ bool InitializeRigRotationsFromImages(
...
+    }
+  } else {
+    // Check if rotations are valid.
+    for (const auto& [rig_id, rig] : reconstruction.Rigs()) {
+      for (const auto& [sensor_id, sensor_from_rig] : rig.NonRefSensors()) {
+        THROW_CHECK(sensor_from_rig.has_value() &&
+                    !sensor_from_rig->rotation().coeffs().hasNaN())
+            << "sensor_from_rig has NaN rotation but "
+               "refine_sensor_from_rig=false (rig="
+            << rig_id << " sensor=" << sensor_id.id << ")";
```
ahojnnes: uber nit for consistency - same for other instances in this PR.
```suggestion
               "refine_sensor_from_rig=false (rig_id="
            << rig_id << ", sensor_id=" << sensor_id.id << ")";
```

ahojnnes (approved):
Thanks a lot. Looks good to me. I'll let @B1ueber2y do the final sign off.

src/colmap/sfm/global_mapper.h
B1ueber2y: The concern was that ``option.rotation_averaging`` support both setter and getter, but its actually not consistent with ``option.RotationAveraging()``, which can be very confusing if someone does for example ``global_mapper.RotationAveraging(options.rotation_averaging)``, which is actually incorrect in the current setup as refine_sensor_from_rig is not propagated in this field. 

A clean fix is to follow the ``IncrementalPipelineOptions`` and expand these struct to some small option fields like ``ra_xxx``, ``gp_xxx``, ``ba_xxx``, and only let the getter interface such as ``RotationAveraging()`` handle this. This can also benefit from moving ``Solve()`` to the upper level of ``GlobalMapperController`` to align with the design of incremental mapper controller.

I agree that this should be addressed in a separate PR. Let us know if you would love to take this : )

src/colmap/sfm/global_mapper_test.cc
```diff
@@ -152,5 +152,18 @@ TEST(GlobalMapper, WithNoiseAndOutliers) {
                                  /*num_obs_tolerance=*/0.02));
 }
 
+TEST(GlobalMapperOptions, RefineSensorFromRigPropagatesToSubOptions) {
```
B1ueber2y, line 155: Following up to the other discussion, the behavior of this test aligns with our expectation, but is not a good design in general imo. We would love to drop this test in the other PR.

B1ueber2y (approved):
Thank you!

src/colmap/sfm/global_mapper_test.cc
TannerGilbert: Makes sense. Thanks for the feedback

TannerGilbert:
Implemented the nit naming change and merged main.

src/colmap/sfm/global_mapper.h
TannerGilbert: Yeah I can try this. Currently also working on adding the prior functionality to the global mapper. Will create a initial PR for this after this one is merged to ask for some feedback. Probably would tackle the update to make things consistent with the incremental pipeline after that.

PR #4379, opened 2026-05-06
Add GPU selection to Caspar BA

tordnat:
- cuda.h/cc: Split SetBestCudaDevice into FindBestCudaDevice() (returns index) + SetBestCudaDevice (sets it). Caspar needs the index directly to pass to CreateSolver.
- Added int gpu_index = -1 to CasparBundleAdjustmentOptions.
- bundle_adjustment_caspar.cc: Use FindBestCudaDevice() when no option provided.
-  Registered BundleAdjustmentCaspar.gpu_index in the option manager and wired it into the BA widget UI. For the pipeline, added ba_caspar_gpu_index to IncrementalPipelineOptions, propagated it in local and global BA methods.

### DIFF
--- a/src/colmap/controllers/incremental_pipeline.h
+++ b/src/colmap/controllers/incremental_pipeline.h
@@ -139,6 +139,9 @@ struct IncrementalPipelineOptions {
   bool ba_use_gpu = false;
   std::string ba_gpu_index = "-1";
 
+  // GPU device index for Caspar bundle adjustment (-1 = auto-select).
+  int ba_caspar_gpu_index = -1;
+
   // Bundle adjustment solver backend for local bundle adjustment.
   BundleAdjustmentBackend ba_local_backend = BundleAdjustmentBackend::CERES;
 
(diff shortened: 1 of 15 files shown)


tordnat:
@ahojnnes Let me know if you can test this on your multi-GPU rig. I'm very curious to see the performance on a larger dataset.

src/colmap/controllers/incremental_pipeline.h
```diff
@@ -139,6 +139,9 @@ struct IncrementalPipelineOptions {
   bool ba_use_gpu = false;
   std::string ba_gpu_index = "-1";
```
ahojnnes: Can we just reuse this one? Potentially make it an int instead of a string.

ahojnnes (commented):
Thanks, looks good to me. Will test this later today on a multi-GPU node.

ahojnnes:
@tordnat Changed it back to a string option due to some breaking changes. Tested the latest version on my multi-GPU node and it works well.

ahojnnes (approved):

Issue #4376, opened 2026-05-05
GLOMAP on ETH3D terrains multi-camera rig: significantly worse ground-truth rotation accuracy than COLMAP despite similar sparse reconstruction quality


DuVogel87:
<h2>Summary</h2>

<p>
I am comparing COLMAP and GLOMAP sparse reconstruction quality on the ETH3D low-res many-view
<code>terrains</code> dataset with a synchronized 4-camera rig setup.
</p>

<p>
GLOMAP reconstructs the subset successfully and is fast. The sparse reconstruction size and reprojection error are reasonably close to COLMAP.
However, the aligned ground-truth rotation error is much worse for GLOMAP, while the position error is relatively close.
</p>

<p>
I would like to understand whether this is expected for GLOMAP on calibrated multi-camera rig datasets, or whether I may be missing a setting / conversion step related to rig calibration, <code>sensor_from_rig</code>, relative poses, rotation averaging, global positioning, or bundle adjustment.
</p>

<hr>

<h2>Dataset</h2>

<ul>
  <li><strong>Dataset:</strong> ETH3D low-res many-view training, sequence <code>terrains</code></li>
  <li><strong>Images:</strong> undistorted PINHOLE</li>
  <li><strong>Rig:</strong> 4 synchronized cameras</li>
  <li><strong>Full dataset:</strong> 165 frames / 660 images</li>
  <li><strong>Subset used here:</strong> 50 evenly sampled synchronized frames / 200 images</li>
  <li><strong>Masks:</strong> none</li>
  <li><strong>Rig setup:</strong> generated via <code>rig_configurator</code></li>
  <li><strong>GT evaluation:</strong> aligned pose metrics against the ETH3D calibration / ground-truth reference</li>
</ul>

<hr>

<h2>Builds used</h2>

<ul>
  <li><strong>COLMAP Ceres reference:</strong> local Ceres baseline build</li>
  <li><strong>COLMAP Caspar:</strong> local Caspar f32 build, generated/fixed-rig setup</li>
  <li><strong>GLOMAP:</strong> local branch <code>glomap-eth3d-relpose-import-fix-20260504-141430</code>, commit <code>0edb1b8435e0f9a594318908b81a31f078a51bf7</code></li>
  <li><strong>GLOMAP notes:</strong> this local build includes ETH3D relative-pose import fixes and final external Caspar BA support</li>
</ul>

<hr>

<h2>Subset results</h2>

<table>
  <thead>
    <tr>
[...]

PR #4380, opened 2026-05-06
Caspar: Raise error on unsupported multi-camera rig

tordnat:
Address #4376

### DIFF
--- a/src/colmap/estimators/bundle_adjustment_caspar.cc
+++ b/src/colmap/estimators/bundle_adjustment_caspar.cc
@@ -255,10 +255,11 @@ class CasparBundleAdjuster : public BundleAdjuster {
     // project(CamFromRig * RigFromWorld * point, calib) as a variable-pose
     // factor. See caspar/caspar_generate.py
     if (!image.IsRefInFrame() && pose_var) {
-      LOG(WARNING) << "Image " << image.ImageId()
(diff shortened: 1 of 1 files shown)


B1ueber2y:
>> Similarly applied to the other warnings in the implementation like https://github.com/colmap/colmap/blob/main/src/colmap/estimators/bundle_adjustment_caspar.cc#L247, where I think we should also throw.

Thanks. Can you also address this?

B1ueber2y:
Not sure if there are other such cases.

src/colmap/estimators/bundle_adjustment_caspar.cc
```diff
@@ -255,10 +255,11 @@ class CasparBundleAdjuster : public BundleAdjuster {
     // project(CamFromRig * RigFromWorld * point, calib) as a variable-pose
     // factor. See caspar/caspar_generate.py
     if (!image.IsRefInFrame() && pose_var) {
-      LOG(WARNING) << "Image " << image.ImageId()
-                   << ": non-ref rig camera with variable pose; CASPAR cannot "
-                      "chain CamFromRig * RigFromWorld into a single pose "
```
B1ueber2y, line 260: I think its not that caspar cannot do it, but rather we currently do not support it (due to shared memory issues). Maybe improve a bit the log here : )

tordnat:
Addressed in #4385

PR #4383, opened 2026-05-06
Avoid rehashing by reserving unordered map + cosmetics

ahojnnes:
Some profiling revealed that creating the factors contributes significant runtime. This shaves off a few percent from the runtime by avoiding rehashing. cc: @tordnat

### DIFF
--- a/src/colmap/estimators/bundle_adjustment_caspar.cc
+++ b/src/colmap/estimators/bundle_adjustment_caspar.cc
@@ -5,6 +5,8 @@
 #include "colmap/scene/camera.h"
 #include "colmap/scene/image.h"
 #include "colmap/sensor/models.h"
+
+#include <optional>
 #ifdef CASPAR_ENABLED
 #include "colmap/estimators/caspar/caspar_model_adapter.h"
 #endif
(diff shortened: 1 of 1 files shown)


B1ueber2y (approved):

tordnat:
Thanks! Did you measure how it scales? Hopefully it doesn't account for that much runtime. Some of the routines might also be embarrassingly parallel.

ahojnnes:
The profiler showed me that the CreateFactors() takes ~3% of total runtime for a reconstruction of ~500 images.

ahojnnes:
@tordnat Just asked Claude about optimization opportunities in Caspar. No idea if this makes sense, simply FYI:
```
The biggest single bottleneck. caspar::GraphSolver::DoUpdateR + DoUpdateRFirst spend ~12.6% of cycles in blocking cudaMemcpy DtoH to read R back to host every iteration.                   
                                                                                                                                     
  Options, smallest to largest scope:                                                                                                
  - Pinned host memory for the residual destination (cudaMallocHost/cudaHostAlloc). Drop-in change, typically 2–3× faster DtoH bandwidth.                                                                                                                         
  - cudaMemcpyAsync + event so the readback overlaps with the next kernel (DoBeta/DoNormalize). Host inspects R one iter behind.
  - GPU-side convergence check: a tiny kernel evaluates ‖R‖ and writes a single byte/flag; only copy R to host every N iterations or on convergence. Eliminates the per-iter readback entirely.                                                                         
                                                                                                                                     
  These are caspar-internal changes, but the payoff is real and isolated.
```

tordnat:
Thanks! I'll check out the recommendations

Issue #2976, opened 2024-12-02
Document how to populate the database with pose priors and use the PosePriorBundleAdjuster


sarlinpe:
As requested by @mattiasmar in https://github.com/colmap/colmap/pull/2660#issuecomment-2511676995, having a documentation page on this would be useful. Which dataset has GPS data for enough images to demo and evaluate this feature?

bbzh:
Most public driving datasets like KITTI and Nuscenes have GPS data. They are good to test this feature.

StonerLing:
UAV images are typical datasets that carry GPS information. Some public datasets can be found there: [pix4d cloud demo](https://cloud.pix4d.com/demo), [wingtra datasets](https://wingtra.com/mapping-drone-wingtraone/aerial-map-types/), [arcgis sample data](https://doc.arcgis.com/en/drone2map/latest/help/sample-data.htm)

ahojnnes:
For people to unblock themselves in the meanwhile, I went through the steps with the arcgis sample above:
```
# Since the EXIF data contains no information about the measurement precision,
# we need to manually specify it. Notice that, for simplicity, we use the same
# precision in x/y/z but typically altitude direction has larger variance than
# lat/lon when dealing with GPS measurements.
prior_position_std=2 # meters

# Automatically extracts pose priors from EXIF data, if available.
colmap feature_extractor \
    --database_path Small_Buildings/database.db \
    --image_path Small_Buildings/images

# Alternative for larger datasets: spatial_matcher
colmap exhaustive_matcher \
    --database_path Small_Buildings/database.db

mkdir -p Small_Buildings/sparse
colmap pose_prior_mapper \
    --database_path Small_Buildings/database.db \
    --image_path Small_Buildings/images \
    --output_path Small_Buildings/sparse \
    --prior_position_std_x $prior_position_std \
    --prior_position_std_y $prior_position_std \
    --prior_position_std_z $prior_position_std \
    --overwrite_priors_covariance 1

# COLMAP reconstructs the model in normalized coordinates. If you want to
# convert it to GPS coordinates, you need to perform a final model alignment.
# Notice that the COLMAP viewer cannot visualize models effectively, if they
# are not normalized, so don't attempt to view the output in COLMAP but use
# another viewer e.g. Meshlab by converting the output using the model_converter.
mkdir -p Small_Buildings/sparse-aligned
colmap model_aligner \
    --input_path Small_Buildings/sparse/0 \
[...]

PR #4384, opened 2026-05-06
feat: pose prior support in global mapper

TannerGilbert:
## Summary

Wires pose priors (from the `pose_priors` table) into GLOMAP's BA, GP, and iterative retriangulation + refinement, so the global mapper can use GPS priors the same way the incremental mapper already does.

## Discussion / future work

This PR plumbs `use_prior_position` into BA and gates the post-GP `Normalize()`, but the GP solve itself does not yet consume position priors. A few follow-ups that would make the prior more useful:

1. **Prior-position residual inside GP.** Currently GP solves freely and BA drags the result back toward the priors. Adding a per-frame prior residual (CAUCHY-wrapped, optionally weighted by prior covariance) inside `GlobalPositioner` would converge directly to a prior-consistent solution, removing the asymmetry where BA respects priors but GP doesn't.
2. **Warm-start `frame_centers_` from priors** in `GlobalPositioner::InitializeRandomPositions` instead of random 100m vectors when priors are available


ahojnnes:
Thank you very much, LGTM after a first look and will make a more thorough pass later.

@lpanaf I remember you tried the warm-starting in the past. Any insights or guidance on what approach to take for (2) in the PR description?

lpanaf:
Thanks for the PR! I have a question regarding the algorithm: results after global positioning often vary significantly in scale and orientation than the GPS results. Since I don't see an explicit alignment step, I'm curious if this will function as intended.

This is precisely why adding a warm start (or using prior) for global positioning is non-trivial; it essentially requires the rotation alignment to be known or estimated beforehand.

TannerGilbert:
Thanks for the input. What you're saying makes sense. Unlike for the incremental mapper here for the global mapper we should definitely take the scale into account. Will look into this in a bit.

ahojnnes:
@TannerGilbert This could be a nice feature to add for the next release. Do you have time wrapping this up? Thank you!

TannerGilbert:
@ahojnnes sorry for the delay. Will try to pick this back up next week.

ahojnnes:
@TannerGilbert Gentle ping about whether you intend to look at this? I'll create a new 4.1 release in the next few days and this is at the edge of making it still into this release or the next one. Thanks!

TannerGilbert:
@ahojnnes sorry for not getting to this. I'm currently finishing up my semester project so I'll not have time for this until at least the end of this week

TannerGilbert:
@ahojnnes I finally got the time to look into this again and I thought about how to use the priors in global positioning itself (not just BA) and came up with the following approach for now.

The tricky part (like @lpanaf pointed out) is that global positioning only fixes the relative geometry — the result is off by an unknown scale/rotation/translation, and the rotation comes from rotation averaging in some arbitrary frame. So just adding a `‖camera - prior‖` residual doesn't really work, since the cameras and the GPS priors aren't in the same frame yet.

What I ended up doing is adding that unknown transform as an explicit Sim3 "gauge" that gets optimized together with GP:
1. run the normal translation averaging to get the relative structure
2. robustly fit the Sim3 from the frame centers to the priors
3. add a per-frame prior residual through that gauge (`scale*R*center + t - prior`, Cauchy + covariance weighted) and solve again
4. transform everything into the prior frame at the end

It's basically the same align-then-penalize trick the existing pose-prior BA already does, just at the GP stage. I also exposed `GlobalMapper.use_prior_position` (+ robust loss / loss scale) on the CLI and in pycolmap.

To check it actually works I ran it on the Baja California ArcGIS sample from #2976. Full pipeline with priors on, comparing the camera centers to the GPS priors (so this is agreement with the input GPS, not real ground truth). It comes out basically on par with the incremental `pose_prior_mapper`:

| mapper | reg | reproj [px] | GPS RMSE [m] | horiz [m] | vert [m] |
|---|---:|---:|---:|---:|---:|
| global (this PR)                  | 610 | 0.868 | 0.356 | 0.285 | 0.214 |
| incremental (`pose_prior_mapper`) | 610 | 0.901 | 0.347 | 0.286 | 0.197 |
| global, prior off                 | 120 |   —   | 182.9 |   —   |   —   |

And to make sure the prior is really doing something in GP and not just in BA: if I turn BA and retriangulation off, GP-only lands also ~0.3 m of the GPS with the prior on, vs ~183 m with it off.

One thing that's still missing: in GP the prior only constrains the reference sensor of each frame. If a prior is attached to a non-reference camera in a multi-cam rig, GP skips it for now. Happy to add full rig support in a follow-up if you think it's worth it.

Global mapper result: Screenshot from 2026-06-29 16-44-49
Incremental mapper result: Screenshot from 2026-06-29 17-13-19

ahojnnes:
Fantastic, thanks for the update. The numbers look quite promising. I'll make a more detailed pass in the next days. Regarding the pose priors for non-reference cameras: It's probably less critical for the GPS-based pose prior scenario but could become more relevant for something like OptiTrack. If it's not too much work, it could be nice to support it.

lpanaf:
Hi @TannerGilbert , thanks for the PR! I tested this locally and sent a quick email to your address to discuss a potential variation on the design direction and next steps. Talk soon!

ichsan2895:
I think this PR combined with https://github.com/colmap/colmap/pull/4619 will be a great addition 👍

src/colmap/estimators/global_positioning.cc
```diff
@@ -418,6 +455,111 @@ void GlobalPositioner::ParameterizeVariables(Reconstruction& reconstruction) {
...
+    const Image& image = reconstruction.Image(pose_prior.corr_data_id.id);
+    if (!image.IsRefInFrame()) {
+      continue;
+    }
+    const auto it = frame_centers_.find(image.FrameId());
+    if (it == frame_centers_.end() ||
+        !problem_->HasParameterBlock(it->second.data())) {
+      continue;
+    }
+    const Eigen::Matrix3d position_cov =
```
StonerLing, line 515: Just a thought, hope it helps.

Since the BATA residual is defined in normalized image coordinates, should we also consider covariance weighting with `σ = 1 / f` assuming 1-pixel observation uncertainty? Otherwise, the BATA and prior position costs may not be properly balanced. The reprojection residual in `PosePriorBundleAdjuster` implicitly assumes unit standard deviation.

sarlinpe: This is now being addressed in https://github.com/colmap/colmap/pull/4765 :)

PR #4385, opened 2026-05-07
Add rig support to Caspar BA

tordnat:
Extends Caspar BA to correctly handle multi-camera rig datasets. sensor_from_rig is added as a ConstantShared parameter to all existing kernels. Identity is passed for single-camera datasets (some overhead), so there is no behavioral change for the common case. Non-ref rig cameras, previously skipped with a warning, are now included in the optimization. refine_sensor_from_rig = true remains unsupported.

TL;DR:
- Rig support with constant cam from rig now working
- *refine_sensor_from_rig=false*, and remains unsupported
- ok performance on single track trajectories,  great on large internet collections of data (1DSfM)
- Improved setup time in Caspar
- Convergence tests added
- Added BA backend options to automatic reconstruction

Edit:
Comment on accuracy in single-track trajectories: Doing one final global pass with Ceres resulted in AUC@1.0 50% of that of Ceres in challenging scenes. This could be considered as an option. The degradation in accuracy is due to Caspar using Conjugate Gradient, which make intrinsics challenging to refine in long camera chains.


ahojnnes:
Thank you, looks good to me. If I understand correctly, this now incurs an extra SE3 composition for each residual evaluation no matter if the sensor_from_rig pose is constant/trivial or not. Since you say the problem is compute bound, I am wondering whether this would incur any significant runtime overhead?

tordnat:
Yes. That's correct. There may be some clever optimisation avoiding the transform in the identity case, but I'm not sure. Potentially, we could generate kernels for the trivial case as well, however, this would double the dispatch and generated code.

tordnat:
I could try to implement some masking here to avoid the composition by passing a flag `rig_trivial`

tordnat:
Seems the safest way is to generate the trivial case as well. I'll make a rough implementation and let you decide if it's too messy of an approach.

ahojnnes:
The main question is whether the runtime gain is worth the extra ~100K lines of generated code. Do you have any sense of how much slowdown we get by doing the unnecessary SE3 composition in the trivial case?

tordnat:
Got Claude to write some plots to compare the runtime difference of having separate rig factors or just using the combined one for both trivial rig and non-trivial rig. I find the results a bit confusing. Seems having one combined factor for trivial and non-trivial rig is better? Maybe this is because of the extra dispatch code. 

image

Edit: Also had to change these parameters:
```c++
  double solver_rel_decrease_min = 0.9;
  int solver_iter_max = 200;
  int pcg_iter_max = 50;
```

tordnat:
Tried to plot this a bit cleaner, but seems having separate factors wins in larger and denser reconstructions.

image

ahojnnes:
Thanks for the plots. Interesting results. Seems like there is no clear winner and not a clear win for the extra complexity?

tordnat:
Yes, I'll revert to the previous version. I think our time is better spent optimizing the factor creation. I'll do some profiling today and see if I can make it faster, or at least figure out the main bottleneck. I wonder if some of it could be parallelized.

DuVogel87:
Thanks for working on this. I have been testing Caspar BA locally on a synchronized dual-camera THIN_PRISM_FISHEYE rig dataset, and this PR is very relevant to what I observed.

A few notes that may be useful:

- Supporting constant `sensor_from_rig` for multi-camera rigs is important. In my tests, silently skipping or not properly handling the non-reference rig camera can still produce a reconstruction that looks plausible, but it is not reliable for downstream use.
- The most stable production-style variant I found was fixed intrinsics, fixed `sensor_from_rig`, fixed rig/image poses, and optimizing points3D only.
- I also tested an active-pose variant with fixed intrinsics and fixed `sensor_from_rig`. It can be made numerically sane, but in my tests it did not clearly improve over the points-only variant, so I would not make it the default without stronger validation.
- For THIN_PRISM_FISHEYE, real-data validation was important. Synthetic tests alone were not enough. I had to check residual parity, point Jacobian parity, pose Jacobian parity, aggregate `JtR`, and that THIN_PRISM observations were actually included rather than skipped.
- One issue I hit in generated THIN_PRISM_FISHEYE factors was that tangential/thin-prism distortion must be applied after the fisheye mapping, in fisheye coordinates, not directly in normalized pinhole coordinates.

Overall, this PR seems directionally aligned with what I found: constant `sensor_from_rig` support for multi-camera rigs is important, while keeping `refine_sensor_from_rig=true` unsupported for now seems reasonable.

Happy to share more details about the validation/parity checks if useful.

tordnat:
@ahojnnes keeping this a draft for now, but hoping to finish it early next week. I'm implementing some new benchmarking tools to make it easier to compare solvers and validate this implementation. Currently, the benchmarking tools do not report loss over time. This would be a much better metric than time to convergence since different solvers have different exit criteria.

DuVogel87:
Thanks for working on this. This matches something I ran into while testing calibrated multi-camera rig data locally.

A few validation points that were useful in my experiments:

- It is important to explicitly verify that non-reference rig sensors are included and not silently skipped.
- `sensor_from_rig` being fixed/constant is the main production case I care about. Keeping `refine_sensor_from_rig=true` unsupported for now seems reasonable.
- I found it useful to test invariants after BA:
  - intrinsics preserved
  - `sensor_from_rig` preserved
  - rig/frame definitions preserved
  - expected pose variables fixed or active depending on the test
  - observation counts unchanged unless explicitly expected
- For generated factors, synthetic parity tests were useful but not sufficient. A small real multi-camera rig subset caught issues that synthetic tests did not.
- The checks that helped most were residual parity, point Jacobian parity, pose Jacobian parity, aggregate `JtR` parity, and explicit inclusion checks for all rig sensors.

One small note from THIN_PRISM_FISHEYE testing: projection/convention bugs can look fine in simple tests but fail on real rig data. In my case, real-data parity was necessary to catch a projection-order issue.

Overall, the direction of this PR looks very useful for real multi-camera rig datasets. Happy to share more details about the validation checks if useful.

tordnat:
Convergence tests and runtime tests revealed that adding factors in optimal order is nearly insignificant for runtime. I therefore simplified the factor dispatch code.

tordnat:
Scene: track_length=20 num_frames=200 num_points3D=15000 max_ceres_iterations=100 max_caspar_iterations=200

Bildschirmfoto 2026-05-13 um 18 12 45

Seems like Ceres has quite the setup time. Is this expected? @ahojnnes @B1ueber2y

B1ueber2y:
> > Seems like Ceres has quite the setup time. Is this expected?

The first step of the ceres optimization for bundle adjustment has a bunch of symbolic analysis and reordering logic, so the setup time overhead can make sense in my opinion. However, I also have to admit that I am not familiar with Caspar does the setup.

tordnat:
@ahojnnes @B1ueber2y Now ready for review:
- Rig support with constant cam from rig now working
- Tested substantially on ETH3D datasets, ok performance on single track trajectories, performs very well on large internet collections of data (1DSfM)
- Improved setup time in Caspar due to simplifications 
- Convergence tests added to compare Ceres and Caspar: Caspar demonstrates extremely fast convergence and little setup time
- Added BA backend options to automatic reconstruction 

Edit: 
Comment on accuracy in single-track trajectories: Doing one final global pass with Ceres resulted in AUC@1.0 50% of that of Ceres in challenging scenes. This could be considered as an option. The degradation in accuracy is due to Caspar using Conjugate Gradient, which make intrinsics challenging to refine in long camera chains.

src/colmap/estimators/caspar/caspar_model_adapter.h
```diff
@@ -441,6 +464,8 @@ class SimpleRadialAdapter : public ICasparModelAdapter {
           FIXED_POSE_FIXED_FOCAL_AND_EXTRA_FIXED_PRINCIPAL_POINT:
         s.SetSimpleRadialSplitFixedPoseFixedFocalAndDistortionFixedPrincipalPointNum(
             n);
+        s.SetSimpleRadialSplitFixedPoseFixedFocalAndDistortionFixedPrincipalPointSensorFromRigDataFromStackedHost(
```
B1ueber2y: The function name uses "FocalAndDistortion" while the enum uses "FOCAL_AND_EXTRA". I suggest to fix it from the enum side.

src/colmap/estimators/caspar/caspar_model_adapter.h
```diff
@@ -454,6 +479,8 @@ class SimpleRadialAdapter : public ICasparModelAdapter {
         break;
       case FactorVariant::FIXED_POSE_FIXED_FOCAL_AND_EXTRA_FIXED_POINT:
         s.SetSimpleRadialSplitFixedPoseFixedFocalAndDistortionFixedPointNum(n);
+        s.SetSimpleRadialSplitFixedPoseFixedFocalAndDistortionFixedPointSensorFromRigDataFromStackedHost(
```
B1ueber2y: Same applies here and other places.

src/colmap/estimators/bundle_adjustment_caspar.cc
```diff
@@ -217,147 +239,113 @@ class CasparBundleAdjuster : public BundleAdjuster {
...
-               principal_point_var && !point_var) {
-      v = FactorVariant::FIXED_POSE_FIXED_FOCAL_AND_EXTRA_FIXED_POINT;
-    } else if (!effective_pose_var && focal_and_extra_var &&
-               !principal_point_var && !point_var) {
-      v = FactorVariant::FIXED_POSE_FIXED_PRINCIPAL_POINT_FIXED_POINT;
-    } else {  // pose && !focal_and_extra && !principal_point && !point
-      v = FactorVariant::
-          FIXED_FOCAL_AND_EXTRA_FIXED_PRINCIPAL_POINT_FIXED_POINT;
-    }
+    // 4-bit key: bit3=pose_var, bit2=fae_var, bit1=pp_var, bit0=pt_var.
```
B1ueber2y, line 289: Nice improvement. Thank you!

src/colmap/estimators/bundle_adjustment_caspar.h
```diff
@@ -104,7 +105,7 @@ namespace colmap {
 // round-trip through OptionManager regardless of the float/double build.
 // Also includes GPU index selection option
 struct CasparBundleAdjustmentOptions {
-  int solver_iter_max = 100;
+  int solver_iter_max = 200;
```
B1ueber2y, line 108: Just wondering, how often does the solver reach max iteration? In other words, how does this change affect runtime in practice?

src/colmap/estimators/bundle_adjustment_caspar_test.cc
```diff
@@ -1012,5 +1099,188 @@ TEST(DefaultBundleAdjuster,
   CheckConstantPoint(reconstruction.Point3D(1), orig_reconstruction.Point3D(1));
 }
 
+TEST(DefaultBundleAdjuster, MultiCameraRigMatchesCeres) {
```
B1ueber2y: I personally think this test is unnecessary. We already have tests against the groundtruth in our test setup, and do not really need to test against a different backend.

src/colmap/estimators/bundle_adjustment_caspar_test.cc
```diff
@@ -145,6 +145,95 @@ TEST(DefaultBundleAdjuster, Nominal) {
                                  /*num_obs_tolerance=*/0.0));
 }
 
+TEST(DefaultBundleAdjuster, MultiCameraRigNominal) {
```
B1ueber2y: I think in this test file, we should explicit mention sensor_from_rig being fixed (e.g., ``NominalMultiCameraRigConstantSensorFromRig``) and try to match the naming convention used in the ceres tests. 

Maybe we can also add a small test with sensor_from_rig being enabled and expect a throw, and remove it once it is supported?

src/colmap/estimators/bundle_adjustment_caspar_test.cc
```diff
@@ -516,7 +605,7 @@ TEST(DefaultBundleAdjuster, MergedCalibConvergence) {
 
   EXPECT_THAT(gt_reconstruction,
               ReconstructionNear(reconstruction,
-                                 /*max_rotation_error_deg=*/0.2,
+                                 /*max_rotation_error_deg=*/0.35,
```
B1ueber2y: Is there any reason why this is relaxed?

B1ueber2y (commented):
Thank you for the efforts! Overall looks good to me and I only left some minor comments. Would be nice to update the PR title to mention sensor_from_rig being fixed and update the PR description.

src/colmap/estimators/bundle_adjustment_caspar.h
tordnat: After testing, 100 iterations was a bit too strict in some scenarios causing it to reach max iter a bit too frequently. 200 iterations is better for a more general tune, in my opinion. The runtime is not impacted for most scenarios I've tested.

src/colmap/estimators/caspar/caspar_model_adapter.h
tordnat: Fixed.

src/colmap/estimators/caspar/caspar_model_adapter.h
tordnat: Fixed.

src/colmap/estimators/bundle_adjustment_caspar_test.cc
tordnat: Removed.

src/colmap/estimators/bundle_adjustment_caspar_test.cc
tordnat: Fixed.

src/colmap/estimators/bundle_adjustment_caspar_test.cc
tordnat: No, fixed.

B1ueber2y (approved):
Thanks! LGTM. The formatting is however failing in the CI runs. Can you run the formatting on both C++ and Python with the tools here? https://github.com/colmap/colmap/tree/main/scripts/format

tordnat:
Is there anything remaining here, or is this PR good to go?

benchmark/runtime/bundle_adjustment.cc
```diff
@@ -86,6 +86,7 @@ class BM_BundleAdjustment : public benchmark::Fixture {
     config_.FixGauge(BundleAdjustmentGauge::TWO_CAMS_FROM_WORLD);
 
     options_.print_summary = false;
+    options_.refine_sensor_from_rig = false;
```
ahojnnes: Suggest either removing this, adding a brief comment about why it is disabled, or (ideally) making it configurable with gflags?

benchmark/runtime/bundle_adjustment_convergence.cc
```diff
@@ -0,0 +1,312 @@
...
+      ParseIntFlag(argc, argv, "--max_ceres_iterations=", -1);
+  const int max_caspar_iterations =
+      ParseIntFlag(argc, argv, "--max_caspar_iterations=", -1);
+  const int track_length = ParseIntFlag(argc, argv, "--track_length=", 20);
+  const int num_frames = ParseIntFlag(argc, argv, "--num_frames=", 50);
+  const int num_points3D = ParseIntFlag(argc, argv, "--num_points3D=", 5000);
+  const int num_cameras_per_rig =
+      ParseIntFlag(argc, argv, "--num_cameras_per_rig=", 1);
+  const bool refine_focal_length =
+      ParseBoolFlag(argc, argv, "--refine_focal_length=", false);
```
ahojnnes: You could use the "OptionManager::AddDefaultOption(...)" here instead of manual parsing.

src/colmap/controllers/automatic_reconstruction.h
```diff
@@ -124,6 +125,10 @@ class AutomaticReconstructionController : public Thread {
     // comma, e.g., "0,1,2,3". For single-GPU stages only the first GPU will be
     // used. By default, all available GPUs will be used in all stages.
     std::string gpu_index = "-1";
+
+    // Bundle adjustment solver backend for local and global BA.
+    BundleAdjustmentBackend ba_local_backend = BundleAdjustmentBackend::CERES;
+    BundleAdjustmentBackend ba_global_backend = BundleAdjustmentBackend::CERES;
```
ahojnnes: Suggest simplifying this here in the automatic reconstructor and only exposing a single `ba_backend` option and then setting it to  both local/global in the underlying mappers.

src/colmap/estimators/caspar/caspar_model_adapter.h
```diff
@@ -80,10 +80,10 @@ class ICasparModelAdapter {
 
   // Number of floats in the focal_and_distortion/focal and principal_point
   // node arrays per camera.
-  virtual size_t FocalAndExtraSize() const = 0;
+  virtual size_t FocalAndDistortionSize() const = 0;
```
ahojnnes: nit: colmap calls these "Extra" and not "Distortion". Was there a specific reason to change it now?

src/colmap/estimators/bundle_adjustment_caspar.cc
```diff
@@ -163,36 +140,81 @@ class CasparBundleAdjuster : public BundleAdjuster {
...
+    for (const image_t image_id : config_.Images()) {
+      sorted_image_ids.push_back(image_id);
     }
-  }
+    // Sort camera-first so all factors for the same calibration are contiguous.
+    // This improves float32 numerical quality in the GPU gradient summation.
+    std::sort(sorted_image_ids.begin(),
+              sorted_image_ids.end(),
+              [&](image_t a, image_t b) {
+                const camera_t cam_a = reconstruction_.Image(a).CameraId();
```
ahojnnes: nit: this is probably a bit slower as compared to first extracting `std::vector<std::pair<camera_t, image_t>>` and sorting on it, because `reconstruction.Image()` incurs a hashmap lookup.

ahojnnes (approved):
Thank you, looks good to me. Left a few more minor comments to address, otherwise we are good to go from my side.

Dawars:
If I understand correctly, currently incremental mapper has duplicate/exposed some _ceres_ specific BA params that are later set in global/local BA for BundleAdjustmentOptions.ceres.
These are: `ba_local_function_tolerance, ba_local_max_num_iterations` for local and `ba_global_function_tolerance, ba_global_max_num_iterations, num_threads, ba_min_num_residuals_for_cpu_multi_threading, ba_use_gpu, ba_gpu_index` for global BA.

The rest of the ceres options are not exposed, BundleAdjustmentOptions.ceres cannot be directly set.

https://github.com/colmap/colmap/blob/e6cc736ed17509b411f6f77a27f33b9bde6f3189/src/colmap/controllers/incremental_pipeline.cc#L167-L187

My use-case: I'd like to set the solver type thresholds (e.g. max_num_images_direct_sparse_gpu_solver) without rebuilding the codebase depending on the problem configuration (rigs, fixed intrinsics, etc).

Would this be a good time to unify BA options used in incremental_mapper, etc? Is there another way to set these at runtime?

src/colmap/estimators/caspar/caspar_model_adapter.h
tordnat: Changed them because of @B1ueber2y 's comment. Can revert it to follow the naming convention of the rest of the code base.

tordnat:
@Dawars the issue of propagating parameters is a bit more nuanced. Every solver is different and requires different parameters. Unifying them is therefore not possible without simplification. Unified parameters for solvers is an issue that should be solved, but would require a more sophisticated way of handling these variations across solvers. I would consider it out of scope for this PR (which is already quite large). I suggest you submit an issue on this describing the problem in more detail.

src/colmap/estimators/caspar/caspar_model_adapter.h
tordnat: Reverted it, but I'll let you do the final call [28cb3c1](https://github.com/colmap/colmap/pull/4385/commits/28cb3c1d3d00e8d65671191863b4a0fe7548e661).

ahojnnes (approved):

src/colmap/estimators/caspar/caspar_model_adapter.h
B1ueber2y: I am fine with going either way. My comment for this PR was about the inconsistency of using "extra" and "distortion". But given that we have changed all the naming to extra, it wont be a problem : )

B1ueber2y (approved):
Thanks! Lets not forget to run the formatting to be able to pass the CI.

ahojnnes:
Seems like there is a CI failure preventing this PR from being merged @tordnat

tordnat:
on it.

PR #4386, opened 2026-05-07
Clip extreme pixels in fisheye undistortion

lszhuhaichao:
## Summary

Fix undistortion boundary scaling for fisheye cameras with off-center principal points.

When tracing image borders to determine the undistorted output size, some border pixels can map to camera-space rays with very large norm (`tan(theta)`), especially near the edge of the valid fisheye circle. Those extreme samples can blow up the estimated undistorted image dimensions.

This change adds an optional `max_cam_point_norm` threshold to skip such pathological rays during border tracing.

## Changes

- add `max_cam_point_norm` to `UndistortCameraOptions`
- ignore boundary samples whose camera-space norm exceeds the threshold
- expose the new option in `pycolmap`

## Notes

- the check is disabled by default (`max_cam_point_norm = -1`), so existing behavior is unchanged unless the option is set
- this is intended as a geometry-based safeguard, rather than requiring users to hand-tune an ROI crop

### DIFF
--- a/src/colmap/image/undistortion.cc
+++ b/src/colmap/image/undistortion.cc
@@ -103,6 +103,18 @@ Camera UndistortCamera(const UndistortCameraOptions& options,
   // Scale in order to match the boundary of the undistorted image.
   if (roi_enabled || (camera.model_id != SimplePinholeCameraModel::model_id &&
                       camera.model_id != PinholeCameraModel::model_id)) {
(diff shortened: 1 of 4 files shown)


ahojnnes:
Thank you Haichao. Ideally, we can add a small unit test to cover this new functionality.

ahojnnes (approved):
@lszhuhaichao I added a test for you.

PR #4389, opened 2026-05-10
Refit missing E/F/H in MaybeDecomposeRelativePoses for old databases

### LINKED ISSUE #4387: `global_mapper` crashes when `TwoViewGeometry` has config but missing E/F/H matrices
### Describe the bug

`EstimateTwoViewGeometryPose` in `src/colmap/estimators/two_view_geometry.cc` crashes with a `THROW_CHECK` failure when processing `TwoViewGeometry` records where the configuration type is set (e.g., `CALIBRATED`, `UNCALIBRATED`, `PLANAR`) but the corresponding matrix (`E`, `F`, or `H`) is `std::nullopt`.

This happens when `MaybeDecomposeRelativePoses` iterates over image pairs from older databases that were created by earlier COLMAP versions, where two-view geometry configurations were stored but the actual matrix decomposition data (`E`, `F`, `H`) was not persisted.

### To Reproduce

1. Use an older COLMAP database (e.g., the `south-building` dataset from the COLMAP website).
2. Run `colmap global_mapper` with the database.
3. The command crashes during `MaybeDecomposeRelativePoses` with one of the following errors:

```
[two_view_geometry.cc:528] Check failed: geometry->E.has_value()
```

or

```
[two_view_geometry.cc:539] Check failed: geometry->F.has_value()
```

or

```
[two_view_geometry.cc:552] Check failed: geometry->H.has_value()
```

### Expected behavior

`EstimateTwoViewGeometryPose` should gracefully handle the case where the configuration type is set but the corresponding matrix is missing, returning `false` instead of crashing. The caller (`MaybeDecomposeRelativePoses`) already handles `false` return values by counting the decomposition as failed and continuing with the remaining pairs.

### Actual behavior

COLMAP crashes with a fatal assertion error, making it impossible to run `global_mapper` on databases that lack the full two-view geometry matrix data.

### Environment

- COLMAP version: 4.1.0.dev0 (commit 5e548461)
- OS: Linux
- Dataset: south-building (128 images, database from ~2016)

### Proposed fix

Replace the three `THROW_CHECK` assertions in `EstimateTwoViewGeometryPose` with early `return false` guards:

```cpp
if (geometry->config == TwoViewGeometry::ConfigurationType::CALIBRATED) {
  if (!geometry->E.has_value()) {
    return false;
  }
  // ...
} else if (geometry->config == TwoViewGeometry::ConfigurationType::UNCALIBRATED) {
  if (!geometry->F.has_value()) {
    return false;
  }
  // ...
} else if (...) {
  if (!geometry->H.has_value()) {
    return false;
  }
  // ...
}
```

This is consistent with how other invalid/degenerate configurations are handled elsewhere in the codebase.

ahojnnes:
- Fixes #4387: `global_mapper` crashed on databases written by older COLMAP versions, which stored the two-view geometry `config` (CALIBRATED / UNCALIBRATED / PLANAR / ...) but persisted the corresponding `E`/`F`/`H` matrix as NULL. `EstimateTwoViewGeometryPose` then `THROW_CHECK`ed the missing matrix and aborted.
- `MaybeDecomposeRelativePoses` now refits the missing matrix from the cached inlier matches before calling `EstimateTwoViewGeometryPose`:
  - `CALIBRATED` → `EssentialMatrixEightPointEstimator` on inlier camera rays
  - `UNCALIBRATED` → `FundamentalMatrixEightPointEstimator` on inlier image points
  - `PLANAR` / `PANORAMIC` / `PLANAR_OR_PANORAMIC` → `HomographyMatrixEstimator` (DLT) on inlier image points
- If the inlier set is too small for the estimator's `kMinNumSamples` or the fit returns no model, the pair is counted as a failed decomposition and skipped (no crash).
- Leaves `EstimateTwoViewGeometryPose`'s `THROW_CHECK`s in place: they still catch programmer errors at non-old-DB call sites such as `incremental_mapper_impl.cc`.

### DIFF
--- a/src/colmap/estimators/two_view_geometry.cc
+++ b/src/colmap/estimators/two_view_geometry.cc
@@ -925,6 +925,113 @@ TwoViewGeometry TwoViewGeometryFromKnownRelativePose(
   return geometry;
 }
 
+namespace {
+
+// Fits the E/F/H matrix matching `geometry->config` from the existing
+// inlier matches when it was not persisted in the database (e.g. databases
+// from older COLMAP versions that stored only the configuration). Returns
+// false if no matrix is needed for the configuration, the relevant matrix
+// is already present, or the fit failed (insufficient inliers / degenerate
+// configuration).
+bool MaybeFitMissingTwoViewGeometryMatrix(
+    const Camera& camera1,
+    const std::vector<Eigen::Vector2d>& points1,
+    const Camera& camera2,
+    const std::vector<Eigen::Vector2d>& points2,
+    TwoViewGeometry* geometry) {
+  switch (geometry->config) {
+    case TwoViewGeometry::ConfigurationType::CALIBRATED: {
+      if (geometry->E.has_value()) {
+        return true;
+      }
+      std::vector<Eigen::Vector3d> cam_rays1;
+      std::vector<Eigen::Vector3d> cam_rays2;
+      cam_rays1.reserve(geometry->inlier_matches.size());
+      cam_rays2.reserve(geometry->inlier_matches.size());
+      for (const FeatureMatch& match : geometry->inlier_matches) {
+        const std::optional<Eigen::Vector2d> cam_point1 =
+            camera1.CamFromImg(points1[match.point2D_idx1]);
+        const std::optional<Eigen::Vector2d> cam_point2 =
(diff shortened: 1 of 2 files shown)


B1ueber2y (approved):
Thanks!

Issue #4390, opened 2026-05-11
Caspar GPU BA backend does not support pose priors


braydenhamilton:
The Caspar (cuDSS-accelerated GPU) bundle-adjustment backend currently rejects pose priors at [`src/colmap/estimators/bundle_adjustment.cc:389`](https://github.com/colmap/colmap/blob/main/src/colmap/estimators/bundle_adjustment.cc#L389):

```cpp
case BundleAdjustmentBackend::CASPAR:
#ifdef CASPAR_ENABLED
  LOG(FATAL_THROW) << "Caspar BA backend does not support pose priors";
```

This means GNSS/RTK-anchored workflows can't use the GPU BA path today — they either get Caspar GPU BA without prior anchoring (`global_mapper`) or Ceres CPU BA with priors (`pose_prior_mapper`), never both.

Is there a planned path to wiring pose-prior support into Caspar? Happy to contribute if there's directional guidance on how the maintainers would like that shaped.

braydenhamilton:
Closing — for our use case (V100 GPUs, sm_70) Caspar cannot run regardless of this gap, and the actual GPU-BA blocker we hit is in the Ceres CUDA_SPARSE path with `pose_prior_mapper` (silent fallback for prior-augmented problems). Filing that as a separate issue. Happy for the maintainers to keep this open if it's still useful for Caspar-targeted hardware.

PR #4391, opened 2026-05-11
Caspar BA: pose-prior support (RTK/GNSS-anchored GPU BA)

braydenhamilton:
Resolves the gap flagged in #4390 — Caspar bundle adjustment can now consume pose priors instead of throwing `FATAL_THROW`.

## What this changes

- **New Symforce factor** in `caspar_generate.py` for `SimpleRadial` and `Pinhole`: whitened position-prior residual `sqrt_info * (world_T_cam.t - prior_position)`, mirroring the Ceres-side `AbsolutePosePositionPriorCostFunctor`. Regenerated kernels live under `src/thirdparty/Symforce-Caspar/generated/{f32,f64}/`.
- **Adapter plumbing**: `ICasparModelAdapter` gains `SetPriorFactors(solver, PriorFactorData&)`. `CasparSolverSizing` gains `num_{simple_radial,pinhole}_pose_prior_core`. Each adapter forwards to the generated `Set*PosePriorCore*` methods.
- **`CasparBundleAdjuster`** gets a pose-prior-aware constructor overload. When `pose_priors` is non-empty, `BuildPriorFactors()` walks each prior, looks up the per-model pose index, packs the position and an upper-triangular sqrt-info into the per-model `PriorFactorData`, and `SetupSolverData()` invokes the adapter's `SetPriorFactors` alongside the variant factors.
- **New `CasparPosePriorBundleAdjuster`** wraps the base adjuster with the same alignment + normalisation dance as the Ceres `PosePriorBundleAdjuster`: filter priors → RANSAC-align with `AlignReconstructionToPosePriors` → `reconstruction_.Normalize(/*fixed_scale=*/true)` (critical for the float32 Caspar build) → transform priors into the normalised frame → solve → denormalise.
- **Dispatch** in `CreatePosePriorBundleAdjuster` routes `BundleAdjustmentBackend::CASPAR` to the new wrapper instead of throwing.
- **`PosePriorBundleAdjusterBackendTest`** is extended to instantiate against `CASPAR` when `CASPAR_ENABLED` is on, exercising the synthetic-dataset round-trip across both backends.

## Implementation notes

- `BuildPriorFactors` drops priors for non-ref-frame images with a `VLOG(2)`. Caspar can't yet express the `CamFromRig * RigFromWorld` chain that the Ceres-side `AbsoluteRigPosePositionPriorCostFunctor` handles — matches Caspar's existing limitation for variable-pose factors on non-ref rig cameras.
- Covariance whitening uses `Eigen::LLT`; a non-PD covariance logs a warning and drops the prior. When `position_covariance` is unset, falls back to a 1 m diagonal.
- Two small upkeep changes hitchhike on the regen and need a maintainer eye:
  - `Solve()` in `CasparBundleAdjuster` moves to the `public:` section so the wrapper can call it via composition. It overrides the public virtual `BundleAdjuster::Solve()`, so this is what it should have been all along; calling it through a derived-class reference previously hit access-control.
  - `CreateSolver` in `caspar_model_adapter.h` now calls `cudaSetDevice(device_id)` before constructing the `GraphSolver`. Current upstream Symforce-Caspar (as of `symforce-org/symforce@main`) dropped the `device_id` constructor parameter; the device must be selected externally. This is also why the regenerated `solver.{h,cc}` and `caspar_mappings.{cu,h}` show large diffs.
- The bulk of the diff under `src/thirdparty/Symforce-Caspar/generated/` is cosmetic codegen-style churn from regenerating against current Symforce main (those files are excluded from `scripts/format/c++.sh` precisely because they're machine-generated). The substantive additions are:
  - the 16 new `kernel_*_pose_prior_core_*.{cu,h}` files per precision (32 files total);
  - registration entries in `caspar_mappings.{cu,h}`;
  - new method declarations + bodies in `solver.{h,cc}`.

## Test plan

- `PosePriorBundleAdjusterBackendTest` parameterised across `CERES` and `CASPAR`. Synthetic 7-frame / 100-point dataset, noised by `point2D_stddev=0.5`, `prior_position_stddev=0.05`; accuracy thresholds `max_rotation_error_deg=0.1`, `max_proj_center_error=0.1`.
- Compile-validated on a clean DGX build with `-DCASPAR_ENABLED=ON -DCMAKE_CUDA_ARCHITECTURES=75 -DCeres_DIR=...`; `colmap_estimators` and the full `colmap` binary both link cleanly.
- **I have not been able to run the runtime tests** — my DGX has V100 (compute capability 70), which is below Caspar's minimum supported arch of 75. Would appreciate maintainer help validating the runtime path on Turing+ hardware. The compile is clean and the dispatch + factor wiring matches the Ceres reference, so I'm reasonably confident, but the test ought to be exercised before merge.

## Open questions (mirroring those in #4390)

1. Does the Symforce factor + adapter approach align with how you'd extend Caspar architecturally, or would you prefer a different shape?
2. Is the per-camera-model `PriorFactorData` pool layout (3-float position + 6-float packed sqrt-info per factor) the right granularity, or would you prefer full 3x3 sqrt-info matrices (12 floats) for symmetry with the Ceres `CovarianceWeightedCostFunctor`?
3. Should the `device_id` constructor-param drop be split out as a separate prep PR? I caught it as part of the regen but happy to extract.


braydenhamilton:
Closing — investigation today showed our V100 hardware (sm_70) cannot run Caspar regardless of pose-prior support, and we have a separate Ceres+cuDSS silent-fallback bug in `pose_prior_mapper` (the actual production blocker on V100 hardware) being filed as a new issue. Happy for a maintainer or another contributor with Turing+ access to revisit this PR if useful for Caspar users.

PR #4393, opened 2026-05-11
Fix mesh_texturer occlusion check edge case

Dawars:
The occlusion check in the texturing code has a bug causing texture projected "behind" objects. (Wall behind chair)

The current implementation incorrectly assumes that if the intersection between camera and target triangle is the target triangle, there is no occlusion possible. This is false because the intersection check used is `any_intersection` which is not ordered and another intersection could occur between the camera and triangle.

https://github.com/colmap/colmap/blob/1169d14d98a8d09bf8dd87ce3df44a186043068f/src/colmap/mvs/texture_mapping.cc#L247-L254

I added a small offset to exclude the target triangle from the checked range. This potentially has some numerical issues but doesn't have an effect on performance.

The numerically correct option would be to check the first intersection in this case but that adds complexity and impacts runtime.

Current implementation:
texturing_bug

Fix with offset:
texturing_offset

Fix with first_intersection:
texturing_first_intersection

There are some minor differences but I propose to use the simple offset based fix until other major issues are encountered.

### DIFF
--- a/src/colmap/mvs/texture_mapping.cc
+++ b/src/colmap/mvs/texture_mapping.cc
@@ -236,10 +236,12 @@ struct OcclusionTester {
     const Eigen::Vector3f dir = vertex - camera_center;
     const float dist = dir.norm();
     if (dist < kEps) return false;
+    const Eigen::Vector3f target_offset = vertex - dir.normalized() * kEps;
 
     const CGALPoint origin(
         camera_center.x(), camera_center.y(), camera_center.z());
-    const CGALPoint target(vertex.x(), vertex.y(), vertex.z());
+    const CGALPoint target(
+        target_offset.x(), target_offset.y(), target_offset.z());
     const CGALSegment segment(origin, target);
 
     const auto intersection = tree.any_intersection(segment);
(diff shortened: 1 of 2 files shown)


ahojnnes (approved):
Thanks, LGTM.

Dawars:
@ahojnnes I think the failing test is incorrect because the closer face is occluding visibility to all other sides of the cube.
I think this test accidentally passed because the shared corners were hit which were counted as non-occluding. This is not desirable.

How should I proceed?

ahojnnes:
@Dawars Yes, it looks buggy to me as well. Ideally you can update it to match the correct behavior? Thanks!

Dawars:
@ahojnnes what's the purpose of this test?
Is it supposed to only chech for visibility or neighborhood?

Currently, the cube faces point inwards. Is it enough if I flip them to point outwards and maybe move the camera to see multiple faces?

ahojnnes:
Sorry for the slow response. Had some time to look properly at this again. Assuming the camera and the cube are set up correctly, I am not sure I understand why this test should be wrong? We have one camera looking at a cube at the origin. In any case, we would expect at least 2 faces to be assigned, no? Can you elaborate what's exactly wrong with the test? Btw, it seems to pass on Mac.

Dawars:
The mac situation is interesting, I don't know why.

The test itself seem correct. It's the setup I'm unsure about. 

Which 2 faces do you expect to be assigned?
Currently, the closer face is filtered through the angle threshold (facing away from the camera/inside the cube). The further side (and all other) is occluded by the closer face.

ahojnnes:
OK, got it. How about just flipping the orientation of all sides?

PR #4396, opened 2026-05-12
Remove unused preconditioner setting for SPARSE_SCHUR

StonerLing:
Remove the explicit `CLUSTER_TRIDIAGONAL` preconditioner setting when using `ceres::SPARSE_SCHUR`.

According to Ceres Solver documentation/comments, visibility-clustering preconditioners are only valid with `ITERATIVE_SCHUR`:

https://github.com/ceres-solver/ceres-solver/blob/0ba987acaf9e8674070f116ed624edf017d2b630/include/ceres/types.h#L100-L119

```cpp
  // Note: The following four preconditioners can only be used with
  // the ITERATIVE_SCHUR solver. They are well suited for Structure
  // from Motion problems.


  // Block diagonal of the Schur complement. This preconditioner may
  // only be used with the ITERATIVE_SCHUR solver.
  SCHUR_JACOBI,


  // Use power series expansion to approximate the inversion of Schur complement
  // as a preconditioner.
  SCHUR_POWER_SERIES_EXPANSION,


  // Visibility clustering based preconditioners.
  //
  // The following two preconditioners use the visibility structure of
  // the scene to determine the sparsity structure of the
  // preconditioner. This is done using a clustering algorithm. The
  // available visibility clustering algorithms are described below.
  CLUSTER_JACOBI,
  CLUSTER_TRIDIAGONAL,
```

### DIFF
--- a/src/colmap/estimators/global_positioning.cc
+++ b/src/colmap/estimators/global_positioning.cc
@@ -416,10 +416,8 @@ void GlobalPositioner::ParameterizeVariables(Reconstruction& reconstruction) {
   // Do not use iterative solvers, for its suboptimal performance.
   if (reconstruction.NumPoints3D() > 0) {
     options_.solver_options.linear_solver_type = ceres::SPARSE_SCHUR;
-    options_.solver_options.preconditioner_type = ceres::CLUSTER_TRIDIAGONAL;
   } else {
     options_.solver_options.linear_solver_type = ceres::SPARSE_NORMAL_CHOLESKY;
-    options_.solver_options.preconditioner_type = ceres::JACOBI;
   }
 }
 
--- a/src/colmap/estimators/global_positioning.cc
+++ b/src/colmap/estimators/global_positioning.cc
@@ -414,11 +414,9 @@ void GlobalPositioner::ParameterizeVariables(Reconstruction& reconstruction) {
 
   // Set up the options for the solver
   // Do not use iterative solvers, for its suboptimal performance.
-  if (reconstruction.NumPoints3D() > 0) {
-    options_.solver_options.linear_solver_type = ceres::SPARSE_SCHUR;
-  } else {
-    options_.solver_options.linear_solver_type = ceres::SPARSE_NORMAL_CHOLESKY;
-  }
+  // TODO: Investigate whether the direct solver should be chosen
+  // adaptively based on problem scale.
+  options_.solver_options.linear_solver_type = ceres::SPARSE_SCHUR;
 }
 
 void GlobalPositioner::ConvertBackResults(Reconstruction& reconstruction) {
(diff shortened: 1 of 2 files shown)


src/colmap/estimators/global_positioning.cc
```diff
@@ -416,10 +416,8 @@ void GlobalPositioner::ParameterizeVariables(Reconstruction& reconstruction) {
   // Do not use iterative solvers, for its suboptimal performance.
   if (reconstruction.NumPoints3D() > 0) {
     options_.solver_options.linear_solver_type = ceres::SPARSE_SCHUR;
-    options_.solver_options.preconditioner_type = ceres::CLUSTER_TRIDIAGONAL;
   } else {
     options_.solver_options.linear_solver_type = ceres::SPARSE_NORMAL_CHOLESKY;
-    options_.solver_options.preconditioner_type = ceres::JACOBI;
```
StonerLing, line 422: `ceres::JACOBI` is already the default preconditioner.

src/colmap/estimators/global_positioning.cc
```diff
@@ -416,10 +416,8 @@ void GlobalPositioner::ParameterizeVariables(Reconstruction& reconstruction) {
...
   }
```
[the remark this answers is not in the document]
StonerLing: I think it would be better to make this conditional configurable(and use observations instead): 
```cpp
if (reconstruction.NumPoints3D() > options_.min_num_observations_schur_solver) {
    options_.solver_options.linear_solver_type = ceres::SPARSE_SCHUR;
} else {
    options_.solver_options.linear_solver_type = ceres::SPARSE_NORMAL_CHOLESKY;
}
```

src/colmap/estimators/global_positioning.cc
ahojnnes: Let's explicitly set it in all the branches to avoid different behavior when Ceres decides to change the behavior? Same below in the global mapper.

StonerLing: Sorry, I didn’t explain it clearly earlier — this is not really about the default value.

As far as I understand, preconditioners in Ceres are only relevant for iterative linear solvers, namely `ITERATIVE_SCHUR` and `CGNR`.

More specifically:
- `SCHUR_JACOBI`, `SCHUR_POWER_SERIES_EXPANSION`, `CLUSTER_JACOBI`, and `CLUSTER_TRIDIAGONAL` are only supported with `ITERATIVE_SCHUR`;
- `SUBSET` is only supported with `CGNR`;
- `JACOBI` can be used with both.

So for direct solvers such as `SPARSE_SCHUR` and `SPARSE_NORMAL_CHOLESKY`, setting a preconditioner probably has no effect.

ahojnnes: Oh, yes, you are right. Let's keep it as is.

src/colmap/estimators/global_positioning.cc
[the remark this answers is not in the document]
ahojnnes: I suggest to go with the Gemini suggestion for now or did you do an experiment to find a good threshold for Schur vs. normal cholesky?

ahojnnes (approved):

src/colmap/estimators/global_positioning.cc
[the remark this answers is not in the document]
StonerLing: Not yet... I'll add a TODO here similar to the one in `global_mapper.h`.

PR #4400, opened 2026-05-13
Fix NaN failures in rotation averaging from non-PD Cholesky

### LINKED ISSUE #4362: nan error when running glomap on 4.0.3
**Describe the bug**
When running `global_mapper` from the cli, it is returning a `nan` error on rotation averaging. 
I was able to fix a similar rotation averaging `nan` error by upgrading from 4.0.2 to 4.0.3, but this one is still showing up. 

Here is the output at the time of the error:
```
I20260425 01:09:51.632642  2171 global_mapper.cc:465] === Running rotation averaging ===
CHOLMOD warning: matrix not positive definite. file: ./CHOLMOD/Supernodal/t_cholmod_super_numeric_worker.c line: 1090
E20260425 01:09:51.726070  2171 rotation_averaging_impl.cc:679] nan error
```

**To Reproduce**
1. `colmap global_mapper --database_path /work/test/glomap/colmap/database.db --image_path /work/test/glomap/images --output_path /work/test/glomap/colmap/sparse`

I'm checking now to see if I'm able to share the frames/files used to create this error; I wasn't sure if the error output above could be enough to find the root cause or not. 

**Expected behavior**
No `nan` errors when running glomap. This worked prior to colmap v4. 

**Environment:**
 - OS: ubuntu 24.04
 - COLMAP Version: 4.0.3
 - Capture Device: iPhone 17 Pro
 - CUDA Version: 13.0
 - Nvidia Driver Version: 580.126.16

ahojnnes:
Fixes #4362.

## Summary

The L1 ADMM and IRLS Cholesky factorizations of A^T (W) A in rotation averaging silently produced NaN steps when CHOLMOD's supernodal LLT reported "matrix not positive definite" on poorly conditioned but mathematically PD pose graphs (e.g., long sequential video chains from a single camera). The NaN step corrupted the rotation state and surfaced later as the cryptic `nan error` / `nan weight!` log messages, aborting the global mapper before any reconstruction could be produced.

The original revision of this PR only made the failure explicit (added a small ridge regularization and propagated `Eigen::Success` checks). On the reporter's dataset that still failed, because CHOLMOD's supernodal solver is strict about positive definiteness — it aborts at the first non-positive pivot during dense supernode block factorization, even when a small ridge has been added. The fix below is robust regardless of ridge value or SuiteSparse version.

## Changes

- New `SparseCholeskyWithFallbackSolver` in `src/colmap/optim/sparse_cholesky.{h,cc}`. Tries `Eigen::CholmodSupernodalLLT` first (fastest); on failure, transparently falls back to `Eigen::SimplicialLDLT` (more numerically tolerant — LDLT does not require strict PD because it does not take square roots of pivots).
- `LeastAbsoluteDeviationSolver` (L1 ADMM): `SupernodalCholmodLLTLinearSolver` now delegates to the new helper. `Valid()` reflects success of either path. `Solve()` returns false instead of NaN if both fail.
- `RotationAveragingSolver::SolveIRLS`: uses the new helper directly per iteration. Detects NaN steps explicitly before applying `UpdateState`.
- `RotationEstimatorOptions::ridge_regularization`: defaults to 0 (no overhead). `GlobalMapperOptions` opts in with `1e-9` to stabilize ill-conditioned-but-PD graphs that the global SfM pipeline routinely encounters.

## Tests

- `sparse_cholesky_test.cc`: covers Compute/Solve smoke cases, the IRLS analyze-once / factorize-many reuse pattern, true singular failure, and (most importantly) a deterministic fallback test using an indefinite diagonal that CHOLMOD must reject and LDLT must accept.
- `rotation_averaging_test.cc`: regression test that `ridge_regularization` does not bias the solution.
- `least_absolute_deviations_test.cc`: regression test for the ridge path.

Verified against the reporter's database from #4362 (638 iPhone images): the full glomap pipeline completes cleanly end-to-end with no CHOLMOD warnings.

### DIFF
--- a/src/colmap/optim/least_absolute_deviations.cc
+++ b/src/colmap/optim/least_absolute_deviations.cc
@@ -29,12 +29,12 @@
 
 #include "colmap/optim/least_absolute_deviations.h"
 
+#include "colmap/optim/sparse_cholesky.h"
 #include "colmap/util/eigen_alignment.h"
 #include "colmap/util/logging.h"
 
 #include <memory>
 
-#include <Eigen/CholmodSupport>
 #include <Eigen/SparseCholesky>
 
 namespace colmap {
@@ -92,17 +92,15 @@ struct SupernodalCholmodLLTLinearSolver
       : ridge_regularization_(ridge_regularization) {}
 
   bool Compute(const Eigen::SparseMatrix<double>& A) override {
-    linear_solver_.compute(NormalEquations(A, ridge_regularization_));
-    return linear_solver_.info() == Eigen::Success;
+    return solver_.Compute(NormalEquations(A, ridge_regularization_));
   }
 
   bool Solve(const Eigen::VectorXd& b, Eigen::VectorXd* x) override {
-    x->noalias() = linear_solver_.solve(b);
-    return linear_solver_.info() == Eigen::Success;
+    return solver_.Solve(b, x);
   }
 
  private:
-  Eigen::CholmodSupernodalLLT<Eigen::SparseMatrix<double>> linear_solver_;
+  SparseCholeskyWithFallbackSolver solver_;
   const double ridge_regularization_;
 };
 
--- a/src/colmap/optim/least_absolute_deviations.cc
+++ b/src/colmap/optim/least_absolute_deviations.cc
@@ -57,10 +57,11 @@ Eigen::SparseMatrix<double> NormalEquations(
     const Eigen::SparseMatrix<double>& A, double ridge_regularization) {
   Eigen::SparseMatrix<double> AtA = A.transpose() * A;
   if (ridge_regularization > 0) {
-    Eigen::SparseMatrix<double> ridge(AtA.cols(), AtA.cols());
-    ridge.setIdentity();
-    ridge *= ridge_regularization;
-    AtA += ridge;
+    // The diagonal of A^T A is populated whenever the corresponding column of
+    // A has any non-zero entry, so coeffRef is cheap (no insertion).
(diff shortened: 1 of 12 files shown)


B1ueber2y (approved):
Thanks! LGTM. However, the issue does not seem to currently be resolved, which I suspect is due to a even more ill-posed structure from the input. I left a comment there.

ahojnnes:
@B1ueber2y I think I found one more potential improvement/mitigation to the issue.

ahojnnes:
@lpanaf PTAL and share your thoughts.

src/colmap/optim/least_absolute_deviations.cc
```diff
@@ -53,10 +53,26 @@ Eigen::VectorXd Shrinkage(const Eigen::VectorXd& a, const double kappa) {
   return a_plus_kappa.cwiseMin(0) + a_minus_kappa.cwiseMax(0);
 }
 
+Eigen::SparseMatrix<double> NormalEquations(
```
lpanaf, line 56: maybe use the name NormalMatrix?

lpanaf (commented):
Thanks for the PR! I remembered encountering the NaN error myself before, but I thought it usually happens when the graph is not connected. Do you have any examples where it is connected, but still has this problem?

ahojnnes:
> Thanks for the PR! I remembered encountering the NaN error myself before, but I thought it usually happens when the graph is not connected. Do you have any examples where it is connected, but still has this problem?

I only have the example from the reference issue. It seems like you don't have a concern. I guess, it is nice to be able to not fail and still compute a solution in case of weakly conditioned or disconnected components.

PR #4403, opened 2026-05-14
fix: rotation averaging crash with multi-image rigs

TannerGilbert:
`RunRotationAveraging`'s post-filter loop iterated `RegImageIds()` to drop images outside the connected component. For multi-camera rigs each registered frame appears once per image. `DeRegisterFrame()` resets the frame's pose internally, so the second visit of the same `frame_id` from a different `image_id` finds `HasPose() == false` and the `THROW_CHECK` fires.

### DIFF
--- a/src/colmap/estimators/rotation_averaging_test.cc
+++ b/src/colmap/estimators/rotation_averaging_test.cc
@@ -29,7 +29,6 @@
 
 #include "colmap/estimators/rotation_averaging.h"
 
-#include "colmap/estimators/rotation_averaging_impl.h"
 #include "colmap/math/math.h"
 #include "colmap/math/random.h"
 #include "colmap/scene/database_cache.h"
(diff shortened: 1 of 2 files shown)


src/colmap/estimators/rotation_averaging_test.cc
```diff
@@ -303,6 +303,74 @@ TEST(RotationAveraging, EmptyPoseGraph) {
...
+  ASSERT_EQ(frame_ids.size(), 4);
+  const frame_t isolated_frame_id = frame_ids.back();
+
+  // 1. Collect every image_id that belongs to the isolated frame.
+  std::unordered_set<image_t> isolated_image_ids;
+  for (const auto& data_id :
+       data.reconstruction.Frame(isolated_frame_id).ImageIds()) {
+    isolated_image_ids.insert(data_id.id);
+  }
+  ASSERT_GE(isolated_image_ids.size(), size_t{2})
```
B1ueber2y: This assertion feels meaningless and should be removed.

B1ueber2y (approved):
Thank you for the catch!

src/colmap/estimators/rotation_averaging_test.cc
TannerGilbert: Thanks for the review. I removed the assert

PR #4405, opened 2026-05-15
Fix SIGABRT in RegisterNextStructureLessImage when NumRegImages < 2

### LINKED ISSUE #901: hierarchical_mapper crash.
Ubuntu 18.04.4 LTS

Start Colmap from comandline:

```
colmap hierarchical_mapper \
--database_path /home/thomasd/3D/haus_v2/database.db \
--image_path /home/thomasd/3D/haus_v2/images \
--output_path /home/thomasd/3D/haus_v2/sparse
```

```
==============================================================================
Registering image #1653 (2)
==============================================================================

  => Image sees 67 / 9257 points
F0613 23:21:20.432286 20212 incremental_mapper.cc:347] Check failed: reconstruction_->NumRegImages() >= 2 (1 vs. 2) 
*** Check failure stack trace: ***
    @     0x7f344dfca0cd  google::LogMessage::Fail()
    @     0x7f344dfcbf33  google::LogMessage::SendToLog()
    @     0x7f344dfc9c28  google::LogMessage::Flush()
    @     0x7f344dfcc999  google::LogMessageFatal::~LogMessageFatal()
    @     0x557fc4abe4f7  (unknown)
    @     0x557fc497a156  (unknown)
    @     0x557fc497c67b  (unknown)
    @     0x557fc4b2d9bc  (unknown)
    @     0x7f344a8a16df  (unknown)
    @     0x7f344af806db  start_thread
    @     0x7f3449d2f88f  clone
Abgebrochen (Speicherabzug geschrieben)
```

amonszpart:
## Summary

Replace a hard `THROW_CHECK_GE` in `RegisterNextStructureLessImage` with a graceful `LOG(WARNING) + return false`, preventing a crash when the mapper attempts structure-less registration with fewer than 2 registered images.

Fixes #901.

## Context

On scenes with degenerate geometry the incremental mapper may register only 0–1 images before attempting structure-less registration of a candidate. The existing `THROW_CHECK_GE(reconstruction_->NumRegImages(), 2)` at line 620 calls `std::abort()` (SIGABRT) in this case.

The caller in `incremental_pipeline.cc` (~line 553) already handles `return false` gracefully — it tries other candidates or falls back to different registration strategies. The check is a precondition that should be a soft guard, not a fatal assertion.

## Precedent

PR #4344 applied the identical `THROW_CHECK → if + return false` pattern to `AdjustGlobalBundle` (line ~1084) for the analogous `ba_config.NumImages() < 2` case, merged April 2026. This is the same class of fix for a different function in the same file.

## Test plan

- [x] Verified on a real scene (urban rig capture) that previously SIGABRT'd at this CHECK with only 1 registered image — now logs a warning and the mapper continues gracefully
- [x] No behavior change for scenes that register ≥ 2 images (the guard is never hit)

### DIFF
--- a/src/colmap/sfm/incremental_mapper.cc
+++ b/src/colmap/sfm/incremental_mapper.cc
@@ -619,9 +619,9 @@ bool IncrementalMapper::RegisterNextStructureLessImage(const Options& options,
   THROW_CHECK_NOTNULL(reconstruction_);
   THROW_CHECK_NOTNULL(obs_manager_);
   if (reconstruction_->NumRegImages() < 2) {
-    LOG(WARNING) << "Structure-less registration requires at least 2 registered "
-                    "images; only "
-                 << reconstruction_->NumRegImages() << " available";
+    VLOG(2) << "Structure-less registration requires at least 2 registered "
+               "images; only "
+            << reconstruction_->NumRegImages() << " available";
     return false;
   }
 
--- a/src/colmap/sfm/incremental_mapper.cc
+++ b/src/colmap/sfm/incremental_mapper.cc
@@ -618,7 +618,12 @@ bool IncrementalMapper::RegisterNextStructureLessImage(const Options& options,
                                                        const image_t image_id) {
   THROW_CHECK_NOTNULL(reconstruction_);
   THROW_CHECK_NOTNULL(obs_manager_);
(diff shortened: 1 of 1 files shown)


B1ueber2y (approved):
Thank you! The Gemini comment makes sense to me. Should we address before merging : )

ahojnnes:
Thanks!

PR #4410, opened 2026-05-18
Add CI config for Ubuntu 26.04

ahojnnes:


B1ueber2y (approved):

src/colmap/estimators/solvers/homography_matrix.cc
```diff
@@ -37,6 +37,29 @@
 #include <Eigen/SVD>
 
 namespace colmap {
+namespace {
+
+bool HasCollinearTriplet(const std::vector<Eigen::Vector2d>& points) {
+  constexpr double kMinNormalizedAreaSquared = 1e-24;
+  for (size_t i = 0; i < points.size(); ++i) {
```
B1ueber2y: If this is only used for homography minimal sampling pruning, does it make sense to just list the four combinations? Especially given that this function is cubic and does not scale well.

PR #4416, opened 2026-05-20
Implement EUCM camera model

Ph03n1xdust:
Implements the [EUCM](https://hal.science/hal-01722264v1/document) camera model, which is quite flexible to represent various distortions with only 2 extra parameters.

https://github.com/tulzoltan also helped with the implementation.


ahojnnes:
Thanks for the PR! The math matches Khomutenko/Usenko. A few items before merging:

**Correctness**
- `HasBogusExtraParams` (inherited) treats α, β symmetrically with `|p| < max_extra_param`, but EUCM requires `α ∈ [0, 1]` and `β > 0`. Bundle adjustment can drive them out of range without detection and produce NaNs in `√(rho2)`. Consider overriding it for this model.
- In `CamFromImg`, `if (std::abs(helper) < epsilon)` admits negative `helper` (= mz), which then sign-flips `u, v` on division. Prefer `if (helper < epsilon) return false;`.

**Style**
- Comment block: missing space after `//` on several lines, stray superscript in `Bogdan Khomutenko1`, and the trailing `///////` divider after `CamFromImg` isn't used elsewhere (other models start with a divider, not end with one).
- `InitializeParams` uses int literals `0`/`1` and has odd brace placement; please run clang-format. `InitializePrincipalPointIdxs` signature is line-broken unlike the other models.

**Tests**
Only one parameter combo is exercised. Suggest adding at least: `α=0, β=1` (pinhole reduction), `α=0.5, β=1` (classic UCM), and a wider-FOV case that stresses the `radicand` near zero.

**Docs**
Please add EUCM to `doc/cameras.rst` alongside the other wide-FOV models.

**Question**
Should `CameraModelIsFisheye` return true for EUCM? It uses `BaseCameraModel` rather than `BaseFisheyeCameraModel`, which is fine, but EUCM is most useful for wide-FOV/fisheye lenses and downstream code may want to treat it as such.

B1ueber2y:
> Should CameraModelIsFisheye return true for EUCM? It uses BaseCameraModel rather than BaseFisheyeCameraModel, which is fine, but EUCM is most useful for wide-FOV/fisheye lenses and downstream code may want to treat it as such.

I would vote for returning false here. ``CameraModelIsFisheye`` has only one caller in ``models_test.cc``, where there is explicit check whether a pixel is valid in the fisheye mapping. Since the projection of EUCM is not equidistant-then-distort, it does not fit here. An improvement here would be to abstract some methods like ``CameraModelIsValidPixel`` and use it in the test.

ahojnnes:
I am going to admit that the comments were drafted by AI and refined by me. I agree that this suggestion doesn't make much sense.

Ph03n1xdust:
Thank you for the review, I hope I managed to correct all things.

> **Correctness**
> 
> * `HasBogusExtraParams` (inherited) treats α, β symmetrically with `|p| < max_extra_param`, but EUCM requires `α ∈ [0, 1]` and `β > 0`. Bundle adjustment can drive them out of range without detection and produce NaNs in `√(rho2)`. Consider overriding it for this model.

I added the ExtraParam check, but had to fiddle with the base class a bit so the EUCM implementation is actually called. I am not sure if this is the correct way to do it (I am not too familiar with modern C++), but as far as I understood with the original way, the unique implementation of the EUCM param check is not used.

```
  return HasBogusPrincipalPoint(params, width, height) ||
         HasBogusFocalLength(params,
                             width,
                             height,
                             min_focal_length_ratio,
                             max_focal_length_ratio) ||
         CameraModel::HasBogusExtraParams(params, max_extra_param);
}
```

ahojnnes:
Thank you. Your added `CameraModel::` should be the correct fix. I suggest adding the same for the HasBogusPrincipalPoint and HasBogusFocalLength for consistency (even though practically not used today) to avoid surprises in the case we add customization for those in the future.

ahojnnes (approved):

Issue #4422, opened 2026-05-24
pycolmap: calling `pycolmap.Database()` will throw a runtime error


Jacky56:
```text
python version: 3.10.12
pycolmap version: 4.0.4
```

For some whatever reason, my scripts will crash once `pycolmap.Database()` is no longer used/linked to memory. Which it tries to call a method `Database::Close` which probably does not exist anymore.

sample code in python shell:
```python
import pycolmap 

def e():
   pycolmap.Database()

e()
```

error logs:

```text
terminate called after throwing an instance of 'std::runtime_error'
  what():  Tried to call pure virtual function "Database::Close"
*** Aborted at 1779639009 (unix time) try "date -d @1779639009" if you are using GNU date ***
PC: @     0x7f8b313c69fc pthread_kill
*** SIGABRT (@0x221) received by PID 545 (TID 0x7f8b3132f1c0) from PID 545; stack trace: ***
    @     0x7f8b313c9ee8 (/usr/lib/x86_64-linux-gnu/libc.so.6+0x99ee7)
    @     0x7f8b265b3dd8 (/usr/local/lib/python3.10/dist-packages/pycolmap/_core.cpython-310-x86_64-linux-gnu.so+0x137bdd7)
    @     0x7f8b31372520 (/usr/lib/x86_64-linux-gnu/libc.so.6+0x4251f)
    @     0x7f8b313c69fc pthread_kill
    @     0x7f8b31372476 raise
    @     0x7f8b313587f3 abort
    @     0x7f8b24676b9e (/usr/lib/x86_64-linux-gnu/libstdc++.so.6.0.30+0xa2b9d)
    @     0x7f8b2468220c (/usr/lib/x86_64-linux-gnu/libstdc++.so.6.0.30+0xae20b)
    @     0x7f8b246811e9 (/usr/lib/x86_64-linux-gnu/libstdc++.so.6.0.30+0xad1e8)
    @     0x7f8b24681959 __gxx_personality_v0
    @     0x7f8b30c40884 (/usr/lib/x86_64-linux-gnu/libgcc_s.so.1+0x16883)
    @     0x7f8b30c40f41 _Unwind_RaiseException
    @     0x7f8b246824cb __cxa_throw
    @     0x7f8b25ddbdb8 (/usr/local/lib/python3.10/dist-packages/pycolmap/_core.cpython-310-x86_64-linux-gnu.so+0xba3db7)
[...]

PR #4423, opened 2026-05-24
Implement Close method for Database

Jacky56:
Copy and pasted in the method from previous commit https://github.com/colmap/colmap/blob/0676c9f282fe83e46f2b1b73a375befeb611a737/src/colmap/scene/database.cc

related to: #4422


ahojnnes:
Thanks, I think a better fix is implemented here: #4424

PR #4426, opened 2026-05-26
Fix determinism check in rotation_averaging_test

chpatrick:
We found that on NixOS aarch64 `angularDistance` returns `1e-17` for the same quaternion (probably due to FMA or similar?). If we check equality explicitly, the test passes.

### DIFF
--- a/src/colmap/estimators/rotation_averaging_test.cc
+++ b/src/colmap/estimators/rotation_averaging_test.cc
@@ -278,8 +278,16 @@ TEST(RotationAveraging, DeterministicRandomSeed) {
   EXPECT_TRUE(RunRotationAveraging(
       options, pose_graph2, reconstruction2, data.pose_priors));
 
-  ExpectEqualRotations(
-      reconstruction1, reconstruction2, /*max_rotation_error_deg=*/0);
+  for (const image_t image_id : reconstruction1.RegImageIds()) {
+    const Eigen::Quaterniond q1 =
+        reconstruction1.Image(image_id).CamFromWorld().rotation();
+    const Eigen::Quaterniond q2 =
+        reconstruction2.Image(image_id).CamFromWorld().rotation();
+    // In the presence of optimizations like FMA, q.angularDistance(q) can be
+    // near-zero instead of zero, so check equality explicitly instead of with
+    // ExpectEqualRotations.
+    EXPECT_EQ(q1.coeffs(), q2.coeffs());
+  }
 }
 
 TEST(RotationAveraging, RidgeRegularizationDoesNotBiasSolution) {


src/colmap/estimators/rotation_averaging_test.cc
```diff
@@ -278,8 +278,16 @@ TEST(RotationAveraging, DeterministicRandomSeed) {
...
-      reconstruction1, reconstruction2, /*max_rotation_error_deg=*/0);
+  const std::vector<image_t> reg_image_ids = reconstruction1.RegImageIds();
+  for (const image_t image_id : reg_image_ids) {
+    const Eigen::Quaterniond q1 =
+        reconstruction1.Image(image_id).CamFromWorld().rotation();
+    const Eigen::Quaterniond q2 =
+        reconstruction2.Image(image_id).CamFromWorld().rotation();
+    // In the presence of optimizations like FMA, x.dot(x) can be near-zero instead of zero,
+    // so check equality explicitly instead of with ExpectEqualRotations.
+    EXPECT_EQ(q1.coeffs(), q2.coeffs());
```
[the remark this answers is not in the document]
chpatrick: It compiles for me.

src/colmap/estimators/rotation_averaging_test.cc
```diff
@@ -278,8 +278,16 @@ TEST(RotationAveraging, DeterministicRandomSeed) {
...
 
-  ExpectEqualRotations(
-      reconstruction1, reconstruction2, /*max_rotation_error_deg=*/0);
+  const std::vector<image_t> reg_image_ids = reconstruction1.RegImageIds();
+  for (const image_t image_id : reg_image_ids) {
+    const Eigen::Quaterniond q1 =
+        reconstruction1.Image(image_id).CamFromWorld().rotation();
+    const Eigen::Quaterniond q2 =
+        reconstruction2.Image(image_id).CamFromWorld().rotation();
+    // In the presence of optimizations like FMA, x.dot(x) can be near-zero instead of zero,
```
B1ueber2y: The comment is wrong. It should be ``acos(|q1.dot(q2)|)``.

src/colmap/estimators/rotation_averaging_test.cc
```diff
@@ -278,8 +278,16 @@ TEST(RotationAveraging, DeterministicRandomSeed) {
   EXPECT_TRUE(RunRotationAveraging(
       options, pose_graph2, reconstruction2, data.pose_priors));
 
-  ExpectEqualRotations(
-      reconstruction1, reconstruction2, /*max_rotation_error_deg=*/0);
+  const std::vector<image_t> reg_image_ids = reconstruction1.RegImageIds();
```
B1ueber2y: nit: local copy is not needed. Can iterate directly.

B1ueber2y (approved):
Thanks for the fix! LGTM.

src/colmap/estimators/rotation_averaging_test.cc
chpatrick: The critical part is that a dot product of a quaternion with itself is non-1, I updated the comment to use `angularDistance`.

src/colmap/estimators/rotation_averaging_test.cc
chpatrick: Done.

src/colmap/estimators/rotation_averaging_test.cc
```diff
@@ -278,8 +278,16 @@ TEST(RotationAveraging, DeterministicRandomSeed) {
   EXPECT_TRUE(RunRotationAveraging(
       options, pose_graph2, reconstruction2, data.pose_priors));
 
-  ExpectEqualRotations(
-      reconstruction1, reconstruction2, /*max_rotation_error_deg=*/0);
+  for (const image_t image_id : reconstruction1.RegImageIds()) {
```
ahojnnes, line 281: Can we use instead?
```
EXPECT_THAT(reconstruction1, ReconstructionEq(reconstruction2));
```

ahojnnes: Defined in `src/colmap/scene/reconstruction_matchers.h`

B1ueber2y: This is a deterministic test for rotation averaging, so only checks on rotation are needed. I am not sure if using ``ReconstructionEq`` is overkilling here.

ahojnnes: But the other fields of the reconstruction should also be identical?

B1ueber2y: Conceptually (not really for this test case), I would assume an output translation of NaN being reasonable as well for this rotation averaging module, which would however break the equality : )

PR #4433, opened 2026-05-28
Add SPHERE equirectangular camera model with CamFromImgRay and sphere_to_cubic

awm33:
# Add SPHERE equirectangular camera model with CamFromImgRay and sphere_to_cubic

## Summary

Adds a `SPHERE` (equirectangular panorama) camera model to COLMAP 4.x and the supporting 3D-bearing-vector ingress + triangulation path, so full-360° panoramas can be reconstructed natively through the existing sparse + dense pipeline (including GLOMAP). Also adds `sphere_to_cubic`, the handoff tool that bridges SPHERE sparse reconstructions into the standard pinhole dense MVS stack (`image_undistorter` → `patch_match_stereo` → `stereo_fusion`, or OpenMVS `InterfaceCOLMAP`).

Validated end-to-end on a real-world dataset (101 Cyclomedia equirectangular streetview panoramas of Philadelphia Center City): **101/101 images registered in a single unified sparse model with 10,778 points via `global_mapper` (GLOMAP), handed off through `sphere_to_cubic` to `image_undistorter` + `patch_match_stereo` + `stereo_fusion`, producing a 1.74M-point dense cloud in ~70 minutes on a single Tesla T4 GPU.** Same-dataset baseline via the SphereSfM fork (COLMAP 3.8) reaches only 68/101 across 2 disjoint fragments with 1,441 sparse points and no dense handoff.

## Commits (6)

1. **`Add SPHERE equirectangular camera model and CamFromImgRay interface`**
   - New `SphereCameraModel` (`CameraModelId::kSphere = 16`, model name `SPHERE`). Parameters are `(width, height)` — an equirectangular projection is fully specified by the image dimensions, no focal length or distortion.
   - `ImgFromCam` accepts any non-zero 3D direction (full 4π steradians, unlike pinhole/fisheye which require w > 0).
   - `CamFromImg` returns 2D normalized coords for the forward hemisphere only — the 2D `(u, v, 1)` representation cannot encode rays with Z ≤ 0, which is why the next change adds a 3D bearing-vector interface.
   - New `CamFromImgRay` method added to `BaseCameraModel` (default: `CamFromImg` → `homogeneous().normalized()`, bit-equivalent to the pattern that existed throughout the geometry code). `SphereCameraModel` overrides it to return valid unit bearings for any pixel in the panorama, including the back hemisphere.
   - Overrides `HasBogusParams` (params are image dimensions, always valid) and `CamFromImgThreshold` (no focal length; uses angular resolution at the equator).
   - Tests in `src/colmap/sensor/models_test.cc`: 9 new tests covering metadata, cardinal directions, round-trips over the forward hemisphere and full sphere, back-hemisphere rejection, angular thresholds, and `Perspective.CamFromImgRayMatchesLegacy` — verifying the new default delegate is bit-equivalent to the pre-existing `CamFromImg(p)->homogeneous().normalized()` pattern across SimplePinhole, Pinhole, SimpleRadial, and OpenCV.

2. **`Migrate geometry estimators to 3D bearing CamFromImgRay ingress`**
   - Replaces `CamFromImg(p)->homogeneous().normalized()` with `CamFromImgRay(p)` at every 2D→3D ray conversion site in the geometry estimators: `two_view_geometry.cc` (6 sites), `generalized_pose.cc` (5 sites), `pose.cc` (1 site), `global_positioning.cc` (1 site).
   - No behavior change for perspective cameras — the default `BaseCameraModel::CamFromImgRay` delegates through the same math. For SPHERE and future omnidirectional models, this is the path that reaches back-hemisphere features.
   - `EstimateTwoViewGeometryPose` guards non-perspective cameras (cameras with zero `FocalLengthIdxs()`) to the CALIBRATED essential-matrix branch only. The UNCALIBRATED / PLANAR / PANORAMIC / PLANAR_OR_PANORAMIC branches call `camera.CalibrationMatrix()`, which is undefined when there is no pinhole K and used to SIGSEGV. The guard causes those pairs to be skipped rather than crash.

3. **`Skip F/H estimation for non-perspective cameras in two_view_geometry`**
   - Root-cause fix for why SPHERE pairs ended up in PLANAR / UNCALIBRATED configs in the first place: in `EstimateCalibratedTwoViewGeometry`, F and H were being estimated on raw pixel coordinates of the ERP image. But equirectangular pixels don't lie on a flat image plane, so F and H fit there are not geometrically meaningful — inliers are essentially accidental pattern matches.
   - When either camera in a pair has no focal-length parameter (SPHERE or any future omnidirectional model), the F/H RANSAC is skipped entirely and the pair is committed to CALIBRATED based on the bearing-vector essential-matrix inliers (which *are* correct for these cameras).
   - Measured on the Cyclomedia 101-panorama dataset:
     - CALIBRATED pairs: **1,028 → 3,343 (3.25×)**
     - UNDEFINED (failed geometric verification): 4,655 → 1,707 (−63%)
     - Matcher wall-clock: 8.7 min → 6.7 min (−23%)
     - No change in reconstruction quality at the GLOMAP output (101/101 registered both before and after, reprojection error unchanged).

4. **`Support bearing-vector triangulation for omnidirectional cameras`**
   - `geometry/triangulation.{h,cc}`: new overload `TriangulateMultiViewPoint(span<const Matrix3x4d>, span<const Vector3d>, Vector3d*)`. The existing projector-based DLT implementation already normalizes internally via `cam_point.homogeneous().normalized()`; the new overload consumes bearings directly. Equivalent for perspective cameras, correct for SPHERE (back-hemisphere rays can contribute).
   - `estimators/triangulation.{h,cc}`: adds `Eigen::Vector3d cam_ray` to `TriangulationEstimator::PointData` alongside the legacy `cam_point`. In `Estimate()`, detects whether any observation has a populated `cam_ray` and routes to the bearing-based overload; otherwise preserves the legacy 2D path exactly. Angular-error residual uses `cam_ray` directly.
   - `EstimateTriangulation` populates `cam_ray` via `Camera::CamFromImgRay` alongside the legacy `cam_point` via `Camera::CamFromImg`.
   - Two new tests in `src/colmap/geometry/triangulation_test.cc`: `TriangulateMultiViewPointFromBearings.NominalEquivalentTo2D` asserts bit-equivalence with the 2D overload on perspective correspondences; `.BackHemisphereRays` constructs a three-camera setup where all observations are in the back hemisphere and validates the bearing path triangulates the point correctly.

5. **`Add sphere_to_cubic subcommand for dense MVS handoff on SPHERE models`**
   - New `colmap sphere_to_cubic` subcommand in `src/colmap/exe/sphere.{h,cc}`.
   - Converts a SPHERE sparse reconstruction to a PINHOLE (cubic) reconstruction: six perspective cube faces (F/B/L/R/U/D, configurable size and FOV) rendered per panorama via bilinear ERP resampling, each with its own pose composed from the panorama pose and a canonical cube-face rotation. Track elements are reprojected onto whichever cube face contains them.
   - COLMAP 4.x `Rig` / `Frame` / `Reconstruction` integration: each face image lives in its own `Frame` belonging to a single shared single-sensor `Rig`.
   - On the Exp 8 dataset: 101 SPHERE images → 606 face images in ~90 s on the CPU; 42,133 observations reassigned (0 dropped), 10,778 / 10,778 points preserved, mean reprojection error unchanged at 0.00165. Subsequent `image_undistorter` + `patch_match_stereo` + `stereo_fusion` produces a 1.74M-point dense cloud (validated separately).
   - Equivalent in intent to SphereSfM's `sphere_cubic_reprojecer` but reimplemented fresh against the current COLMAP 4.x API rather than ported from 3.8.

6. **`Add unit tests for bearing-vector triangulation and cube-face geometry`**
   - Five new tests in `src/colmap/exe/sphere_test.cc` covering the cube-face rotations (count, orthogonality + det +1, canonical direction mapping, full-sphere coverage, forward-ray projects to face F center).
   - Refactor: exposes `SphereCubeFaceSpecs` / `SphereCubeFaceSpec` from `sphere.h` so tests can exercise the same rotation data used by the subcommand. `sphere.h` / `sphere.cc` now also live in the `colmap_exe` library (`COLMAP_EXE_SRCS`), matching the feature / sfm / model pattern.
   - Total `ctest` count: 145 → 147 (all passing).

## Validation

- Full `ctest` suite: **147 / 147 passing** (+2 from the baseline, for the new bearing-vector and cube-face tests; all existing tests unchanged).
- `src/colmap/sensor/models_test`: **25 / 25** (16 existing + 9 new), including the `Perspective.CamFromImgRayMatchesLegacy` cross-model equivalence check.
- End-to-end on 101 Cyclomedia ERP panoramas: `feature_extractor` → `exhaustive_matcher` → `global_mapper` (GLOMAP) → `sphere_to_cubic` → `image_undistorter` → `patch_match_stereo` → `stereo_fusion` produces a **1,740,836-point fused dense PLY** in ~70 min on a Tesla T4.

## Scope boundary / known follow-ups

- **PLANAR / PANORAMIC pose-from-homography for SPHERE**: 988 / 4,493 stored two-view geometries in the Exp 8 database end up in non-CALIBRATED configs under the pre-fix behavior because their H-on-ERP inliers outnumber E inliers. With commit 3 those configs are no longer produced for SPHERE pairs, so the issue disappears at the source; the decomposition-side guard in commit 2 remains as a defense-in-depth for legacy databases. A proper ray-based homography decomposition for omnidirectional cameras would let those pairs contribute again; scoped for a follow-up.
- **`SphereCameraModel` focal-length reporting**: the `feature_extractor` log line prints `Focal Length: -nanpx (Prior)` for SPHERE. Cosmetic only — downstream code correctly sees `FocalLengthIdxs().size() == 0`. A small conditional in `feature_extraction.cc` to skip the focal-length line for non-perspective cameras would clean this up.
- **Dense MVS on cube faces with no stereo neighbor**: on the Exp 8 run, 235 of 606 faces produce empty depth maps (predominantly U/D/B faces with sky / ground / no adjacent panorama behind). Expected, not a fork issue.

Happy to split into multiple smaller PRs if preferred — the commits are already logically independent.

### DIFF
--- a/src/colmap/controllers/incremental_pipeline.cc
+++ b/src/colmap/controllers/incremental_pipeline.cc
@@ -141,6 +141,8 @@ IncrementalMapper::Options IncrementalPipelineOptions::Mapper() const {
   options.use_prior_position = use_prior_position;
   options.use_robust_loss_on_prior_position = use_robust_loss_on_prior_position;
   options.prior_position_loss_scale = prior_position_loss_scale;
+  options.use_robust_loss_on_prior_rotation = use_robust_loss_on_prior_rotation;
+  options.prior_rotation_loss_scale = prior_rotation_loss_scale;
   options.random_seed = random_seed;
   return options;
 }
@@ -244,6 +246,7 @@ bool IncrementalPipelineOptions::Check() const {
   CHECK_OPTION_GE(ba_global_max_refinement_change, 0);
   CHECK_OPTION_GE(snapshot_frames_freq, 0);
   CHECK_OPTION_GT(prior_position_loss_scale, 0.);
+  CHECK_OPTION_GT(prior_rotation_loss_scale, 0.);
   CHECK_OPTION_GE(num_threads, -1);
   CHECK_OPTION_GE(random_seed, -1);
   CHECK_OPTION(Mapper().Check());
--- a/src/colmap/controllers/incremental_pipeline.h
+++ b/src/colmap/controllers/incremental_pipeline.h
@@ -149,6 +149,13 @@ struct IncrementalPipelineOptions {
   // (chi2 for 3DOF at 95% = 7.815).
   double prior_position_loss_scale = 7.815;
 
+  // Whether to use a robust (Cauchy) loss on prior camera rotation.
+  bool use_robust_loss_on_prior_rotation = false;
+
+  // Threshold on the residual for the robust rotation prior loss
+  // (chi2 for 3DOF at 95% = 7.815).
+  double prior_rotation_loss_scale = 7.815;
+
   // Path to a folder with reconstruction snapshots during incremental
   // reconstruction. Snapshots will be saved according to the specified
   // frequency of registered images.
--- a/src/colmap/estimators/bundle_adjustment.cc
+++ b/src/colmap/estimators/bundle_adjustment.cc
@@ -349,6 +349,7 @@ PosePriorBundleAdjustmentBackendOptions::operator=(
 
 bool PosePriorBundleAdjustmentOptions::Check() const {
   CHECK_OPTION_GT(prior_position_fallback_stddev, 0);
+  CHECK_OPTION_GT(prior_rotation_fallback_stddev, 0);
   return THROW_CHECK_NOTNULL(ceres)->Check();
 }
 
--- a/src/colmap/estimators/bundle_adjustment.h
+++ b/src/colmap/estimators/bundle_adjustment.h
@@ -248,6 +248,11 @@ struct PosePriorBundleAdjustmentOptions
   // Fallback if no prior position covariance is provided.
   double prior_position_fallback_stddev = 1.0;
 
+  // Fallback if no prior rotation covariance is provided. Interpreted as the
+  // standard deviation per axis of the angle-axis rotation prior residual
+  // (radians). Default 0.1 rad ≈ 5.7°.
+  double prior_rotation_fallback_stddev = 0.1;
+
   // Sim3 alignment options.
   RANSACOptions alignment_ransac_options;
 
--- a/src/colmap/estimators/bundle_adjustment_ceres.cc
+++ b/src/colmap/estimators/bundle_adjustment_ceres.cc
@@ -233,6 +233,7 @@ bool CeresBundleAdjustmentOptions::Check() const {
 
 bool CeresPosePriorBundleAdjustmentOptions::Check() const {
   CHECK_OPTION_GT(prior_position_loss_scale, 0);
+  CHECK_OPTION_GT(prior_rotation_loss_scale, 0);
   return true;
 }
 
@@ -890,12 +891,14 @@ class PosePriorBundleAdjuster : public CeresBundleAdjuster {
         reconstruction_(reconstruction) {
     THROW_CHECK(prior_options_.Check());
 
-    // Filter irrelevant pose priors.
+    // Filter irrelevant pose priors. Keep rows that contribute at least one
+    // of position or rotation — a rotation-only prior still constrains yaw.
     pose_priors_.erase(
         std::remove_if(pose_priors_.begin(),
                        pose_priors_.end(),
                        [this](const auto& pose_prior) {
-                         return !pose_prior.HasPosition() ||
+                         return (!pose_prior.HasPosition() &&
+                                 !pose_prior.HasRotation()) ||
                                 pose_prior.corr_data_id.sensor_id.type !=
                                     SensorType::CAMERA ||
                                 !config_.HasImage(pose_prior.corr_data_id.id);
@@ -922,6 +925,9 @@ class PosePriorBundleAdjuster : public CeresBundleAdjuster {
       prior_loss_function_ = CreateLossFunction(
           prior_options_.ceres->prior_position_loss_function_type,
           prior_options_.ceres->prior_position_loss_scale);
+      prior_rotation_loss_function_ = CreateLossFunction(
+          prior_options_.ceres->prior_rotation_loss_function_type,
+          prior_options_.ceres->prior_rotation_loss_scale);
 
       // Only consider parameterized images for pose priors. Notice that some
       // images may be configured to be included in the BA problem but have no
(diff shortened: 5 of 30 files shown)


zalo:
Claude review comments as well, in the context of the previous Equirectangular PR ( https://github.com/colmap/colmap/pull/4014 ) I just closed for this one.

---

## 1. (Correctness) Back-hemisphere points are triangulated but then rejected by chirality / negative-depth checks

This PR correctly teaches the **ingress** and the **angular residual** to handle
Z ≤ 0 observations (great — `EstimateTriangulation` populates `cam_ray` via
`CamFromImgRay`, and `TriangulationEstimator::Residuals` prefers `cam_ray` over
the 2D path). But the **chirality/positive-depth gates are still applied
unconditionally**, so the same back-hemisphere observations that now produce
valid bearings and valid residuals get discarded before they can become points.
The net effect is silent loss of valid panorama points rather than a crash, so
it won't show up as a failure in an end-to-end run — only as reduced coverage.

There are three spots:

### 1a. `src/colmap/estimators/triangulation.cc` — two-view branch (the bearing path you added)

```cpp
if (triangulated &&
    HasPointPositiveDepth(pose_data[0].cam_from_world, xyz) &&
    HasPointPositiveDepth(pose_data[1].cam_from_world, xyz) &&
    CalculateTriangulationAngle(...) >= min_tri_angle_) {
```

For a `SPHERE` camera a correctly-triangulated point can legitimately lie behind
the camera (Z ≤ 0), so requiring positive depth in **both** views drops valid
back-hemisphere points. Suggest gating the depth check on perspective-ness, e.g.:

```cpp
const bool cam0_perspective = pose_data[0].camera->FocalLengthIdxs().size() > 0;
const bool cam1_perspective = pose_data[1].camera->FocalLengthIdxs().size() > 0;
if (triangulated &&
    (!cam0_perspective ||
     HasPointPositiveDepth(pose_data[0].cam_from_world, xyz)) &&
    (!cam1_perspective ||
     HasPointPositiveDepth(pose_data[1].cam_from_world, xyz)) &&
    CalculateTriangulationAngle(...) >= min_tri_angle_) {
```

### 1b. `src/colmap/estimators/triangulation.cc` — multi-view branch

The existing per-pose chirality loop (not modified by this PR) has the same
issue:

```cpp
for (const auto& pose : pose_data) {
  if (!HasPointPositiveDepth(pose.cam_from_world, xyz)) {
    return;
  }
}
```

Same suggested guard:

```cpp
for (const auto& pose : pose_data) {
  if (pose.camera->FocalLengthIdxs().size() > 0 &&
      !HasPointPositiveDepth(pose.cam_from_world, xyz)) {
    return;
  }
}
```

### 1c. `src/colmap/sfm/observation_manager.cc` — `FilterObservationsWithNegativeDepth()`

This PR doesn't touch `observation_manager.cc`, but this filter runs during
incremental mapping / cleanup (and `incremental_mapper` *is* in scope here). It
removes any observation whose 3D point has negative depth — which for a `SPHERE`
camera deletes every back-hemisphere observation regardless of how it was
triangulated:

```cpp
const Eigen::Matrix3x4d cam_from_world = image.CamFromWorld().ToMatrix();
for (...) {
  if (!HasPointPositiveDepth(cam_from_world, point3D.xyz)) {
    // observation removed
  }
}
```

Suggest skipping the filter for cameras without a focal length:

```cpp
const Camera& camera = *image.CameraPtr();
if (camera.FocalLengthIdxs().empty()) {
  continue;  // omnidirectional camera: all directions are valid
}
```

> Why this matters: in our PR, leaving these gates in place was the difference
> between losing most back-hemisphere structure and keeping it. Because the
> failure mode is "fewer points," not a crash, the 101/101 + 10,778-point
> validation here wouldn't reveal it — the reconstruction still succeeds, just
> with less coverage than it could have.

---

## 2. (Latent crash) `Camera::Rescale` aborts on a zero-focal model

`Camera::Rescale` → `Camera::ScaleFocalLengths` (`src/colmap/scene/camera.cc`)
does:

```cpp
} else {
  LOG(FATAL_THROW)
      << "Camera model must either have 1 or 2 focal length parameters.";
}
```

A `SPHERE` camera has 0 focal params, so any code path that rescales it aborts.
This PR's own dense handoff converts to `PINHOLE` via `sphere_to_cubic` first, so
the happy path avoids it — but `image_undistorter`/resize paths and downstream
tooling can still call `Rescale` on a SPHERE camera. Suggest a no-op branch for
the zero-focal case:

```cpp
} else if (FocalLengthIdxs().empty()) {
  // Omnidirectional cameras have no focal length to rescale; the principal
  // point rescale performed by the caller is sufficient.
} else {
  LOG(FATAL_THROW) << "...";
}
```

---

## 3. (Question) GLOMAP integration for SPHERE

The validation runs through `global_mapper` (GLOMAP), but no GLOMAP code is
changed here. In our PR, SPHERE cameras going through GLOMAP needed guards in
view-graph calibration (don't try to estimate a focal length for a camera that
has none) and sensible fallbacks anywhere `Focal()` / `CalibrationMatrix()` is
called on a non-perspective camera. Did SPHERE work through GLOMAP unmodified
here, or is there a companion GLOMAP change? If it worked as-is, it'd be useful
to note why (e.g., view-graph calibration already skips zero-focal cameras),
since that's a natural place for the same class of "no K / no focal" bug.

ahojnnes:
Thanks for your contribution. It's a bit difficult to see the forest for the trees in this PR. I assume significant portions of it are AI generated. The introduction of CamFromImgRay (or I'd rather call it CamRayFromImg) is a good one independent of the spherical camera. I would need to spend a bit more time to look into all the edge cases, which is a bit difficult given that the rotation prior is entangled into the changes. To make progress here, I suggest first extracting the CamRayFromImg changes in a separate PR and then disentangle the spherical camera from rotation prior changes.

Issue #4436, opened 2026-05-31
model_analyzer reports normalized point3D.error as px for COLMAP 4 global_mapper output


DuVogel87:
## Bug

`colmap model_analyzer` appears to report a normalized `point3D.error` value as `px` for models generated with COLMAP 4 `global_mapper`.

## Environment

- OS: Ubuntu Linux
- COLMAP Version: 4.0.4, commit `9c23f694`
- Capture Device: Insta360 camera / dual-fisheye video frames
- CUDA: enabled
- Command used: `colmap global_mapper`

## Observed

`colmap model_analyzer --path sparse/0` reports:

```txt
Mean reprojection error: 0.002582px
```

## Expected

The value should be the actual pixel reprojection error, or it should not be labeled as `px`.

I independently recomputed the reprojection error from `cameras.bin`, `images.bin`, and `points3D.bin` and got approximately:

```txt
Mean per-point pixel reprojection error: ~0.604 px
Mean per-observation pixel reprojection error: ~0.623 px
```

For comparison, an older COLMAP/GLOMAP model on the same dataset reports and recomputes consistently around:

```txt
~0.618 px
```

## Suspected Cause

`model_analyzer` calls `Reconstruction::ComputeMeanReprojectionError()`, which averages stored `point3D.error` values.

In COLMAP 4 `global_mapper`, these `point3D.error` values appear to be updated during normalized reprojection filtering, not as pixel reprojection errors. The analyzer still prints the value with `px`.

Relevant files:

```txt
src/colmap/exe/model.cc
src/colmap/scene/reconstruction.cc
src/colmap/sfm/observation_manager.cc
src/colmap/sfm/global_mapper.cc
```

This looks like a reporting/labeling bug or regression in `model_analyzer` for `global_mapper` outputs.

B1ueber2y:
Nice catch. Thanks! Should be fixed here: https://github.com/colmap/colmap/pull/4442

DuVogel87:
Thanks for the quick fix!

PR #4441, opened 2026-06-04
Add Spherical Camera Model

zalo:
This is not necessarily a PR for merging, just expanding on #4440 and putting it through its paces to find bugs.

Here is a test of this PR's spherical camera model implemented on top of it:

https://github.com/user-attachments/assets/ae416c79-87e1-4b82-b55f-87abdb1ab164

This PR is now detecting way more points than my previous PR, so I think it's a sign that it's operating more natively in equirect space.

#4440 had a bug where 

I'm including the script I used to generate this in the Python examples folder for easier validation.

If we did want to merge my attempt at the Spherical Mapper, here's what I need to finish:
- [x] Test with GPU Acceleration (CuDSS Ceres)
- [x] Solve why only the glomap path is losing all the points in one hemisphere
- [x] Make the example script use pycolmap instead of the CLI (keeping the super fast FFMPEG-based extraction, etc.)

# Native SPHERICAL SfM — CPU vs GPU benchmark

Controlled benchmark of the native equirectangular (`SPHERICAL`) SfM pipeline
(`run_spherical_sfm.py`) measuring full wall-clock time from **video
decomposition → TXT model**, for the incremental (`mapper`) and global
(`global_mapper`) backends, on CPU vs GPU.

## Setup

- **Hardware:** NVIDIA RTX 4090 (sm_89, 24 GB); 8 build/compute threads.
- **colmap:** `feature/spherical-on-cam-ray` (commit 8ee2a7b8), built with CUDA
  (g++-10 CUDA host compiler, sm_89). Ceres 2.3.0 with CUDA + cuDSS (CUDA-12
  variant) for GPU bundle adjustment.
- **Input:** `output2.mp4` — 8K equirectangular (7680×3840) 360° video,
  rotations canceled during stitching.
- **Frames:** 66 frames extracted (every 20th), downscaled 4× → 1920×960.
- **Camera model:** native `SPHERICAL` (id 17), `--ImageReader.single_camera 1`.
- CPU runs use `--FeatureExtraction/FeatureMatching.use_gpu 0`; GPU runs use
  GPU SIFT + matching and `--Mapper.ba_use_gpu 1` (incremental).

## Timing: video → TXT model (seconds)

| Stage | CPU incremental | CPU global | GPU incremental | GPU global |
|---|---|---|---|---|
| Video → frames (ffmpeg) | 35.0 | 35.0 | 14.0 | 14.0 |
| Feature extraction (SIFT) | 55.9 | 56.3 | **2.6** | **2.6** |
| Matching (sequential) | 28.2 | 28.1 | **3.8** | **3.9** |
| Mapping | 30.8 | 23.4 | 29.1 | 25.6 |
| TXT export | ~0.5 | ~0.5 | ~0.5 | ~0.5 |
| **Total (video→txt)** | **~150 s** | **~143 s** | **~50 s** | **~46 s** |

Running both modes in a single invocation (one shared video decode):
**261 s on CPU vs 106 s on GPU.**

## Quality (largest reconstructed sub-model)

| Metric | CPU incremental | CPU global | GPU incremental | GPU global |
|---|---|---|---|---|
| Registered images | 66 / 66 | 66 / 66 | 66 / 66 | 66 / 66 |
| 3D points | 24,950 | 28,288 | 24,730 | 27,746 |
| Mean track length | 5.42 | 5.28 | 5.37 | 5.24 |
| Mean reprojection error | 0.545 px | 0.0015 px | 0.576 px | 0.0015 px |

## Takeaways

- **GPU vs CPU:** ~**2.1×** end-to-end per mode, with **identical quality**
  (same registration, ±1% points, comparable error — the small deltas are just
  GPU vs CPU SIFT producing slightly different features).
- **The front-end is where the GPU wins:** SIFT 56 s → 2.6 s (~21×), matching
  28 s → 3.8 s (~7×). Mapping is largely CPU-bound graph/triangulation work and
  barely moves (GPU bundle adjustment helps a little: dense-CUDA for
  incremental, cuDSS sparse for global).
- **Incremental vs global:** both register all 66 images; global yields ~13%
  more points. The reprojection-error figures are **not directly comparable
  across mappers** — global's 0.0015 px reflects its tighter angular/normalized
  inlier filtering for the no-focal model, not a 360× accuracy gain. Use
  registration count and point count as the fairer quality signals here.

## Notes

- For this rotation-stabilized footage, `--GlobalMapper.ba_skip_joint_optimization_stage 1`
  (the modern equivalent of the old glomap `--BundleAdjustment.optimize_rotations 0`)
  shaves ~14% off global mapping (25.8 s → 22.1 s) with no quality change, but
  it was intentionally left out of the committed script since it assumes
  pre-canceled rotations and would degrade general (moving-camera) inputs.
  The old `--skip_rotation_averaging 1` is unsupported in colmap's
  `global_mapper` (aborts in track establishment), and
  `--BundleAdjustment.optimize_intrinsics 0` is a no-op for `SPHERICAL`
  (no focal length; `(w,h)` already pinned constant in BA).


zalo:
I've been putting this PR through its paces; it's working fantastically!  Here are three more random samples from the set I've tested on:

https://github.com/user-attachments/assets/974e28bb-88a9-427f-90c7-2cc5bed39b38

I haven't come across a scene yet where it's failed (though, my whole set has rotation compensation on when exporting from Insta360, so I'll still need to try it on equirects with the original rotations intact).

zalo:
@ahojnnes The core additions here are ready to be reviewed; I've recently validated it on many many many equirectangular videos, and results seem to be as good or better than rig-based methods.

I can adjust the included python examples to taste as well; I figured some FFMPEG frame extraction utility would help people get started with Insta360 Studio outputs.

ichsan2895:
Hello, some question.. 
1. Does your code supports Caspar?
```
>> colmap mapper --database_path distorted/database.db \
    --image_path input/ \
    --Mapper.ba_use_gpu 1 \
    --Mapper.ba_local_backend CASPAR \
    --Mapper.ba_global_backend CASPAR \
    --output_path distorted/colmap_caspar
```
W0610 00:16:59.406899 53716 bundle_adjustment_caspar.cc:54] Skipping image 5 with unsupported camera model: SPHERICAL
W0610 00:16:59.407387 53716 bundle_adjustment_caspar.cc:54] Skipping image 4 with unsupported camera model: SPHERICAL
W0610 00:16:59.407419 53716 bundle_adjustment_caspar.cc:894] No residuals to optimize

2. Does your code support multiple camera setup?
For example, I want to join dataset that consist of equirectangular images and simple_radial images. Then I train it with GUT in LFS or 3DGUT.

3. Suggestion. I know equirectangular can not be undistorted. Please gives clear error log when user accidentially runs colmap image_undistorter.
```
I0610 00:21:53.760371 58635 image.cc:368] === Reading reconstruction ===
I0610 00:21:53.775105 58635 image.cc:371] => Reconstruction with 8 images and 4296 points
I0610 00:21:53.775125 58635 undistorters.cc:166] === Image undistortion ===
I0610 00:21:53.846256 58635 undistorters.cc:205] Undistorting image [1/8]
*** Aborted at 1781050914 (unix time) try "date -d @1781050914" if you are using GNU date ***
PC: @                0x0 (unknown)
*** SIGSEGV (@0x0) received by PID 58635 (TID 0x7984689fc000) from PID 0; stack trace: ***
    @     0x7984efef6046 (unknown)
    @     0x79847af47520 (unknown)
    @          0x1436ce3 colmap::UndistortCamera()
    @          0x143a209 colmap::UndistortImage()
    @           0x6ab89f colmap::COLMAPUndistorter::Undistort()
    @           0x6b26f9 _ZNSt17_Function_handlerIFSt10unique_ptrINSt13__future_base12_Result_baseENS2_8_DeleterEEvENS1_12_Task_setterIS0_INS1_7_ResultIbEES3_EZNS1_11_Task_stateISt5_BindIFMN6colmap17COLMAPUndistorterEKFbjEPSD_jEESaIiEFbvEE6_M_runEvEUlvE_bEEE9_M_invokeERKSt9_Any_data
    @           0x5f3b2b std::__future_base::_State_baseV2::_M_do_set()
    @     0x79847af9eee8 (unknown)
    @           0x6b3f67 std::__future_base::_Task_state<>::_M_run()
    @           0x6b5914 _ZNSt17_Function_handlerIFSt8functionIFvvEEvEZN6colmap10ThreadPool7AddTaskIMNS4_17COLMAPUndistorterEKFbjEJPS7_RKjEEESt13shared_futureINSt13invoke_resultIT_JDpT0_EE4typeEEOSF_DpOSG_EUlvE_E9_M_invokeERKSt9_Any_data
    @           0x7dca11 colmap::ThreadPool::WorkerFunc()
    @     0x79847b21a064 execute_native_thread_routine
    @     0x79847af99ac3 (unknown)
    @     0x79847b02aa04 clone
 ```

**Comment by zalo** [comment]
> 1. Does your code supports Caspar?

No, but I'm having my Claude take a look in a followup PR, in-case it's easy.

> 2. Does your code support multiple camera setup?
>    For example, I want to join dataset that consist of equirectangular images and simple_radial images. Then I train it with GUT in LFS or 3DGUT.

In theory, this should work, so long as your other images have a relatively known calibration.

> 3. Suggestion. I know equirectangular can not be undistorted. Please gives clear error log when user accidentially runs colmap image_undistorter.

Just added this.

**Comment by zalo** [comment]
@ichsan2895
> 1. Does your code supports Caspar?

Claude just gave me an implementation for Caspar on top of this branch; the additional algorithm code is only a few hundred lines, but Symforce autogenerates like 20k lines and a hundred additional committed files to include it:
https://github.com/zalo/colmap/compare/feature/spherical-on-cam-ray...zalo:colmap:feature/spherical-caspar-ba

Claude's Caspar Benchmark Summary for a new/later PR:

## Caspar GPU bundle adjustment — SPHERICAL benchmark

GPU bundle-adjustment backends compared on a native equirectangular
(`SPHERICAL`) reconstruction. Same input, same shared GPU front-end; only the
mapping/BA path differs.

**Setup:** `output2.mp4` — 8K equirectangular 360° video, 66 frames extracted
(every 20th) and downscaled 4× → 1920×960, native `SPHERICAL` camera model,
single camera. RTX 4090 (sm_89). Shared GPU front-end: SIFT feature extraction
2.5 s + sequential matching 3.6 s = **6.1 s**. Times below are the mapping stage
and the end-to-end total (excluding the ~35 s video decode, which is identical
for all three).

| Mapping path | Mapping | Total (video→model) | Reg. images | 3D points | Mean reproj. error |
|---|---|---|---|---|---|
| Incremental + **Ceres** GPU BA | 28.1 s | ~34 s | 66 / 66 | 24,748 | 0.576 px |
| Incremental + **Caspar** GPU BA | **13.2 s** | **~19 s** | 66 / 66 | 24,730 | 0.587 px |
| **Global** mapper (cuDSS GPU BA) | 22.6 s | ~29 s | 66 / 66 | 27,738 | 0.535 px |

### Summary

- **Caspar makes the incremental mapper ~2.1× faster** (28.1 s → 13.2 s) at
  essentially identical quality (24,730 vs 24,748 points; 0.587 vs 0.576 px).
  Bundle adjustment is a large share of incremental wall-clock, so moving it
  fully onto the GPU is the win.
- With the CPU free, **Caspar-incremental (13.2 s) even beats the global mapper
  (22.6 s)**. (On a loaded machine the opposite appears true — incremental's
  CPU-bound registration/triangulation gets starved, while the global mapper's
  GPU-bound cuDSS BA is less affected.)
- The **global mapper** still produces the densest model with the lowest
  reprojection error (27.7k points, 0.535 px) — a result of its
  angular/normalized track filtering for the no-focal `SPHERICAL` model — but it
  is slower than Caspar-incremental here.
- In isolation the BA step itself is far faster on Caspar: a standalone
  `bundle_adjuster` run on a 521-image `SPHERICAL` model took **1.4 s** with
  Caspar vs **41 s** with Ceres (~30×), converging to the same reprojection
  error (1.085 px, matching to < 1e-4 px).

**Fastest path to a model for this clip: incremental + Caspar BA (~19 send-to-end).**

**Comment by ichsan2895** [comment]
Cools! Hopefully it get merged. 😀

**Comment by ahojnnes** [comment]
How does this PR relate to #4433 ?

File: src/colmap/estimators/generalized_pose.cc [hunk 1]
```diff
@@ -223,15 +223,21 @@ bool EstimateGeneralizedRelativePose(
     Rigid3d cam2_from_cam1;
     std::vector<Eigen::Vector3d> cam_rays1(num_points);
     std::vector<Eigen::Vector3d> cam_rays2(num_points);
+    // Precompute the (few) inverse rig rotations once instead of per point.
+    std::vector<Eigen::Quaterniond> rig_from_cam_rotations(
+        cams_from_rig.size());
+    for (size_t i = 0; i < cams_from_rig.size(); ++i) {
+      rig_from_cam_rotations[i] = cams_from_rig[i].rotation().inverse();
+    }
```
ahojnnes: I am a bit skeptical that this leads to a speedup. Inverting a quanternion just involves negating the sign, which should be very fast. Did you profile this?

src/colmap/estimators/triangulation.cc
```diff
@@ -55,16 +57,65 @@ void TriangulationEstimator::Estimate(const std::vector<X_t>& point_data,
 
   models->clear();
 
+  // A camera with no focal length (omnidirectional, e.g. SPHERICAL) sees the
+  // full sphere: its observations can lie in the back hemisphere, so neither
+  // the 2D (u, v, 1) representation nor the positive-depth (cheirality)
+  // constraint apply. Detect these via an empty focal-length parameter set.
+  const auto is_perspective = [](const Camera* camera) {
+    return camera != nullptr && camera->FocalLengthIdxs().size() > 0;
+  };
```
ahojnnes: I'd rather introduce this as a static flag on the underlying camera model and then expose it through the Camera class as a method, as the same logic is reused in different places.

src/colmap/estimators/triangulation.cc
```diff
@@ -164,6 +225,17 @@ bool EstimateTriangulation(const EstimateTriangulationOptions& options,
...
+    // omnidirectional models (e.g. SPHERICAL) where back-hemisphere
+    // observations produce nullopt from CamFromImg but still yield a valid
+    // unit ray from CamRayFromImg.
+    if (const std::optional<Eigen::Vector3d> cam_ray =
+            cameras[i]->CamRayFromImg(points[i]);
+        cam_ray) {
+      point_data[i].cam_ray = *cam_ray;
+    } else {
+      point_data[i].cam_ray.setZero();
+    }
```
ahojnnes: .value_or() let's you write this in a single line.

src/colmap/geometry/triangulation.cc
```diff
@@ -130,6 +130,35 @@ bool TriangulateMultiViewPoint(
   return true;
 }
 
+bool TriangulateMultiViewPoint(
```
ahojnnes, line 178: Needs corresponding unit tests.

src/colmap/scene/camera.cc
```diff
@@ -73,6 +81,10 @@ Eigen::Matrix3d Camera::CalibrationMatrix() const {
 
 double Camera::MeanFocalLength() const {
   const span<const size_t> focal_length_idxs = FocalLengthIdxs();
+  // Omnidirectional cameras (e.g. SPHERICAL) have no focal length.
+  if (focal_length_idxs.size() == 0) {
```
ahojnnes: .empty()

src/colmap/sensor/models.h
```diff
@@ -604,6 +606,73 @@ struct EUCMCameraModel : public BaseCameraModel<EUCMCameraModel> {
...
+  // compute the ray directly from the azimuth/elevation parametrization.
+  static inline bool CamRayFromImg(const double* params,
+                                   double x,
+                                   double y,
+                                   double* rx,
+                                   double* ry,
+                                   double* rz) {
+    const double width = params[0];
+    const double height = params[1];
+    const double theta = 2.0 * M_PI * (x / width - 0.5);
```
ahojnnes: Let's use EIGEN_PI instead here and above.

src/colmap/sensor/models.h
```diff
@@ -604,6 +606,73 @@ struct EUCMCameraModel : public BaseCameraModel<EUCMCameraModel> {
...
+                                    size_t height,
+                                    T min_focal_length_ratio,
+                                    T max_focal_length_ratio,
+                                    T max_extra_param) {
+    (void)params;
+    (void)width;
+    (void)height;
+    (void)min_focal_length_ratio;
+    (void)max_focal_length_ratio;
+    (void)max_extra_param;
```
ahojnnes:
```suggestion
  static inline bool HasBogusParams(const std::vector<T>& /*params*/,
                                    size_t /*width*/,
                                    size_t height*/,
                                    T /*min_focal_length_ratio*/,
                                    T /*max_focal_length_ratio*/,
                                    T /*max_extra_param*/) {
```

src/colmap/sensor/models.h
```diff
@@ -2515,6 +2584,90 @@ bool EUCMCameraModel::CamFromImg(const double* params,
   return true;
 }
 
+////////////////////////////////////////////////////////////////////////////////
+// SphericalCameraModel
+
+std::string SphericalCameraModel::InitializeParamsInfo() { return "w, h"; }
```
ahojnnes:
```suggestion
std::string SphericalCameraModel::InitializeParamsInfo() { return "w,h"; }
```

src/colmap/sensor/models.h
```diff
@@ -604,6 +606,73 @@ struct EUCMCameraModel : public BaseCameraModel<EUCMCameraModel> {
...
+// Spherical (equirectangular panorama) camera model.
+//
+// Maps the full 360°x180° sphere onto an equirectangular image: the azimuth
+// spans the image width and the elevation spans the image height. The model
+// is fully specified by the image dimensions, so the two parameters are the
+// width and height; there is no focal length and no lens distortion.
+//
+// Parameter list is expected in the following order:
+//
+//    w, h
```
ahojnnes, line 743: I'd rather encode these as a principal point instead of extra parameters.

src/colmap/sensor/models.h
```diff
@@ -2515,6 +2584,90 @@ bool EUCMCameraModel::CamFromImg(const double* params,
...
+}
+
+// SPHERICAL has no lens distortion. The macro declares a Distortion function;
+// this no-op satisfies the interface.
+template <typename T>
+void SphericalCameraModel::Distortion(
+    const T* /*extra_params*/, const T& /*u*/, const T& /*v*/, T* du, T* dv) {
+  *du = T(0);
+  *dv = T(0);
+}
```
ahojnnes: Not needed.

src/colmap/estimators/triangulation.h
```diff
@@ -57,14 +57,28 @@ class TriangulationEstimator {
...
+        : img_point(img_point), cam_point(cam_point), cam_ray(cam_ray) {}
     // Image observation in pixels. Only needs to be set for REPROJECTION_ERROR.
-    Eigen::Vector2d img_point;
-    // Normalized camera coordinates. Must always be set.
-    Eigen::Vector2d cam_point;
+    Eigen::Vector2d img_point = Eigen::Vector2d::Zero();
+    // Normalized camera coordinates (u = X/Z, v = Y/Z). Must be set for
+    // perspective cameras. For omnidirectional (SPHERICAL) cameras, back-
+    // hemisphere observations may leave this zero and rely on cam_ray only.
+    Eigen::Vector2d cam_point = Eigen::Vector2d::Zero();
```
ahojnnes: Unless I misread the changes, we could just change everything to cam_ray and get rid of cam_point?

src/colmap/estimators/two_view_geometry.cc
```diff
@@ -320,12 +320,30 @@ TwoViewGeometry EstimateTwoViewGeometry(
...
+    };
+    if (is_calibrated(camera1) && is_calibrated(camera2)) {
+      return EstimateCalibratedTwoViewGeometry(
+          camera1, points1, camera2, points2, matches, options);
+    } else if (camera1.FocalLengthIdxs().size() == 0 ||
+               camera2.FocalLengthIdxs().size() == 0) {
+      // A non-perspective camera (e.g. SPHERICAL) paired with an uncalibrated
+      // perspective one: F/H on the non-perspective image are not meaningful
+      // and there is no calibration to fall back on, so bail out rather than
+      // waste RANSAC iterations on a meaningless model.
```
ahojnnes: It should be possible to solve this case, not sure if there is an existing suitable solver though.

src/colmap/estimators/two_view_geometry.cc
```diff
@@ -508,8 +526,17 @@ bool EstimateTwoViewGeometryPoseFromCamRays(
...
+  // valid for them. The matcher commits such pairs to CALIBRATED, but guard
+  // here as well so a stale or unexpected config can never reach the
+  // CalibrationMatrix() calls below (which would read out of bounds).
+  const bool non_perspective = camera1.FocalLengthIdxs().size() == 0 ||
+                               camera2.FocalLengthIdxs().size() == 0;
+
   Rigid3d cam2_from_cam1;
-  if (geometry->config == TwoViewGeometry::ConfigurationType::CALIBRATED) {
+  if (geometry->config == TwoViewGeometry::ConfigurationType::CALIBRATED ||
+      (non_perspective && geometry->E.has_value())) {
```
ahojnnes: Do we really need this expression? If the cameras are indeed spherical, the geometry should be labelled as CALIBRATED?

src/colmap/exe/image.cc
```diff
@@ -383,6 +383,23 @@ int RunImageUndistorter(int argc, char** argv) {
     }
   }
 
+  // Reject non-perspective (e.g. SPHERICAL) cameras up front with a clear
+  // message: undistortion produces a pinhole image, which is undefined for
+  // omnidirectional models. This avoids launching the worker pool only to
+  // fail per-image inside UndistortCamera.
+  for (const auto& [camera_id, camera] : reconstruction.Cameras()) {
+    if (camera.FocalLengthIdxs().size() == 0) {
+      LOG(ERROR) << "Camera " << camera_id << " uses model "
```
ahojnnes: I'd turn this into a WARNING instead of an error and simply skip over those images or copy them.

src/colmap/geometry/triangulation.cc
```diff
@@ -130,6 +130,35 @@ bool TriangulateMultiViewPoint(
...
+  // Same projector-based DLT as the 2D overload, but starting directly from
+  // a 3D bearing. Works for any camera model whose unprojection returns a
+  // unit ray, including omnidirectional (SPHERICAL) cameras where the legacy
+  // 2D (u, v, 1) representation can't encode back-hemisphere rays.
+  Eigen::Matrix4d A = Eigen::Matrix4d::Zero();
+  for (size_t i = 0; i < cam_rays.size(); ++i) {
+    const Eigen::Vector3d point = cam_rays[i].normalized();
+    const Eigen::Matrix3x4d term =
+        cams_from_world[i] - point * point.transpose() * cams_from_world[i];
+    A += term.transpose() * term;
```
ahojnnes: We could extract this as a small helper and reuse here and above.

src/colmap/sensor/models.h
```diff
@@ -604,6 +606,73 @@ struct EUCMCameraModel : public BaseCameraModel<EUCMCameraModel> {
...
+// width and height; there is no focal length and no lens distortion.
+//
+// Parameter list is expected in the following order:
+//
+//    w, h
+//
+// Named "SPHERICAL" and parametrized by (w, h) to match PoseLib's
+// SphericalCameraModel, so cameras map across the COLMAP/PoseLib boundary by
+// model name.
+struct SphericalCameraModel : public BaseCameraModel<SphericalCameraModel> {
```
ahojnnes: Needs to be covered in the unit tests.

src/colmap/sfm/observation_manager.cc
```diff
@@ -536,6 +542,24 @@ size_t ObservationManager::FilterPoints3DWithLargeReprojectionError(
...
+                camera.CamRayFromImg(point2D.xy);
+            if (!cam_ray.has_value()) {
+              should_filter = true;
+              break;
+            }
+            const double squared_error =
+                (point3D_in_cam.normalized() - *cam_ray).squaredNorm();
+            should_filter = squared_error > max_squared_error;
+            observation_error = std::sqrt(squared_error);
+            break;
```
ahojnnes: would instead do else {}

src/colmap/sfm/observation_manager.cc
```diff
@@ -536,6 +542,24 @@ size_t ObservationManager::FilterPoints3DWithLargeReprojectionError(
...
+            const std::optional<Eigen::Vector3d> cam_ray =
+                camera.CamRayFromImg(point2D.xy);
+            if (!cam_ray.has_value()) {
+              should_filter = true;
+              break;
+            }
+            const double squared_error =
+                (point3D_in_cam.normalized() - *cam_ray).squaredNorm();
+            should_filter = squared_error > max_squared_error;
+            observation_error = std::sqrt(squared_error);
```
ahojnnes: Can just directly compute the norm() instead of squaredNorm()

python/examples/panorama_sfm_native.py
```diff
@@ -0,0 +1,559 @@
+#!/usr/bin/env python3
```
ahojnnes: Suggest removing this script as it is redundant to the pycolmap one below.

ahojnnes (commented):
Thanks, some initial comments.

ahojnnes:
Do you have an A/B comparison to how this performs relative to running the existing panorama_sfm on the same data? For the reported results, it shows 0.0015px mean reprojection error for the global pipeline, which sounds suspiciously low. Are you sure there is no mistake?

ichsan2895:
> For the reported results, it shows 0.0015px mean reprojection error for the global pipeline, which sounds suspiciously low. Are you sure there is no mistake?

Seems that is a bug from this one? https://github.com/colmap/colmap/issues/4436. The bugfixes is this PR: https://github.com/colmap/colmap/pull/4442

zalo:
Thank you for the review!  Here are Claude’s comments; I’ll rerun my test set with these too:

### `models.h`
- **Encode `(w, h)` as principal-point params** — changed the model to
  `(0 focal, 2 pp, 0 extra)`. They stay constant in BA via the existing
  zero-focal guard in `ParameterizeCameras` (the whole param block is pinned
  for cameras with no focal length), so `refine_principal_point` can't move
  them.
- **`EIGEN_PI`** instead of `M_PI` everywhere in the model. ✅
- **`InitializeParamsInfo` → `"w,h"`**. ✅
- **`HasBogusParams`** — commented out the unused argument names, dropped the
  `(void)` casts. ✅
- **Dropped the no-op `Distortion`** override (matches `SimplePinhole`, which
  omits it now that there are no extra params). ✅
- **Unit tests** — added `TEST(Spherical, Nominal)` covering the metadata
  (focal idxs empty, pp = `{0, 1}`, no extra), `IsPerspective() == false`,
  `InitializeParams`, full-sphere `CamRayFromImg → ImgFromCam` round-trips, and
  the back-hemisphere case where `CamFromImg` fails but `CamRayFromImg`
  succeeds.

### `is_perspective` flag
Introduced `is_perspective` as a `static constexpr` on every camera model (in
the `CAMERA_MODEL_DEFINITIONS` macro), exposed via `CameraModelIsPerspective()`
and `Camera::IsPerspective()`. Replaced the duplicated
`FocalLengthIdxs().size()` checks in `triangulation`, `two_view_geometry`,
`observation_manager`, and `bundle_adjustment_ceres` with this method.
(Implemented as `num_focal_params > 0` so it applies to all models without
touching each definition — happy to switch to an explicit per-model argument if
you'd prefer the two concepts decoupled.)

### Triangulation
- **Unified on `cam_ray`** — removed the redundant 2D `cam_point` from
  `PointData` and the dual 2D/bearing code paths in the estimator. The ray is
  now set with `CamRayFromImg().value_or(...)`. The 2-view path consequently
  goes through the bearing DLT rather than the dedicated `TriangulatePoint`;
  all triangulation tests still pass.
- **`.value_or()`** one-liner for the ray. ✅
- **Helper reuse** — the 2D `TriangulateMultiViewPoint` overload now converts
  points → rays and delegates to the 3D-bearing overload, so the DLT lives in
  one place.
- **Unit tests** — added `TEST(TriangulateMultiViewPoint, Bearings)`, including
  back-hemisphere (negative-Z) points the 2D overload can't represent.

### `two_view_geometry`
Dropped the redundant `non_perspective` expression in the pose-from-cam-rays
path — spherical pairs are already committed to `CALIBRATED` upstream, so the
`CALIBRATED` branch handles them.

### `generalized_pose`
Reverted the inverse-rotation precompute — you're right, the quaternion inverse
is just a conjugate; back to the inline `.inverse()` per point.

### `image_undistorter`
Turned the hard error into a **warning + skip**: non-perspective cameras are
skipped with a warning and the original image is copied through, and
`UndistortReconstruction` leaves those cameras and their observations
unchanged. `UndistortCamera` keeps a defensive `THROW_CHECK` so any unforeseen
path fails clearly instead of segfaulting.

### `observation_manager`
Switched the perspective filter path to an `else { ... }` branch.

### `util/types.h`
Added `span::empty()` (needed for the `.empty()` cleanup in `camera.cc`).

---

### Not addressed: F/H-free relative pose for non-perspective + uncalibrated
> It should be possible to solve this case, not sure if there is an existing
> suitable solver though.

I left this returning `DEGENERATE` for now — there's no suitable solver in the
tree for a bearing camera paired with an unknown-focal pinhole, and adding one
felt out of scope for this PR. In practice the perspective side usually has a
focal prior (EXIF, `--ImageReader.camera_params`, or `view_graph_calibrator`),
in which case both are treated as calibrated and the bearing essential-matrix
path handles the pair. Happy to file a follow-up if you'd like this solved
properly.

Validation: all affected unit suites pass (models, geometry/triangulation,
estimators/{triangulation,two_view_geometry,generalized_pose}, scene/camera,
sfm/observation_manager), plus the two new tests. SPHERICAL incremental SfM
re-validated end-to-end (66/66 images, 0.576 px mean reprojection error,
unchanged), and the model round-trips to text as `SPHERICAL w h`.

src/colmap/util/types.h
```diff
@@ -239,6 +239,7 @@ class span {
   T const& operator[](size_t i) const noexcept { return ptr_[i]; }
 
   size_t size() const noexcept { return size_; }
+  bool empty() const noexcept { return size_ == 0; }
```
ahojnnes, line 242: Let's cover this new method in the unit tests.

src/colmap/sfm/observation_manager.cc
```diff
@@ -536,23 +542,41 @@ size_t ObservationManager::FilterPoints3DWithLargeReprojectionError(
...
-          if (point3D_in_cam.z() < kMinDepth) {
-            should_filter = true;
-            break;
-          }
-          const std::optional<Eigen::Vector2d> cam_point =
-              camera.CamFromImg(point2D.xy);
-          if (!cam_point.has_value()) {
-            should_filter = true;
-            break;
+          if (!camera.IsPerspective()) {
```
ahojnnes: Let's swap the order of the branches and avoid the "!".

src/colmap/sensor/models.h
```diff
@@ -99,57 +99,62 @@ MAKE_ENUM_CLASS_OVERLOAD_STREAM(CameraModelId,
...
+                                 has_img_from_cam_with_jac_val)                \
+  static constexpr size_t num_params =                                         \
+      (num_focal_params_val) + (num_pp_params_val) + (num_extra_params_val);   \
+  static constexpr size_t num_focal_params = num_focal_params_val;             \
+  static constexpr size_t num_pp_params = num_pp_params_val;                   \
+  static constexpr size_t num_extra_params = num_extra_params_val;             \
+  /* Perspective models have a focal length and a finite pinhole image      */ \
+  /* plane (positive-depth cheirality applies). Omnidirectional models such */ \
+  /* as SPHERICAL have no focal length and are not perspective.             */ \
+  static constexpr bool is_perspective = (num_focal_params_val) > 0;           \
```
ahojnnes: I suggest to pass an explicit boolean flag as a macro parameter instead of automatically deriving this from the focal length parameters.

src/colmap/sensor/models.h
```diff
@@ -604,6 +610,69 @@ struct EUCMCameraModel : public BaseCameraModel<EUCMCameraModel> {
...
+// distortion parameters) since they are intrinsic image geometry.
+//
+// Parameter list is expected in the following order:
+//
+//    w, h
+//
+// Named "SPHERICAL" and parametrized by (w, h) to match PoseLib's
+// SphericalCameraModel, so cameras map across the COLMAP/PoseLib boundary by
+// model name.
+struct SphericalCameraModel : public BaseCameraModel<SphericalCameraModel> {
```
ahojnnes: Another cleaner alternative could be to have a BaseSpherical and BasePerspective camera model class similar to the fisheye camera model base class as suggested offline by @B1ueber2y.

src/colmap/scene/camera.cc
```diff
@@ -61,6 +61,14 @@ Camera Camera::CreateFromModelName(camera_t camera_id,
 }
 
 Eigen::Matrix3d Camera::CalibrationMatrix() const {
+  // Omnidirectional cameras (no focal length, e.g. SPHERICAL) have no pinhole
+  // calibration matrix; fail loudly rather than reading out of bounds on the
+  // empty focal-length/principal-point spans. Callers that handle such
+  // cameras use the bearing interface (CamRayFromImg) instead.
+  THROW_CHECK_GT(FocalLengthIdxs().size(), 0)
```
ahojnnes:
```suggestion
  THROW_CHECK(!FocalLengthIdxs().empty() && !PrincipalPointIdxs().empty())
```

src/colmap/image/undistortion.cc
```diff
@@ -49,6 +49,18 @@ Camera UndistortCamera(const UndistortCameraOptions& options,
   THROW_CHECK_LT(options.roi_min_x, options.roi_max_x);
   THROW_CHECK_LT(options.roi_min_y, options.roi_max_y);
 
+  // Undistortion produces a pinhole image, which is only well-defined for
+  // perspective cameras. Omnidirectional models without a focal length (e.g.
+  // SPHERICAL) have no pinhole image plane and cannot be undistorted; reject
+  // them here with a clear message instead of dereferencing the (empty) focal
+  // length parameters below.
+  THROW_CHECK_GT(camera.FocalLengthIdxs().size(), 0)
```
ahojnnes: THROW_CHECK(camera.IsPerspective());

src/colmap/image/undistortion.cc
```diff
@@ -268,7 +280,9 @@ void UndistortReconstruction(const UndistortCameraOptions& options,
   const std::unordered_map<camera_t, Camera> distorted_cameras =
       reconstruction->Cameras();
   for (const auto& camera : distorted_cameras) {
-    if (camera.second.IsUndistorted()) {
+    // Non-perspective cameras (e.g. SPHERICAL) cannot be undistorted to a
+    // pinhole; leave them and their observations unchanged.
+    if (camera.second.IsUndistorted() || !camera.second.IsPerspective()) {
```
ahojnnes: Maybe the IsUndistorted() should return true for non-perspective.

ichsan2895:
A simple test. Your PR vs SphereSFM.
8 Equirectangular images with 10000 x 5000 resolution.

Duration (minutes) | Sphere SFM | This   PR_Incremental | This PR_Global
-- | -- | -- | --
colmap   feature_extractor | 0,078 | 0,238 | 0,238
colmap   exhaustive_matcher | 0,004 | 0,021 | 0,021
colmap mapper | 0,133 | 0,092 |  0
colmap   global_mapper | 0 | 0  | 0,120
Total | 0,215 | 0,351 | 0,379

```
$ colmap model_analyzer --path distorted/spheresfm_ceres/0
Cameras: 1
Images: 8
Registered images: 8
Points: 763
Observations: 3310
Mean track length: 4.338139
Mean observations per image: 413.750000
Mean reprojection error: 1.226867px
```

```
$ colmap model_analyzer --path distorted/colmap_ceres/0
I0611 07:58:50.036085 58656 model.cc:440] Rigs: 1
I0611 07:58:50.036592 58656 model.cc:441] Cameras: 1
I0611 07:58:50.036602 58656 model.cc:442] Frames: 8
I0611 07:58:50.036607 58656 model.cc:443] Registered frames: 8
I0611 07:58:50.036613 58656 model.cc:445] Images: 8
I0611 07:58:50.036623 58656 model.cc:446] Registered images: 8
I0611 07:58:50.036629 58656 model.cc:448] Points: 4289
I0611 07:58:50.036636 58656 model.cc:449] Observations: 15995
I0611 07:58:50.036644 58656 model.cc:451] Mean track length: 3.729308
I0611 07:58:50.036654 58656 model.cc:453] Mean observations per image: 1999.375000
I0611 07:58:50.036662 58656 model.cc:456] Mean reprojection error: 1.147644px
```

```
$ colmap model_analyzer --path distorted/glomap/0
I0611 07:58:51.532375 59436 model.cc:440] Rigs: 1
I0611 07:58:51.532840 59436 model.cc:441] Cameras: 1
I0611 07:58:51.532850 59436 model.cc:442] Frames: 8
I0611 07:58:51.532855 59436 model.cc:443] Registered frames: 8
I0611 07:58:51.532861 59436 model.cc:445] Images: 8
I0611 07:58:51.532868 59436 model.cc:446] Registered images: 8
I0611 07:58:51.533104 59436 model.cc:448] Points: 4165
I0611 07:58:51.533113 59436 model.cc:449] Observations: 15680
I0611 07:58:51.533124 59436 model.cc:451] Mean track length: 3.764706
I0611 07:58:51.533138 59436 model.cc:453] Mean observations per image: 1960.000000
I0611 07:58:51.533149 59436 model.cc:456] Mean reprojection error: 1.127161px
```

ahojnnes:
@ichsan2895 Thanks, so it looks like this is better than SphereSfM, which is not a surprise. How about this PR vs. the existing panorama_sfm ?

src/colmap/geometry/triangulation.cc
```diff
@@ -110,11 +110,31 @@ bool TriangulateMultiViewPoint(
     const span<const Eigen::Vector2d>& cam_points,
     Eigen::Vector3d* xyz) {
   THROW_CHECK_EQ(cams_from_world.size(), cam_points.size());
+  // Delegate to the 3D-bearing overload, which holds the shared DLT.
+  std::vector<Eigen::Vector3d> cam_rays(cam_points.size());
+  for (size_t i = 0; i < cam_points.size(); ++i) {
+    cam_rays[i] = cam_points[i].homogeneous().normalized();
+  }
```
ahojnnes: I didn't mean to suggest extracting the rays in a vector but simply have a small helper function that computes the right hand side of `A += term.transpose() * term;`.

ichsan2895:
> @ichsan2895 Thanks, so it looks like this is better than SphereSfM, which is not a surprise. How about this PR vs. the existing panorama_sfm ?

Okay, I update the previous benchmark..
Duration | Sphere SFM | This   PR_Incremental | This PR_Global | panorama_sfm_incremental | panorama_sfm_global
-- | -- | -- | -- | -- | --
extracting   images | 0 | 0 | 0 | 1,483 | 1,37
colmap   feature_extractor | 0,078 | 0,238 | 0,238 | 0,228 | 0,235
colmap   exhaustive_matcher | 0,004 | 0,021 | 0,021 | 0,51 | 0,575
rig   verification | 0 | 0 | 0 | 0,264 | 0,273
colmap mapper | 0,133 | 0,092 | 0 | 0,609 | 0
colmap   global_mapper | 0 | 0 | 0,120 | 0 | 0,75
Total | 0,215 | 0,351 | 0,379 | 3,094 | 3,200

```
$ colmap model_analyzer --path distorted/colmap_ceres/0
I0611 07:58:50.036085 58656 model.cc:440] Rigs: 1
I0611 07:58:50.036592 58656 model.cc:441] Cameras: 1
I0611 07:58:50.036602 58656 model.cc:442] Frames: 8
I0611 07:58:50.036607 58656 model.cc:443] Registered frames: 8
I0611 07:58:50.036613 58656 model.cc:445] Images: 8
I0611 07:58:50.036623 58656 model.cc:446] Registered images: 8
I0611 07:58:50.036629 58656 model.cc:448] Points: 4289
I0611 07:58:50.036636 58656 model.cc:449] Observations: 15995
I0611 07:58:50.036644 58656 model.cc:451] Mean track length: 3.729308
I0611 07:58:50.036654 58656 model.cc:453] Mean observations per image: 1999.375000
I0611 07:58:50.036662 58656 model.cc:456] Mean reprojection error: 1.147644px

$ colmap model_analyzer --path distorted/glomap/0
I0611 07:58:51.532375 59436 model.cc:440] Rigs: 1
I0611 07:58:51.532840 59436 model.cc:441] Cameras: 1
I0611 07:58:51.532850 59436 model.cc:442] Frames: 8
I0611 07:58:51.532855 59436 model.cc:443] Registered frames: 8
I0611 07:58:51.532861 59436 model.cc:445] Images: 8
I0611 07:58:51.532868 59436 model.cc:446] Registered images: 8
I0611 07:58:51.533104 59436 model.cc:448] Points: 4165
I0611 07:58:51.533113 59436 model.cc:449] Observations: 15680
I0611 07:58:51.533124 59436 model.cc:451] Mean track length: 3.764706
I0611 07:58:51.533138 59436 model.cc:453] Mean observations per image: 1960.000000
I0611 07:58:51.533149 59436 model.cc:456] Mean reprojection error: 1.127161px
```
```
$ panorama_sfm_incremental
num_rigs = 1
num_cameras = 12
num_frames = 8
num_reg_frames = 8
num_images = 96
num_points3D = 14318
num_observations = 55162
mean_track_length = 3.85263
mean_observations_per_image = 574.604
mean_reprojection_error = 1.00233

$ panorama_sfm_global
num_rigs = 1
num_cameras = 12
num_frames = 8
num_reg_frames = 8
num_images = 96
num_points3D = 14153
num_observations = 56166
mean_track_length = 3.96849
mean_observations_per_image = 585.062
mean_reprojection_error = 0.992227
```

src/colmap/sensor/models.h
B1ueber2y: Thanks Johannes. Here is a detailed restructuring proposal:

```
BaseCameraModel
    - BasePerspectiveCameraModel (previously the ``BaseCameraModel``)
        - bunch of current perspective models
        - BasePerspectiveFisheyeCameraModel (previously the ``BaseFisheyeCameraModel``)
            - bunch of our current fisheye models
    - SphereicalCameraModel
```

python/examples/panorama_sfm_native.py
ahojnnes: I think we should remove both scripts and instead add the spherical camera model as a new mode in the existing panorama_sfm.py.

src/colmap/sensor/models.h
zalo: Since there are multiple projections of a sphere, I wouldn’t be against a “base” spherical camera.

There are lower distortion projections of spherical images, like octahedral mapping, which might be desirable to add in future PRs.

zalo:
Sweet; 
- Seems like a lot fewer points, but this is to be expected, SIFT and all of the other image-space logic assumes rectilinear images.  It still works quite well despite the distortion of the images.
- It’s tough to compare mean reprojection errors across two projections (I suppose the “right” way to do it would be to take the panorama_sfm results and compute the reprojection error in equirectangular space, to get an apples to apples)
- the whole reason I did this PR (speed!) seems to be a huge victory!   I’m surprised that SphereSFM is even faster, but it seems to be entirely in the feature extraction step, so there’s more to do there.  I wonder if the COLMAP version here was on the GPU?
- Perhaps we can add a frontend to the feature extractor that extracts locally rectilinear patches from the equirectangular image to improve matching performance in a future PR…

ichsan2895:
Another dataset: 20 equirectangular images with 7680x3840 resolution:
Duration | Sphere SFM | This   PR_Incremental | This PR_Global | panorama_sfm_incremental
-- | -- | -- | -- | --
extracting   images | 0 | 0 | 0 | 3,600
colmap   feature_extractor | 0,133 | 0,2 | 0,2 | 0,399
colmap   sequential_matcher | 0,027 | 0,045 | 0,045 | 2,433
rig   verification | 0 | 0 | 0 | 0,93
colmap mapper | 0,217 | 0,188 | 0 | 1,05
colmap   global_mapper | 0 | 0 | 0,333 | 0
Total | 0,377 | 0,433 | 0,578 | 8,409 |

```
$ colmap model_analyzer --path distorted/sphere_sfm/0
Cameras: 1
Images: 20
Registered images: 20
Points: 993
Observations: 5278
Mean track length: 5.315206
Mean observations per image: 263.900000
Mean reprojection error: 1.031232px
```
```
$ colmap model_analyzer --path distorted_zalo/colmap_ceres/0
I0611 09:55:21.734493 195293 model.cc:440] Rigs: 1
I0611 09:55:21.735345 195293 model.cc:441] Cameras: 1
I0611 09:55:21.735357 195293 model.cc:442] Frames: 20
I0611 09:55:21.735365 195293 model.cc:443] Registered frames: 20
I0611 09:55:21.735373 195293 model.cc:445] Images: 20
I0611 09:55:21.735378 195293 model.cc:446] Registered images: 20
I0611 09:55:21.735383 195293 model.cc:448] Points: 6131
I0611 09:55:21.735389 195293 model.cc:449] Observations: 30094
I0611 09:55:21.735399 195293 model.cc:451] Mean track length: 4.908498
I0611 09:55:21.735415 195293 model.cc:453] Mean observations per image: 1504.700000
I0611 09:55:21.735425 195293 model.cc:456] Mean reprojection error: 1.024820px

$ colmap model_analyzer --path distorted_zalo/glomap/0
I0611 09:55:22.270066 195550 model.cc:440] Rigs: 1
I0611 09:55:22.270673 195550 model.cc:441] Cameras: 1
I0611 09:55:22.270684 195550 model.cc:442] Frames: 20
I0611 09:55:22.270689 195550 model.cc:443] Registered frames: 20
I0611 09:55:22.270695 195550 model.cc:445] Images: 20
I0611 09:55:22.270701 195550 model.cc:446] Registered images: 20
I0611 09:55:22.270706 195550 model.cc:448] Points: 6274
I0611 09:55:22.270711 195550 model.cc:449] Observations: 30852
I0611 09:55:22.270720 195550 model.cc:451] Mean track length: 4.917437
I0611 09:55:22.270730 195550 model.cc:453] Mean observations per image: 1542.600000
I0611 09:55:22.270740 195550 model.cc:456] Mean reprojection error: 0.990568px

$ panorama_sfm_incremental
num_rigs = 1
num_cameras = 12
num_frames = 20
num_reg_frames = 20
num_images = 240
num_points3D = 21111
num_observations = 109259
mean_track_length = 5.17545
mean_observations_per_image = 455.246
mean_reprojection_error = 0.719734
```

src/colmap/estimators/triangulation.cc
```diff
@@ -157,13 +146,11 @@ bool EstimateTriangulation(const EstimateTriangulationOptions& options,
...
-        cam_point) {
-      point_data[i].cam_point = *cam_point;
-    } else {
-      point_data[i].cam_point.setZero();
-    }
+    // Unit bearing in the camera frame. CamRayFromImg yields a valid ray for
+    // any camera model, including omnidirectional (SPHERICAL) back-hemisphere
+    // observations that CamFromImg cannot represent.
+    point_data[i].cam_ray =
+        cameras[i]->CamRayFromImg(points[i]).value_or(Eigen::Vector3d::Zero());
```
ahojnnes: **Possible regression (affects perspective cameras, not just SPHERICAL):** when `CamRayFromImg` fails, `value_or(Eigen::Vector3d::Zero())` stores a zero ray. Downstream in `TriangulationDltTerm` (`geometry/triangulation.cc`), `cam_ray.normalized()` on a zero vector is `0/0 = NaN`, which poisons the accumulated `A` matrix and the eigen-solve. The previous code stored a zeroed 2D `cam_point` whose `.homogeneous().normalized()` was a defined `(0,0,1)` bearing and never NaN'd. Consider skipping observations with a failed/zero ray, or guarding `normalized()` against a near-zero norm. (SPHERICAL itself never fails here, so this only bites perspective cameras whose unprojection fails.)

src/colmap/estimators/triangulation.cc
```diff
@@ -55,61 +57,47 @@ void TriangulationEstimator::Estimate(const std::vector<X_t>& point_data,
...
+  }
 
-    M_t xyz;
-    if (!TriangulateMultiViewPoint(
-            span<const Eigen::Matrix3x4d>(cams_from_world.data(),
-                                          cams_from_world.size()),
-            span<const Eigen::Vector2d>(cam_points.data(), cam_points.size()),
-            &xyz)) {
+  M_t xyz;
+  if (!TriangulateMultiViewPoint(
```
ahojnnes: **Behavioral + perf change for all camera models:** removing the `point_data.size() == 2` branch drops the closed-form 2-view `TriangulatePoint` path — every sample now runs this projector-DLT + `SelfAdjointEigenSolver<Matrix4d>`. Since `kMinNumSamples == 2`, the minimal RANSAC sample in `EstimateTriangulation` is 2-view, so this swaps the inner-loop estimator for **all** reconstructions to a heavier eigendecomposition with slightly different numerics. Could you benchmark runtime + point count / reprojection error on a standard perspective dataset to confirm no regression — or keep the closed-form 2-view path for perspective cameras?

src/colmap/estimators/triangulation.cc
```diff
@@ -55,61 +57,47 @@ void TriangulationEstimator::Estimate(const std::vector<X_t>& point_data,
...
+          span<const Eigen::Vector3d>(cam_rays.data(), cam_rays.size()),
+          &xyz)) {
+    return;
+  }
+
+  // Cheirality: points must lie in front of perspective cameras. Skipped for
+  // omnidirectional cameras (e.g. SPHERICAL), which legitimately observe
+  // points behind their local +Z axis.
+  for (const auto& pose : pose_data) {
+    if (pose.camera != nullptr && pose.camera->IsPerspective() &&
```
ahojnnes: **Cheirality silently skipped when `pose.camera == nullptr`.** `PoseData`'s default ctor sets `camera = nullptr`, and the old multi-view path applied cheirality unconditionally. Any caller that leaves `camera` unset now silently loses the cheirality check (a quiet quality regression rather than a crash). Suggest `THROW_CHECK_NOTNULL(pose.camera)` in `Estimate`, or documenting that `camera` must be set for correct cheirality behavior.

src/colmap/sensor/models.h
```diff
@@ -2515,6 +2605,81 @@ bool EUCMCameraModel::CamFromImg(const double* params,
...
+  // focal_length is ignored: an equirectangular projection is fully specified
+  // by the image dimensions.
+  return {static_cast<double>(width), static_cast<double>(height)};
+}
+
+// Projects camera-frame point (u, v, w) onto the equirectangular image plane.
+// Unlike pinhole/fisheye models that require w > 0, SPHERICAL accepts any
+// non-zero direction — all 4π sr of the sphere are representable.
+template <typename T>
+bool SphericalCameraModel::ImgFromCam(
```
ahojnnes: **Equirectangular seam discontinuity in bundle adjustment.** `ImgFromCam` wraps azimuth at ±π, so the pixel x-coordinate jumps from ~0 to ~`w` across the seam. The standard reprojection-error BA cost functor calls `ImgFromCam`, so an observation straddling the seam produces a ~`w`-pixel residual and a bad gradient. Could you confirm SPHERICAL BA uses an angular residual (the PR notes mention angular/normalized inlier filtering, but the BA cost path doesn't appear changed), or otherwise document the seam as a known limitation? Worth checking empirically how many tracks land near x∈{0,w}.

ahojnnes (commented):
Thorough review of the SPHERICAL camera model PR. Overall the design is sound — threading a 3D unit-bearing representation (`CamRayFromImg`) through triangulation / two-view geometry / angular error, gated by `Camera::IsPerspective()`, is the right approach, and the test coverage (back-hemisphere triangulation, seam/pole avoidance, `span::empty()`) is good.

The main concerns are that a few **shared** code paths were changed in ways that affect all camera models, not just SPHERICAL. Inline comments cover H1–H3 and M1; the remaining items are below.

**M2 — `MeanFocalLength()` now returns 0 for SPHERICAL.** Please audit the call sites: `estimators/view_graph_calibration.cc:217` (view-graph calibration — relevant to the global mapper path this PR exercises), `image/undistortion.cc:348` (`K(0,0)=min(f1,f2)` → 0), and the Bundler/NVM/PLY exporters in `scene/reconstruction_io.cc` (will emit focal=0). The exporters are probably acceptable, but worth confirming view_graph_calibration excludes no-focal cameras.

**M3 — F/H RANSAC still runs for SPHERICAL pairs.** The new comment in `EstimateTwoViewGeometry` says non-perspective pairs "bail out rather than waste RANSAC iterations on a meaningless model," but `EstimateCalibratedTwoViewGeometry` still runs the F- and H-matrix RANSAC on raw equirectangular pixels; only `config` is overridden to CALIBRATED afterwards. Consider actually skipping F/H estimation for non-perspective cameras to realize the claimed savings.

**L1 — docs.** `doc/cameras.rst` enumerates every camera model; please add SPHERICAL (and a CHANGELOG entry).

**L2 — nit.** In `CAMERA_MODEL_DEFINITIONS` (`sensor/models.h`), the new `is_perspective` comment block has a misaligned line-continuation backslash on the middle `/* … */` line.

**L3 — naming (optional).** Fisheye models (incl. >180° FOV like `OPENCV_FISHEYE`) are tagged `is_perspective = true` and keep the positive-depth cheirality gate. That preserves existing behavior, but "perspective" reads oddly for fisheye — the distinction actually used everywhere is "has a focal length / pinhole image plane." A name like `IsOmnidirectional()` (or deriving it from `FocalLengthIdxs().empty()`) might read truer.

_Review assembled with Claude Code._

src/colmap/estimators/triangulation.cc
ahojnnes: I guess a solution can be to use `.value_or(Eigen::Vector3d::UnitZ());`

src/colmap/sensor/models.h
ahojnnes: Maybe we should clamp points near the poles during feature extraction or upon loading points?

src/colmap/sensor/models.h
ahojnnes: I suggest to go with @B1ueber2y's suggestion.

src/colmap/scene/camera.cc
```diff
@@ -61,6 +61,13 @@ Camera Camera::CreateFromModelName(camera_t camera_id,
 }
 
 Eigen::Matrix3d Camera::CalibrationMatrix() const {
+  THROW_CHECK(!FocalLengthIdxs().empty())
```
B1ueber2y: Can these be replaced simply with ``IsSpherical``?

src/colmap/scene/camera.cc
```diff
@@ -73,6 +80,10 @@ Eigen::Matrix3d Camera::CalibrationMatrix() const {
 
 double Camera::MeanFocalLength() const {
   const span<const size_t> focal_length_idxs = FocalLengthIdxs();
+  // Omnidirectional cameras (e.g. SPHERICAL) have no focal length.
+  if (focal_length_idxs.empty()) {
```
B1ueber2y: Same here. Replace with ``IsSpherical`` for better readability?

src/colmap/sensor/models.h
```diff
@@ -596,14 +628,79 @@ struct FisheyeCameraModel : public BaseFisheyeCameraModel<FisheyeCameraModel> {
...
                                          T max_extra_param);
 };
 
+// Spherical (equirectangular panorama) camera model.
+//
+// Maps the full 360°x180° sphere onto an equirectangular image: the azimuth
+// spans the image width and the elevation spans the image height. The model
+// is fully specified by the image dimensions, so the two parameters are the
+// width and height; there is no focal length and no lens distortion. The two
+// parameters are classified as principal-point parameters (not extra/
```
B1ueber2y: The width and height as principal points is quite confusing. How about we put all three parts as {} for spherical models, and add a small spherical branch in the ``Rescale`` method, which I believe was the main motivation for this design.

src/colmap/estimators/two_view_geometry_test.cc
```diff
@@ -386,6 +386,119 @@ TwoViewGeometryTestData CreateTwoViewGeometryTestData(
   return data;
 }
 
+TEST(EstimateTwoViewGeometry, Spherical) {
```
B1ueber2y, line 392: Should we check the resulting cam2_from_cam1 in the tvg tests?

src/colmap/estimators/triangulation.cc
```diff
@@ -55,59 +57,67 @@ void TriangulationEstimator::Estimate(const std::vector<X_t>& point_data,
...
-    // Check for cheirality constraint.
-    for (const auto& pose : pose_data) {
-      if (!HasPointPositiveDepth(pose.cam_from_world, xyz)) {
-        return;
-      }
+  // Cheirality: points must lie in front of perspective cameras. Skipped for
+  // omnidirectional cameras (e.g. SPHERICAL), which legitimately observe
+  // points behind their local +Z axis.
+  for (const auto& pose : pose_data) {
+    if (pose.camera->IsPerspective() &&
```
B1ueber2y: Probably it makes sense to generalize the logic here to spherical models? something like (cam_from_world · X).dot(cam_ray) > 0?

src/colmap/estimators/two_view_geometry.cc
```diff
@@ -320,12 +416,22 @@ TwoViewGeometry EstimateTwoViewGeometry(
...
-    return EstimateCalibratedTwoViewGeometry(
-        camera1, points1, camera2, points2, matches, options);
   } else {
-    return EstimateUncalibratedTwoViewGeometry(
-        camera1, points1, camera2, points2, matches, options);
+    // A spherical (omnidirectional) camera has no pinhole image plane, so only
+    // the bearing-based essential matrix is meaningful. Otherwise, use the
+    // calibrated path if both cameras have a known focal length, and the
+    // uncalibrated path otherwise.
+    if (camera1.IsSpherical() || camera2.IsSpherical()) {
```
B1ueber2y, line 431: The mixed case also comes to this branch and is probably not reliable, if the prior focal length is very off.

src/colmap/controllers/undistorters.cc
```diff
@@ -233,11 +233,14 @@ bool COLMAPUndistorter::Undistort(const image_t image_id) const {
...
-  // scaling is needed
-  if (camera.IsUndistorted() && camera_options_.max_image_size < 0 &&
-      ExistsFile(input_image_path)) {
-    LOG(INFO) << "Copying already distorted image to location: "
+  // Already-undistorted cameras (undistorted perspective and all
+  // non-perspective models, e.g. SPHERICAL) are left unchanged by
+  // UndistortReconstruction, so copy their images through unchanged to keep the
+  // exported image and its camera the same size. max_image_size is
+  // intentionally not applied here, since the camera dimensions are not
+  // rescaled.
```
B1ueber2y: I dont understand this new comment. With this change, a smaller max_image_size on already undistorted images would not work here for the undistortion for MVS, and the images are still directly copied. This does not feel right to me (maybe I misunderstood sth).

B1ueber2y (commented):
Thanks for the great efforts! Left some comments : )

zalo:
Thank you for the review!

src/colmap/scene/camera.cc
zalo: Yep!

src/colmap/sensor/models.h
zalo: Claude says:

> This is the one I'd like to confirm direction on before changing, since it's a
maintainer call and it reverses an earlier request. The current
principal-point encoding was chosen so `Camera::Rescale` scales `(w, h)`
automatically via the existing principal-point scaling (no model-specific
branch). Your alternative — empty focal/principal-point/extra index groups plus
a spherical branch in `Rescale` — is cleaner semantically, but it needs
`num_params` to be decoupled from the sum of the index-group sizes in
`CAMERA_MODEL_DEFINITIONS` (today `num_params = num_focal + num_pp + num_extra`),
so it touches the shared macro for all models. @ahojnnes had earlier suggested
the principal-point encoding (vs. extra params), so I want to make sure you two
are aligned before I change it. Happy to implement the empty-groups + `Rescale`
branch version if that's the consensus.

src/colmap/estimators/triangulation.cc
zalo: Fixed. Perspective cameras still require positive depth; spherical cameras now
require the triangulated point to lie in the half-space the observed bearing
points toward — `(cam_from_world * X).dot(cam_ray) > 0` — instead of skipping
the check. End-to-end on a 360 clip this rejects a few back-of-ray
mis-triangulations and lowers the mean reprojection error (0.576 → 0.513 px),
with the same registration.

src/colmap/estimators/two_view_geometry.cc
zalo: Good point. `EstimateSphericalTwoViewGeometry` now treats a mixed pair as
degenerate when the perspective camera lacks a focal prior
(`has_prior_focal_length`), since its bearings (from `CamFromImg`) would be
unreliable. Spherical/spherical and spherical/calibrated-perspective pairs are
unaffected.

src/colmap/controllers/undistorters.cc
zalo: Oops; fixed.

src/colmap/estimators/two_view_geometry_test.cc
zalo: Yep; good point!

src/colmap/estimators/two_view_geometry.cc
```diff
@@ -276,6 +276,113 @@ TwoViewGeometry EstimateMultipleTwoViewGeometries(
...
+
+  TwoViewGeometry geometry;
+
+  const size_t min_num_inliers = static_cast<size_t>(options.min_num_inliers);
+  if (matches.size() < min_num_inliers) {
+    geometry.config = TwoViewGeometry::ConfigurationType::DEGENERATE;
+    return geometry;
+  }
+
+  // The bearing-based essential matrix needs reliable bearings from both
```
ahojnnes: This logic seems a bit too strict. In my experience, there is still a significant fraction of essential matrices that can be estimated as long as the focal length is not completely off. I'd let the inlier count be the decider on whether to set DEGENERATE or not in this case.

cc: @B1ueber2y

src/colmap/sensor/models.h
ahojnnes: I am not sure, I understand how this is supposed to work. How does the ImgFromCam and CamFromImg get the width and height, if we are not storing it somewhere in the model's parameters?

ahojnnes: @B1ueber2y

B1ueber2y: I was thinking of still storing it as model parameters but assign it to a new group, for example, metadata? and then have num_params = num_focal + num_pp + num_extra + num_metadata, just to avoid the confusion of tagging it as principal point parameters. Wdyt?

ahojnnes: OK, maybe extra_params is semantically more meaningful? It's meant to be anything but focal or principal point. WDYT?

B1ueber2y: extra_params sounds good. Thanks!

ahojnnes (approved):
Thank you for iterating on this with us, LGTM.

src/colmap/scene/camera.cc
```diff
@@ -146,6 +169,9 @@ void Camera::ScaleFocalLengths(double scale_x, double scale_y) {
   } else if (num_focal_params == 2) {
     SetFocalLengthX(scale_x * FocalLengthX());
     SetFocalLengthY(scale_y * FocalLengthY());
+  } else if (num_focal_params == 0) {
```
B1ueber2y: Is this necessary? Should we throw instead? As we should not call SetFocalLengths for spherical models. This similarly applies to SetPrincipalPointX/Y which was already protected with THROW_CHECK_EQ.

ahojnnes: Yeah, good catch. I am actually not too happy with the current rescaling. I think it can be better to implement this at the camera model level with a CameraModelRescale(model_id, params) function. WDYT?

B1ueber2y: Sounds good to me. 

Actually, I think the cleanest way is to make some updates to the macro of CAMERA_MODEL_DEFINITIONS. For example, the following block should only belong to the perspective models:
```
static constexpr size_t num_params =                                        \                                                                                                                                                                   
  113       (num_focal_params_val) + (num_pp_params_val) + (num_extra_params_val);  \                                                                                                                                                                   
  114   static constexpr size_t num_focal_params = num_focal_params_val;            \                                                                                                                                                                   
  115   static constexpr size_t num_pp_params = num_pp_params_val;                  \                                                                                                                                                                   
  116   static constexpr size_t num_extra_params = num_extra_params_val;            \
```

We can ideally make some shared definitions and some perspective/spherical specific definitions, and define by composition of macros, as how we did for fisheye models.

B1ueber2y: And then we can make new group abstraction like image_size_param and extra_param for spherical models and handle Rescale function cleanly with CameraModelRescale. This also depends on whether we plan to extend to more spherical models in the future, otherwise I agree that it probably does not worth it : )

ahojnnes: Went with the model-level approach: added `CameraModelRescale(model_id, scale_x, scale_y, params)` that dispatches to a per-model `Rescale`. `BasePerspectiveCameraModel::Rescale` scales the focal length(s) and principal point; `BaseSphericalCameraModel::Rescale` scales the `(w, h)` image-size params. `Camera::Rescale` now just computes the per-axis scale factors, updates `width`/`height`, and delegates — so the `IsSpherical()` special-casing and the `Camera::ScaleFocalLengths` helper are gone, and we no longer call `SetFocalLength`/`SetPrincipalPoint` on spherical models at all.

On the heavier macro refactor (pulling `num_params`/group counts out of the shared `CAMERA_MODEL_DEFINITIONS` and adding a dedicated `image_size_param` group, composed like the fisheye split): I left it out for now. With a single spherical model it felt like more machinery than it buys us, since the per-model `Rescale` already removes the focal/principal-point special-casing cleanly. Happy to do the fuller composition if/when we add more spherical models — agree it's the right shape then. WDYT?

B1ueber2y: Sounds good to me. Thank you!

zalo:
Awesome!  I’ll look at why the tests aren’t passing…

ahojnnes (approved):
Thank you very much for your contribution!

B1ueber2y (approved):

PR #4445, opened 2026-06-08
Add GlobalMapper.keep_max_num_tracks option to bound established tracks

### LINKED ISSUE #4438: COLMAP 4 global_mapper: replacement for GLOMAP TrackEstablishment.max_num_tracks / OOM protection?
## Question

I am migrating a workflow from standalone GLOMAP to COLMAP 4 `global_mapper`.

In older standalone GLOMAP, we used the following option to avoid excessive memory usage / OOM issues on large image sets:

```txt
--TrackEstablishment.max_num_tracks 1500000
```

This was important for our workflow because the old default could allow up to roughly 10M tracks. On larger datasets, this could lead to very large track sets and eventually OOM during mapping or bundle adjustment.

In COLMAP 4.0.4, GLOMAP is now integrated as:

```bash
colmap global_mapper
```

However, I cannot find an equivalent option for the old standalone GLOMAP flag:

```txt
TrackEstablishment.max_num_tracks
```

The closest COLMAP 4 options I found are:

```txt
--GlobalMapper.track_required_tracks_per_view arg (=2147483647)
--GlobalMapper.track_min_num_views_per_track arg (=3)
```

From reading the source, it looks like COLMAP 4 no longer uses a global maximum track count. Instead, it appears to sort tracks by length and keep adding tracks until each image / view has enough tracks according to `track_required_tracks_per_view`.

Relevant option definitions:

```cpp
int track_required_tracks_per_view = std::numeric_limits<int>::max();
int track_min_num_views_per_track = 3;
```

Relevant track selection logic:

```cpp
const bool should_add = std::any_of(
    point3D.track.Elements().begin(),
    point3D.track.Elements().end(),
    [&](const auto& obs) {
      return tracks_per_image[obs.image_id] <=
             static_cast<size_t>(options.track_required_tracks_per_view);
    });
```

With the default value:

```txt
--GlobalMapper.track_required_tracks_per_view 2147483647
```

this seems to effectively mean that there is no early stopping based on per-image track coverage.

## Environment

```txt
COLMAP version: 4.0.4
COLMAP commit: 9c23f6942fe69962e06030905e77067c8673382f
CUDA enabled: yes

Binary:
/home/ubu/dev/install/colmap_4_0_4_cuda_cudss/bin/colmap

OS: Ubuntu Linux
GPU: NVIDIA GeForce RTX 3090
```

## Questions

What is the recommended COLMAP 4 replacement for the old standalone GLOMAP option?

```txt
--TrackEstablishment.max_num_tracks
```

Specifically:

1. Is `--GlobalMapper.track_required_tracks_per_view` intended to replace the old global track cap?
2. If yes, how should users choose a safe value for large datasets?
3. Is there still a way to enforce a global maximum number of established tracks?
4. If not, would it make sense to expose such an option again for OOM protection?
5. Are there other recommended settings to avoid OOM in COLMAP 4 `global_mapper` when using GPU / CUDA / cuDSS bundle adjustment?

## Why this matters

For large datasets, the old standalone GLOMAP default track cap could still lead to very large track sets and OOM. In our pipeline, we reduced it to around:

```txt
--TrackEstablishment.max_num_tracks 1500000
```

This made mapping much more stable.

With COLMAP 4 `global_mapper`, the old flag is no longer available. The current default:

```txt
--GlobalMapper.track_required_tracks_per_view 2147483647
```

appears to effectively mean “no early stopping” unless the user manually lowers it.

We would like to migrate safely to COLMAP 4, but need to understand the intended OOM-protection strategy for large datasets.

ahojnnes:
Fixes https://github.com/colmap/colmap/issues/4438

## Problem

When migrating from standalone GLOMAP to COLMAP 4's `global_mapper`, there was no equivalent of the old `--TrackEstablishment.max_num_tracks` flag, which capped the total number of established tracks to bound memory usage and avoid OOM on large datasets. The existing `GlobalMapper.track_required_tracks_per_view` option only controls per-view early stopping and defaults to no capping.

## Change

Adds a new `GlobalMapper.keep_max_num_tracks` option (default: unlimited). During track establishment, tracks are already sorted by decreasing length; the selection loop now stops once this budget is reached, keeping the longest (most valuable) tracks. The option is exposed via the C++ `GlobalMapperOptions` struct, the CLI option manager, and the pycolmap bindings.

### DIFF
--- a/src/colmap/controllers/option_manager.cc
+++ b/src/colmap/controllers/option_manager.cc
@@ -760,6 +760,8 @@ void OptionManager::AddGlobalMapperOptions() {
                    &global_mapper->mapper.track_required_tracks_per_view);
   AddDefaultOption("GlobalMapper.track_min_num_views_per_track",
                    &global_mapper->mapper.track_min_num_views_per_track);
+  AddDefaultOption("GlobalMapper.keep_max_num_tracks",
+                   &global_mapper->mapper.keep_max_num_tracks);
 
   // Global positioning options.
   AddDefaultOption("GlobalMapper.gp_use_gpu",
--- a/src/colmap/sfm/global_mapper.cc
+++ b/src/colmap/sfm/global_mapper.cc
@@ -235,7 +235,13 @@ void GlobalMapper::EstablishTracks(const GlobalMapperOptions& options) {
 
   std::unordered_map<image_t, size_t> tracks_per_image;
   size_t images_left = image_id_to_keypoints.size();
+  const size_t max_num_tracks =
+      static_cast<size_t>(options.keep_max_num_tracks);
   for (const auto& [track_length, point3D_id] : track_lengths) {
+    // Stop once the global track budget is exhausted. As tracks are sorted by
+    // decreasing length, this keeps the longest tracks and bounds memory usage.
+    if (reconstruction_->NumPoints3D() >= max_num_tracks) break;
+
     auto& point3D = candidate_points3D.at(point3D_id);
 
     // Check if any image in this track still needs more observations.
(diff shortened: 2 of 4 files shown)


B1ueber2y (approved):
Thanks!

PR #4457, opened 2026-06-17
Replace libqt6svg6-dev with qt6-svg-dev

laszlovandenhoek:
When following the installation instructions for Ubuntu 24.04, in the step `cmake .. -GNinja -DBLA_VENDOR=Intel10_64lp`, I get this error:

```
CMake Error at cmake/FindDependencies.cmake:424 (find_package):
  Found package configuration file:

    /usr/lib/x86_64-linux-gnu/cmake/Qt6/Qt6Config.cmake

  but it set Qt6_FOUND to FALSE so package "Qt6" is considered to be NOT
  FOUND.  Reason given by package:

  Failed to find required Qt component "Svg".

  Expected Config file at
  "/usr/lib/x86_64-linux-gnu/cmake/Qt6Svg/Qt6SvgConfig.cmake" does NOT exist
```

Installing `qt6-svg-dev` in addition fixes that problem.

### DIFF
--- a/.github/workflows/build-ubuntu.yml
+++ b/.github/workflows/build-ubuntu.yml
@@ -149,7 +149,7 @@ jobs:
           if [ "${{ matrix.config.qtVersion }}" == "5" ]; then
             qt_packages="qtbase5-dev libqt5opengl5-dev libqt5svg5-dev libcgal-qt5-dev"
           elif [ "${{ matrix.config.qtVersion }}" == "6" ]; then
-            qt_packages="qt6-base-dev libqt6opengl6-dev libqt6openglwidgets6 libqt6svg6-dev"
+            qt_packages="qt6-base-dev libqt6opengl6-dev libqt6openglwidgets6 qt6-svg-dev"
           fi
 
           sudo apt-get update && sudo apt-get install -y \
--- a/.github/workflows/build-ubuntu.yml
+++ b/.github/workflows/build-ubuntu.yml
@@ -149,7 +149,14 @@ jobs:
           if [ "${{ matrix.config.qtVersion }}" == "5" ]; then
             qt_packages="qtbase5-dev libqt5opengl5-dev libqt5svg5-dev libcgal-qt5-dev"
           elif [ "${{ matrix.config.qtVersion }}" == "6" ]; then
-            qt_packages="qt6-base-dev libqt6opengl6-dev libqt6openglwidgets6 qt6-svg-dev"
+            qt_packages="qt6-base-dev libqt6opengl6-dev libqt6openglwidgets6"
+            # The Svg dev package is named qt6-svg-dev on Ubuntu 24.04+ but
+            # libqt6svg6-dev on Ubuntu 22.04.
+            if [ "${{ matrix.config.os }}" == "ubuntu-22.04" ]; then
+              qt_packages="$qt_packages libqt6svg6-dev"
+            else
+              qt_packages="$qt_packages qt6-svg-dev"
+            fi
           fi
 
           sudo apt-get update && sudo apt-get install -y \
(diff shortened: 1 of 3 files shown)


ahojnnes:
Thanks! Is this a fresh install of Ubuntu 24.04? It's a bit surprising that the CI build doesn't fail with the same issues?

laszlovandenhoek:
The issue is that the CI build adds flags that are not in the installation instructions. The offending one is here:

https://github.com/colmap/colmap/blob/main/.github/workflows/build-ubuntu.yml#L229

`-DGUI_ENABLED=${{ matrix.config.guiEnabled }}`

...where `guiEnabled` is `false` in all matrix entries. (_**EDIT: this is not true, but that doesn't matter so much - see my next comment.**_)

You can verify with:

```
docker run --rm -it --platform linux/amd64 ubuntu:24.04 bash
qt_packages="qt6-base-dev libqt6opengl6-dev libqt6openglwidgets6 libqt6svg6-dev"
apt update
apt-get install -y \
            build-essential \
            cmake \
            ninja-build \
            libboost-program-options-dev \
            libboost-graph-dev \
            libboost-system-dev \
            libeigen3-dev \
            libceres-dev \
            libopenimageio-dev \
            openimageio-tools \
            libsuitesparse-dev \
            libmetis-dev \
            libgoogle-glog-dev \
            libgtest-dev \
            libgmock-dev \
            libsqlite3-dev \
            libglew-dev \
            $qt_packages \
            libcgal-dev \
            libgl1-mesa-dri \
            libunwind-dev \
            libcurl4-openssl-dev \
            libmkl-full-dev \
            xvfb
mkdir -p /usr/include/opencv4
apt-get install -y \
    nvidia-cuda-toolkit \
    nvidia-cuda-toolkit-gcc
git clone https://github.com/colmap/colmap.git
cd colmap
mkdir build
cd build

# this fails (as in the documentation):
cmake .. -GNinja -DBLA_VENDOR=Intel10_64lp

# this succeeds (as in the build file):
cmake .. -GNinja -DBLA_VENDOR=Intel10_64lp -DGUI_ENABLED=false
```

The `--platform linux/amd64` is important on Apple silicon because of availability of the MKL library. Probably not relevant for this issue, but just ensuring an apples-to-apples (🤓) comparison.

ahojnnes:
Thanks for your detailed response. I probably misunderstand but the very first entry has guiEnabled=true and the fourth entry has guiEnabled=false. I guess there are some other side-effects that prevent the issue?

laszlovandenhoek:
You're right, I missed that one. I thought they were all `false`. Nevertheless, the reproducer stands, and I would argue it's representative for the issue I ran into, i.e. trying to build colmap on Ubuntu 24.04 following the manual.

In any case, I've done some further digging, and it seems the `libqt6svg6-dev` Debian package was renamed to `qt6-svg-dev` with the 6.4.1 release in 2022. From the [release notes](https://debian.pkgs.org/12/debian-main-amd64/qt6-svg-dev_6.4.2-2_amd64.deb.html):

> Rename libqt6svg6-dev to qt6-svg-dev.

The QT documentation also refers to `qt6-*dev*` as the correct package name format:

https://doc.qt.io/qt-6/linux-alternative-installers.html#installing-qt-on-ubuntu-and-debian-based-systems

As for the question why the CI build has no issues: the _GitHub_ Ubuntu 24.04 image has quite a different lineage from the _vanilla_ Ubuntu 24.04 image. They start with [canonical:ubuntu-24_04-lts:server](https://marketplace.microsoft.com/en-us/product/canonical.ubuntu-24_04-lts) from the Azure Marketplace, and then add a slew of packages:

https://github.com/actions/runner-images/blob/main/images/ubuntu/Ubuntu2404-Readme.md

Considering the big difference I think it's a tall order to find out what, exactly, is causing the build to _succeed_ despite not installing `qt6-svg-dev`. In any case, I think it would be cleanest to entirely replace `libqt6svg6-dev` with `qt6-svg-dev` in both the documentation and the build script. I'll amend the PR.

ahojnnes (approved):
Thanks!

PR #4459, opened 2026-06-18
Add global mapper to panorama_sfm

sarlinpe:
Also create an enum for the matcher type.


ichsan2895:
Just an idea, since caspar is supported in rig mode, maybe you can add the new args for selecting ceres or caspar (global & local bundle adjustment)

ahojnnes (approved):
Thanks!

PR #4462, opened 2026-06-22
Fix SiftGPU set device to include device 0

nharbiso:
SiftGPU sets the GPU device to use based on that passed to `ProgramCU::CheckCudaDevice`. However, this function only calls `cudaSetDevice` for devices with an ID greater than 0. Fixes to allow setting to device 0.

### DIFF
--- a/src/thirdparty/SiftGPU/ProgramCU.cu
+++ b/src/thirdparty/SiftGPU/ProgramCU.cu
@@ -1387,7 +1387,7 @@ int ProgramCU::CheckCudaDevice(int device)
 
 		}
     }
-    if(device >0 && device < count)
+    if(device >= 0 && device < count)
     {
         cudaSetDevice(device);
         CheckErrorCUDA("cudaSetDevice\n");


ahojnnes (approved):
Makes sense — correct, minimal bug fix. Device `0` is a valid CUDA index, so `device > 0` silently skipped `cudaSetDevice` for it, falling back to CUDA's implicit per-thread default. That only *happens* to be 0 on a fresh thread; if the thread's current device was previously set elsewhere, extraction would run on the wrong GPU (and trip the `device != device_used` warning just below).

The value comes from `-cuda <index>` (`sift.cc:585-588`), defaulting to `0` (`sift.cc:581-583`), so device 0 is the common case. Note the **matching** path already sidesteps this via `SetBestCudaDevice(gpu_index)` (`feature_matching_utils.cc:79`), but **extraction** relies solely on `CheckCudaDevice` — so this fix brings extraction in line.

Risk is negligible: in the single-GPU / already-on-0 case, `cudaSetDevice(0)` is a harmless one-time no-op, and the `device < 0` sentinel and `device < count` bound are preserved.

PR #4463, opened 2026-06-22
Add EQUIRECTANGULAR support to the Caspar GPU bundle adjuster

zalo:
This is a follow-up to #4441 to add Caspar solver support, which is ~2x faster than the global mapper due to the way it fuses the kernels.

The contributed algorithm is relatively small, but SymForce autogenerates hundreds of files and nearly 20,000 lines, bloating the apparent size of this PR.   Wondering if there's a cleaner way to do Caspar PRs in the future... but for now, I've added a .gitattributes file that marks the symforce generated files, so they should be collapsed in future Github diff previews 😅 

More explorations on the speed of EQUIRECTANGULAR Caspar here: https://github.com/colmap/colmap/pull/4441#issuecomment-4666195356

## Summary

Caspar's GPU bundle adjuster only shipped Symforce kernels for `PINHOLE`
and `SIMPLE_RADIAL`, and silently skipped every other model. As a result an
all-`EQUIRECTANGULAR` reconstruction optimized nothing (`No residuals to
optimize`) under `--BundleAdjustment.backend CASPAR`. This PR adds the
EQUIRECTANGULAR residual kernels and adapter so 360° / panoramic
reconstructions can be bundle-adjusted on the GPU.

`EQUIRECTANGULAR` is the simplest model to add: its `(w, h)` parameters are
the image dimensions and are held **constant** in BA (the model has no focal
length, principal point, or distortion). So it needs none of the
tunable-calibration machinery — only the camera pose and the 3D point are
optimized, with `(w, h)` carried along as a per-factor constant node.

## What's included

- **`caspar_generate.py`** — `spherical_core`, the equirectangular
  reprojection residual mirroring `EquirectangularCameraModel::ImgFromCam`:
  `theta = atan2(X, Z)`, `phi = atan2(-Y, sqrt(X² + Z²))`,
  `pixel = ((theta/2π + ½)·w, (½ − phi/π)·h)`. Registered with
  `FIXABLE = {pose, point}` → 3 reachable variants (pose+point, point-only,
  pose-only). A numeric `pi` is used because Caspar's expression lowering has
  no entry for the symbolic `Pi` constant.
- **`caspar_model_adapter.h`** — `SphericalAdapter`: calib / focal /
  principal-point methods are no-ops; `(w, h)` is appended as the constant
  `Wh` node; the factory maps `CameraModelId::kEquirectangular`. The model is
  the `EQUIRECTANGULAR` member of COLMAP's spherical-camera family
  (`BaseSphericalCameraModel`), so the internal kernels keep the `spherical`
  name.
- **`bundle_adjustment_caspar.{h,cc}`** — `VariantData::const_camera_params`
  holds `(w, h)`; the focal/principal-point refine toggles are gated on
  `FocalAndExtraSize() / PrincipalPointSize() > 0` so a model with no such
  parameters is never marked variable by the `refine_*` defaults;
  `BuildSizing` handles `kEquirectangular`.
- **Generated `f32` / `f64` kernels** — clang-formatted with the repo style;
  existing kernels untouched.
- **`bundle_adjustment_caspar_test.cc`** — `EquirectangularMatchesCeres`:
  bundle-adjusts a synthetic EQUIRECTANGULAR dataset with both the Ceres and
  Caspar backends and asserts the recovered poses/points agree, and that the
  `(w, h)` intrinsics are left untouched.

## Validation

- **Unit tests:** all 28 Caspar tests pass, including the new
  `EquirectangularMatchesCeres`.
- **Real data:** `bundle_adjuster --BundleAdjustment.backend CASPAR` on a
  66-image EQUIRECTANGULAR reconstruction yields **0.515386 px** mean
  reprojection error, matching the **CERES** backend (**0.515455 px**) to
  < 1e-4 px.

## Notes

- Regenerating kernels requires Symforce; the generated kernels are committed,
  so building only needs CUDA (`-DCASPAR_ENABLED=ON`).
- No changes to any non-Caspar code path.


ahojnnes:
Thanks for the contribution. I am not opposed to eventually merging this, but I'd like to first see this feature being used more broadly by the community before introducing specialized CUDA kernels, because it significantly increase compile-time and binary size.

zalo:
By this feature, do you mean the Equirect support or Caspar?

COLMAP Equirect support was just added to Lichtfeld as well, so I suspect we’re going to start seeing the basic trickle of Gaussian Splats come out of that…. Already some people in their Discord server are reporting better trajectories from the equirectangular solver.

IMG_0531
IMG_0530

Though, this also does call out the fundamental weakness with “ERPs” (Equirectangular Panoramas) glossing over the sub-camera offset issue and the optical flow seam issue… I’m not sure I have a good solution for this one other than maybe the generative priors that have been coming out lately 😅

sarlinpe:
> better trajectories from the equirectangular solver.

I highly doubt so and would like to see evidence of it. This smells like misconfiguration of the rig constraints, which are properly handled in [panorama_sfm.py](https://github.com/colmap/colmap/blob/main/python/examples/panorama_sfm.py). This script also properly propagates EXIF GPS data, unlike reported by the user.

ahojnnes:
> By this feature, do you mean the Equirect support or Caspar?

Both, as they are just being introduced as new features in the upcoming release.

ichsan2895:
Some users amazed with direct alignment from equirectangular images
image

Also, he noted it seems faster than metashape
image

zalo:
@sarlinpe Just had Claude run a 66-image smoke test with both the original panorama_sfm.py and the new direct equirectangular.  The panorama_sfm.py path does indeed appear to score very slightly worse in angular reprojection accuracy (possibly within the noise floor, but it does seem consistent)…

My primary two guesses:
- It’s the extra resampling to cube maps.  The footage was already resampled once to equirectangular, so this is just layering more resampling on top.
- It could also be that panorama_sfm.py detects slightly more points, which are maybe of worse quality and/or conflicting and pull the reprojection accuracy down?   

IMG_0548

I also ran the CASPAR path here relative to the direct equirect incremental mapper, and it does seem to be slightly worse, but still better than panorama_sfm.py:

IMG_0549

250s panorama_sfm.py -> 50s CASPAR wall-clock time might be _way_ faster than metashape, which seems like a big win to me 😄

I’ll run it on more datasets now…

EDIT1: It seems to hold true at 8k resolution (represented as “D2” vs Dataset 1 here), except you need to account for the fact that the cubemap gets a feature budget per sub-image:

IMG_0550
IMG_0551

Trying on one more dataset now…

EDIT2: Still holds on my third dataset; the global mapper is generally higher quality for both.

IMG_0552

ichsan2895:
**FYI,** Some people (especially who have interest with 3D gaussian splats, including me) likes this feature (Direct equirectangular SfM). He also creates [a tutorial about creating 3DGS from equirectangular dataset](https://lilea-net.translate.goog/lab/colmap-equirectangular-to-lichtfeld-studio-3dgs/?_x_tr_sl=ko&_x_tr_tl=id&_x_tr_hl=id&_x_tr_pto=wapp)
image

ahojnnes:
It will be great to understand the source of groundtruth for the above experiments or repeat them on a standard 360 benchmark dataset with high-quality GT poses to make a more informed decision about next steps.

sarlinpe:
Thank you for sharing the results! IIUC these datasets don't have any ground truth poses, hence why reprojection errors are reported. IMO they don't tell the full pictures, it would be interesting to report the average track length as well (in combination with the number of 3D points). In parallel, I'll try to get a 360 dataset added to the our benchmark.

ichsan2895:
More and more 3DGS software that supports direct training from equirectangular without splitting it. This one 3DGS software created by Mykhailo Moroz.

image
image

Another one is Lichtfeld Studio that supports that.

ahojnnes:
@ichsan2895 Thanks for sharing. Again: Running SFM and training GS are two independent steps. The former can use virtual perspective rigs and the former can still train on equirectangular images. In fact, the latest panorama_sfm.py supports this scenario easily by producing a sparse reconstruction for the equirectangular images even if performing SFM on the perspective rig. We will reconsider this PR once we have conclusive benchmarking results.

Otherwise, you can check out the latest commits in the master branch, where we just landed analytical Jacobians for the equirectangular camera model, which should provide a incremental speed improvement for Ceres.

zalo:
(Specifically, would you like me to make a synthetic equirect dataset with known ground truth to accuracy benchmark the solvers on?)

ichsan2895:
@ahojnnes Ah thats a good things. I already use the new panorama_sfm.py that also creates sparse_equirectangular.

PR #4465, opened 2026-06-23
Estimate separate fx/fy in p4pf for two-focal camera models

B1ueber2y:
EstimateAbsolutePose previously broadcast a single P4PF focal length to all focal indices, forcing fx == fy as the seed even for models with independent focal lengths (e.g., PINHOLE, OPENCV, OPENCV_FISHEYE). This used PoseLib's averaging p4pf overload, whose aspect-ratio filter also rejected exactly the non-square-pixel solutions wanted for such models.

Make P4PFEstimator model-aware: keep the shared-focal overload for single-focal models (unchanged behavior) and use PoseLib's separate-fx/fy overload for two-focal models, assigning each focal length to its corresponding parameter index. The estimator now carries a share_focal_length flag and stores fx/fy as a vector; residuals scale per axis accordingly.

cc: @vlarsson

### DIFF
--- a/src/colmap/estimators/pose_test.cc
+++ b/src/colmap/estimators/pose_test.cc
@@ -47,18 +47,22 @@ struct AbsolutePoseProblem {
   std::vector<Eigen::Vector3d> points3D;
 };
 
-AbsolutePoseProblem CreateAbsolutePoseTestData() {
+AbsolutePoseProblem CreateAbsolutePoseTestData(
+    CameraModelId camera_model_id = CameraModelId::kSimpleRadial,
+    const std::vector<double>& camera_params = {1280, 512, 384, 0.05}) {
   AbsolutePoseProblem problem;
   SyntheticDatasetOptions synthetic_dataset_options;
   synthetic_dataset_options.num_rigs = 1;
   synthetic_dataset_options.num_cameras_per_rig = 1;
   synthetic_dataset_options.num_frames_per_rig = 2;
   synthetic_dataset_options.num_points3D = 100;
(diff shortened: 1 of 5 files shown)


src/colmap/estimators/pose_test.cc
```diff
@@ -119,6 +123,33 @@ TEST(EstimateAbsolutePose, EstimateFocalLength) {
   EXPECT_THAT(inlier_mask, testing::Each(testing::Eq(true)));
 }
 
+TEST(EstimateAbsolutePose, EstimateSeparateFocalLengths) {
+  // PINHOLE camera with distinct fx and fy (fx, fy, cx, cy).
+  const AbsolutePoseProblem problem = CreateAbsolutePoseTestData(
```
ahojnnes, line 128: Would it make sense to generate a synthetic camera with separate focal lengths?

ahojnnes (approved):
Looks like the theoretically principled fix. I am wondering about practical relevance? Do we expect the focal length to differ sufficiently for the pose estimator to reliably recover it through RANSAC? Curious if you have seen this being an issue in practice?

B1ueber2y:
No experience. I also agree that the practical impact is likely limited. Just came across this when developing something else and thought this was a strictly better fix, as the initialization for the pose refinement should be strictly better for two focal cameras.

One potential benefit is that this would fix cases when images are rescaled without ratio kept. Without this PR the focal length may be rejected via the ratio threshold from PoseLib, or wrongly initialized as the mean.

ahojnnes (approved):

PR #4467, opened 2026-06-24
Fix Windows MSVC test issues

bidwej:
## Summary

Two small Windows MSVC test fixes extracted from a downstream migration branch.

## Changes

- `src/colmap/feature/sift_test.cc`: include `colmap/math/math.h` so `M_PI` is available under MSVC without relying on `_USE_MATH_DEFINES`.
- `src/colmap/util/file_test.cc`: skip the soft-link portion of `FileCopy.Nominal` when the host Windows environment lacks symlink privileges, mirroring the existing skip in `GetNormalizedRelativePath.PreservesSymlinkStructure`.

## Validation

Built and tested on Windows with vcpkg + MSVC 14.44 + Ninja. All affected tests pass.

### DIFF
--- a/src/colmap/feature/sift_test.cc
+++ b/src/colmap/feature/sift_test.cc
@@ -38,6 +38,7 @@
 #include "colmap/feature/sift.h"
 #include "colmap/feature/utils.h"
 #include "colmap/geometry/essential_matrix.h"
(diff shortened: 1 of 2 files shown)


bidwej:
Closing in favor of keeping these changes in the downstream migration only.

PR #4468, opened 2026-06-24
Convert panorama_sfm perspective cameras back to equirectangular

ahojnnes:
## Description

For the `perspective_*` modes in `python/examples/panorama_sfm.py`, this adds a `--convert_to_equirectangular` option that maps the reconstructed rig of perspective virtual cameras back onto the original equirectangular input images, using the native `EQUIRECTANGULAR` camera model.

After reconstructing with the rig of perspective virtual cameras, the script now optionally writes a second reconstruction to `sparse_equirectangular/<idx>` in which:

- Each input panorama is a single `EQUIRECTANGULAR` image referencing the original input filename.
- Frame poses are derived from the rig poses (`pano_from_world = pano_from_cam0 * rig_from_world`, the rig reference sensor being virtual camera 0).
- 3D points and **all** keypoints (including those without a 3D observation) are carried over by re-projecting the perspective keypoints onto the panorama through the same spherical mapping used for rendering.

The result is a valid, bundle-adjustable reconstruction.

## Notes

- The flag only applies to the `perspective_*` modes; combining it with `--pano_render_type spherical` errors out.
- The equirectangular projection convention matches COLMAP's `EquirectangularCameraModel`, so the back-conversion is exact (verified to numerical precision via a standalone round-trip check).

### DIFF
--- a/python/examples/panorama_sfm.py
+++ b/python/examples/panorama_sfm.py
@@ -7,10 +7,13 @@ Two modes are available via --pano_render_type:
     and reconstruct from those.
   * "spherical": reconstruct directly on the panoramas using the native
     equirectangular camera model, without rendering perspective images.
-The "perspective_*" modes are generally more accurate but slower.
+The "perspective_*" modes are generally more accurate but slower. For these
+modes, the script additionally writes a reconstruction that maps the perspective
+virtual cameras back onto the original equirectangular input images.
 """
 
 import argparse
+import collections
 import os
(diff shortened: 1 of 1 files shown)


B1ueber2y (approved):
Thanks! LGTM. The PR description was a bit confusing as there was no newly added argument for this feature. 

Would be nice to have some tests for this or even port it in C++ side in the future.

PR #4471, opened 2026-06-24
Add constraint reweighting for rotation averaging

lpanaf:
## Summary

Adds an optional `RotationEstimatorOptions::weight_by_num_matches` flag that weights each relative-rotation constraint by the number of two-view matches (`PoseGraph::Edge::num_matches`) of the corresponding pose-graph edge.

Weights are applied as a diagonal scaling of the linear system in both the L1 regression and IRLS solvers, so a noise-free (consistent) system yields an identical solution. The weight vector is only allocated when the feature is enabled (`std::nullopt` otherwise, no overhead).

The flag is exposed via the `GlobalMapper.ra_weight_by_num_matches` CLI option and the pycolmap `RotationEstimatorOptions` binding.

## Tests

- **Noise-free invariance**: with non-trivial per-edge match counts (with and without gravity priors), the weighted solver recovers the ground truth and produces a solution identical to the unweighted one.
- **Noisy down-weighting**: injecting rotation noise into relative poses and halving the match count of the noisy edges, the weighted solver yields a lower mean relative-rotation error than the unweighted baseline.


src/colmap/estimators/rotation_averaging_impl.cc
```diff
@@ -445,6 +445,23 @@ void RotationAveragingProblem::BuildConstraintMatrix(
...
   residuals_.resize(curr_row);
+
+  // Optionally build per-row edge weights from the number of two-view matches.
+  // Gauge-fixing rows keep weight 1. Left as nullopt (no allocation) when the
+  // feature is disabled.
+  if (options_.weight_by_num_matches) {
+    Eigen::VectorXd weights = Eigen::VectorXd::Ones(curr_row);
+    for (const auto& [pair_id, constraint] : pair_constraints_) {
+      const double num_matches =
+          static_cast<double>(pose_graph.Edges().at(pair_id).num_matches);
```
ahojnnes: Does it make sense to normalize the weights to [0, 1] for numerical stability?

src/colmap/estimators/rotation_averaging_impl.cc
```diff
@@ -792,6 +828,14 @@ bool RotationAveragingSolver::SolveIRLS(RotationAveragingProblem& problem) {
       return false;
     }
 
+    // For weighted rotation averaging, fold the per-edge weights into the IRLS
+    // weights. They then propagate to both the system matrix at_weight_a and
+    // the right-hand side at_weight * residuals below, scaling both sides of
+    // the linear system by the same weights.
+    if (problem.EdgeWeights().has_value()) {
+      *weights_irls = weights_irls->cwiseProduct(*problem.EdgeWeights());
+    }
```
ahojnnes: Suggest moving this logic into ComputeIRLSWeights().

src/colmap/estimators/rotation_averaging_impl.cc
```diff
@@ -682,8 +712,14 @@ bool RotationAveragingSolver::SolveL1Regression(
...
     step.setZero();
-    if (!l1_solver.Solve(problem.Residuals(), &step) ||
-        step.array().isNaN().any()) {
+    Eigen::VectorXd weighted_residuals;
+    if (weighted) {
+      weighted_residuals =
+          problem.EdgeWeights()->asDiagonal() * problem.Residuals();
+    }
+    const Eigen::VectorXd& b =
+        weighted ? weighted_residuals : problem.Residuals();
```
ahojnnes: Suggest absorbing this in a new problem.WeightedResiduals() method.

src/colmap/estimators/rotation_averaging_impl.cc
```diff
@@ -660,8 +677,21 @@ bool RotationAveragingSolver::SolveL1Regression(
...
+  // constraint matrix. For a noise-free (consistent) system, the minimizer is
+  // unchanged.
+  const bool weighted = problem.EdgeWeights().has_value();
+  Eigen::SparseMatrix<double> weighted_matrix;
+  if (weighted) {
+    weighted_matrix =
+        problem.EdgeWeights()->asDiagonal() * problem.ConstraintMatrix();
+  }
+  const Eigen::SparseMatrix<double>& l1_matrix =
+      weighted ? weighted_matrix : problem.ConstraintMatrix();
```
ahojnnes: Suggest absorbing this in a WeightedConstraintMatrix() method.

src/colmap/estimators/rotation_averaging.h
```diff
@@ -79,6 +79,12 @@ struct RotationEstimatorOptions {
   // When false, treat each non-ref sensor's cam_from_rig rotation as a
   // pre-calibrated constant
   bool refine_sensor_from_rig = true;
+
+  // If true, weight each relative-rotation constraint by the number of
+  // two-view matches (PoseGraph::Edge::num_matches) of the corresponding edge.
+  // Weights are applied as a diagonal scaling of the linear system, so a
+  // noise-free (consistent) system yields an identical solution.
+  bool weight_by_num_matches = false;
```
ahojnnes: Would it make sense to introduce a new enum class using our enum macros as `RotationAveragingWeighting { NONE, NUM_MATCHES }`, so we can more easily add other mechanisms in the future without breaking the contract?

ahojnnes (commented):
Thanks, looks good to me with some suggestions.

src/colmap/estimators/rotation_averaging_impl.cc
B1ueber2y: Agreed. Statistically it would only make sense to put a weight of [0, 1] here, corresponding to the likelihood of this constraint being good.

src/colmap/estimators/rotation_averaging.h
B1ueber2y: Yes. This would be nice. We can also easily add another option to weight by inlier ratio? I am personally not a big fan of match count.

B1ueber2y: Also, maybe we should specify in the variable name and comments that we here refer to number of inlier matches rather than number of matches.

sarlinpe: +1, this paper shows that uncertainties are better weights than the number of matches: https://openaccess.thecvf.com/content/CVPR2023/papers/Zhang_Revisiting_Rotation_Averaging_Uncertainties_and_Robust_Losses_CVPR_2023_paper.pdf

src/colmap/estimators/rotation_averaging_impl.cc
```diff
@@ -445,6 +446,57 @@ void RotationAveragingProblem::BuildConstraintMatrix(
...
+            max_num_matches;
+        if (std::holds_alternative<GravityAligned1DOF>(constraint.constraint)) {
+          reweighting_diagonal[constraint.row_index] = reweighting;
+        } else {
+          reweighting_diagonal.segment<3>(constraint.row_index)
+              .setConstant(reweighting);
+        }
+      }
+    }
+    Eigen::SparseMatrix<double> reweighting(curr_row, curr_row);
```
ahojnnes: Since this is only used as a diagonal rewighting, I suggest we restore the previous implementation, where you defined a dense weights vector with asDiagonal() ?

src/colmap/estimators/rotation_averaging_impl.cc
```diff
@@ -445,6 +446,57 @@ void RotationAveragingProblem::BuildConstraintMatrix(
 
   // Initialize residual vector.
   residuals_.resize(curr_row);
+
+  // Optionally build the residual-space reweighting operator W (see
+  // residual_reweighting_). Weights are normalized to (0, 1] for numerical
+  // stability (a global scale leaves the minimizer unchanged); gauge-fixing
+  // rows keep weight 1.
```
ahojnnes: Where do you handle the Gauge fixing?

src/pycolmap/estimators/motion_averaging.cc
```diff
@@ -121,6 +121,13 @@ void BindRotationEstimator(py::module& m) {
                           .value("HALF_NORM", WeightType::HALF_NORM);
   AddStringToEnumConstructor(PyWeightType);
 
+  auto PyRotationAveragingWeighting =
```
ahojnnes: Let's cover this new enum in the pycolmap tests as well.

src/colmap/estimators/rotation_averaging.h
```diff
@@ -14,6 +15,13 @@
 // to the paper "Gravity Aligned Rotation Averaging"
 namespace colmap {
 
+// Weighting scheme applied to the relative-rotation constraints.
+//   NONE: all constraints are weighted equally.
```
B1ueber2y: nit: how about "UNIFORM"?

src/colmap/estimators/rotation_averaging.h
```diff
@@ -14,6 +15,13 @@
...
+//   INLIER_MATCH_COUNT: weight each constraint by the number of inlier
+//     two-view matches (PoseGraph::Edge::num_matches) of the corresponding
+//     edge, normalized to (0, 1].
+MAKE_ENUM_CLASS(RotationAveragingWeighting, 0, NONE, INLIER_MATCH_COUNT);
```
B1ueber2y: nit: how about reweighting rather than weighting? Similarly applied to the CLI option?

src/colmap/estimators/rotation_averaging_impl.cc
lpanaf: in constructing the problem, the first camera is fixed to have identity as prediction, which fix the gauge freedom. In setting the weight, it is initialized to be 1, and is not touched by the iteration.

src/colmap/estimators/rotation_averaging_impl.h
```diff
@@ -60,6 +60,22 @@ class RotationAveragingProblem {
     return constraint_matrix_;
   }
   const Eigen::VectorXd& Residuals() const { return residuals_; }
+  // Whether a residual-space reweighting operator is configured (i.e. the
+  // reweighting scheme is not UNIFORM).
+  bool HasResidualReweighting() const {
```
ahojnnes: nit: Unused. Let's remove it?

ahojnnes (approved):

PR #4472, opened 2026-06-25
Add documentation and references for Caspar bundle adjustment backend

ahojnnes:


doc/bibliography.rst
```diff
@@ -8,6 +8,10 @@ Bibliography
     "Accurate, dense, and robust multiview stereopsis."
     Transactions on Pattern Analysis and Machine Intelligence, 2010.
 
+.. [caspar] Martens, Emil and Miller, Aaron and Varnum, Matias and Stahl,
```
B1ueber2y, line 11: This was accepted and presented at ICRA 2026.

sarlinpe (approved):

B1ueber2y (approved):

doc/bibliography.rst
ahojnnes: Thanks, updated the citation to reference ICRA 2026 instead of the arXiv preprint in 6819bc0c8.

doc/faq.rst
```diff
@@ -562,6 +567,34 @@ The following describes practical ways to reduce bundle adjustment runtime.
...
   - Shared camera intrinsics.
 
+- **Use the Caspar GPU bundle adjustment backend**
+
+  COLMAP includes Caspar [caspar]_, an experimental GPU-accelerated bundle
+  adjustment backend that can be one to two orders of magnitude faster than the
+  Ceres CUDA solver for medium- to large-scale problems, leading to drastic
+  speedups especially for the incremental mapper. Caspar requires CUDA and is
+  disabled by default; it must be enabled at build time by configuring COLMAP
+  with ``-DCASPAR_ENABLED=ON``.
```
lpanaf, line 577: According to my understanding, Caspar is an inexact solver. Would it make sense to comment on this? From my experience, this generally leads to less accurate reconstruction. Maybe users should be aware of this potential problem as well before switching.

lpanaf (approved):

doc/faq.rst
ahojnnes: Not sure if it's worth point it out, because Ceres also uses an inexact solver for larger problems? I think we showed that the solution quality is roughly on par with the benchmarking results?

B1ueber2y: Yes. In COLMAP we use inexact solver as well for number of images larger than 1000. I think the last benchmark we had was https://github.com/colmap/colmap/pull/4018#issuecomment-4319617528. Not sure if there are any more recent ones.

PR #4473, opened 2026-06-25
Modify the FAQ Reconstruction with pose priors (GPS) section

ferreram:
Current behavior of pose_prior_mapper requires to set the ``--overwrite_priors_covariance`` option to enable using the ``--prior_position_std_XXX`` args value for updating the current priors covariance.  FAQ was inaccurate here.  This PR fixes this.


ahojnnes (approved):
Thanks!

PR #4474, opened 2026-06-25
Set a default identity covariance to pose priors extracted from exif metadata.

ferreram:
Current behavior does not set any covariance on the priors extracted from exif metadata.  This leads to undefined covariance matrix for the created priors.  It feels cleaner to define a default identity covariance here.  Note that there is no direct impact with this modification as the PosePriorBundleAdjuster currently falls back to a ``prior_position_fallback_stddev`` value (set to 1.0) when no valid covariance is defined for a prior.

### DIFF
--- a/src/colmap/controllers/image_reader.cc
+++ b/src/colmap/controllers/image_reader.cc
@@ -304,6 +304,7 @@ ImageReader::Status ImageReader::Next(Rig* rig,
     const std::optional<double> altitude = bitmap->ExifAltitude();
     if (latitude.has_value() && longitude.has_value() && altitude.has_value()) {
       pose_prior->position = Eigen::Vector3d(*latitude, *longitude, *altitude);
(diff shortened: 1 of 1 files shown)


ahojnnes:
Hm, I think the design is intentional, because we don't have an estimate of the covariance, which is why we defined it as an optional field in the PosePrior struct. @B1ueber2y @sarlinpe WDYT?

B1ueber2y:
> Hm, I think the design is intentional, because we don't have an estimate of the covariance, which is why we defined it as an optional field in the PosePrior struct. @B1ueber2y @sarlinpe WDYT?

Agreed that the design is intentional. I would vote for not setting any covariance unless we have a real one.

ahojnnes:
Thanks for the proposal, however, closing the PR as discussed above.

ferreram:
Fair enough 👍

PR #4478, opened 2026-06-27
Tint UI icons to palette for dark theme legibility

### LINKED ISSUE #4477: New Icons Are Hard to Read in Ubuntu's Dark Theme
Image

ahojnnes:
## Problem

Fixes #4477. The Material Symbols SVG icons under `src/colmap/ui/media/` (introduced in #4343) are flat silhouettes whose `<path>` elements have no `fill` attribute. Per the SVG spec, fill then defaults to **solid black**, independent of the Qt palette. On dark themes (e.g. Ubuntu's Yaru-dark) the toolbar/menu icons render black-on-dark and are nearly unreadable.

## Fix

Add a `ThemedIcon()` helper in `qt_utils` that loads the SVG with `QSvgRenderer` and recolors the silhouette to the palette's `WindowText` color via a `CompositionMode_SourceIn` fill, then use it at all icon load sites in `main_window.cc` (replacing `QIcon(":/media/*.svg")`).

This auto-adapts to any theme (light, dark, high-contrast, macOS/Windows dark mode) from a single helper, with no duplicate assets and no new dependencies (QtSvg was already linked). Icons are rendered at 64×64 so `QIcon` stays crisp at toolbar/menu sizes including on high-DPI displays.

## Notes

- The tint is baked in at icon-creation time; a runtime theme switch is only reflected once icons are recreated (documented in the header). This covers the reported scenario, where the theme is set before launch.

### DIFF
--- a/src/colmap/ui/main_window.cc
+++ b/src/colmap/ui/main_window.cc
@@ -374,37 +374,39 @@ void MainWindow::CreateActions() {
   // File actions
   //////////////////////////////////////////////////////////////////////////////
 
-  action_project_new_ =
-      new QAction(ThemedIcon(":/media/project-new.svg"), tr("New project"), this);
+  action_project_new_ = new QAction(
+      ThemedIcon(":/media/project-new.svg"), tr("New project"), this);
   action_project_new_->setShortcuts(QKeySequence::New);
   connect(
       action_project_new_, &QAction::triggered, this, &MainWindow::ProjectNew);
 
-  action_project_open_ =
-      new QAction(ThemedIcon(":/media/project-open.svg"), tr("Open project"), this);
+  action_project_open_ = new QAction(
+      ThemedIcon(":/media/project-open.svg"), tr("Open project"), this);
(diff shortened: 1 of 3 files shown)


kevintsq:
Thanks for your quick response. It seems the issue is still unresolved; the icons remain black even though the theme is dark on Ubuntu.

ahojnnes:
Thanks, can you confirm which Qt and Ubuntu version you are on?

The fix in this PR works on Mac with Qt 6:

image
image

ahojnnes:
Note that you need to close and reopen COLMAP when changing the theme. It will not work on the fly, as the icons are only refreshed upon COLMAP start.

kevintsq:
I'm using Qt version 6.4.2 and Ubuntu 24.04.4 LTS. I didn't change the theme; it's dark by default.

B1ueber2y:
I can confirm that the issue was reproduced with ubuntu 24.04 and Qt 6.4.2 on the main branch and fixed in this PR:

image

B1ueber2y (approved):

PR #4486, opened 2026-06-29
Bypass `IncrementalMapper` to use `max_normalized_reproj_error` in retriangulation

StonerLing:
`IncrementalMapper::IterativeGlobalRefinement` uses pixel-based `filter_max_reproj_error`, which is
incompatible with the global mapper's `max_normalized_reproj_error`. Bypassing the `IncrementalMapper` layer allows the
global mapper to pass its own error threshold directly.

### DIFF
--- a/src/colmap/sfm/global_mapper.cc
+++ b/src/colmap/sfm/global_mapper.cc
@@ -4,13 +4,15 @@
 #include "colmap/estimators/rotation_averaging.h"
 #include "colmap/math/union_find.h"
 #include "colmap/scene/projection.h"
-#include "colmap/sfm/incremental_mapper.h"
+#include "colmap/sfm/incremental_triangulator.h"
 #include "colmap/sfm/observation_manager.h"
 #include "colmap/util/logging.h"
 #include "colmap/util/misc.h"
+#include "colmap/util/string.h"
 #include "colmap/util/timer.h"
 
 #include <algorithm>
+#include <memory>
 
 namespace colmap {
 namespace {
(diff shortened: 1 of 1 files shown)


src/colmap/sfm/global_mapper.cc
```diff
@@ -443,20 +451,60 @@ bool GlobalMapper::IterativeRetriangulateAndRefine(
...
+    // TODO: Skip normalization when position priors are used (similar to
+    // incremental mapper's !use_prior_position condition).
+    reconstruction_->Normalize();
+
+    size_t num_completed = triangulator->CompleteAllTracks(options);
+    size_t num_merged = triangulator->MergeAllTracks(options);
+    size_t num_filtered = obs_manager->FilterPoints3DWithLargeReprojectionError(
+        max_normalized_reproj_error,
+        reconstruction_->Point3DIds(),
+        ReprojectionErrorType::NORMALIZED);
```
StonerLing, line 483: Main change: explicitly pass the correct ReprojectionErrorType::NORMALIZED threshold.

lpanaf:
Hi @StonerLing, thank you for the PR. Originally, the pipeline was designed to leverage the native COLMAP triangulation since I assumed the hyperparameters were already tuned. However, it is possible that making it consistent with the global pipeline will yield better results. Do you have any statistics in support of switching this parameter? For example, have you tried running the ETH3D and IMC benchmarks available inside COLMAP?

StonerLing:
Thanks for the clarification! I only tested on my own data and noticed the inconsistency — I don't have ETH3D/IMC benchmark numbers to support the change currently. 

If the reuse of the incremental triangulation parameters was intentional, I'll close this PR. That said, adding a brief comment or a separate config option for the retriangulation error metric could help avoid future confusion.

PR #4487, opened 2026-06-29
Add mapper selection (incremental/global/hierarchical) to the GUI

ahojnnes:
## Summary

Previously the COLMAP GUI could only run the **incremental** mapper. This PR adds a drop-down in the reconstruction options to select the **incremental**, **global**, or **hierarchical** mapper, with rendering behavior appropriate to each.

## Behavior

- **Incremental** — unchanged live rendering (per initial pair / image / sub-model).
- **Global** — renders after global positioning and after each refinement (each bundle-adjustment iteration and retriangulation). Pause/Step are supported *between stages*.
- **Hierarchical** — renders only the final merged reconstruction (clusters reconstruct in separate managers). Pause/Step are disabled.

The reconstruction options dialog now switches the displayed option tabs based on the selected mapper, exposing dedicated **global** and **hierarchical** option panels (the hierarchical page reuses the incremental tabs, since each cluster is reconstructed with the incremental options, plus a tab for the clustering/worker options).

## Implementation notes

- `MainWindow::mapper_controller_` is now a polymorphic `std::unique_ptr<Thread>`; `CreateControllers()` builds the selected pipeline and wires the appropriate render callbacks.
- Added an optional progress callback to `GlobalMapper::Solve` / `IterativeBundleAdjustment`, and a `GlobalPipeline::MODEL_UPDATE_CALLBACK`. The global pipeline now adds its reconstruction to the manager up front and renders intermediate states (this also removes a full reconstruction copy that previously happened at the end).
- Promoted `HierarchicalPipeline::Options` to a top-level `HierarchicalPipelineOptions` (mirroring `GlobalPipelineOptions`) and registered it in `OptionManager` via `AddHierarchicalMapperOptions()`, shared by the GUI and the `hierarchical_mapper` CLI. **Note:** the `hierarchical_mapper` CLI flags are now `HierarchicalMapper.*`-prefixed (was `--num_workers`, `--image_overlap`, `--leaf_max_num_images`), consistent with `Mapper.*` / `GlobalMapper.*`.
- Added `OptionsWidget::AddOptionIntUnlimited` to handle `INT_MAX` "unlimited" sentinels (e.g. `track_required_tracks_per_view`, `keep_max_num_tracks`), shown/edited as `-1`, fixing a `QSpinBox` overflow.

The mapper selection itself is a GUI-session setting (not written to the project `.ini`), consistent with how the global mapper options are handled.

## Testing

image

### DIFF
--- a/src/colmap/exe/sfm.cc
+++ b/src/colmap/exe/sfm.cc
@@ -423,20 +423,14 @@ int RunGlobalMapper(int argc, char** argv) {
 }
 
 int RunHierarchicalMapper(int argc, char** argv) {
-  HierarchicalPipeline::Options mapper_options;
   std::filesystem::path output_path;
 
   OptionManager options;
   options.AddDatabaseOptions();
-  options.AddRequiredOption("image_path", &mapper_options.image_path);
+  options.AddRequiredOption("image_path",
+                            &options.hierarchical_mapper->image_path);
   options.AddRequiredOption("output_path", &output_path);
-  options.AddDefaultOption("num_threads", &mapper_options.num_threads);
-  options.AddDefaultOption("num_workers", &mapper_options.num_workers);
-  options.AddDefaultOption("image_overlap",
-                           &mapper_options.clustering_options.image_overlap);
-  options.AddDefaultOption(
-      "leaf_max_num_images",
-      &mapper_options.clustering_options.leaf_max_num_images);
+  options.AddHierarchicalMapperOptions();
   options.AddMapperOptions();
   if (!options.Parse(argc, argv)) {
     return EXIT_FAILURE;
@@ -447,10 +441,10 @@ int RunHierarchicalMapper(int argc, char** argv) {
     return EXIT_FAILURE;
   }
 
-  mapper_options.incremental_options = *options.mapper;
+  options.hierarchical_mapper->incremental_options = *options.mapper;
   auto reconstruction_manager = std::make_shared<ReconstructionManager>();
   HierarchicalPipeline hierarchical_mapper(
-      mapper_options,
+      *options.hierarchical_mapper,
       Database::Open(*options.database_path),
       reconstruction_manager);
   hierarchical_mapper.Run();
--- a/src/colmap/sfm/global_mapper.cc
+++ b/src/colmap/sfm/global_mapper.cc
@@ -340,7 +340,8 @@ bool GlobalMapper::IterativeBundleAdjustment(
     double min_tri_angle_deg,
     int num_iterations,
     bool skip_fixed_rotation_stage,
-    bool skip_joint_optimization_stage) {
+    bool skip_joint_optimization_stage,
+    const std::function<bool()>& on_progress) {
   for (int ite = 0; ite < num_iterations; ite++) {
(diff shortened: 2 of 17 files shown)


src/colmap/exe/sfm.cc
```diff
@@ -423,20 +423,14 @@ int RunGlobalMapper(int argc, char** argv) {
...
   options.AddDatabaseOptions();
-  options.AddRequiredOption("image_path", &mapper_options.image_path);
+  options.AddRequiredOption("image_path",
+                            &options.hierarchical_mapper->image_path);
   options.AddRequiredOption("output_path", &output_path);
-  options.AddDefaultOption("num_threads", &mapper_options.num_threads);
-  options.AddDefaultOption("num_workers", &mapper_options.num_workers);
-  options.AddDefaultOption("image_overlap",
-                           &mapper_options.clustering_options.image_overlap);
-  options.AddDefaultOption(
```
B1ueber2y, line 437: This would be considered a breaking change in the CLI.

src/colmap/sfm/global_mapper.cc
```diff
@@ -372,6 +368,12 @@ bool GlobalMapper::IterativeBundleAdjustment(
     // incremental mapper's !use_prior_position condition).
     reconstruction_->Normalize();
 
+    // Report progress for this refinement iteration and stop early if
```
B1ueber2y, line 371: The error at this point is still in normalized space, which may result in inconsistency at intermediate visualizations. Maybe we should call UpdatePoint3DErrors() here as well?

src/colmap/ui/reconstruction_options_widget.h
```diff
@@ -34,12 +34,19 @@
 
 #include <QtCore>
 #include <QtWidgets>
+#include <functional>
 
 namespace colmap {
 
+// Selects which mapper the GUI runs when starting a reconstruction.
+enum class MapperType { INCREMENTAL, HIERARCHICAL, GLOBAL };
```
B1ueber2y, line 42: There seems to be a similar enum in automatic_reconstruction.h. Does it make sense to consolidate?

src/colmap/ui/main_window.cc
```diff
@@ -1730,6 +1779,17 @@ void MainWindow::EnableBlockingActions() {
...
     action->setEnabled(true);
   }
+  UpdateMapperControls();
+}
+
+void MainWindow::UpdateMapperControls() {
+  // The hierarchical mapper runs as a single solve that cannot be stepped or
+  // paused, so disable both controls when it is selected.
+  const bool supports_stepping = mapper_type_ != MapperType::HIERARCHICAL;
+  action_reconstruction_step_->setEnabled(
```
B1ueber2y: I think here the logic is not ideal. Imagine that a user switches in the UI from incremental to hierarchical and then switch back, the step would be disabled.

src/colmap/ui/reconstruction_options_widget.h
ahojnnes: I'd avoid adding a dependency here. It trades one code line with one include.

B1ueber2y (approved):

PR #4489, opened 2026-06-29
Add nanosecond timestamp type and conversion utils

B1ueber2y:
Splitted from https://github.com/colmap/colmap/pull/2625. 

Added timestamp_t type as int64_t, in nanoseconds.

### DIFF
--- a/src/colmap/util/timestamp.h
+++ b/src/colmap/util/timestamp.h
@@ -1,0 +1,53 @@
+// Copyright (c), ETH Zurich and UNC Chapel Hill.
+// All rights reserved.
+//
+// Redistribution and use in source and binary forms, with or without
+// modification, are permitted provided that the following conditions are met:
+//
+//     * Redistributions of source code must retain the above copyright
+//       notice, this list of conditions and the following disclaimer.
+//
+//     * Redistributions in binary form must reproduce the above copyright
+//       notice, this list of conditions and the following disclaimer in the
+//       documentation and/or other materials provided with the distribution.
+//
+//     * Neither the name of ETH Zurich and UNC Chapel Hill nor the names of
+//       its contributors may be used to endorse or promote products derived
+//       from this software without specific prior written permission.
+//
+// THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS "AS IS"
+// AND ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO, THE
+// IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS FOR A PARTICULAR PURPOSE
+// ARE DISCLAIMED. IN NO EVENT SHALL THE COPYRIGHT HOLDERS OR CONTRIBUTORS BE
+// LIABLE FOR ANY DIRECT, INDIRECT, INCIDENTAL, SPECIAL, EXEMPLARY, OR
+// CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF
+// SUBSTITUTE GOODS OR SERVICES; LOSS OF USE, DATA, OR PROFITS; OR BUSINESS
+// INTERRUPTION) HOWEVER CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN
+// CONTRACT, STRICT LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE)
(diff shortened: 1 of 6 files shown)


src/colmap/util/timestamp.h
```diff
@@ -0,0 +1,53 @@
...
+// POSSIBILITY OF SUCH DAMAGE.
+
+#pragma once
+
+#include "colmap/util/types.h"
+
+namespace colmap {
+
+// Convert a nanosecond timestamp to seconds.
+inline double TimestampToSeconds(timestamp_t t) { return t * 1e-9; }
```
sarlinpe: I suggest using AFromB here too, to be consistent with the naming transform variables a_from_b.

sarlinpe (approved):

src/pycolmap/util/timestamp.cc
```diff
@@ -0,0 +1,13 @@
+#include "colmap/util/timestamp.h"
+
+#include <pybind11/pybind11.h>
+
+using namespace colmap;
+using namespace pybind11::literals;
+namespace py = pybind11;
+
+void BindTimestamp(py::module& m) {
+  m.def("seconds_from_timestamp", &SecondsFromTimestamp, "timestamp_ns"_a);
```
ahojnnes: nit: The C++ interface uses "t" as arg name.

src/colmap/util/types.h
```diff
@@ -121,6 +121,13 @@ constexpr point2D_t kInvalidPoint2DIdx = std::numeric_limits<point2D_t>::max();
 using point3D_t = uint64_t;
 constexpr point3D_t kInvalidPoint3DId = std::numeric_limits<point3D_t>::max();
 
+// Timestamp in nanoseconds. Using int64_t rather than double avoids
+// floating-point precision loss when comparing or differencing large absolute
+// timestamps (e.g., Unix epoch in nanoseconds), and enables exact equality
+// checks for use as map keys.
+using timestamp_t = int64_t;
+constexpr timestamp_t kInvalidTimestamp = -1;
```
ahojnnes: Is -1 a good value? Do we expect timestamps to be strictly positive? In this case, we should probably document it here and add some throw_check to the timestamp conversions?

src/colmap/util/timestamp.h
```diff
@@ -0,0 +1,53 @@
...
+namespace colmap {
+
+// Convert a nanosecond timestamp to seconds.
+inline double SecondsFromTimestamp(timestamp_t t) { return t * 1e-9; }
+
+// Convert seconds to a nanosecond timestamp. Truncates sub-nanosecond values.
+// Intended for small durations (e.g., config values), not large absolute
+// timestamps which should be parsed as int64 directly.
+inline timestamp_t TimestampFromSeconds(double s) {
+  return static_cast<timestamp_t>(s * 1e9);
```
ahojnnes: Would std::round be "more accurate"?

src/colmap/util/timestamp.h
```diff
@@ -0,0 +1,53 @@
...
+// POSSIBILITY OF SUCH DAMAGE.
+
+#pragma once
+
+#include "colmap/util/types.h"
+
+namespace colmap {
+
+// Convert a nanosecond timestamp to seconds.
+inline double SecondsFromTimestamp(timestamp_t t) { return t * 1e-9; }
```
ahojnnes: I guess this will run into bad precision issues for integers > 2^53?

src/colmap/util/types.h
B1ueber2y: Good catch. I documented it, added throw checks and some tests.

ahojnnes:
I don't see this comment being addressed:

> I guess this will run into bad precision issues for integers > 2^53?

Btw, technically nothing is preventing the timestamps from being negative and we could just pick any integer > 2^53 as the invalid value?

src/colmap/util/timestamp.h
B1ueber2y: I added a comment about it. Do we want to explicit check and throw when its larger than 2^53?

B1ueber2y:
Here: https://github.com/colmap/colmap/pull/4489/commits/5fc99397b1d08906f940fec9386dab0eee0aaf20, I changed the invalid number to INT_MIN and added warning when the absolute value of the input to ``SecondsFromTimestamp`` exceeds 2^53. In practice this should not happen as the whole point of this new type is to use ``TimestampDiffSeconds`` to achieve better precision.

I think the comments have now been fully addressed. PTAL : ) @ahojnnes

ahojnnes (approved):
Thanks for addressing all the feedback. LGTM.

PR #4499, opened 2026-07-02
Speed up cheirality test with a closed-form alternative

B1ueber2y:
In preparation for wiring the cheirality test into the sampson error residual computation.

### DIFF
--- a/src/colmap/estimators/two_view_geometry.cc
+++ b/src/colmap/estimators/two_view_geometry.cc
@@ -44,6 +44,7 @@
 #include "colmap/util/logging.h"
 #include "colmap/util/timer.h"
 
+#include <algorithm>
 #include <unordered_set>
 
 #include <Eigen/Geometry>
@@ -627,14 +628,15 @@ bool EstimateTwoViewGeometryPoseFromCamRays(
   // CALIBRATED configuration (or DEGENERATE), so they are handled by the
   // CALIBRATED branch below and never reach the CalibrationMatrix() calls.
   Rigid3d cam2_from_cam1;
+  std::vector<int> inlier_idxs;
   if (geometry->config == TwoViewGeometry::ConfigurationType::CALIBRATED) {
     THROW_CHECK(geometry->E.has_value());
     PoseFromEssentialMatrix(*geometry->E,
                             inlier_cam_rays1,
                             inlier_cam_rays2,
                             &cam2_from_cam1,
-                            &points3D);
-    if (points3D.empty()) {
+                            &inlier_idxs);
+    if (inlier_idxs.empty()) {
       return false;
     }
   } else if (geometry->config ==
@@ -643,8 +645,8 @@ bool EstimateTwoViewGeometryPoseFromCamRays(
     const Eigen::Matrix3d E = EssentialFromFundamentalMatrix(
         camera2.CalibrationMatrix(), *geometry->F, camera1.CalibrationMatrix());
     PoseFromEssentialMatrix(
-        E, inlier_cam_rays1, inlier_cam_rays2, &cam2_from_cam1, &points3D);
-    if (points3D.empty()) {
+        E, inlier_cam_rays1, inlier_cam_rays2, &cam2_from_cam1, &inlier_idxs);
+    if (inlier_idxs.empty()) {
       return false;
     }
   } else if (geometry->config == TwoViewGeometry::ConfigurationType::PLANAR ||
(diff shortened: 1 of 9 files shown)


src/colmap/estimators/two_view_geometry.cc
```diff
@@ -685,7 +687,25 @@ bool EstimateTwoViewGeometryPoseFromCamRays(
...
+  if (!valid_indices.empty()) {
+    // Essential-matrix paths (CALIBRATED / UNCALIBRATED): the triangulation
+    // angle is the parallax between the surviving corresponding rays, which
+    // needs no explicit triangulation.
+    const Eigen::Quaterniond cam1_from_cam2_rotation =
+        cam2_from_cam1.rotation().inverse();
+    std::vector<double> tri_angles;
+    tri_angles.reserve(valid_indices.size());
+    for (const int idx : valid_indices) {
+      const double angle = CalculateAngleBetweenVectors(
```
B1ueber2y, line 699: Before the triangulation angle was calculated w.r.t. to the 3D point rather than the ray. Note that there is a logic change here because the two rays do not necessarily intersect at the 3D point, which was computed via mid point triangulation. This change should be conceptually negligible but worth flagging here.

src/pycolmap/estimators/essential_matrix.cc
```diff
@@ -78,19 +78,18 @@ py::typing::Optional<py::dict> PyEstimateAndDecomposeEssentialMatrix(
...
                           &cam2_from_cam1,
-                          &inlier_points3D);
+                          &valid_indices);
 
   py::gil_scoped_acquire acquire;
   return py::dict("E"_a = report.model,
                   "cam2_from_cam1"_a = cam2_from_cam1,
                   "num_inliers"_a = report.support.num_inliers,
-                  "inlier_mask"_a = ToPythonMask(report.inlier_mask),
-                  "inlier_points3D"_a = inlier_points3D);
```
B1ueber2y, line 93: Here is a breaking change: dropping the field inlier_points3D. Through a search on github this field has never been used in any public repo.

src/colmap/geometry/pose.cc
```diff
@@ -245,18 +244,25 @@ Eigen::Matrix3d RightJacobianFromAngleAxis(const Eigen::Vector3d& omega) {
 bool CheckCheirality(const Rigid3d& cam2_from_cam1,
                      const std::vector<Eigen::Vector3d>& cam_rays1,
                      const std::vector<Eigen::Vector3d>& cam_rays2,
-                     std::vector<Eigen::Vector3d>* points3D) {
+                     std::vector<int>* valid_indices) {
   THROW_CHECK_EQ(cam_rays1.size(), cam_rays2.size());
-  points3D->clear();
+  valid_indices->clear();
+  const Eigen::Matrix3d R = cam2_from_cam1.rotation().toRotationMatrix();
+  const Eigen::Vector3d t = cam2_from_cam1.translation();
```
ahojnnes: const ref

src/colmap/geometry/essential_matrix.h
```diff
@@ -61,12 +61,12 @@ void DecomposeEssentialMatrix(const Eigen::Matrix3d& E,
 // @param cam_rays1       First set of corresponding rays.
 // @param cam_rays2       Second set of corresponding rays.
 // @param cam2_from_cam1  Relative camera transformation.
-// @param points3D        Triangulated 3D points infront of camera.
+// @param valid_indices     Indices of correspondences in front of both cameras.
```
ahojnnes:
```suggestion
// @param valid_indices   Indices of correspondences in front of both cameras.
```

src/colmap/estimators/two_view_geometry.cc
ahojnnes: Yes, I think it should be fine.

src/colmap/geometry/pose.cc
ahojnnes: Is it even worth extracting?

src/colmap/geometry/pose.cc
```diff
@@ -245,18 +244,25 @@ Eigen::Matrix3d RightJacobianFromAngleAxis(const Eigen::Vector3d& omega) {
 bool CheckCheirality(const Rigid3d& cam2_from_cam1,
                      const std::vector<Eigen::Vector3d>& cam_rays1,
                      const std::vector<Eigen::Vector3d>& cam_rays2,
-                     std::vector<Eigen::Vector3d>* points3D) {
+                     std::vector<int>* valid_indices) {
   THROW_CHECK_EQ(cam_rays1.size(), cam_rays2.size());
-  points3D->clear();
+  valid_indices->clear();
+  const Eigen::Matrix3d R = cam2_from_cam1.rotation().toRotationMatrix();
```
ahojnnes: nit: cam2_from_cam1_rot

ahojnnes (approved):
Thanks, LGTM

src/colmap/estimators/two_view_geometry.cc
vlarsson: I don't understand how this works? It seems like it does not depend on the translation at all? Maybe I am missunderstanding what it is supposed to do

vlarsson: Ah okay, I thought this was the chirality part.

PR #4501, opened 2026-07-03
Load mapper database lazily instead of at GUI startup

ahojnnes:
## Problem

Starting the COLMAP GUI (with no mapping requested) reads the entire database into a `DatabaseCache`, emitting logs like:

```
incremental_pipeline.cc:278] Loading database
database_cache.cc:72] Loading rigs...
database_cache.cc:90] Loading cameras...
...
```

The `MainWindow` constructor called `CreateControllers()`, which eagerly constructed an `IncrementalPipeline` whose constructor loads the database. This also fired on every New/Open Project and Import (via `ReconstructionReset()`). The work is redundant since the pipeline is rebuilt anyway in `ReconstructionStart()` right before running.

## Fix

Construct the mapper controller lazily — only when a reconstruction is actually started:

- The constructor now just initializes control states via `UpdateMapperControls()` (already null-safe).
- `ReconstructionReset()` calls the new `ResetMapperController()`, which tears down any running controller **without** building a new one (no DB read).
- `CreateMapperController()` (renamed from the inaccurate plural `CreateControllers`) remains the single place that builds the pipeline, invoked from `ReconstructionStart()`.
- Shared `Stop`/`Wait`/`reset` teardown factored into `StopMapperController()`.
- `mapper_controller_` can now be null before the first start, so the relevant checks in `ReconstructionStart()`/`ReconstructionStep()` are null-guarded.

After this change, launching the GUI and opening/importing projects no longer read the database — that only happens when mapping actually starts.

## Testing

- `ninja colmap_ui` builds cleanly.

### DIFF
--- a/src/colmap/ui/main_window.cc
+++ b/src/colmap/ui/main_window.cc
@@ -1393,6 +1393,9 @@ void MainWindow::ReconstructionStep() {
 }
 
 void MainWindow::ReconstructionPause() {
+  if (!mapper_controller_) {
+    return;
+  }
   timer_.Pause();
   mapper_controller_->Pause();
   EnableBlockingActions();
--- a/src/colmap/ui/main_window.cc
+++ b/src/colmap/ui/main_window.cc
@@ -183,7 +183,10 @@ MainWindow::MainWindow(OptionManager options)
   CreateMenus();
   CreateToolbar();
   CreateStatusbar();
-  CreateControllers();
+  // The mapper controller (and its database cache) is created lazily when a
+  // reconstruction is actually started, so opening the GUI does not read the
+  // database. Just initialize the control states here.
+  UpdateMapperControls();
 
   ShowLog();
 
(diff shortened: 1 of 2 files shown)


src/colmap/ui/main_window.cc
```diff
@@ -184,7 +183,10 @@ MainWindow::MainWindow(OptionManager options)
   CreateMenus();
   CreateToolbar();
   CreateStatusbar();
```
B1ueber2y, line 186: Since there is no controller created anymore, there is a potential path for crushing:

on a fresh GUI launch, no project is configured, so options_.Check() fails. And UpdateMapperControls() enables the Step action whenever !running && !finished, which is true with a null controller. So the Step button is enabled and clickable at startup. Click it → ReconstructionStart shows the "invalid project" dialog and returns without building anything → ReconstructionPause() dereferences null → crash.

A fix would be a null guard in ReconstructionPause().

B1ueber2y (approved):

PR #4497, opened 2026-07-02
Manifold-aware tiny solver for fast non-linear geometric refinement of E

B1ueber2y:
This PR adds non-linear geometric refinement of E via ceres TinySolver. TinySolver does not support manifold optimization, so a tangent sampson error impl was added. 

loransac impl was updated to support refinement from initial model. 

cc: @vlarsson 

Some benchmarking numbers on synthetic data by sweeping on local estimator and support measurement, where the trend is consistent: 1) geometric refinement is better than algebraic non-minimal solver; 2) MSAC is better than vanilla ransac. 

There are still some accuracy and performance gap against PoseLib when there are more outliers in the data, for which my current guess is related to the cheirality check. Will try to look more into this part as a follow-up. 

```
colmap/100/100                 792 ms          792 ms            1 fail_pct=0 items_per_second=126.333/s num_trials_mean=1.001k rot_err_mean_deg=0.231557 rot_err_med_deg=0.197138 t_err_mean_deg=0.423347 t_err_med_deg=0.332021
colmap8pt/100/100              780 ms          780 ms            1 fail_pct=0 items_per_second=128.249/s num_trials_mean=1.001k rot_err_mean_deg=0.188561 rot_err_med_deg=0.170085 t_err_mean_deg=0.395329 t_err_med_deg=0.356236
colmap_msac/100/100            773 ms          773 ms            1 fail_pct=0 items_per_second=129.43/s num_trials_mean=1.001k rot_err_mean_deg=0.179189 rot_err_med_deg=0.160359 t_err_mean_deg=0.325814 t_err_med_deg=0.296999
colmap_lm/100/100              857 ms          857 ms            1 fail_pct=0 items_per_second=116.696/s num_trials_mean=1.001k rot_err_mean_deg=0.163713 rot_err_med_deg=0.123556 t_err_mean_deg=0.304265 t_err_med_deg=0.220374
colmap_lm_msac/100/100         838 ms          838 ms            1 fail_pct=0 items_per_second=119.293/s num_trials_mean=1.001k rot_err_mean_deg=0.118874 rot_err_med_deg=0.108084 t_err_mean_deg=0.21372 t_err_med_deg=0.203962
poselib/100/100                957 ms          957 ms            1 fail_pct=0 items_per_second=104.447/s num_trials_mean=1.001k rot_err_mean_deg=0.118034 rot_err_med_deg=0.101039 t_err_mean_deg=0.220989 t_err_med_deg=0.211565
[...]

PR #4504, opened 2026-07-03
Add cheirality check for calculating sampson error at E estimation

B1ueber2y:
The cheirality check at residual evaluation is crucial for accuracy when there is a large presence of outliers, according to the numbers at https://github.com/colmap/colmap/pull/4497#issuecomment-4876271485


ahojnnes (approved):

ahojnnes (commented):
Suggest creating one or two unit tests to cover cheirality and ensure otherwise equality with normal sampson error?

PR #4510, opened 2026-07-05
Add swappable hash map backend (std, boost)

ahojnnes:
## Summary

Speeds up the mapping pipeline by making the hash map implementation used in
the performance-critical scene/SfM containers a **compile-time-selectable
backend**, so a faster open-addressing map can replace `std::unordered_map`
without touching call sites.

The backend is chosen via CMake and defaults to **auto**: `BOOST` when the
available Boost is new enough (`boost::unordered_node_map` requires **Boost >=
1.84**), otherwise `STD`. Also compared ABSEIL and ANKERL implementations,
while BOOST provided the largest speedup and is an existing direct dependency.

```
-DCOLMAP_HASH_MAP_BACKEND=BOOST|STD   # default: auto (empty)
```

Auto keeps apt-based Linux builds working (Ubuntu's Boost is < 1.84, so they use
`STD`), while macOS (brew) and Windows/vcpkg get `BOOST`. Explicitly requesting
`BOOST` with an older Boost is a configuration error.

On synthetic mapping this yields a **1.3–1.8× end-to-end speedup**, and **~1.09×
on the real south-building dataset** (see results below).

## Design

New `src/colmap/util/hash_containers.h` defines two families of aliases in
`namespace colmap`, selected by the backend macro:

- `FlatHashMap` / `FlatHashSet` — open-addressing; fastest and lowest-overhead,
  but invalidate element references/iterators on rehash (and on erase for
  boost). Use only where no long-lived reference to an element is held across a
  mutation, and avoid iterator-based erase loops (erase by key instead).
- `NodeHashMap` / `NodeHashSet` — node-based; element references/pointers stay
  valid across insert/erase of other elements (`std::unordered_*` semantics).

| Backend | Flat | Node |
|---|---|---|
| `STD` | `std::unordered_map/set` | `std::unordered_map/set` |
| `BOOST` (default when Boost >= 1.84) | `boost::unordered_flat_map/set` | `boost::unordered_node_map/set` |

The default hash stays `std::hash<K>` for both backends, so the existing
`std::hash` specializations and `colmap::PairHash` are reused and hashing is held
constant — only the container implementation varies. Both backends are
header-only (boost-unordered comes with `Boost::boost`), so no new linked
dependency is introduced.

## What changed

- **All** `std::unordered_map`/`std::unordered_set` across colmap and pycolmap
  now go through the aliases.
- **Sets → `FlatHashSet`** everywhere: a set element *is* its key, so there are
  no element references to invalidate, and no iterator-based erase loops remain.
- **Maps → `NodeHashMap`** by default; promoted to `FlatHashMap` where proven
  safe (observation-manager stats, correspondence `image_pairs_`, mapper
  registration stats, triangulator caches, `point3D_t` id-maps). The element
  stores (`Reconstruction`/`DatabaseCache`/`CorrespondenceGraph::images_`), the
  LRU cache and the feature-matcher caches stay node-based because they hand out
  element references across mutations. In particular `Reconstruction::points3D_`
  must stay node-based (the triangulator holds a `Point3D&` across a merge that
  erases from the same map).
- **Portable erase:** `IncrementalTriangulator::GetModifiedPoints3D` erases by
  key instead of via an iterator loop, since flat iterator-erase is not portable
  (boost flat erase invalidates iterators).
- **pycolmap:** the opaque bound maps are aliased to `NodeHashMap`, and pybind11
  type casters are added for the flat/node aliases (guarded to `BOOST`; under
  `STD` the aliases are `std::` types pybind11 already handles).
- **Include hygiene:** `hash_containers.h` is included only where an alias is
  used, and the now-unused `<unordered_map>`/`<unordered_set>` includes were
  removed across the tree.

## Benchmark

End-to-end `IncrementalPipeline` runs, single-threaded, fixed seed, Release
builds, one build per backend. Measured with a local benchmark harness (not
part of this PR).

**Synthetic scenes with realistic observation noise** (`point2D_stddev = 1.0`,
1500 points). Reconstructions are equivalent across backends (identical
registered-image counts):

| Images | STD | BOOST | Speedup | Peak RSS STD → BOOST |
|---:|---:|---:|---:|---:|
| 50  | 4.28 s | 3.28 s | 1.30× | 199 → 198 MB (1.00×) |
| 100 | 17.89 s | 10.13 s | 1.77× | 405 → 404 MB (1.00×) |
| 200 | 67.59 s | 48.92 s | 1.38× | 760 → 827 MB (1.09×) |
| 400 | 296.87 s | 168.20 s | 1.77× | 1883 → 2149 MB (1.14×) |

**Real south-building dataset** (128 SIFT images, ~61.5k points), mean of 3 runs;
identical registered-image count (128), 3D-point count within 0.01% (hash-order
tie-breaks):

| | STD | BOOST | Speedup | Peak RSS |
|---|---:|---:|---:|---:|
| south-building | 63.0 s | 57.7 s | 1.09× | 913 → 987 MB (1.08×) |

`boost::unordered` gives a speedup that grows with image count, at a modest
peak-memory cost at scale (+14% at 400 synthetic images). The real-dataset gain
is smaller (~9%) because real data spends proportionally more time in bundle
adjustment, where the container matters less. For reference, a noise-free
synthetic sweep (where BA does almost no work) reaches up to ~2.8× at 400 images.


B1ueber2y (approved):
Thanks! LGTM.

PR #4511, opened 2026-07-06
Remove std::pair hash specialization and make PairHash pack integral pairs losslessly

B1ueber2y:
There was still a std::hash specialization of std::pair<uint32_t, uint32_t> in the codebase, which had the same hazard that PairHash was aimed to avoid. 

Removed and consolidated with PairHash. Now PairHash packs std::pair<uint32_t, uint32_t> in a lossless way.

### DIFF
--- a/src/colmap/util/types.h
+++ b/src/colmap/util/types.h
@@ -31,6 +31,8 @@
 
 #include <functional>
 #include <iterator>
+#include <type_traits>
+#include <utility>
 
 #ifdef _MSC_VER
 #if _MSC_VER >= 1600
@@ -349,6 +351,11 @@ struct filter_view {
 template <typename>
 struct always_false : std::false_type {};
 
+// Folds `value` into `seed`, following the classic boost::hash_combine spread.
+inline std::size_t HashCombine(std::size_t seed, std::size_t value) {
+  return seed ^ (value + 0x9e3779b9 + (seed << 6) + (seed >> 2));
+}
+
 // Hash functor for std::pair, e.g., for use with unordered containers keyed on
 // e.g., std::pair<point3D_t, point3D_t>. Provided as an explicit functor
 // (passed to the container) rather than a std::hash<std::pair<...>>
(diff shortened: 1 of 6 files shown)


ahojnnes:
Let's wait with this until the hash map change goes through, as these will conflict.

src/colmap/util/types.h
```diff
@@ -349,19 +351,31 @@ struct filter_view {
...
+//
+// Packs both halves into disjoint bits when they fit in a size_t (collision-
+// free), else mixes them with HashCombine.
 struct PairHash {
   template <typename T1, typename T2>
   std::size_t operator()(const std::pair<T1, T2>& p) const {
-    const std::size_t h1 = std::hash<T1>{}(p.first);
-    const std::size_t h2 = std::hash<T2>{}(p.second);
-    return h1 ^ (h2 + 0x9e3779b9 + (h1 << 6) + (h1 >> 2));
+    if constexpr (std::is_unsigned_v<T1> && std::is_unsigned_v<T2> &&
```
sarlinpe: Can you explain why this branch cannot use HashCombine as well? Because HashCombine has a (very small) risk of collision? Please add a comment accordingly.

sarlinpe (approved):
Thank you!

src/colmap/util/types.h
B1ueber2y: Yes. Added a comment.

src/colmap/util/types.h
```diff
@@ -349,19 +351,33 @@ struct filter_view {
...
 // e.g., std::pair<point3D_t, point3D_t>. Provided as an explicit functor
 // (passed to the container) rather than a std::hash<std::pair<...>>
 // specialization, since specializing std::hash for the std-owned std::pair type
 // is a global ODR hazard: a downstream translation unit that implicitly
 // instantiates the primary template first would make the later explicit
 // specialization ill-formed.
+//
+// Packs both halves into disjoint bits when they fit in a size_t (collision-
+// free), else mixes them with HashCombine.
 struct PairHash {
```
ahojnnes, line 369: I think we can also cover signed integers?

```suggestion
struct PairHash {
  template <typename T1, typename T2>
  std::size_t operator()(const std::pair<T1, T2>& p) const {
    if constexpr (std::is_integral_v<T1> && std::is_integral_v<T2> &&
                  !std::is_same_v<T1, bool> && !std::is_same_v<T2, bool> &&
                  sizeof(T1) + sizeof(T2) <= sizeof(std::size_t)) {
      // Convert through the unsigned counterpart: the signed->unsigned cast
      // is a bijection modulo 2^n, so negatives map losslessly into exactly
      // their low 8*sizeof(T) bits. Both halves then occupy disjoint bit
      // ranges, making the pack collision-free (unlike HashCombine).
      return (static_cast<std::size_t>(
                  static_cast<std::make_unsigned_t<T1>>(p.first))
              << (8 * sizeof(T2))) |
             static_cast<std::size_t>(
                 static_cast<std::make_unsigned_t<T2>>(p.second));
    } else {
      return HashCombine(std::hash<T1>{}(p.first), std::hash<T2>{}(p.second));
    }
  }
};
```

src/colmap/util/types_test.cc
```diff
@@ -157,7 +157,7 @@ TEST(FeatureMatchHashing, LargeValues) {
 }
 
 TEST(FeatureMatchHashing, Deterministic) {
```
ahojnnes, line 159: Add a test for signed types here and above?

PR #4512, opened 2026-07-06
Fix: rescale already-undistorted images when max_image_size is provided.

### LINKED ISSUE #4508: undistort_images ignores --max_image_size for already-undistorted cameras (PINHOLE/SIMPLE_PINHOLE)
**Describe the bug**
When running colmap image_undistorter with --max_image_size on a reconstruction whose cameras are already pinhole models (e.g. PINHOLE or SIMPLE_PINHOLE), the camera resolution stored in the output reconstruction is not rescaled. The images themselves may be resized, but the camera intrinsics in cameras.bin/cameras.txt retain the original resolution.

Root Cause:
In UndistortReconstruction() (src/colmap/image/undistortion.cc), cameras for which IsUndistorted() returns true are skipped entirely

**To Reproduce**
Have a reconstruction with PINHOLE or SIMPLE_PINHOLE cameras at a high resolution (e.g. 4000×3000).
Run colmap image_undistorter --max_image_size 1000 ....
Inspect the output cameras.bin — the camera width/height are unchanged from the original.

**Expected behavior**
The camera intrinsics in the output reconstruction should be rescaled to match --max_image_size, consistent with the behavior for distorted cameras.

**Environment:**
 - OS: Linux
 - COLMAP Version: main

B1ueber2y:
Fixes https://github.com/colmap/colmap/issues/4508

### DIFF
--- a/src/colmap/image/undistortion.cc
+++ b/src/colmap/image/undistortion.cc
@@ -300,7 +300,7 @@ void UndistortImage(const UndistortCameraOptions& options,
 
 void UndistortReconstruction(const UndistortCameraOptions& options,
                              Reconstruction* reconstruction) {
-  const std::unordered_map<camera_t, Camera> distorted_cameras =
+  const NodeHashMap<camera_t, Camera> distorted_cameras =
       reconstruction->Cameras();
   // Leave a camera unchanged exactly when the image undistortion also copies
   // its images through unchanged, so that the reconstruction stays consistent
@@ -314,18 +314,18 @@ void UndistortReconstruction(const UndistortCameraOptions& options,
   const auto keep_unchanged = [&options](const Camera& camera) {
     return camera.IsUndistorted() && options.max_image_size < 0;
   };
-  for (const auto& camera : distorted_cameras) {
-    if (keep_unchanged(camera.second)) {
+  for (const auto& [camera_id, distorted_camera] : distorted_cameras) {
+    if (keep_unchanged(distorted_camera)) {
       continue;
     }
-    Camera& undistorted_camera = reconstruction->Camera(camera.first);
-    if (camera.second.IsSpherical()) {
+    Camera& undistorted_camera = reconstruction->Camera(camera_id);
+    if (distorted_camera.IsSpherical()) {
       // Only reached with a max_image_size: resize the spherical camera in
       // place, keeping its model.
-      undistorted_camera = camera.second;
+      undistorted_camera = distorted_camera;
       RescaleToMaxImageSize(options, &undistorted_camera);
     } else {
-      undistorted_camera = UndistortCamera(options, camera.second);
+      undistorted_camera = UndistortCamera(options, distorted_camera);
     }
   }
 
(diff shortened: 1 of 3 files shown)


src/colmap/image/undistortion.cc
```diff
@@ -273,11 +273,20 @@ void UndistortReconstruction(const UndistortCameraOptions& options,
...
+  const auto keep_unchanged = [&options](const Camera& camera) {
+    return camera.IsSpherical() ||
+           (camera.IsUndistorted() && options.max_image_size < 0);
+  };
   for (const auto& camera : distorted_cameras) {
-    // IsUndistorted() is true for non-perspective cameras (e.g.
-    // EQUIRECTANGULAR), which cannot be undistorted to a pinhole, so they are
-    // left unchanged.
-    if (camera.second.IsUndistorted()) {
+    if (keep_unchanged(camera.second)) {
```
ahojnnes: Why not also resize spherical images to the target resolution? It seems inconsistent?

B1ueber2y: I think Spherical images cannot be undistorted and before it was gated by IsUndistorted?

ahojnnes: It cannot be undistorted but it can be resized and it makes sense to me to apply a consistent max_image_size across all images -- no matter if already undistorted or not?

I actually already had the same behavior change, as you implement here now when implementing a recent fix to the same code but undid it because of a review comment. Maybe there was a misunderstanding.

B1ueber2y: Sounds good. I have added logic to address rescaling for spherical cameras, with a test. PTAL : )

src/colmap/image/undistortion.cc
ahojnnes: Needs a rebase? Should be NodeHashMap?

src/colmap/image/undistortion.cc
```diff
@@ -275,22 +304,29 @@ void UndistortReconstruction(const UndistortCameraOptions& options,
...
+  // copied through as-is. When a max_image_size is requested, they are still
+  // resized to match the rescaled output images: perspective cameras through
+  // undistortion, spherical cameras by resizing to a smaller image of the same
+  // model.
   const auto keep_unchanged = [&options](const Camera& camera) {
-    return camera.IsSpherical() ||
-           (camera.IsUndistorted() && options.max_image_size < 0);
+    return camera.IsUndistorted() && options.max_image_size < 0;
   };
   for (const auto& camera : distorted_cameras) {
```
ahojnnes: nit: `const auto& [camera_id, distorted_camera] : distorted_cameras`

src/colmap/image/undistortion.cc
```diff
@@ -301,6 +337,26 @@ void UndistortReconstruction(const UndistortCameraOptions& options,
       continue;
     }
     const Camera& undistorted_camera = *image.CameraPtr();
+
+    if (distorted_camera.IsSpherical()) {
```
ahojnnes, line 342: nit: I guess the existing ImgFromCam(CamFromImage()) logic below should still work even though a bit more expensive. Not sure if it's worth the extra code to handle spherical explicitly here?

ahojnnes (approved):

src/colmap/image/undistortion.cc
B1ueber2y: We would need to switch CamRayFromImage somehow for spherical images, as CamFromImage only supports the front hemisphere: https://github.com/colmap/colmap/blob/main/src/colmap/sensor/models.h#L2778

PR #4513, opened 2026-07-06
Analytical reprojection Jacobians for more camera models and fixed-pose BA

ahojnnes:
## Motivation

Bundle adjustment dominates the incremental mapper's runtime on real data. Its reprojection-error cost function had a hand-written **analytical Jacobian** fast path (`AnalyticalReprojErrorCostFunction`), but it was only wired up for a single camera model, **`SIMPLE_RADIAL`**. Every other model — including the calibrated `PINHOLE`/`OPENCV` workflows — and **all fixed-pose observations** (added for neighbour images in local BA after every registration) fell back to Ceres autodiff, which is substantially slower.

## Changes

- Implement `ImgFromCamWithJac()` and enable the analytical path (`has_img_from_cam_with_jac = true`) for `SIMPLE_PINHOLE`, `PINHOLE`, `RADIAL`, `OPENCV` and `FULL_OPENCV`. The existing dispatch in `CreateCameraCostFunction()` routes them automatically — no call-site changes.
- Add `AnalyticalReprojErrorConstantPoseCostFunction` (a fixed-pose variant that bakes in the pose and skips the pose Jacobian) and route `ReprojErrorConstantPoseCostFunctor` through it. This targets the local-BA hot path.
- Rework `benchmark/runtime/cost_functions.cc` into an autodiff-vs-analytical comparison across models.

Analytical Jacobians do not change the optimum, only the speed — the produced reconstructions are byte-for-byte identical (verified below).

## Correctness

- `sensor/models_test` validates every analytical Jacobian against Ceres Jets (autodiff) to 1e-10; it runs automatically for any model with the flag enabled.
- `reprojection_error_test` compares each cost function against numeric differentiation and the autodiff residual, extended to cover all new models and the fixed-pose function.
- All 68 tests across `sensor`/`estimators`/`geometry`/`sfm`/`controllers` pass.

## Benchmark: per-residual cost-function evaluation

`benchmark_cost_functions`, mean time per `Evaluate()` (residual + all Jacobians), single run. `SIMPLE_RADIAL` was already analytical and is shown as a baseline sanity check.

| model | full autodiff | full analytic | speedup | const-pose autodiff | const-pose analytic | speedup |
|---|---:|---:|---:|---:|---:|---:|
| SIMPLE_PINHOLE | 135 ns | 12.3 ns | 11.0× | 44.5 ns | 4.82 ns | 9.2× |
| PINHOLE | 127 ns | 12.2 ns | 10.4× | 66.8 ns | 5.09 ns | 13.1× |
| SIMPLE_RADIAL* | 137 ns | 13.9 ns | 9.9× | 78.0 ns | 7.02 ns | 11.1× |
| RADIAL | 180 ns | 16.5 ns | 10.9× | 82.9 ns | 7.90 ns | 10.5× |
| OPENCV | 232 ns | 20.1 ns | 11.5× | 132 ns | 11.7 ns | 11.3× |
| FULL_OPENCV | 299 ns | 25.2 ns | 11.9× | 215 ns | 16.6 ns | 13.0× |

\* full-variable `SIMPLE_RADIAL` was already analytical before this PR (baseline).

## Benchmark: end-to-end incremental mapping

Measured with the standalone `benchmark_incremental_mapping` harness (colmap/colmap#4514) — synthetic scenes (1000 points, sparse view graph, `point2D_stddev=1.0`), single-threaded, fixed seed. "before" = analytical path disabled for these two models (autodiff); "after" = this PR. `num_reg_images`, `num_points3d` and `mean_track_length` were **identical** across both configs at every size.

| camera model | frames | autodiff | analytic | speedup |
|---|---:|---:|---:|---:|
| PINHOLE | 25 | 1302 ms | 994 ms | 1.31× |
| PINHOLE | 50 | 2883 ms | 2353 ms | 1.23× |
| PINHOLE | 100 | 10696 ms | 8129 ms | 1.32× |
| OPENCV | 25 | 2382 ms | 1542 ms | 1.54× |
| OPENCV | 50 | 3904 ms | 2547 ms | 1.53× |
| OPENCV | 100 | 12902 ms | 9628 ms | 1.34× |

The end-to-end gain is smaller than the per-residual gain because cost-function evaluation is only part of a full solve (Schur complement / linear solves, triangulation, filtering), but it is a consistent ~1.2–1.3× for `PINHOLE` and ~1.35–1.55× for `OPENCV` at no accuracy cost.


src/colmap/sensor/models.h
```diff
@@ -1146,6 +1146,55 @@ bool SimplePinholeCameraModel::ImgFromCam(
   return true;
 }
 
+template <bool Enable, typename std::enable_if<Enable, int>::type>
```
sarlinpe: Since this file is already huge, would it make sense to move these kinds of low-level implementations to a different header?

src/colmap/estimators/cost_functions/reprojection_error.h
```diff
@@ -191,21 +262,37 @@ class ReprojErrorConstantPoseCostFunctor
  public:
   ReprojErrorConstantPoseCostFunctor(const Eigen::Vector2d& point2D,
                                      const Rigid3d& cam_from_world)
-      : cam_from_world_(cam_from_world), reproj_cost_(point2D) {}
+      : point2D_(point2D),
+        cam_from_world_rotation_(cam_from_world.rotation().toRotationMatrix()),
+        cam_from_world_translation_(cam_from_world.translation()) {}
```
sarlinpe, line 278: What is the rationale for this change? It is not documented in the PR description and is different from the statement _route `ReprojErrorConstantPoseCostFunctor` through it_.

sarlinpe (approved):
Thank you, very nice speed up!

src/colmap/estimators/cost_functions/reprojection_error.h
ahojnnes: Will add a comment. The rotation is fixed and 3x3 rotation matrix multiplication is faster than quaternion multiplication. Same applies to the analytical version, which further eliminates quaternion to rotation matrix conversion for the Jacobian.

src/colmap/sensor/models.h
ahojnnes: Good call. Let's split these out.

src/colmap/estimators/cost_functions/reprojection_error_test.cc
```diff
@@ -72,14 +72,41 @@ TEST(ReprojErrorCostFunctor, Nominal) {
...
+#endif
+      numeric_diff_options);
+}
+
+// Validates the fully-variable analytical reprojection error cost function
+// against both the numeric Jacobian and the autodiff residual. The numeric
+// (Ridders) Jacobian comparison uses a looser tolerance than the residual
+// comparison, since finite differences on higher-order distortion models do
+// not reach full double precision; the exact analytical Jacobians are verified
+// separately against Ceres Jets in sensor/models_test.
```
B1ueber2y: Stale comment. These tests were moved to models_jacobian_test.

src/colmap/estimators/cost_functions/reprojection_error_test.cc
```diff
@@ -89,43 +116,126 @@ TEST(ReprojErrorCostFunctor, AnalyticalVersusAutoDiff) {
...
                                Eigen::Vector3d(1, 2, 3));
         Eigen::Vector3d point3D(x, y, z);
-        std::vector<double> simple_radial_params = {200, 100, 120, 0.1};
+        std::vector<double> params = camera_params;
+
+        // Restrict to a realistic field of view. The OpenCV/rational models are
+        // only well-conditioned for moderate normalized radii, where the
+        // finite-difference (Ridders) gradient check is reliable; the exact
+        // analytical Jacobians are validated over the full range against Ceres
+        // Jets in sensor/models_test.
```
B1ueber2y: Same here. Probably make sense to do a full pass on these occurences.

B1ueber2y (approved):
Thank you! LGTM.

PR #4514, opened 2026-07-06
Add end-to-end incremental mapping runtime benchmark

ahojnnes:
## Summary

Adds `benchmark/runtime/incremental_mapping.cc`, an end-to-end incremental mapping runtime benchmark on synthetic data. It runs a full reconstruction (registration, triangulation, bundle adjustment) via `IncrementalPipeline` on a synthetically generated in-memory database and reports wall-clock runtime plus reconstruction statistics.

## Details

- **Deterministic:** fixed random seed and single-threaded, so timings are stable and the produced reconstruction is reproducible across runs.
- **Representative work:** injects realistic observation noise (`point2D_stddev = 1.0`, plus point/pose noise) so bundle adjustment, triangulation, and outlier filtering do meaningful work; uses a sparse-but-connected view graph so per-image cost stays bounded as the image count grows.
- **Parametrized** over camera model (`PINHOLE`, `OPENCV`) and scene size (25 / 50 / 100 frames, 1000 points).
- **Reports** `num_reg_images`, `num_points3d`, and `mean_track_length` as counters alongside the timing, so a change's effect on the reconstruction (not just speed) is visible.

Registered under the existing `BENCHMARK_ENABLED` CMake option.

## Usage

```
cmake .. -DBENCHMARK_ENABLED=ON
ninja benchmark_incremental_mapping
./benchmark/runtime/benchmark_incremental_mapping
```

This is a general-purpose harness for evaluating the runtime impact of mapper / bundle-adjustment changes.

### DIFF
--- a/benchmark/runtime/incremental_mapping.cc
+++ b/benchmark/runtime/incremental_mapping.cc
@@ -78,12 +78,11 @@ void RunPipelineAndReport(benchmark::State& state,
     mapper.Run();
 
     state.PauseTiming();
-    size_t best_size = 0;
     largest = nullptr;
     for (size_t i = 0; i < reconstruction_manager->Size(); ++i) {
       const auto& reconstruction = reconstruction_manager->Get(i);
-      if (reconstruction->NumRegImages() >= best_size) {
-        best_size = reconstruction->NumRegImages();
+      if (largest == nullptr ||
+          reconstruction->NumRegImages() > largest->NumRegImages()) {
         largest = reconstruction;
       }
     }
@@ -151,7 +150,8 @@ void AddArguments(::benchmark::Benchmark* b) {
   }
   b->Unit(benchmark::kMillisecond)
       ->Iterations(1)
-      ->Repetitions(1)
+      ->Repetitions(5)
+      ->ReportAggregatesOnly(true)
       ->UseRealTime();
 }
 
--- a/benchmark/runtime/incremental_mapping.cc
+++ b/benchmark/runtime/incremental_mapping.cc
@@ -1,0 +1,165 @@
+// Copyright (c), ETH Zurich and UNC Chapel Hill.
+// All rights reserved.
+//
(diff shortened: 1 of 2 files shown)


benchmark/runtime/incremental_mapping.cc
```diff
@@ -0,0 +1,165 @@
...
+    auto reconstruction_manager = std::make_shared<ReconstructionManager>();
+    IncrementalPipeline mapper(MakeOptions(), database, reconstruction_manager);
+    mapper.Run();
+
+    state.PauseTiming();
+    size_t best_size = 0;
+    largest = nullptr;
+    for (size_t i = 0; i < reconstruction_manager->Size(); ++i) {
+      const auto& reconstruction = reconstruction_manager->Get(i);
+      if (reconstruction->NumRegImages() >= best_size) {
```
sarlinpe: NumRegImages is cheap to call, so consider calling it directly?
```suggestion
      if (largest == nullptr || 
          reconstruction->NumRegImages() > largest->NumRegImages()) {
```

benchmark/runtime/incremental_mapping.cc
```diff
@@ -0,0 +1,165 @@
...
+}
+
+void AddArguments(::benchmark::Benchmark* b) {
+  b->ArgNames({"num_frames", "num_points3D"});
+  for (const int num_frames : {25, 50, 100}) {
+    b->Args({num_frames, /*num_points3D=*/1000});
+  }
+  b->Unit(benchmark::kMillisecond)
+      ->Iterations(1)
+      ->Repetitions(1)
```
sarlinpe: If the benchmark is fast, should we run multiple repetitions to average out the noise?

```suggestion
      ->Repetitions(5)
      ->ReportAggregatesOnly(true)
```

sarlinpe (approved):
Thank you!

Issue #4382, opened 2026-05-06
Local/Global BA mapper options not propagated to Caspar


ahojnnes:
See IncrementalPipelineOptions::LocalBundleAdjustment() etc.

cc: @tordnat Was this intentional or would it make sense to apply at least the same max_num_iterations?

tordnat:
I have not done any proper hyperparameter tuning on Caspar, other than testing effects on accuracy. I also was a bit unsure whether the best option would be to use Ceres for local BA or Caspar for both. I think it would make sense to reduce the iteration count in local BA, but whether we should mirror the Ceres option here is a bit more difficult to answer, since Ceres can have different solver types based on problem structure and size. We could try to apply the same max num iterations, but I wonder what the impact on accuracy would be on the benchmarks.

tordnat:
I plan to do some more benchmarking soon, but feel free to test out parameters and figure out what works best.

tordnat:
Since Caspar solves the normal equation directly and has different dampening updates and convergence criteria, it does not make sense to propagate these parameters from a global solver. A reasonable compromise is to make the user able to override the default parameters shared across all solvers with the same flag, and let COLMAP propagate them to the right solver internally.

behnamasadi:
A concrete real-world data point on this, plus a closely-related suggestion, since the missing propagation turned into a hard failure for us:

Running COLMAP 4.1.0 with Caspar as the default BA backend, on a ~128-image terrestrial scene the mapper's final **global** bundle adjustment (Caspar) ran for **>20 minutes with no output**, and our CI's no-output hang-watchdog killed the container (SIGKILL, exit 137). Smaller scenes (~10–70 images) finished fine on the same backend.

Two things converged:

1. **This issue** — because the Local/Global BA options (notably `max_num_iterations`) aren't propagated to Caspar, its global BA isn't bounded the way the Ceres path is, so on larger scenes it runs much longer than the equivalent Ceres BA.
2. **No progress output** — the Caspar solve emits nothing per-iteration (the Ceres path logs per-iteration cost/gradient), so a long-but-*working* BA is indistinguishable from a genuine hang to any external monitor / CI watchdog / user tailing the log. That's what turned (1) from "slow" into "killed."

Even a minimal periodic progress line (or just start/finish markers with an iteration count) for the Caspar backend would make long BAs distinguishable from real hangs, and would also help the tuning discussion here — right now there's no visibility into how many iterations Caspar is actually taking.

[...]

PR #4515, opened 2026-07-07
Emit Caspar bundle-adjustment progress under VLOG(1)

behnamasadi:
## Problem

`BundleAdjustmentCaspar::Solve()` calls the Caspar solver with `print_progress` hardcoded to `false`:

```cpp
caspar::SolveResult result = solver.solve(
    /*print_progress=*/false, /*verbose_logging=*/collect_iters);
```

So the Caspar backend produces **no output while it runs**. On larger reconstructions the **global** BA can take many minutes, during which a slow‑but‑working solve is **indistinguishable from a hang** — both to a user tailing the log and to any external monitor (a CI job, or a no‑output watchdog that kills wedged subprocesses).

We hit exactly this running COLMAP 4.1.0 with Caspar as the default backend: on a ~128‑image scene the mapper's final global BA ran **>20 minutes silently** and our no‑output watchdog killed the container (`SIGKILL`, exit 137); smaller scenes finished fine. Data point + context in #4382.

## Change

Gate `print_progress` on `VLOG_IS_ON(1)`, consistent with the `VLOG(1)` diagnostics already used throughout this file:

```cpp
caspar::SolveResult result = solver.solve(
    /*print_progress=*/VLOG_IS_ON(1), /*verbose_logging=*/collect_iters);
```

- **No change to default output** — silent unless run with `--log_level 1` / `GLOG_v=1`.
- With verbose logging on, the Caspar solve reports progress, so a long‑running BA is clearly distinguishable from a real hang.

## Notes / alternatives

Deliberately the smallest, non‑intrusive change. Happy to adjust if you'd prefer to tie it to `BundleAdjustmentOptions::print_summary` (prints by default) or add a dedicated `print_progress` option to `CasparBundleAdjustmentOptions`.

Related: #4382 — the mapper's BA iteration limits aren't propagated to Caspar, which is *why* the global BA runs long in the first place; this PR makes that long run observable.

## Testing

One‑liner substituting an existing glog macro (`VLOG_IS_ON`, already used in this file) for the hardcoded `false`; it doesn't alter control flow. I haven't run a full CUDA+Caspar build locally — CI should confirm the build, and I'm glad to attach sample `-v 1` output from our GPU host.

### DIFF
--- a/src/colmap/estimators/bundle_adjustment_caspar.cc
+++ b/src/colmap/estimators/bundle_adjustment_caspar.cc
@@ -935,7 +935,7 @@ class CasparBundleAdjuster : public BundleAdjuster {
     const bool collect_iters =
         options_.caspar && options_.caspar->collect_iteration_data;
     caspar::SolveResult result = solver.solve(
-        /*print_progress=*/false, /*verbose_logging=*/collect_iters);
+        /*print_progress=*/VLOG_IS_ON(1), /*verbose_logging=*/collect_iters);
     ReadSolverResults(solver);
     WriteResultsToReconstruction();
 


behnamasadi:
@ahojnnes, I think it is an easy one

ahojnnes:
Let's use log level = 2 instead of 1 to be consistent with how Ceres works already. Otherwise, LGTM. Thank you.

PR #4516, opened 2026-07-07
Analytical reprojection Jacobians for all remaining camera models

ahojnnes:
Follow-up to #4513, which added analytic reprojection Jacobians for the pinhole/radial/OpenCV families. This completes the set: every camera model now implements `ImgFromCamWithJac`, so `AnalyticalReprojErrorCostFunction` and its constant-pose variant cover the full model list and no model falls back to Ceres autodiff.

## What's added

`ImgFromCamWithJac` (2×3 `J_uvw` + 2×N `J_params`, row-major) + `has_img_from_cam_with_jac = true` for:

- **FOV**
- **Fisheye family**: SimpleRadialFisheye, RadialFisheye, OpenCVFisheye, ThinPrismFisheye, RadTanThinPrismFisheye, SimpleFisheye, Fisheye
- **Division models**: SimpleDivision, Division
- **EUCM**
- **Equirectangular**

The fisheye models share a few small `internal::` helpers in `models_jacobian.h`:
- `FisheyeProjectionWithJac` — the equidistant `atan` projection and its 2×2 Jacobian (with the `r→0` identity limit), Jacobian output is nullable so the value-only path stays cheap.
- `MatMul2x2` and `UvwJacFromAbJac` — compose per-stage Jacobians and apply the chain rule through `(a,b) = (u/w, v/w)`.

Each fisheye model then composes projection → distortion → linear via those; the division models share `DivisionScaleWithJac`; FOV mirrors its three piecewise distortion branches (`omega²<1e-4`, `radius²<1e-4`, else) with matching derivatives.

## Validation

- `sensor/models_jacobian_test.cc`: every analytic Jacobian validated against Ceres Jets to `1e-10` over a grid (gold standard).
- `estimators/cost_functions/reprojection_error_test.cc`: each cost function validated against numeric (Ridders) + autodiff, for both variable and constant pose (same `kJacEps=1e-4` + FOV-restriction rationale as #4513).

## Benchmark

`benchmark/runtime/cost_functions.cc` is extended to register all 18 models. Per-`Evaluate()` (residual + all Jacobians) speedups, autodiff → analytic, for the newly-covered models:

| Model | Full BA | Const-pose |
|---|---|---|
| FOV | 6.2× | 3.6× |
| SimpleRadialFisheye | 5.3× | 3.3× |
| RadialFisheye | 5.8× | 3.8× |
| OpenCVFisheye | 6.1× | 4.3× |
| ThinPrismFisheye | 7.6× | 5.2× |
| RadTanThinPrismFisheye | 8.4× | 6.0× |
| SimpleDivision | 7.8× | 4.6× |
| Division | 9.8× | 5.9× |
| SimpleFisheye | 6.2× | 2.7× |
| Fisheye | 6.0× | 3.8× |
| EUCM | 8.6× | 5.1× |
| Equirectangular | 4.4× | 3.0× |

Fisheye/rational models sit lower than the pinhole family (~11–13×) because `atan`/`sqrt`/rational terms dominate both the autodiff and analytic paths.

### DIFF
--- a/benchmark/runtime/cost_functions.cc
+++ b/benchmark/runtime/cost_functions.cc
@@ -59,6 +59,80 @@ std::vector<double> NominalCameraParams<FullOpenCVCameraModel>() {
           -0.0005,
           0.0001};
 }
+template <>
+std::vector<double> NominalCameraParams<FOVCameraModel>() {
+  return {1000, 1000, 320, 240, 0.5};
+}
+template <>
+std::vector<double> NominalCameraParams<SimpleRadialFisheyeCameraModel>() {
+  return {1000, 320, 240, 0.01};
+}
+template <>
+std::vector<double> NominalCameraParams<RadialFisheyeCameraModel>() {
+  return {1000, 320, 240, 0.01, 0.001};
+}
+template <>
+std::vector<double> NominalCameraParams<OpenCVFisheyeCameraModel>() {
+  return {1000, 1000, 320, 240, 0.01, 0.001, 0.0001, 0.0001};
+}
+template <>
+std::vector<double> NominalCameraParams<ThinPrismFisheyeCameraModel>() {
+  return {1000,
+          1000,
+          320,
+          240,
+          0.01,
+          0.001,
+          0.0001,
+          0.0001,
+          0.001,
+          0.0005,
+          0.0001,
+          0.0001};
+}
+template <>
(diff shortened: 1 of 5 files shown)


B1ueber2y (approved):
Thanks! LGTM. This should be correct as long as the CI passes.

PR #4521, opened 2026-07-08
Integrate 6pt shared focal solver to two view geometry estimation.

B1ueber2y:
19 single-camera ETH3D dslr scenes — uncalibrated (SIFT, quality=high)

Per-scene average AUC (%). Excludes the 6 multi-camera scenes (courtyard,
electro, facade, playground, terrace, terrains).

tested over five different seeds. Reported the mean and variance. 

### Incremental mapper
  | variant | AUC@0.5° | AUC@1° | AUC@5° | AUC@10° |
  |---|---|---|---|---|
  | base | 29.68 ± 1.21 | 49.59 ± 1.47 | 78.69 ± 1.82 | 83.98 ± 1.75 |
  | sf | 30.62 ± 0.43 | 51.25 ± 0.77 | 81.70 ± 1.60 | 87.08 ± 1.71 |
  | **Δ (sf−base)** | **+0.94 ± 1.45** | **+1.66 ± 1.80** | **+3.02 ± 2.61** | **+3.10 ± 2.64** |

  ### Global mapper
  | variant | AUC@0.5° | AUC@1° | AUC@5° | AUC@10° |
  |---|---|---|---|---|
  | base | 30.63 ± 0.69 | 50.48 ± 1.31 | 80.17 ± 2.42 | 85.55 ± 2.65 |
  | sf | 31.22 ± 1.21 | 51.16 ± 1.94 | 80.97 ± 2.94 | 86.43 ± 3.19 |
  | **Δ (sf−base)** | **+0.60 ± 0.77** | **+0.68 ± 1.34** | **+0.80 ± 2.35** | **+0.88 ± 2.68** |

cc: @vlarsson


src/colmap/estimators/cost_functions/tiny_sampson_error.h
```diff
@@ -87,4 +87,67 @@ class TinySampsonErrorCostFunctor {
...
+// points (u - cx, v - cy), not calibrated rays. The 8 ambient parameters are
+// parameterized directly here; the 6-DoF manifold (rotation on SO(3),
+// translation on the unit sphere, log-focal) is applied by the solver.
+class TinyFocalSampsonErrorCostFunctor {
+ public:
+  using Scalar = double;
+  enum {
+    NUM_RESIDUALS = Eigen::Dynamic,
+    NUM_PARAMETERS = 8,
+  };
```
ahojnnes: What's the benefit of an enum here vs. constexpr static?

src/colmap/estimators/cost_functions/tiny_sampson_error.h
```diff
@@ -87,4 +87,67 @@ class TinySampsonErrorCostFunctor {
...
+
+  int NumResiduals() const { return static_cast<int>(points1_.size()); }
+
+  template <typename T>
+  bool operator()(const T* const params, T* residuals) const {
+    using std::exp;
+    // Build E once from the pose, then scale it into the pixel-space
+    // fundamental matrix F = diag(inv_f, inv_f, 1) * E * diag(inv_f, inv_f, 1).
+    Eigen::Matrix<T, 3, 3> F = EssentialMatrixFromPoseParams(params);
+    const T inv_f = exp(-params[7]);
```
ahojnnes: std::exp

src/colmap/estimators/cost_functions/tiny_sampson_error.h
```diff
@@ -87,4 +87,67 @@ class TinySampsonErrorCostFunctor {
...
+    F.template topLeftCorner<2, 2>() *= inv_f * inv_f;
+    F.template block<2, 1>(0, 2) *= inv_f;
+    F.template block<1, 2>(2, 0) *= inv_f;
+    for (size_t i = 0; i < points1_.size(); ++i) {
+      const Eigen::Matrix<T, 3, 1> point1(
+          T(points1_[i].x()), T(points1_[i].y()), T(1));
+      const Eigen::Matrix<T, 3, 1> point2(
+          T(points2_[i].x()), T(points2_[i].y()), T(1));
```
ahojnnes: .cast<T>().homogeneous() ?

src/colmap/estimators/solvers/relpose_shared_focal.cc
```diff
@@ -0,0 +1,204 @@
...
+Eigen::Matrix3d FundamentalFromEssentialSharedFocal(const Eigen::Matrix3d& E,
+                                                    const double focal) {
+  const double inv_f = 1.0 / focal;
+  const Eigen::DiagonalMatrix<double, 3> K_inv(inv_f, inv_f, 1.0);
+  return K_inv * E * K_inv;
+}
+
+// Focal-calibrated, normalized camera rays (x / f, y / f, 1) from centered
+// image points.
+std::vector<Eigen::Vector3d> CalibratedRays(
```
ahojnnes, line 69: CamRaysFromImg?

src/colmap/estimators/solvers/relpose_shared_focal.cc
```diff
@@ -0,0 +1,204 @@
...
+
+void RelativePoseSharedFocalEstimator::Residuals(
+    const std::vector<X_t>& points1,
+    const std::vector<Y_t>& points2,
+    const M_t& model,
+    std::vector<double>* residuals) {
+  THROW_CHECK_EQ(points1.size(), points2.size());
+  THROW_CHECK_NOTNULL(residuals);
+  if (!(model.focal > 0.0)) {
+    residuals->assign(points1.size(), DBL_MAX);
```
ahojnnes: std::numeric_limits<double>::max()

ahojnnes (commented):
Looks great, initial pass, will look again tomorrow.

src/colmap/estimators/solvers/relpose_shared_focal.cc
B1ueber2y: This will collapse with CamRayFromImg and cause confusion. Keeping CalibratedRays : )

B1ueber2y:
19 single-camera ETH3D dslr scenes tested over five runs. 

incremental pipeline
| | AUC @ 0.5° | AUC @ 1.0° | AUC @ 5.0° | AUC @ 10° |
  |---|---|---|---|---|
  | **base** (5 runs) | 29.56 ± 1.24 | 49.52 ± 1.48 | 78.81 ± 1.84 | 84.10 ± 1.77 |
  | **this PR** (5 runs) | 30.25 ± 0.86 | 50.43 ± 1.16 | 79.95 ± 1.48 | 85.15 ± 1.49 |
  | **Δ (paired)** | +0.69 ± 0.57 | +0.91 ± 0.69 | +1.13 ± 0.89 | +1.06 ± 0.90 |

This PR is ready to go from my side.

src/colmap/estimators/two_view_geometry.cc
```diff
@@ -629,7 +673,12 @@ bool EstimateTwoViewGeometryPoseFromCamRays(
   // CALIBRATED branch below and never reach the CalibrationMatrix() calls.
   Rigid3d cam2_from_cam1;
   std::vector<int> valid_indices;
-  if (geometry->config == TwoViewGeometry::ConfigurationType::CALIBRATED) {
+  // The shared-focal configuration is treated like the calibrated one here: it
+  // carries a metric essential matrix, and the caller has already calibrated
```
sarlinpe: What's a *metric* essential matrix?

src/colmap/scene/two_view_geometry.h
```diff
@@ -49,6 +49,10 @@ struct TwoViewGeometry {
     CALIBRATED_RIG = 9,
     // Fundamental matrix.
     UNCALIBRATED = 3,
+    // Essential matrix with a shared, estimated focal length (both images from
+    // the same uncalibrated camera). Carries E, a fundamental matrix, a metric
+    // relative pose, and `shared_focal_length`.
```
sarlinpe: What's a *metric* relative pose?

src/colmap/scene/two_view_geometry.h
```diff
@@ -75,6 +79,12 @@ struct TwoViewGeometry {
   // Relative pose.
   std::optional<Rigid3d> cam2_from_cam1;
 
+  // Shared focal length, set only for the UNCALIBRATED_SHARED_FOCAL
+  // configuration, where it is estimated jointly with the relative pose. Its
+  // presence discriminates a shared-focal geometry from a plain uncalibrated
+  // one.
+  std::optional<double> shared_focal_length;
```
sarlinpe: Is the focal used in the subsequent view-graph calibration? is it usable by the global mapper as well? wouldn't it be easier to update the the camera after TVG (or in the VGC) rather than in the incremental mapper?

sarlinpe (commented):
Thank you, great work!

src/colmap/scene/two_view_geometry.h
B1ueber2y: Good catch. Sorry. Should have caught it myself.

src/colmap/scene/two_view_geometry.h
B1ueber2y: VGC already encodes same focal constraint in its cost functor, so I do not expect this constraint to improve a lot there. However, if we do it in TVG it can potentially improve the initialization. 

> wouldn't it be easier to update the the camera after TVG (or in the VGC) rather than in the incremental mapper?

Yes this can make sense. I purposely did not update it in TVG to keep it consistent as before, as we did not do it with F. The update is also not trivial in the sense that each image can be associated with multiple pairs. Lets leave it out of the scope of this PR

src/colmap/estimators/solvers/relpose_shared_focal.h
```diff
@@ -0,0 +1,114 @@
...
+                     M_t* model);
+
+  // Squared pixel-space Sampson error of each centered point pair under the
+  // fundamental matrix implied by the model (F = Kinv * E * Kinv).
+  static void Residuals(const std::vector<X_t>& points1,
+                        const std::vector<Y_t>& points2,
+                        const M_t& model,
+                        std::vector<double>* residuals);
+
+  // Score in [0, 1]: |b . (a1 x a2)| for unit optical axes a1 = e_z,
```
B1ueber2y, line 108: I invented this criteria myself with claude, as I did not come across any established criteria. It seems to work well in practice in both the tests and e2e benchmarks. Happy to switch if there are better ones. cc: @vlarsson

B1ueber2y: cc: @kocurvik

src/colmap/scene/two_view_geometry.h
sarlinpe: If each pair gives a focal estimate, could we pick the one with lowest covariance as initialization for the TVG? For this to influence the global mapper we'd need to use it in TVG, right?

B1ueber2y: It can already influences global mapper by providing better relative poses in the motion averaging if VGC is not on. If VGC is on, the focals will be anyway reestimated and this only gives a better initialization. 

I think updating focal lengths in TVG is interesting but out of scope of this PR, because it can be risky in the sense that we never did it in the previous TVG with F (though we could also do it with decomposition). I suggest we properly study this in a separate thread.

B1ueber2y:
numbers for global pipeline:

| | AUC @ 0.5° | AUC @ 1.0° | AUC @ 5.0° | AUC @ 10° |
  |---|---|---|---|---|
  | **base** (5 runs) | 30.64 ± 0.69 | 50.43 ± 1.31 | 80.01 ± 2.45 | 85.41 ± 2.66 |
  | **sf (this PR)** (5 runs) | 30.92 ± 1.06 | 51.02 ± 1.89 | 81.03 ± 2.95 | 86.52 ± 3.12 |
  | **Δ** (paired) | +0.28 ± 0.51 | +0.59 ± 0.93 | +1.02 ± 1.47 | +1.11 ± 1.53 |

src/colmap/estimators/solvers/relpose_shared_focal.h
kocurvik: I think that this criterion is not correct for the case of shared focal lengths. In that case the the coplanarity is a necessary, but not a sufficient condition for singularity. The singularity occurs if the principal axes are parallel, or if they intersect and the two camera centers are the same distance away from the point of intersection (e.g. the centers + intersection point form an isosceles triangle) - for more see p. 5 of [Stewenius et al. IVC 2008 paper](https://www.researchgate.net/profile/Frederik-Schaffalitzky/publication/4156344_A_minimal_solution_for_relative_pose_with_unknown_focal_length/links/0c96053581b9a6b2bb000000/A-minimal-solution-for-relative-pose-with-unknown-focal-length.pdf?__cf_chl_tk=_8P79J3_9F2KdGVQusur_h_MXsB_LcfMJeH2waJfoCA-1784057272-1.0.1.1-4dL1jhAFft2QmjbGq9zSTbQSh2_WS0TG6kG39kbWhSw).

The criterion as defined would fit the degeneracies for the more general case with two unknown focal lengths - this is the degeneracy the comments are actually recommending. This could use the estimator from my [PoseLib PR](https://github.com/PoseLib/PoseLib/pull/205). I like the design of the proposed scoring criterion since it is reasonably geometric (cosine of the angle) and easy to compute. For the shared focal case it could be modified to also measure how close the distances are with some ratio or something. This could be a two step thing where under some score for this condition you calculate the score for the distances to the intersection point.

I am not aware of any published criteria in this regard. I think there are other possible meaningful criteria which would not be so easy to compute, but could take into account also the point distribution, but I did not have to time validate any of them yet. We can discuss them in greater detail if you are interested, but maybe they would fall outside of the scope of this PR.

src/colmap/estimators/cost_functions/tiny_sampson_error.h
```diff
@@ -87,4 +85,62 @@ class TinySampsonErrorCostFunctor {
...
+// converted to the fundamental matrix implied by the focal so the Sampson error
+// is measured in *pixel* space:
+//
+//   F = diag(1/f, 1/f, 1) * E * diag(1/f, 1/f, 1),   f = exp(log_f).
+//
+// The inputs points1/points2 are therefore principal-point-centered image
+// points (u - cx, v - cy), not calibrated rays. The 8 ambient parameters are
+// parameterized directly here; the 6-DoF manifold (rotation on SO(3),
+// translation on the unit sphere, log-focal) is applied by the solver.
+class TinyFocalSampsonErrorCostFunctor {
```
ahojnnes, line 105: Should we have a unit test for this in tiny_sampson_error_test.cc ?

src/colmap/estimators/solvers/relpose_shared_focal.cc
```diff
@@ -0,0 +1,220 @@
...
+  // normalization PoseLib applies before running the solver.
+  double scale = 0.0;
+  for (size_t i = 0; i < points1.size(); ++i) {
+    scale += points1[i].norm() + points2[i].norm();
+  }
+  scale /= 2.0 * points1.size();
+  if (!(scale > 0.0)) {
+    return;
+  }
+  const double inv_scale = 1.0 / scale;
```
ahojnnes: nit: Extract as small helper in anonymous namespace?

src/colmap/estimators/solvers/relpose_shared_focal.cc
```diff
@@ -0,0 +1,220 @@
...
+
+  Eigen::Matrix<double, 8, 1> x;
+  x.head<4>() = cam2_from_cam1.rotation().normalized().coeffs();
+  x.segment<3>(4) = cam2_from_cam1.translation().normalized();
+  x[7] = std::log(model->focal);
+  solver.Solve(f, &x, options);
+
+  // Keep the refined estimate only if the solve stayed finite; otherwise fall
+  // back to the decomposed pose and seed focal.
+  if (x.allFinite()) {
```
ahojnnes: Should we check that translation is non-zero?

src/colmap/estimators/two_view_geometry.cc
```diff
@@ -923,6 +988,171 @@ TwoViewGeometry EstimateCalibratedTwoViewGeometry(
...
+                                      &inlier_cam_rays1,
+                                      &inlier_cam_rays2);
+      Rigid3d cam2_from_cam1;
+      std::vector<int> valid_indices;
+      PoseFromEssentialMatrix(*geometry.E,
+                              inlier_cam_rays1,
+                              inlier_cam_rays2,
+                              &cam2_from_cam1,
+                              &valid_indices);
+      constexpr double kMinFocalIdentifiability = 0.05;
```
ahojnnes: A brief note on how this threshold was chosen?

src/colmap/estimators/two_view_geometry.cc
```diff
@@ -426,12 +427,23 @@ TwoViewGeometry EstimateTwoViewGeometry(
...
+    // adjustment. Distortion is not a barrier here (as in the
+    // fundamental-matrix path, it is absorbed by the epipolar fit and later
+    // refined). Otherwise, use the calibrated path if both cameras have a known
+    // focal length, and the uncalibrated path otherwise.
     if (camera1.IsSpherical() || camera2.IsSpherical()) {
       return EstimateSphericalTwoViewGeometry(
           camera1, points1, camera2, points2, matches, options);
+    } else if (camera1.camera_id == camera2.camera_id &&
+               !camera1.has_prior_focal_length && camera1.IsPerspective() &&
+               !CameraModelIsFisheye(camera1.model_id)) {
```
ahojnnes: nit: add a Camera::IsFisheye method?

src/colmap/estimators/two_view_geometry.cc
```diff
@@ -614,6 +626,38 @@ void ExtractInlierCamRays(const Camera& camera1,
...
+  inlier_cam_rays2->resize(inlier_matches.size());
+  for (size_t i = 0; i < inlier_matches.size(); ++i) {
+    const FeatureMatch& match = inlier_matches[i];
+    const Eigen::Vector2d& point1 = points1[match.point2D_idx1];
+    const Eigen::Vector2d& point2 = points2[match.point2D_idx2];
+    (*inlier_cam_rays1)[i] =
+        Eigen::Vector3d((point1.x() - principal_point.x()) * inv_f,
+                        (point1.y() - principal_point.y()) * inv_f,
+                        1.0)
+            .normalized();
```
ahojnnes:
```suggestion
        (inv_f * (point1 - principal_point)).homogeneous().normalized();
```

src/colmap/estimators/two_view_geometry.cc
ahojnnes: nit: define a shared constant in RelativePoseSharedFocalEstimator::kMinFocalIdentifiability, since we use the same value everywhere or redefining the method to simply return a boolean?

src/colmap/scene/two_view_geometry.h
```diff
@@ -75,6 +80,12 @@ struct TwoViewGeometry {
...
```
ahojnnes: For future flexibility, would it make sense to store an optional camera1 and camera2, so we can also store distortion and other calibration parameters for other two-view solvers?

ahojnnes (commented):
Thanks, looks great.

src/colmap/scene/two_view_geometry.h
B1ueber2y: Good idea. Upon integrating this I realized we did not handle the new field in the database. This is fixed in the new commit.

src/colmap/sfm/incremental_mapper.cc
```diff
@@ -1431,8 +1439,20 @@ bool IncrementalMapper::EstimateInitialTwoViewGeometry(
...
+  const bool success = IncrementalMapperImpl::EstimateInitialTwoViewGeometry(
+      options,
+      *database_cache_,
+      image_id1,
+      image_id2,
+      cam2_from_cam1,
+      shared_focal_length);
+  if (success && shared_focal_length.has_value()) {
+    // Seed the shared camera's focal length, leaving it optimizable for BA.
+    reconstruction_->Image(image_id1).CameraPtr()->SetFocalLength(
```
ahojnnes: How about returning optional camera1 and camera2 here?

src/colmap/sfm/incremental_mapper_impl.cc
```diff
@@ -743,6 +765,12 @@ bool IncrementalMapperImpl::EstimateInitialTwoViewGeometry(
 
   cam2_from_cam1 = *two_view_geometry.cam2_from_cam1;
 
+  // Surface the shared focal length (if it was estimated) so the caller can
+  // seed the camera before registering the initial pair.
+  if (two_view_geometry.camera1.has_value()) {
```
ahojnnes: Suggest returning camera1 and camera2 instead.

src/pycolmap/scene/two_view_geometry.cc
```diff
@@ -24,6 +24,8 @@ void BindTwoViewGeometryScene(py::module& m) {
       .value("CALIBRATED", TwoViewGeometry::CALIBRATED)
       .value("CALIBRATED_RIG", TwoViewGeometry::CALIBRATED_RIG)
       .value("UNCALIBRATED", TwoViewGeometry::UNCALIBRATED)
+      .value("UNCALIBRATED_SHARED_FOCAL",
+             TwoViewGeometry::UNCALIBRATED_SHARED_FOCAL)
```
ahojnnes: This name is a bit confusing, imo. In other cases, when E has sufficient inliers relative the F, we call it CALIBRATED. My suggestion: Just reuse "CALIBRATED" in this case and then use the existence of the newly optional cameras to decide in downstream tasks whether the focal length is known or not?

B1ueber2y: Okay sounds good. I would use UNCALIBRATED though, as we anyway rely on recomposed F rather than E in this case for guided matching and VGC. The setup itself is also an uncalibrated setup in the sense that we do not have full information on the calibration.

B1ueber2y: I have carefully navigated this tricky update: https://github.com/colmap/colmap/pull/4521/commits/481921dc2abddb24b8842f4eee1c4278ed27b283. This is indeed much cleaner than before and the next step: adding one-side focal solver becomes straightforward and naturally included in this design. PTAL : )

ahojnnes:
Actually, my proposal was to use CALIBRATED if the shared focal length solver succeeds. We can compute E from decomposed F (based on shared focal length and centered principal point) in this case, no?

B1ueber2y:
> Actually, my proposal was to use CALIBRATED if the shared focal length solver succeeds. We can compute E from decomposed F (based on shared focal length and centered principal point) in this case, no?

Yes. However, in guided matching and VGC we are both using F for this case, so I think UNCALIBRATED is a better choice here, especially that we currently do not propagate the resulting cameras from this solver in two-view geometry estimation, which is strictly needed to use E for guided matching.

src/colmap/sfm/incremental_mapper_impl.cc
```diff
@@ -698,8 +713,22 @@ bool IncrementalMapperImpl::EstimateInitialTwoViewGeometry(
...
+  TwoViewGeometry two_view_geometry;
+  if (camera1.camera_id == camera2.camera_id &&
+      !camera1.has_prior_focal_length && camera1.IsPerspective() &&
+      !camera1.IsFisheye()) {
+    two_view_geometry = EstimateSharedFocalTwoViewGeometry(
+        camera1, points1, points2, matches, two_view_geometry_options);
+  } else {
+    two_view_geometry = EstimateCalibratedTwoViewGeometry(
+        camera1, points1, camera2, points2, matches, two_view_geometry_options);
+  }
```
ahojnnes: Wondering whether we should maybe just call EstimateTwoViewGeometry here? The logic for selecting which Estimate* function is the correct one is now duplicated here and there (and potentially not fully mirrored e.g. for the spherical cameras?).

src/colmap/sfm/incremental_mapper_impl.cc
```diff
@@ -195,8 +195,12 @@ bool IncrementalMapperImpl::FindInitialImagePair(
     FlatHashSet<image_pair_t>& init_image_pairs,
     image_t& image_id1,
     image_t& image_id2,
-    Rigid3d& cam2_from_cam1) {
+    Rigid3d& cam2_from_cam1,
+    std::optional<Camera>& estimated_camera1,
+    std::optional<Camera>& estimated_camera2) {
```
ahojnnes: nit: could potentially now expose the InitInfo struct and return it instead of the output args.

src/colmap/sfm/incremental_mapper.cc
```diff
@@ -1431,8 +1445,26 @@ bool IncrementalMapper::EstimateInitialTwoViewGeometry(
...
+                                                            camera1,
+                                                            camera2);
+  if (success) {
+    // Seed the estimated intrinsics, leaving them optimizable for BA.
+    if (camera1.has_value()) {
+      *reconstruction_->Image(image_id1).CameraPtr() = *camera1;
+    }
+    if (camera2.has_value()) {
+      *reconstruction_->Image(image_id2).CameraPtr() = *camera2;
+    }
```
ahojnnes: We are now setting these redundantly here and in FindInitialImagePair? We shouldn't set it in the other though?

src/colmap/scene/database_sqlite.cc
```diff
@@ -1385,6 +1461,20 @@ class SqliteDatabase : public Database {
       SQLITE3_CALL(sqlite3_bind_null(sql_stmt_write_two_view_geometry_, 9));
       SQLITE3_CALL(sqlite3_bind_null(sql_stmt_write_two_view_geometry_, 10));
     }
+
+    // Write NULL for camera1/camera2 if not set. The blob strings must outlive
+    // sqlite3_step because they are bound with SQLITE_STATIC.
+    const std::string
+        camera1_blob =  // NOLINT(bugprone-unused-local-non-trivial-variable)
```
ahojnnes, line 1468: could use [[maybe_unused]] instead?

src/colmap/scene/database_sqlite.cc
```diff
@@ -227,6 +229,68 @@ void WriteStaticMatrixBlob(sqlite3_stmt* sql_stmt,
...
+  camera.has_prior_focal_length = ReadBinaryLittleEndian<uint8_t>(&stream) != 0;
+  const uint64_t num_params = ReadBinaryLittleEndian<uint64_t>(&stream);
+  camera.params.resize(num_params);
+  ReadBinaryLittleEndian<double>(&stream, &camera.params);
+  return camera;
+}
+
+// Bind an optional camera to a statement column, writing NULL when unset. The
+// returned blob must outlive the sqlite3_step call (bound with SQLITE_STATIC).
+std::string BindOptionalCamera(sqlite3_stmt* sql_stmt,
```
ahojnnes: nit: WriteOptionalCamera for consistency.

ahojnnes (approved):
Great work, thank you very much!

src/colmap/scene/database_sqlite.cc
B1ueber2y: bugprone-unused-local-non-trivial-variable doesn't respect [[maybe_unused]] under our clang-tidy (it fails with -warnings-as-errors), so the // NOLINT has to stay.

See https://github.com/colmap/colmap/actions/runs/29507494444/job/87651959540

src/colmap/sfm/incremental_mapper.cc
B1ueber2y: This is not a double set, as the main logic was

```
if (!options_->IsInitialPairProvided()) {
    // (a) No pair specified → search for a good one.
    mapper.FindInitialImagePair(mapper_options, image_id1, image_id2, cam2_from_cam1);
  } else {
    // (b) User pinned init_image_id1 / init_image_id2 → just estimate that pair.
    mapper.EstimateInitialTwoViewGeometry(mapper_options, image_id1, image_id2, cam2_from_cam1);
  }
```

B1ueber2y: Nonetheless I extracted a helper to further consolidate.

src/colmap/sfm/incremental_mapper_impl.cc
B1ueber2y: Good point. This is updated. However, note that there is a behavior change here for two view geometry from different uncalibrated cameras, from 1.2 * max_dim + E to F. From the discussion thread here https://github.com/PoseLib/PoseLib/pull/205#issuecomment-4205402944 this is supposed to be an improvement.

src/colmap/sfm/incremental_mapper_impl.h
```diff
@@ -88,12 +106,15 @@ class IncrementalMapperImpl {
       const Reconstruction& reconstruction);
 
   // Implement IncrementalMapper::EstimateInitialTwoViewGeometry
+  // On success, returns true and populates `init_info` with the estimated
+  // two-view geometry; returns false if the pair is unsuitable for
+  // initialization.
   static bool EstimateInitialTwoViewGeometry(
```
ahojnnes: Final nit, feel free to ignore: Maybe return the std::optional<InitInfo> here and for other methods? In any case, please then just merge :-)

kocurvik (commented):
Sorry, forgot to submit my comments after writing them and now only realized this once it was merged.

Also what you won't be able to catch with this check is the degeneracy of coplanar points - or scenes with one dominant plane, but you generally only need very few off-plane points to recover some reasonably correct focals (unless the camera positions are also degenerate).

B1ueber2y:
> Sorry, forgot to submit my comments after writing them and now only realized this once it was merged.
> 
> Also what you won't be able to catch with this check is the degeneracy of coplanar points - or scenes with one dominant plane, but you generally only need very few off-plane points to recover some reasonably correct focals (unless the camera positions are also degenerate).

Thanks for the valuable feedback and references! This is vastly helpful and I will try to add a distance check (probably sometime next week) as a follow up. Until this new shared focal estimator, colmap does not have any estimator decomposing focal length natively, as it stopped at estimating F matrix before. So this is the only place that this degeneracy check is useful until now. I agree that a more structured check makes a lot of sense. 

Regarding the point distribution and coplanar-points degeneracy, I guess we can do something similar as DEGENSAC for this shared focal estimator. However, I consider this as low priority on my side for now. I will add a TODO along with the fixing of the degeneracy check : )

kocurvik:
> I guess we can do something similar as DEGENSAC for this shared focal estimator

We have implemented a [DEGENSAC version of the shared focal solver](https://link.springer.com/chapter/10.1007/978-3-642-22819-3_19) into PoseLib for our paper [Three-view Focal Length Recovery From Homographies](https://arxiv.org/pdf/2501.07499). You can see how it performs on synth data in Fig. 3 a) of the paper - fEf is the standard solver fEf + PP is DEGENSAC.  The problem is that the even the DEGENSAC version still needs to have some off-plane points to work well. In practice you either have enough off-plane points and then the difference is small (although DEGENSAC may terminate faster) or you do not have enough and then it won't really help you.

Anyways, if you need the solver that is used for DEGENSAC we have an implementation provided by Zuzana Kukelova [here](https://github.com/kocurvik/Poselib-hf/blob/main/PoseLib/solvers/plane_parallax_shared_focal.cc). It is used in [these calls](https://github.com/kocurvik/Poselib-hf/blob/main/PoseLib/robust/estimators/relative_pose.cc#L367) when degeneracy is detected. It is not very clean since for that implementation I did not want to modify the main PoseLib ransac implementation.

PR #4525, opened 2026-07-10
Replace Eigen random number generator with COLMAP PRNG for cross-platform determinism

### LINKED ISSUE #4431: reconstruction_pruning_test is fragile
We found on NixOS that the `reconstruction_pruning_test` fails on macOS. I think the real reason is that the test only coincidentally passes currently on the sequence produced by `SetPRNGSeed(1);` on Linux and the sequence generated by `Eigen::Vector3d::Random()` is different on mac (although still deterministic).

If I change it to `SetPRNGSeed(0xDEADBEEF)`, it fails also on Linux:

```
[==========] Running 4 tests from 1 test suite.
[----------] Global test environment set-up.
[----------] 4 tests from FindRedundantPoints3D
[ RUN      ] FindRedundantPoints3D.Empty
[       OK ] FindRedundantPoints3D.Empty (0 ms)
[ RUN      ] FindRedundantPoints3D.VaryingCoverageGain
/home/chpatrick/dev/colmap/src/colmap/scene/reconstruction_pruning_test.cc:65: Failure
Expected: (redundant_point3D_ids.size()) > (prev_num_redundant_points3D), actual: 0 vs 0

[  FAILED  ] FindRedundantPoints3D.VaryingCoverageGain (0 ms)
[ RUN      ] FindRedundantPoints3D.VaryingTrackLength
[       OK ] FindRedundantPoints3D.VaryingTrackLength (0 ms)
[ RUN      ] FindRedundantPoints3D.VaryingSpatialDistribution
[       OK ] FindRedundantPoints3D.VaryingSpatialDistribution (0 ms)
[----------] 4 tests from FindRedundantPoints3D (0 ms total)

[----------] Global test environment tear-down
[==========] 4 tests from 1 test suite ran. (1 ms total)
[  PASSED  ] 3 tests.
[  FAILED  ] 1 test, listed below:
[  FAILED  ] FindRedundantPoints3D.VaryingCoverageGain

 1 FAILED TEST
```

If it's fine that it only works on a specific sequence then I think Eigen random functions should be replaced with ones that use the COLMAP PRNG instead of `rand()` so it's the same on every platform. Or the test should be fixed so it works with any sequence.

behnamasadi:
## Problem

`SynthesizeDataset` (and `SynthesizeNoise`) generate random unit vectors with `Eigen::Vector3d::Random()`. Eigen's `Random()` draws from the C library `rand()`, **not** from COLMAP's seeded `PRNG`. As a result, `SetPRNGSeed(...)` does not make dataset synthesis reproducible across platforms.

This is the root cause of #4431: `reconstruction_pruning_test` only passes because of the specific `rand()` sequence produced on Linux with `SetPRNGSeed(1)`. It fails on macOS/NixOS, and even on Linux it fails under a different seed (e.g. `SetPRNGSeed(0xDEADBEEF)`), as reported in the issue.

## Fix

Add a small helper that draws each component from `RandomUniformReal<double>(-1, 1)` using the COLMAP `PRNG`, matching the `[-1, 1]` range of `Eigen::Vector3d::Random()`, and use it at the three call sites. Synthesized datasets are now deterministic for a given seed on every platform, which is what the tests already assume when they call `SetPRNGSeed`.

`RandomUniformReal<double>(...)` is already used elsewhere in this same file (e.g. for 2D point sampling), so this only makes the randomness source consistent.

## Testing

`reconstruction_pruning_test` (and the other synthetic-dataset-based tests) now depend only on the COLMAP PRNG, making them platform-independent for a fixed seed. Relying on CI to run the full test suite across platforms.

Fixes #4431


behnamasadi:
Verified locally in a container (Ubuntu 24.04, CUDA/GUI/CGAL disabled, apt-installed Ceres): the branch builds and `reconstruction_pruning_test` passes with all 4 cases green under the existing `SetPRNGSeed(1)`:

```
[==========] Running 4 tests from 1 test suite.
[ RUN      ] FindRedundantPoints3D.Empty
[       OK ] FindRedundantPoints3D.Empty
[ RUN      ] FindRedundantPoints3D.VaryingCoverageGain
[       OK ] FindRedundantPoints3D.VaryingCoverageGain
[ RUN      ] FindRedundantPoints3D.VaryingTrackLength
[       OK ] FindRedundantPoints3D.VaryingTrackLength
[ RUN      ] FindRedundantPoints3D.VaryingSpatialDistribution
[       OK ] FindRedundantPoints3D.VaryingSpatialDistribution
[  PASSED  ] 4 tests.
```

Since `SynthesizeDataset` now draws its random unit vectors from the COLMAP PRNG rather than `rand()`, the sequence is identical on every platform for a given seed, so this should also resolve the macOS/NixOS failures reported in #4431.

ahojnnes:
Thanks, I think the reason is not that it's not seeded as SetPRNGSeed() calls srand() but that the rng implementation differs across platforms. Your proposed fix still makes sense but the PR description is a bit misleading. Will review the changes in a bit.

B1ueber2y:
Thanks. This is a good improvement to have. Would be nice to have a thorough pass across the repo and propagate. I assume the usage of Eigen3d::Random() does not only limit to synthetic.cc.

ahojnnes (approved):
Thanks, I generalized this across the repo to replace Eigen::X::Random().

src/colmap/math/random_eigen.h
```diff
@@ -0,0 +1,147 @@
...
+  Eigen::VectorXd vector(size);
+  return internal::SetRandomEigen(vector);
+}
+
+inline Eigen::VectorXf RandomEigenVectorXf(const Eigen::Index size) {
+  Eigen::VectorXf vector(size);
+  return internal::SetRandomEigen(vector);
+}
+
+inline Eigen::Quaterniond EigenRandomQuaterniond() {
```
B1ueber2y: Thanks! Should we make it consistent with either RandomEigen or EigenRandom?

B1ueber2y (approved):

behnamasadi:
@ahojnnes, @B1ueber2y  Thank you for the follow-up and updating the code-base. I was too scared to make such a big change, cheers

B1ueber2y:
This seems to cause an error in the main branch CI. Is it concerning? @ahojnnes https://github.com/colmap/colmap/actions/runs/29492488322/job/87601685448

B1ueber2y:
Hopefully be fixed with https://github.com/colmap/colmap/pull/4545

Issue #4398, opened 2026-05-12
Quaternion format confusion wxyz <-> xyzw in Python bindings


michael-hoss:
**Describe the bug**

It looks like there are conflicting conventions for quaternions across the codebase and it seems to me that a proper conversion between them is missing. 

xyzw in the Python binding:
https://github.com/colmap/colmap/blob/6cfbc0404cb973ca79a6835bb88bb020ed96bc8f/src/pycolmap/geometry/eigen.cc#L21

wxyz in C++ with Eigen:
https://github.com/colmap/colmap/blob/6cfbc0404cb973ca79a6835bb88bb020ed96bc8f/src/colmap/scene/rig.cc#L284

**To Reproduce**

1. Save "rig_config.json": 

```json
[
  {
    "cameras": [
      {
        "image_prefix": "camera_from_json_config_1/",
        "ref_sensor": true
      },
      {
        "image_prefix": "camera_from_json_config_2/",
        "cam_from_rig_rotation": [
            0.7071067811865475,
            0.0,
            0.7071067811865476,
            0.0
        ],
        "cam_from_rig_translation": [
            0,
            0,
            0
        ]
      }
    ]
  }
]
```

2. Execute the following pytest:

```python
def test_obtain_rig_config_from_json_file_pass() -> None:
    rig_config: list[RigConfig] = pycolmap.read_rig_config("rig_config.json")
    expected_rotation_quat: np.ndarray = np.array([0.7071067811865475, 0.0, 0.7071067811865476, 0.0])
    assert np.allclose(rig_config[0].cameras[1].cam_from_rig.rotation.quat, expected_rotation_quat)
```

3. Obtain test failure about the ordering of quaternion entries :

```
AssertionError: assert False
9.055  +  where False = <function allclose at 0x73e3f0b19730>(array([0.        , 0.70710678, 0.        , 0.70710678]), array([0.70710678, 0.        , 0.70710678, 0.        ]))
9.055  +    where <function allclose at 0x73e3f0b19730> = np.allclose
9.055  +    and   array([0.        , 0.70710678, 0.        , 0.70710678]) = Rotation3d(xyzw=[0, 0.707107, 0, 0.707107]).quat
9.055  +      where Rotation3d(xyzw=[0, 0.707107, 0, 0.707107]) = Rigid3d(rotation_xyzw=[0, 0.707107, 0, 0.707107], translation=[0, 0, 0]).rotation
[...]

PR #4526, opened 2026-07-10
Document quaternion order (wxyz vs xyzw) in read_rig_config binding

behnamasadi:
## Problem

Reported in #4398. There are two intentional but different quaternion layouts in the codebase, and the Python side does not document the difference:

- The rig config JSON `cam_from_rig_rotation` is read in **`[w, x, y, z]`** order — see the `ReadRigConfig` header docs ("The rotation is expected in the order [w, x, y, z]") and `Eigen::Quaterniond(w, x, y, z)` in `rig.cc`.
- `pycolmap.Rotation3d.quat` / `Rigid3d.rotation.quat` expose the quaternion in **`[x, y, z, w]`** order (Eigen coefficient order), as their own docstrings state.

Because the `read_rig_config` binding docstring was just "Read the rig configuration from a .json file.", a Python user naturally assumes the JSON quaternion and `.quat` share the same layout and compares them directly, getting a surprising mismatch (the reporter's `assert np.allclose(...)` fails with the components swapped).

## Change

Documentation only — no behavior change. Spell out both conventions in the `read_rig_config` docstring so the difference is discoverable via `help(pycolmap.read_rig_config)`.

Addresses #4398


ahojnnes (approved):
Thanks!

PR #4527, opened 2026-07-10
Propagate max_num_iterations to Caspar BA backend (fixes #4382)

### LINKED ISSUE #4382: Local/Global BA mapper options not propagated to Caspar

behnamasadi:
## Problem

`IncrementalPipelineOptions::LocalBundleAdjustment()` and `GlobalBundleAdjustment()` build the Ceres backend options with the configured `max_num_iterations`, tolerances, thread count, loss function, etc. — but the parallel `if (options.caspar)` block only copied `gpu_index`:

```cpp
if (options.caspar) {
  options.caspar->gpu_index = ba_gpu_index;   // everything else dropped
}
```

So `ba_local_max_num_iterations` / `ba_global_max_num_iterations` were silently ignored when the Caspar backend was selected, leaving the Caspar solve effectively unbounded relative to the Ceres path. On larger scenes this manifested as a global BA running far longer than the equivalent Ceres BA (in one real case, >20 min with no per-iteration output, which a CI watchdog killed).

## Change

As discussed in the issue, Caspar solves the normal equations directly with its own dampening and convergence criteria, so most Ceres solver parameters do not map. The iteration bound is the sensible exception. This forwards the respective `max_num_iterations` to `CasparBundleAdjustmentOptions::solver_iter_max` in both builders.

Added a unit test (`IncrementalPipelineOptions.PropagatesMaxNumIterationsToCaspar`) asserting the propagation for both the local and global builders. `options.caspar` is always allocated (independent of the `CASPAR_ENABLED` build flag), so the test runs headless without the Caspar backend compiled in.

## Testing

Built with CUDA/GUI/CGAL disabled and ran the full `incremental_pipeline_test` suite — all 20 tests pass, including the new one and the existing end-to-end reconstruction tests that exercise the BA path.

Fixes #4382

### DIFF
--- a/src/colmap/controllers/incremental_pipeline.cc
+++ b/src/colmap/controllers/incremental_pipeline.cc
@@ -188,6 +188,11 @@ BundleAdjustmentOptions IncrementalPipelineOptions::LocalBundleAdjustment()
     options.ceres->gpu_index = ba_gpu_index;
   }
   if (options.caspar) {
+    // Caspar has different dampening and convergence criteria than Ceres, so
+    // most solver parameters are not meaningful to propagate. The maximum
+    // number of iterations is the exception: bounding it keeps the Caspar
+    // backend from running unboundedly long compared to the Ceres path.
+    options.caspar->solver_iter_max = ba_local_max_num_iterations;
     options.caspar->gpu_index = ba_gpu_index;
   }
   return options;
@@ -228,6 +233,9 @@ BundleAdjustmentOptions IncrementalPipelineOptions::GlobalBundleAdjustment()
     options.ceres->gpu_index = ba_gpu_index;
   }
   if (options.caspar) {
+    // See LocalBundleAdjustment(): only the iteration bound is propagated to
+    // the Caspar backend, whose other solver parameters do not map to Ceres.
+    options.caspar->solver_iter_max = ba_global_max_num_iterations;
     options.caspar->gpu_index = ba_gpu_index;
   }
   return options;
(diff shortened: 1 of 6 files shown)


ahojnnes:
I think we deliberately did not propagate these until further evaluation, because Caspar has different convergence behavior, if I remember correctly @tordnat ?

tordnat:
Yes, that's right. However, if it is possible to propagate the parameters without overwriting the solver-specific defaults that could be a possible solution.

ahojnnes:
This only works as long as the user didn't actually request for the default number of iterations explicitly. I guess the safer path is to change the default in the incremental pipeline options to "-1" and then set it to the equivalent ceres/caspar defaults.

behnamasadi:
Implemented as suggested: `ba_local/global_max_num_iterations` now default to -1, which resolves to the equivalent backend defaults (Ceres keeps the previous 25/50, Caspar keeps its own tuned `solver_iter_max`), while any explicitly set value >= 0 is propagated to the active backend. The quality presets now scale the effective value via new `EffBa*MaxNumIterations()` getters (same pattern as `EffMaxImageSize()`) — Ceres preset results are unchanged; note this means presets now also scale Caspar's default (e.g. low quality → 100), which seemed like the intended preset semantics but happy to leave Caspar untouched there if preferred.

ahojnnes (approved):
Thank you!

PR #4519, opened 2026-07-07
Handle directory output_path for stereo_fusion PLY output

alsolex:
## Summary

This PR makes `stereo_fusion` fail early when `output_type` is `PLY` and `output_path` is invalid.

Previously, passing a directory as `output_path` made COLMAP run stereo fusion first and then crash in the PLY writer.

## Changes

- Reject directory paths for PLY output before running stereo fusion.
- Validate the `.ply` extension and parent directory early.
- Preserve BIN/TXT directory output semantics, but validate invalid directories before running stereo fusion.

## Validation

- Reproduced the original crash with `--output_type PLY` and `--output_path` set to the dense workspace directory.
- Verified that the same invalid `output_path` now fails immediately with a clear error.
- Verified that an explicit `.ply` output path still writes the PLY and visibility files.

## Related issues

Fixes #3117.

PR #4530, opened 2026-07-13
Validate stereo fusion output path before running fusion

### LINKED ISSUE #3117: stereo_fusion crash if folder is used as output_path, rather than creating a file inside the directory
**Describe the bug**

Program argument inconsistency. The code uses --output_path as a file rather than creating a file in the directory like all the other pipeline steps.

See (it's also the same for the other output file formats)

https://github.com/colmap/colmap/blob/cf8f116197ff0e2ac869d04d3f41792ca18da449/src/colmap/util/ply.cc#L327

It would be better to check if the file can be created before the pipeline runs, before I go crazy losing hours of compute time.

**To Reproduce**

Set --output_path during stereo_fusion to a folder

**Expected behavior**

Colmap creates files in the folder, like all the other times I used --output_path. Please make this consistent.

**Screenshots**
```
I20250121 00:01:42.671082 240820 timer.cc:91] Elapsed time: 326.787 [minutes]
I20250121 00:01:54.919654 240820 mvs.cc:203] Writing output: ./dense/
E20250121 00:01:54.926281 240820 ply.cc:375] Check failed: (text_file).is_open() Could not open ./dense/. Is the path a directory or does the parent dir not exist?
terminate called after throwing an instance of 'std::invalid_argument'
  what():  [ply.cc:375] Check failed: (text_file).is_open() Could not open ./dense/. Is the path a directory or does the parent dir not exist?
*** Aborted at 1737446514 (unix time) try "date -d @1737446514" if you are using GNU date ***
PC: @                0x0 (unknown)
*** SIGABRT (@0x3e80003acb4) received by PID 240820 (TID 0x764a09369000) from PID 240820; stack trace: ***
    @     0x764a3032b206 (unknown)
    @     0x764a0a445320 (unknown)
    @     0x764a0a49eb1c pthread_kill
    @     0x764a0a44526e gsignal
    @     0x764a0a4288ff abort
    @     0x764a0a8a5ff5 (unknown)
    @     0x764a0a8bb0da (unknown)
    @     0x764a0a8a5a55 std::terminate()
    @     0x764a0a8bb391 __cxa_throw
    @     0x5ac8e3888c6c _ZN6colmap20WriteBinaryPlyPointsERKNSt7__cxx1112basic_stringIcSt11char_traitsIcESaIcEEERKSt6vectorINS_8PlyPointESaIS9_EEbb.cold
    @     0x5ac8e3946df2 colmap::RunStereoFuser()
    @     0x5ac8e38ef0d0 main
    @     0x764a0a42a1ca (unknown)
    @     0x764a0a42a28b __libc_start_main
    @     0x5ac8e38f5105 _start
Aborted (core dumped)
```

**Environment:**
 - OS: Ubuntu 24.04 LTS
 - COLMAP Version latest git
 - Capture Device A54 IMX778

behnamasadi:
RunStereoFuserImpl validated workspace_format, input_type and output_type up front, but never checked output_path until the results were written by WriteBinaryPlyPoints / Reconstruction::Write* after fusion completed. On large datasets fusion can run for hours, so a misspecified output_path (e.g. an existing directory when output_type is PLY, or a non-existent parent directory) aborted the whole run at the very end with a deferred THROW_CHECK_FILE_OPEN, discarding all the fusion work (issue #3117).

Add an early, output-type-aware validation of output_path in RunStereoFuserImpl before constructing the fuser:
  - PLY writes a single file, so reject an existing directory and require the parent directory to exist.
  - BIN/TXT write a directory of files, so require it to already exist (matching the THROW_CHECK_DIR_EXISTS in Reconstruction::Write*).

Also add a ply_test case asserting that WriteTextPlyPoints / WriteBinaryPlyPoints on a directory path fail cleanly, documenting the underlying deferred-failure behavior the early check guards against.

Fixes #3117

### DIFF
--- a/src/colmap/exe/mvs.cc
+++ b/src/colmap/exe/mvs.cc
@@ -352,6 +352,26 @@ Reconstruction RunStereoFuserImpl(const std::filesystem::path& output_path,
       << "Invalid `output_type` " << output_type
       << " - supported values are 'bin', 'ply' and 'txt'.";
 
+  // Validate the output path up front, before the expensive fusion, so that a
+  // misspecified path fails immediately rather than after fusion completes
+  // (see issue #3117). For 'bin'/'txt' the output is a directory of files; for
+  // 'ply' it is a single file.
+  if (output_type == "ply") {
+    THROW_CHECK(!ExistsDir(output_path))
+        << "Invalid `output_path` " << output_path
+        << " - for output_type 'PLY' this must be a file path, but an "
+           "existing directory was given.";
+    const std::filesystem::path output_parent_dir = GetParentDir(output_path);
+    THROW_CHECK(output_parent_dir.empty() || ExistsDir(output_parent_dir))
+        << "Invalid `output_path` " << output_path
+        << " - the parent directory " << output_parent_dir
+        << " does not exist.";
+  } else {
+    THROW_CHECK(ExistsDir(output_path))
+        << "Invalid `output_path` " << output_path << " - for output_type '"
+        << output_type << "' this must be an existing directory.";
+  }
+
   mvs::StereoFusion fuser(
       options, workspace_path, workspace_format, pmvs_option_name, input_type);
 
--- a/src/colmap/exe/mvs.cc
+++ b/src/colmap/exe/mvs.cc
@@ -353,6 +353,27 @@ Reconstruction RunStereoFuserImpl(const std::filesystem::path& output_path,
       << "Invalid `output_type` " << output_type
       << " - supported values are 'bin', 'ply' and 'txt'.";
 
+  // Validate the output path up front, before the expensive fusion, so that a
+  // misspecified path fails immediately rather than after fusion completes
+  // (see issue #3117). For 'bin'/'txt' the output is a directory of files; for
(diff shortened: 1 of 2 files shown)


behnamasadi:
Flagging an overlap I should have caught earlier: #4519 by @alsolex predates this PR by about a week and fixes the same issue (#3117) with essentially the same up-front validation. Two open PRs for one bug is probably part of why neither has moved, so to make the choice easy:

I've just added the `THROW_CHECK_HAS_FILE_EXTENSION(output_path, ".ply")` check from #4519, which mine was missing. This branch is now a strict superset of #4519:

- same up-front validation of `output_path` before fusion starts,
- plus the `.ply` extension check from #4519,
- plus a directory-vs-file distinction for the `bin`/`txt` output types,
- plus a regression test in `src/colmap/util/ply_test.cc` covering the directory-as-output-path case,
- plus error messages that name the offending option.

@alsolex — no ownership claim here, you got there first. I'm equally happy for #4519 to be merged and to send my test as a follow-up PR on top of it; whichever the maintainers prefer. @ahojnnes, either path closes #3117, so this just needs a pick.

PR #4533, opened 2026-07-13
Add CoreML execution provider for ONNX inference on macOS

ahojnnes:
## Motivation

COLMAP only registers the **CUDA** execution provider for ONNX models (`onnx_utils.cc`). On Apple platforms there is no CUDA, so all ONNX inference (ALIKED, LightGlue, and any future ONNX models) runs on the **CPU** regardless of `use_gpu` — there is currently no accelerated path on macOS.

This PR registers the **CoreML** execution provider on macOS, which offloads supported subgraphs to the GPU / Apple Neural Engine.

## Changes

- **cmake**: define `COLMAP_COREML_ENABLED` on macOS when onnxruntime is available (the prebuilt macOS arm64 binaries bundle the CoreML EP and `coreml_provider_factory.h`).
- **onnx_utils**: factor the common session setup into `ConfigureSessionOptions()`, then append the CoreML EP (ML Program format) when `use_gpu` is set — mapping `use_gpu` onto CoreML the same way it maps onto CUDA. Unsupported nodes fall back to CPU automatically, and if a model cannot be compiled by CoreML at all, the session is transparently rebuilt on the CPU (so this can never make a model fail to load).

Selected automatically whenever GPU use is requested (`use_gpu` defaults to `true`); set `use_gpu=false` to force pure CPU execution.

## Validation

- Builds and links on macOS arm64; existing ONNX tests (`aliked_test`) still pass.
- Verified end-to-end through the `ONNXModel` C++ class that `use_gpu=true` routes to CoreML and accelerates inference. On a **statically-shaped** LightGlue-style matcher (2048×2048 keypoints) the CoreML path is **~4× faster** than CPU (961 ms → 233 ms).

## Note on dynamic shapes

The current ALIKED/LightGlue models are exported with **dynamic** keypoint dimensions. CoreML cannot compile subgraphs with unbounded dimensions, so with these models most nodes fall back to CPU and the net speedup is currently marginal — the ~4× above requires static input shapes. This PR is the enabling backend change; realizing the speedup for the shipped models is a follow-up (export with static shapes / free-dimension overrides). The change is safe and beneficial regardless: it never regresses correctness and immediately benefits any statically-shaped model.

### DIFF
--- a/cmake/FindDependencies.cmake
+++ b/cmake/FindDependencies.cmake
@@ -490,6 +490,14 @@ endif()
 if(TARGET onnxruntime::onnxruntime)
     list(APPEND COLMAP_COMPILE_DEFINITIONS COLMAP_ONNX_ENABLED)
     message(STATUS "Enabling ONNX support")
+    # The prebuilt macOS onnxruntime binaries ship with the CoreML execution
+    # provider, which accelerates ONNX inference on the GPU / Apple Neural
+    # Engine. Enable it as the GPU backend on Apple platforms (CUDA is
+    # unavailable there).
+    if(IS_MACOS)
+        list(APPEND COLMAP_COMPILE_DEFINITIONS COLMAP_COREML_ENABLED)
+        message(STATUS "Enabling ONNX CoreML execution provider")
+    endif()
 endif()
 
 if(GUI_ENABLED)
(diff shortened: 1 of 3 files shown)


sarlinpe (commented):
Thank you!

sarlinpe (approved):

PR #4536, opened 2026-07-13
Modernize documentation: pydata theme + redesigned landing page

ahojnnes:
## Summary

Replaces the basic `sphinx_rtd_theme` docs with a modern, consistent design across the **entire site**, and rebuilds the landing page in the spirit of pytorch.org.

- **Sitewide theme swap** to `pydata-sphinx-theme` (top navbar, light/dark mode). Every page — install, tutorial, CLI, and the PyCOLMAP API pages — now shares the same modern chrome, colors, and dark mode.
- **Redesigned landing page** (`doc/index.rst`):
  - Hero section with tagline, description, and call-to-action buttons.
  - **Interactive install selector** (pytorch.org-style): pick OS / package / compute (CUDA vs CPU) and it composes the exact command or download link live, with a copy button. Covers pip, conda, docker, binary, homebrew, vcpkg, and source; invalid combinations (e.g. CUDA on macOS, CPU-only Windows/macOS wheels) are disabled.
  - Feature card grid (via `sphinx-design`) linking to the main sections.
- **Branding**: generated SVG logo (light/dark variants) and favicon (none existed before).

Autodoc API pages, the hidden `toctree` navigation, and the `pycolmap._core` → `pycolmap` remapping are all preserved.

## Dependencies

`doc/requirements.txt` now uses `pydata-sphinx-theme==0.20.0` and `sphinx-design==0.7.0` in place of `sphinx-rtd-theme` (both verified compatible with the pinned `sphinx==9.1.0`).

## Verification

Built locally with the documented flow (`pip install -r doc/requirements.txt && make -C doc html`):

- Clean build — no new Sphinx warnings or errors.
- All pages render with the new theme (no residual RTD markup); the PyCOLMAP API page builds with 1000+ signatures and zero `pycolmap._core` leaks.
- Install-selector matrix/normalization logic validated (OS→method→compute transitions, disabled cells, command/link outputs).
- Logo and favicon load in both light and dark mode; sphinx-design assets included.

Interactive/visual behavior (selector clicks, copy button, dark-mode toggle) should be spot-checked in a browser from the built `doc/_build/html/index.html`.

## Notes

- Independent of the docs-deploy PR (#4535); no overlapping files.
- This changes only the documentation; no library/source code is affected.

PR #4539, opened 2026-07-14
Docs SEO: canonical URLs, sitemap, Open Graph, faster images

ahojnnes:
Branches off `docs-modern-landing` (PR #4536). Implements the SEO improvements discussed.

## Changes
- **Canonical URLs** — set `html_baseurl = https://colmap.github.io/`, so every page emits `<link rel="canonical">`. Important because the hosted `legacy/<version>/` copies are near-duplicate content.
- **Sitemap** — added `sphinx-sitemap`; generates `sitemap.xml` (24 URLs) at the site root.
- **robots.txt** — added via `html_extra_path`, allows crawling and points to the sitemap.
- **Open Graph / Twitter cards** — added `sphinxext-opengraph`: `og:*`/`twitter:card` tags, per-page `<meta name="description">`, and a raster social-card image (`_static/og-image.png`, 1200×630). The landing page gets an explicit keyword-rich description (the hero text is raw HTML, so auto-extraction would otherwise grab a figure caption).
- **Descriptive homepage `<title>`** — "Structure-from-Motion and Multi-View Stereo — COLMAP" (via a small `htmltitle` template override); the visible hero H1 stays "COLMAP". Subpage titles unchanged.
- **Faster images (Core Web Vitals / LCP)** — converted the hero (`sparse`) and tutorial (`incremental-sfm`) screenshots from PNG to WebP: 1054 KB → 322 KB and 129 KB → 36 KB (~70% smaller). Removed the now-unused PNGs; `dense.png` was already unreferenced and isn't shipped.

New deps (verified against the pinned `sphinx==9.1.0`): `sphinx-sitemap==2.9.0`, `sphinxext-opengraph==0.13.0`.

## Verification (local build, clean)
- `sitemap.xml` present with correct absolute URLs; `robots.txt` at root referencing it.
- `<link rel="canonical">`, `og:title/description/image`, `twitter:card`, and `<meta name="description">` all present and correct on the landing page.
- WebP images copied and referenced; homepage `<title>` descriptive; hero H1 intact; subpage titles unaffected.

## Off-code follow-ups (need account access)
- Verify the site in **Google Search Console** and submit the sitemap. I can add the verification `<meta>` tag if you provide a token.


doc/conf.py
```diff
@@ -164,6 +166,32 @@ def get_git_revision():
 # Follow the visitor's OS/browser preference for light/dark mode by default.
 html_context = {"default_mode": "auto"}
 
+# -- SEO -----------------------------------------------------------------
+
+# Canonical site URL. Enables <link rel="canonical"> on every page (important
```
B1ueber2y: I think this comment is misleading. The canonical site URL would only work if we update the legacy html to link it.

doc/index.rst
```diff
@@ -1,4 +1,8 @@
 :html_theme.sidebar_secondary.remove: true
+:og:description: COLMAP is a free, open-source Structure-from-Motion (SfM) and Multi-View Stereo (MVS) pipeline with a graphical and command-line interface for 3D reconstruction from ordered and unordered image collections.
+
+.. meta::
```
B1ueber2y, line 4: I think this serves the same purpose as og:description and is directly droppable.

B1ueber2y (approved):

doc/conf.py
ahojnnes: Good catch — reworded to drop the overclaim. `html_baseurl` only emits a self-referential `<link rel="canonical">` on the current build and sets the base URL for the sitemap/OG tags; it does not dedupe the hosted `legacy/<version>/` copies (those would each need their own HTML updated to point here). Fixed in 0a29ca02d.

doc/index.rst
ahojnnes: I looked into this and it turns out they are not interchangeable, so I kept it. `sphinxext-opengraph` populates the two tags from independent sources and does not cross-fill: `:og:description:` drives `og:description`, while the `.. meta:: :description:` directive drives the standard `<meta name="description">`. I verified with a local build of the landing page:

| Source present | `og:description` | `<meta name="description">` |
|---|---|---|
| Both (current) | correct | correct |
| Drop `.. meta::` | correct | **garbage** — auto-extracted: *"Sparse reconstruction of central Rome… Install COLMAP: Select your platform…"* |
| Drop `:og:description:` | **garbage** | correct |

Because the hero is raw HTML, ogp's auto-extraction for `<meta name="description">` grabs the figure caption + install blurb, which is exactly why the explicit `.. meta::` block is there. `<meta name="description">` (Google snippet) and `og:description` (social cards) also serve different consumers. Happy to drop it if you feel the search-snippet description is not worth the duplicated string, but flagging the regression first.

B1ueber2y: I see. Thanks for the clarification. Let’s keep both then.

PR #4540, opened 2026-07-14
Docs: open external links in a new tab + restore build commit stamp

ahojnnes:
Addresses two pieces of landing-page review feedback from Shaohui Liu.

## 1. Open external links in a new tab
- Adds `doc/_static/external_links.js`, registered via `html_js_files`. On page load it gives every anchor pointing to a **different host** (GitHub repo, release downloads, `demuc.de`, contributor profiles, etc.) `target="_blank"` plus `rel="noopener noreferrer"`.
- **Same-site links are left untouched** — including absolute `https://colmap.github.io/` URLs (e.g. the "build from source" link) and all relative/in-page links.
- Covers both reStructuredText links and raw-HTML anchors (the hero **GitHub** button is raw HTML, which a build-time link-rewriting extension would miss — hence JS, which also avoids a new pinned dependency).

## 2. Restore the build commit stamp
The old sidebar surfaced the project version (with git hash + date); the pydata theme does not, so the stamp survived only in an invisible `<meta name="docsearch:version">` tag.
- Adds a footer `version` component (`doc/_templates/version.html`) shown in `footer_center`, rendering `COLMAP <release> | <hash> (<date>)` on every page. The value comes from the existing `get_git_revision()` in `conf.py`.

## Verification
- Local `sphinx-build` is clean (no new warnings/errors).
- `external_links.js` copied to `_static/` and referenced from `index.html`; host-comparison logic verified against the real landing-page links (external → new tab; `colmap.github.io`/relative → unchanged).
- Footer stamp renders visibly on both the landing page and subpages: `COLMAP 4.2.0.dev0 | <hash> (<date>)`.

> Stacked on #4539 (base branch `docs-seo`); retarget to `main` once that merges.


doc/conf.py
```diff
@@ -216,7 +217,7 @@ def get_git_revision():
 # so a file named "default.css" will overwrite the builtin "default.css".
 html_static_path = ["_static"]
 html_css_files = ["custom.css", "landing.css"]
-html_js_files = ["install_selector.js"]
+html_js_files = ["install_selector.js", "external_links.js"]
```
B1ueber2y, line 220: From claude:

This one-shot pass misses the install selector's links, which the description lists as covered ("release downloads"). install_selector.js creates its result anchor with createElement in render(), and re-renders on every click — and the default selection (Linux/pip/CUDA) is a command, so at DOMContentLoaded the selector has no anchors at all. So "Download from GitHub Releases" (Windows/Binary) and the Repology link (Linux/Binary) never get target="_blank". Could pull the scan into a function and re-run it from a MutationObserver on document.body, or just set target/rel in install_selector.js where the anchor is built.

B1ueber2y (approved):

doc/conf.py
ahojnnes: Good catch. Fixed in a13a9077e: the scan is now a reusable function re-run from a MutationObserver on document.body, so the install selector's download anchors (built in render() and rebuilt on each click) also get target="_blank" + rel="noopener noreferrer".

PR #4542, opened 2026-07-15
Add optional FAISS GPU SIFT matching

POBIM:
## Summary

This PR adds an opt-in FAISS GPU backend for SIFT descriptor matching. The
normal matching path uses `GpuIndexFlatL2`, so nearest-neighbor search remains
exact. Guided matching first asks the same index for a bounded number of
descriptor candidates and then applies COLMAP's epipolar or homography filter,
avoiding the dense feature-by-feature distance matrix.

The existing matcher remains the default. FAISS GPU code is compiled only with
`-DFAISS_GPU_MATCHING_ENABLED=ON`, and it is selected at runtime with
`--SiftMatching.faiss_gpu_matcher=1`.

## Motivation

This started from a practical scaling limit: high-resolution images produced
roughly 70k-125k SIFT descriptors per image, while the existing GPU matcher had
to be clamped to 32,768 descriptors or failed during matcher creation. The goal
was not to make an approximate matcher look faster; it was first to replace the
quadratic-memory matching step with an exact search whose index and result
buffers scale linearly with the descriptor count.

That changes the useful trade-off. More descriptors can be retained without
exhausting an 8 GiB GPU, at the cost of additional search time. The backend is
therefore explicit and experimental rather than a new default.

## Implementation

- Enables the GPU portion of the existing FAISS 1.14.1 dependency only when
  `FAISS_GPU_MATCHING_ENABLED` is set.
- Adds an exact `GpuIndexFlatL2` SIFT matcher with an LRU cache of per-image
  descriptor indices.
- Keeps ordinary matching exact with two returned neighbors for the ratio test.
- Makes guided matching memory-bounded with a configurable top-K candidate set
  (default 32), followed by essential/fundamental/homography filtering and the
  existing ratio/cross-check logic.
- Exposes cache size, temporary GPU workspace, and guided candidate count in
  the CLI and pycolmap options.
- Supports COLMAP's existing `gpu_index=-1` best-device selection and the
  one-GPU-per-matcher-worker model.
- Adds option-validation, exact matching, guided essential-matrix, CLI, and
  pycolmap binding coverage.

IVF is deliberately not included. In the controlled experiment it was slower
than Flat exact (72.60 s vs 18.55 s) and retained fewer verified matches, so it
would broaden the API without supporting the main result.

## Results

Hardware: NVIDIA RTX 4060 Laptop GPU (8,188 MiB), Linux, one FAISS matcher
worker unless otherwise noted.

### Equal-input comparison

The same database, 132 images, 528 image pairs, 2,136,348 descriptors, matching
thresholds, and geometric verification were used on both sides.

| Metric | Existing GPU matcher | FAISS GPU Flat exact | Change |
|---|---:|---:|---:|
| Matching wall time | 11.65 s | 18.55 s | +6.91 s |
| Incremental peak VRAM | 2,353 MiB | 506 MiB | -78.50% |
| Raw matches | 1,446,576 | 1,437,716 | 99.39% retained |
| Verified matches | 1,415,246 | 1,407,281 | 99.44% retained |
| Non-empty verified pairs | 522 | 522 | unchanged |

The correspondence-level Jaccard score was 97.92% for raw matches and 97.69%
after geometric verification. The difference comes from backend numeric and
tie behavior; the FAISS index itself is not approximate.

### High-feature stress pair

The two images contained 121,600 and 124,579 descriptors.

| Matcher | Descriptors used | Time | Incremental VRAM | Verified matches |
|---|---:|---:|---:|---:|
| Existing GPU matcher | 32,768 per image | 0.75 s | 6,249 MiB | 852 |
| Existing GPU matcher | full | failed to create (out of memory) | - | 0 |
| FAISS GPU Flat | full | 2.41 s | 970 MiB | 2,391 |

### End-to-end reconstruction

A fixed 2,830-pair graph was used for 256 images. The baseline retained 16k
descriptors per image; the FAISS run used 23,288,831 descriptors in total
(65,777-137,801 per image).

| Metric | 16k baseline | Full-feature FAISS GPU |
|---|---:|---:|
| Registered images | 256 / 256 | 256 / 256 |
| Sparse points | 225,849 | 282,315 |
| Observations | 891,162 | 1,140,102 |
| Mean reprojection error | 1.2003 px | 0.9570 px |
| Fixed-pair matching time | 76.44 s | 2,726.95 s |
| Overall peak VRAM | 3,573 MiB | 2,353 MiB |

This is intentionally not presented as a speed win: full exact matching was
35.67x slower because it searched 5.69x as many descriptors. It demonstrates
that the previous memory ceiling can be crossed while still completing a full
reconstruction. The baseline used two GPU matcher workers while the FAISS run
used one, so the absolute peak-VRAM comparison also includes that topology
difference.

The detailed benchmark protocol, controlled-pair results, stress artifacts,
and reproduction commands are recorded here:
[FAISS-GPU baseline proof](https://github.com/POBIM/ProBimSplatting/blob/experiment/ann-vram-20260713/docs/FAISS_GPU_BASELINE_PROOF.md).

## Guided matching semantics

Ordinary FAISS matching is exact. The guided path is intentionally bounded and
therefore approximate with respect to COLMAP's dense guided matcher: geometry
is tested only among the nearest `faiss_gpu_guided_num_neighbors` descriptor
candidates. With K=32 in a separate 133-image, 1,378-pair integration run using
up to 95k features per image, matching completed in 474.8 s at 3,157 MiB peak
VRAM; the mapper registered 133/133 images and produced 360,416 sparse points
at 0.9778 px mean reprojection error.

The candidate count is exposed because the appropriate speed/recall point can
vary by dataset. The default is based on the tested integration run, not a claim
that K=32 is universally optimal.

## Validation

- Ported and built against `colmap/colmap` main at `328d6838` (4.2.0.dev0).
- Built `colmap_main` with GCC 13, CUDA 12.6, and FAISS 1.14.1.
- Full `sift_test`: 24/24 passed on the RTX 4060, including the new exact and
  guided FAISS GPU tests.
- `option_manager_test`: 14/14 passed.
- Verified all new options in `colmap exhaustive_matcher -h`.
- Formatted with the repository-pinned clang-format 22.1.5 and Ruff 0.15.20;
  `git diff --check` passes.

## Build and use

```bash
cmake -S . -B build \
  -DCUDA_ENABLED=ON \
  -DFAISS_GPU_MATCHING_ENABLED=ON
cmake --build build

build/src/colmap/exe/colmap exhaustive_matcher \
  --database_path path/to/database.db \
  --FeatureMatching.use_gpu=1 \
  --SiftMatching.faiss_gpu_matcher=1
```

For guided matching, also pass `--FeatureMatching.guided_matching=1`. The
default cache is 16 image indices, the temporary FAISS workspace is 256 MiB,
and guided matching evaluates 32 candidates per query.


ahojnnes:
Thanks for the contribution. I am a bit hesitant to merge this tbh. I am not sure there is a practical setting where so many features per image are a good choice. It'll usually lead to mismatches and/or dropping matches due to accidental triggering of the ratio test. Can you share a bit more about your use case?

PR #4520, opened 2026-07-08
Add DEGENSAC for fundamental matrix estimation robust to a dominant plane

ahojnnes:
## Summary

Implements **DEGENSAC** (Chum, Werner, Matas, *"Two-View Geometry Estimation Unaffected by a Dominant Plane"*, CVPR 2005) as an **opt-in** fundamental-matrix estimator that is robust to a dominant scene plane.

When the scene contains a dominant plane, a random minimal sample frequently contains a majority of coplanar correspondences. The fundamental matrix from such an *H-degenerate* sample is compatible with the plane homography but its epipole is essentially unconstrained — yet it still accrues high inlier support (all plane points fit it) and makes RANSAC terminate early on a wrong epipolar geometry. COLMAP previously only detected this *after the fact* (H/F inlier-ratio → `PLANAR_OR_PANORAMIC`) and discarded `F` rather than recovering it.

## Design

DEGENSAC is a minimal/non-minimal solver injected into the existing LO-RANSAC. It receives a sample and holds pointers to the full correspondence set, so inside `Estimate()` it can:

1. Fit `F` (7-point for a minimal sample, 8-point otherwise).
2. Detect H-degeneracy: build the plane homography compatible with `F` from triplets (all C(7,3) for a minimal sample, randomly sampled triplets otherwise) and check the coplanar fraction, using a **plane-consistency threshold** (default `sqrt(3)·max_error`) that is deliberately looser than the inlier threshold.
3. **Complete** a degenerate sample via plane-and-parallax: refit the plane homography on its plane inliers, recover the epipole from **clean, high-parallax off-plane** correspondences (those beyond an off-plane margin, default `10·max_error`, so near-plane low-parallax points don't destabilize the epipole), form `F = [e2]ₓH`, and finally **refine `F` with the 8-point algorithm on mixed samples of plane + off-plane inliers** (so the epipole is constrained by several off-plane points, not just two). If no usable off-plane parallax exists, the sample is dropped.

The DEGENSAC estimator is used as **both the hypothesis and the local-optimization solver** (`LORANSAC<DegensacEstimator, DegensacEstimator>`), so LO cannot re-fit a plane-corrupted model on the plane-dominated inlier set.

[...]

PR #4543, opened 2026-07-15
Enable DEGENSAC by default for two-view geometry estimation

ahojnnes:
## Summary

Flips the default of `TwoViewGeometryOptions::use_degensac` from `false` to `true`, making DEGENSAC (Chum et al., CVPR 2005) the default fundamental matrix estimator instead of plain LO-RANSAC.

DEGENSAC makes fundamental matrix estimation robust to a dominant scene plane, a common degenerate configuration. It was introduced in #4520 as an opt-in option; this change enables it by default.

## Notes

- The pycolmap binding (`use_degensac`) inherits this default via `def_readwrite`, so the CLI and Python interfaces pick it up automatically.
- Comment updated to reflect the new default ("On by default").

## Test plan

- Existing `two_view_geometry_test.cc` and `fundamental_matrix_degensac_test.cc` cover the DEGENSAC path.

### DIFF
--- a/src/colmap/estimators/two_view_geometry.h
+++ b/src/colmap/estimators/two_view_geometry.h
@@ -99,9 +99,9 @@ struct TwoViewGeometryOptions {
   bool force_H_use = false;
 
   // Use DEGENSAC (Chum et al., CVPR 2005) for the fundamental matrix instead of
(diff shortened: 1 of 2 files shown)


sarlinpe (approved):
Thank you!

PR #4544, opened 2026-07-15
Add configurable global descriptor image retrieval (MixVPR, MegaLoc) replacing / complementing vocabulary tree

Vincentqyw:
## Summary

Add deep-learning-based global descriptor image retrieval as a modern alternative to the classic vocabulary tree + Hamming embedding. Each image is represented by a single compact descriptor (e.g. 4096-dim for MixVPR, 8448-dim for MegaLoc), and nearest-neighbor search is performed via FAISS `IndexFlatIP` (cosine similarity on L2-normalized vectors).

Models are defined in a **registry** (`GlobalDescriptorModel`) — adding a new model requires only a one-struct change, no GUI or pipeline code modifications.

## Key features

- **Generic model abstraction** (`GlobalDescriptorModel`): config struct with input size, normalization, ONNX I/O shapes, default download URI. Ships with MixVPR (320×320, 4096d) and MegaLoc (518×518, 8448d, DINOv2).
- **Auto-download from HuggingFace**: models automatically downloaded and cached (SHA256 verified) when `model_path` is empty.
- **Descriptor caching**: extracted descriptors serialized to disk next to the database, avoiding redundant ONNX inference on subsequent runs.
- **CoreML execution provider**: ONNX inference on macOS uses CoreML EP for ANE/GPU acceleration, with transparent CPU fallback. Reference: [colmap/colmap@2fe2b41](https://github.com/colmap/colmap/commit/2fe2b41158fa5932deab0b9ac2492f12af5c386e)
- **Sequential loop detection**: `SequentialPairGenerator` supports global descriptors (MixVPR or MegaLoc) for loop closure detection, alongside the vocabulary tree option.
- **Dedicated GlobalDescriptor matching tab** in the Feature Matching dialog, parallel to VocabTree.
- **Full backwards compatibility**: vocabulary tree path is untouched; global descriptors are opt-in via `model_path`/`model_type` fields.

## Screenshots

**Sequential Matching — Loop detection type selector (MixVPR / MegaLoc / Vocabulary Tree / None):**

Sequential tab

**Feature Matching — New GlobalDescriptor tab (model type, num_images, batch_size, model path):**

GlobalDescriptor tab

## Files changed

| Area | Files |
|------|-------|
| Retrieval core | `global_descriptor_index.h/cc`, `global_descriptor_model.h/cc`, `resources.h`, `CMakeLists.txt` |
| Pairing / pipeline | `pairing.h/cc`, `automatic_reconstruction.h/cc`, `feature_matching.h/cc`, `CMakeLists.txt` |
| Option management | `option_manager.h/cc` |
| ONNX / CoreML | `onnx_utils.h/cc`, `FindDependencies.cmake` |
| CLI | `feature.h/cc`, `sfm.cc`, `colmap.cc` |
| GUI | `automatic_reconstruction_widget.h/cc`, `feature_matching_widget.cc` |

## Usage

```bash
# Standalone global descriptor matching (model auto-downloaded)
colmap global_descriptor_matcher \
  --database_path /path/to/db \
  --image_path /path/to/images

# Sequential matching with global descriptor loop detection
colmap sequential_matcher \
  --database_path /path/to/db \
  --image_path /path/to/images \
  --SequentialMatching.loop_detection true

# Automatic reconstructor with global descriptors
colmap automatic_reconstructor \
  --workspace_path /path/to/workspace \
  --image_path /path/to/images \
  --global_descriptor_path /path/to/model.onnx  # omit for auto-download
```

## Adding a new model

One entry in `src/colmap/retrieval/global_descriptor_model.cc`:

```cpp
{ .name = "NewModel",
  .input_width = 512, .input_height = 512,
  .mean = {0.5f, 0.5f, 0.5f}, .std = {0.25f, 0.25f, 0.25f},
  .expected_input_shape = {-1, 3, 512, 512},
  .expected_output_shape = {-1, 2048},
  .descriptor_dim = 2048,
  .supports_batching = true,
  .default_model_uri = "https://.../model.onnx;filename;sha256",
},
```

## Architecture

```
Image files → Bitmap::Read() → Rescale → Normalize (per model config)
  → ONNXModel (CoreML-accelerated on macOS)
  → L2-normalized descriptor vector
  → GlobalDescriptorIndex::Add()
  → Prepare() [FAISS IndexFlatIP]
  → Query() [cosine similarity]
  → Top-K image pairs → FeatureMatcherController → SfM
```


ahojnnes:
Thanks very much @Vincentqyw for the new contribution. It will definitely be great to have support for this. Some high-level questions/comments before I will dive deeper into reviewing this in the next days:
* I wonder whether we shouldn't find a common abstraction for the newly added global descriptor methods and the vocabulary tree. Conceptually, they perform the same operation in that they efficiently find similar images for matching. If we find a common abstraction, then the vocabulary tree becomes just another implementation of a global descriptor (which Bag-of-Words technically is) and we can transparently inject them into various different consumers instead of if/else branching.
* Is your initial selection of global descriptor methods backed by some benchmarking in terms of precision/recall/runtime? How fast or slow is it to extract the descriptor for one image?

Vincentqyw:
@ahojnnes Thanks for the thoughtful feedback! Let me address both points.

## On abstraction

Conceptually, I agree — BoW is technically a type of global descriptor, and a unified interface would be cleaner. Two practical concerns though:

1. **Input data differs.** `VisualIndex` needs local features (already in DB); `GlobalDescriptorIndex` needs raw image pixels → ONNX. A shared `Add()` signature would be leaky.
2. **Lifecycle differs.** Vocab tree needs a separate `Build()` phase; global descriptors go straight to FAISS inference.

That said, a full index-level abstraction may be overkill for 2 implementations. What about a lighter approach:

- Add a `RetrievalPairingOptions` struct with an enum (`kVocabTree` / `kGlobalDescriptor`) and a factory method `CreatePairGenerator()`.
- Consumers see a single `retrieval_matcher` instead of two separate matchers + if/else.
- The existing `VocabTreePairGenerator` / `GlobalDescriptorPairGenerator` internals stay untouched.

What scope do you have in mind — full Build/Index/Query lifecycle, or is a consumer-level unification sufficient?

## On benchmarking

Fair point. I'll prepare a table with precision/recall and per-image extraction times for MixVPR and MegaLoc and share it here. In my experience, MixVPR is ~5–10ms/image on a consumer GPU, but I'll get concrete numbers.

sarlinpe:
Thank you for this great contribution! I agree with Johannes that a tighter integration with the vocab tree matcher is preferred. I'm mostly concerned with minimizing the complexity for the user, so generalizing the existing vocab-tree interfaces (pairing, matching, etc.) to a more generic retrieval interface (with specializations vocab tree vs global descriptors) would be sufficient. What you suggest makes sense for me -- ideally we'd have a RetrievalFeatureMatcher and RetrievalPairingOptions and update the sequential matching accordingly.

Vincentqyw:
@sarlinpe Sounds good — this is now implemented and pushed:

- **`RetrievalPairingOptions`** (method enum `VOCAB_TREE` / `GLOBAL_DESCRIPTOR` + shared and per-method options) and **`RetrievalPairGenerator`**, dispatching internally to the existing vocab-tree / global-descriptor implementations.
- **`CreateRetrievalFeatureMatcher()`** replaces both previous factories; one `retrieval_matcher` command with a `RetrievalMatching.*` namespace and one "Retrieval" GUI tab replace the two previous surfaces.
- **Sequential matching** embeds a `RetrievalPairingOptions loop_detection_options` and selects the method explicitly via `--SequentialMatching.loop_detection_method` (existing option names and defaults preserved), removing the implicit "empty `vocab_tree_path` selects global descriptors" convention.
- **pycolmap**: `match_retrieval` + `RetrievalPairingOptions` replace the vocab-tree bindings; docs updated.
- Per your point that interface-level generalization is sufficient, `retrieval::VisualIndex` / `GlobalDescriptorIndex` internals stay separate.

GUI:

The unified "Retrieval" tab — the method selector lists the vocabulary tree and the registered global descriptor models:

image

Selecting a model shows its options (precision variant, batch size) with the model path pre-filled with the variant's default download URI:
image



Sequential matching selects the loop detection method the same way:
image
image


This is currently a clean break (no `vocab_tree_matcher` alias) — happy to add a deprecated alias for a release cycle if you prefer. Benchmark numbers for Johannes's question follow in a separate comment.

gmberton:
I saw this only now, thanks for doing this @Vincentqyw! It will be great to see MegaLoc in COLMAP.
I tried your branch and ran into a few small things, mostly around how MegaLoc wants its inputs. I opened a PR against your branch with fixes (+136/-115), feel free to squash it into your own commits: https://github.com/Vincentqyw/colmap-gps/pull/3

- Added the ImageNet normalization that MegaLoc expects.
- Re-exported the ONNX from the latest MegaLoc code (which I updated on Jul 27 with little optimizations) to [the HF repo](https://huggingface.co/gberton/MegaLoc/tree/main) (437 MiB, happy to move it to a COLMAP release).
- The descriptor cache was only validated by image count, so a different database in the same folder reused the old descriptors.
- An unreadable image was skipped, then the run died with `Query image 8 not found in index`.
- The model list uses designated initializers (C++20), which should break MSVC in C++17 mode.
- Without `COLMAP_DOWNLOAD_ENABLED` no model was registered, so even a local model file was rejected.
- `automatic_reconstructor` never set `model_type`, so MegaLoc could not be selected. Added `--global_descriptor_model`.

Benchmarks: ETH3D DSLR, 13 scenes, 454 images, SIFT, 10 retrieved images per query, AUC on relative pose over all GT pairs.

| Method | AUC@5 | AUC@10 | Registered |
|---|---|---|---|
| Vocab tree | 0.763 | 0.815 | 427/454 |
| MixVPR | 0.767 | 0.811 | 416/454 |
| MegaLoc | 0.784 | 0.837 | 424/454 |

Retrieval quality alone, all 454 images pooled, a pair counting as correct if the two images share at least 50 GT points:

| Method | P@5 | P@10 | R@20 |
|---|---|---|---|
| MixVPR | 0.910 | 0.774 | 0.665 |
| MegaLoc | 0.927 | 0.824 | 0.730 |

MegaLoc takes 10 ms per image on an RTX 3090, 171 ms on CPU, plus image decoding.

PS: I used a council of LLMs (Fable, Astra, Gemini 3.8) for this PR to write and evaluate the edits.

Vincentqyw:
@gmberton Thank you so much for trying the branch and for the fixes!

I've merged your PR (Vincentqyw#3) into the `add-image-retrieval` branch, so all of your fixes are now part of this PR. The ImageNet normalization in particular was a real bug on my side; thanks for catching it.

Thanks a lot as well for re-exporting the ONNX model and for running the ETH3D experiments.

PR #4547, opened 2026-07-16
doc: Document populating the database with custom pose priors

behnamasadi:
## Summary

Expands the "Reconstruction with pose priors (GPS)" FAQ section to address the recurring questions in #2976 (as requested there, and repeatedly asked in its comment thread):

- **Populating the database with custom pose priors** (e.g., from RTK/GNSS logs, SLAM, or motion capture) via pycolmap's database API, including:
  - association of a prior with an image's sensor data via `corr_data_id`
  - the coordinate system convention (`WGS84` vs `CARTESIAN`) and that the covariance is always in metric Cartesian units
  - clarification that prior positions are **camera centers in the world frame** (a repeated point of confusion in the issue thread)
- **Running reconstruction with pose priors from pycolmap** via `IncrementalPipelineOptions(use_prior_position=True)`
- States the main benefit of pose priors (robustness, slight accuracy gains), as asked in the thread.

The pycolmap example was validated end-to-end against a real database (priors written and read back correctly, matching the current `PosePrior` bindings).

Docs build cleanly with Sphinx and the new section renders correctly.

Addresses #2976

### DIFF
--- a/doc/faq.rst
+++ b/doc/faq.rst
@@ -144,12 +144,58 @@ the ``pose_prior_mapper``::
 
 The ``pose_prior_mapper`` is essentially the incremental mapper with prior
 position constraints enabled. You can override the priors covariance (uncertainty)
-using ``--overwrite_priors_covariance``.  The new covariance will be built based 
+using ``--overwrite_priors_covariance``.  The new covariance will be built based
 on the values of ``--prior_position_std_x``, ``--prior_position_std_y``, and
-``--prior_position_std_z`` (default: 1.0 meter each).
+``--prior_position_std_z`` (default: 1.0 meter each). Using pose priors mainly
+improves the robustness of the reconstruction process and may also slightly
+improve its accuracy.
 
 For geo-registration of an already reconstructed model (without using priors
 during mapping), see the `Geo-registration`_ section.
+
+**Populating the database with custom pose priors**
+
+If your prior positions do not come from EXIF metadata but from another source
+(e.g., RTK/GNSS logs, SLAM, or motion capture), you can write them directly to
+the ``pose_priors`` table in the database using pycolmap. Each prior is
+associated with the sensor data of an image via ``corr_data_id`` and stores the
+position of the sensor in the world coordinate system (i.e., the camera center
+in world coordinates, not the world origin in camera coordinates). The position
+must be either in the ``WGS84`` (latitude, longitude, altitude) or ``CARTESIAN``
+(arbitrary metric frame) coordinate system, while the position covariance is
+always specified in metric Cartesian units. For example, after running the
+``feature_extractor``::
+
+    import numpy as np
+    import pycolmap
+
+    # Custom prior positions of the camera centers in the world frame,
+    # e.g., from RTK/GNSS logs or a SLAM system: image name -> 3D position.
+    positions = {"image0001.jpg": np.array([1.0, 2.0, 3.0]), ...}
+    std = 2.0  # Measurement uncertainty in meters.
+
+    db = pycolmap.Database.open("path/to/database.db")
+    for image in db.read_all_images():
+        prior = pycolmap.PosePrior()
+        prior.corr_data_id = image.data_id
+        prior.position = positions[image.name]
+        prior.position_covariance = std**2 * np.eye(3)
+        prior.coordinate_system = pycolmap.PosePriorCoordinateSystem.CARTESIAN
+        db.write_pose_prior(prior)
+    db.close()
+
+The reconstruction with pose priors can also be run directly from pycolmap
+using the incremental mapping pipeline::
+
+    import pycolmap
+
+    options = pycolmap.IncrementalPipelineOptions(use_prior_position=True)
+    reconstructions = pycolmap.incremental_mapping(
+        database_path="path/to/database.db",
+        image_path="path/to/images",
+        output_path="path/to/sparse",
+        options=options,
+    )
 
 
 .. _faq-share-intrinsics:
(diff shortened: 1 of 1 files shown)


StonerLing:
Hi @behnamasadi! I'm also working on addressing this issue in #4537. A few thoughts from a slightly different design perspective:

- Documenting direct use of `pycolmap.Database.Write*` APIs may lead users toward low-level Database interfaces when importing pose priors. Similar to rig import, where `Database::WriteRig` exists but the recommended workflow is `RigConfig` + `ApplyRigConfig()`, a higher-level abstraction may provide a cleaner and more consistent interface.

- Pose prior import is a reconstruction-level functionality rather than a Python-specific workflow. Providing the import path through `pycolmap` may leave C++ API, CLI, or GUI workflows without a similarly accessible workflow.

- External pose prior sources (e.g., GNSS/RTK, SLAM, motion capture) often have different metadata and conventions. A format-independent intermediate representation can decouple these external formats from COLMAP's internal database schema and provide a consistent place for validation, conversion, and future extensions.

These are just some design considerations from my side. I think both approaches could be useful, depending on whether the goal is to provide a quick import utility or a more general API-level integration.

behnamasadi:
Thanks @StonerLing — this is a fair point, and #4537 looks like the right long-term direction. I fully agree that a higher-level abstraction (mirroring the `RigConfig` + `ApplyRigConfig()` pattern) is cleaner than pointing users at the raw `Database.Write*` APIs, and that pose-prior import belongs at the reconstruction level with a format-independent intermediate representation for validation and conversion.

My intent here was narrower: documenting what's possible with the API that exists today, since users are already hand-populating the database (and in some cases hex-editing SQLite blobs) to get pose priors in. I'm happy to treat this as a stopgap and fold it into your approach once #4537 lands — e.g. reduce this to a short pointer to the high-level workflow and keep only the convention/units notes that stay relevant. Let me know how you'd like to coordinate; I'm glad to defer to #4537 as the primary path.

Irving87:
> Thanks @StonerLing — this is a fair point, and #4537 looks like the right long-term direction. I fully agree that a higher-level abstraction (mirroring the `RigConfig` + `ApplyRigConfig()` pattern) is cleaner than pointing users at the raw `Database.Write*` APIs, and that pose-prior import belongs at the reconstruction level with a format-independent intermediate representation for validation and conversion.
> 
> My intent here was narrower: documenting what's possible with the API that exists today, since users are already hand-populating the database (and in some cases hex-editing SQLite blobs) to get pose priors in. I'm happy to treat this as a stopgap and fold it into your approach once #4537 lands — e.g. reduce this to a short pointer to the high-level workflow and keep only the convention/units notes that stay relevant. Let me know how you'd like to coordinate; I'm glad to defer to #4537 as the primary path.

@behnamasadi Hi, i would like ask some question about how to populate pose_priors table in current colmap 4.2.0 version. There are 4 NOTNULL column in pose_priors table: **pose_prior_id, corr_data_id, corr_sensor_id, corr_sensor_type**, can you tell me what's the difference between **pose_prior_id** and **corr_data_id**? I have only one camera, is the **pose_prior_id** equal to **corr_data_id** and they are same with **image_id** in images table?

behnamasadi:
@Irving87 Sorry for the slow reply. `pose_prior_id` and `corr_data_id` are unrelated, and neither is automatically the `image_id`:

- **`pose_prior_id`** is just the table's own primary key — the identity of the *row*. If you write via `Database::WritePosePrior` with `use_pose_prior_id = false` (the default in the pycolmap binding), it is bound to `NULL` and SQLite assigns the next rowid. It carries no meaning and you should not try to make it match anything.
- **`corr_data_id` + `corr_sensor_id` + `corr_sensor_type`** are the three columns that say *what the prior is attached to*. Together they are a `data_t`, i.e. a `(sensor_t{type, sensor_id}, data_id)` triple. For an image, `Image::DataId()` is `data_t(sensor_t(SensorType::CAMERA, camera_id), image_id)` (`src/colmap/scene/image.h:201`), so:
  - `corr_data_id` = your **`image_id`**
  - `corr_sensor_id` = your **`camera_id`**
  - `corr_sensor_type` = **0** (`SensorType::CAMERA`; `IMU` is 1 — `src/colmap/util/types.h:143`)

So for your single-camera case: `corr_data_id` equals `image_id`, `corr_sensor_id` is the one `camera_id`, `corr_sensor_type` is 0, and `pose_prior_id` is whatever SQLite hands out.

From pycolmap you do not have to assemble the triple by hand — `image.data_id` gives you the whole `data_t`, which is what the example in this PR uses:

```python
prior = pycolmap.PosePrior()
prior.corr_data_id = image.data_id   # (sensor_t(CAMERA, camera_id), image_id)
prior.position = position            # camera center in world coordinates
prior.position_covariance = std**2 * np.eye(3)
prior.coordinate_system = pycolmap.PosePriorCoordinateSystem.CARTESIAN
db.write_pose_prior(prior)
```

Two things worth knowing: there is a `UNIQUE INDEX` on `(corr_data_id, corr_sensor_id, corr_sensor_type)` (`src/colmap/scene/database_sqlite.cc:2093`), so there is one prior per image and a second write for the same image is rejected rather than silently replacing; and `position`/`position_covariance` are raw little-endian float64 BLOBs (3 and 9 doubles), with the covariance always in the **Cartesian** frame even when `coordinate_system` is `WGS84`.

@StonerLing @ahojnnes — questions like this one are exactly why I opened this PR, and they keep arriving while #4537 is still open. I remain happy to fold this into the `PosePriorArchive` workflow once that lands, and to reduce this entry to a pointer plus the convention/units notes that stay relevant either way. Until then it documents the API that exists today.

Irving87:
@behnamasadi Thank you for your so detailed reply

Issue #4367, opened 2026-04-28
Question about min_num_inliers in TwoViewGeometryOptions


moh-ake:
Hello,

I have a question about the min_num_inliers parameter in geometric verification.

I set it to 30, but after processing I can still see image pairs in the database with a geometric verification inlier count below 30, as shown in the image below.

Could you please clarify why image pairs with fewer than 30 inliers are still being kept in the database?

Thank you in advance,

<img width="1921" height="546" alt="Image" src="https://github.com/user-attachments/assets/99b1ffc8-d072-492c-bb16-ddb60c77ca72" />

ahojnnes:
Have you been running matching/verification multiple times with different parameters or with custom imported inlier matches? When re-running matching with different parameters, existing two-view geometries will be skipped and not touched. Does this explain the behavior?

moh-ake:
Thanks for your response @ahojnnes 

The problem appears even when reconstructing from scratch. I think that when the sequence is made of frames, and each frame is composed of many images linked by a rig, min_num_inliers applies between frames rather than to individual images. Am I right?

Thank you in advance

ahojnnes:
No, min_num_inliers applies on a per image level except when you explicitly set rig_verification=true.

PR #4548, opened 2026-07-16
doc: Add FAQ entry on re-running feature matching with different options

behnamasadi:
## Summary

Adds a FAQ entry documenting a behavior that repeatedly confuses users (most recently in #4367):

- When re-running feature matching on an existing database, image pairs with existing two-view geometries are skipped and not updated (`feature_matching_utils.cc`), so changed matching/verification options (e.g. `--TwoViewGeometry.min_num_inliers`) only affect previously unprocessed pairs.
- How to re-run matching from scratch using `colmap database_cleaner` with `--type two_view_geometries` / `matches` / `features`.
- Clarifies that `min_num_inliers` applies per image pair unless `--FeatureMatching.rig_verification` is enabled (per-frame-pair verification), as explained by @ahojnnes in #4367.

Docs build cleanly with Sphinx and the section renders correctly.

Related to #4367

### DIFF
--- a/doc/features.rst
+++ b/doc/features.rst
@@ -78,6 +78,19 @@ In the GUI, open ``Processing > Feature matching``, select any matching tab
 (Exhaustive, Sequential, etc.), and choose the matcher type from the "Type"
 dropdown in the shared options section.
 
+When re-running matching on an existing database, image pairs that already have
+a two-view geometry are skipped, so changing ``--FeatureMatching.type`` on such
+a database leaves previously matched pairs untouched. To force re-matching, first
+clear the existing results with ``database_cleaner``::
+
+    $ colmap database_cleaner \
+        --database_path $DATASET_PATH/database.db \
+        --type two_view_geometries
+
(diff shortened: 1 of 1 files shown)


sarlinpe (commented):
Thank you! Isn't this too low-level for the FAQ? do we really expect users to play with parameters like `min_num_inliers`? I think that only `FeatureExtraction.type` and `FeatureMatching.type` are relevant. I suggest to mention `database_cleaner` in https://colmap.github.io/features.html only, not in the FAQ.

behnamasadi:
Good point, thanks — agreed it's too low-level for the FAQ. I've moved it to the Feature Extraction and Matching page (`features.rst`) and trimmed it to the essentials: image pairs with an existing two-view geometry are skipped when re-running matching, so switching `--FeatureMatching.type` (or re-extracting with a different `--FeatureExtraction.type`) requires clearing prior results with `database_cleaner`. Dropped the `min_num_inliers` / `rig_verification` details. PTAL.

sarlinpe (approved):
Thank you!

PR #4551, opened 2026-07-17
Fix misspelled pycolmap enum name GPSTransfromEllipsoid

behnamasadi:
## Summary
The Python binding of `GPSTransform::Ellipsoid` is exposed under the misspelled name `GPSTransfromEllipsoid` (Transfrom → Transform). This renames it to `GPSTransformEllipsoid` and keeps the old name as an alias for backwards compatibility, so existing user code does not break. Tests updated to the new name, with one test asserting the alias still resolves.

Happy to drop the compatibility alias if you prefer a clean break.

### DIFF
--- a/src/pycolmap/geometry/gps.cc
+++ b/src/pycolmap/geometry/gps.cc
@@ -14,11 +14,11 @@ using namespace pybind11::literals;
 namespace py = pybind11;
 
 void BindGPS(py::module& m) {
-  py::enum_<GPSTransform::Ellipsoid> PyGPSTransfromEllipsoid(
-      m, "GPSTransfromEllipsoid");
-  PyGPSTransfromEllipsoid.value("GRS80", GPSTransform::Ellipsoid::GRS80)
+  py::enum_<GPSTransform::Ellipsoid> PyGPSTransformEllipsoid(
+      m, "GPSTransformEllipsoid");
+  PyGPSTransformEllipsoid.value("GRS80", GPSTransform::Ellipsoid::GRS80)
       .value("WGS84", GPSTransform::Ellipsoid::WGS84);
-  AddStringToEnumConstructor(PyGPSTransfromEllipsoid);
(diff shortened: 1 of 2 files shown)


B1ueber2y (approved):

src/pycolmap/geometry/gps.cc
```diff
@@ -14,11 +14,13 @@ using namespace pybind11::literals;
...
-      m, "GPSTransfromEllipsoid");
-  PyGPSTransfromEllipsoid.value("GRS80", GPSTransform::Ellipsoid::GRS80)
+  py::enum_<GPSTransform::Ellipsoid> PyGPSTransformEllipsoid(
+      m, "GPSTransformEllipsoid");
+  PyGPSTransformEllipsoid.value("GRS80", GPSTransform::Ellipsoid::GRS80)
       .value("WGS84", GPSTransform::Ellipsoid::WGS84);
-  AddStringToEnumConstructor(PyGPSTransfromEllipsoid);
+  AddStringToEnumConstructor(PyGPSTransformEllipsoid);
+  // Deprecated alias for the misspelled name kept for backwards compatibility.
+  m.attr("GPSTransfromEllipsoid") = PyGPSTransformEllipsoid;
```
B1ueber2y: I would suggest to just drop it and tag breaking.

B1ueber2y (changes requested):

src/pycolmap/geometry/gps.cc
behnamasadi: Done — dropped the alias and removed its test in the latest push. The commit is now marked `BREAKING CHANGE` (the old `pycolmap.GPSTransfromEllipsoid` name is removed with no fallback). I don't have permission to add the `breaking` label myself, so could you apply it? Happy to adjust if you tag breaking changes differently.

B1ueber2y (approved):

ahojnnes:
Thank you!

PR #4559, opened 2026-07-18
Use optional cameras if available when decomposing relative poses.

B1ueber2y:
A follow-up fix of https://github.com/colmap/colmap/pull/4521. Scanned the repo and confirmed that this was the only site that we missed to handle.

### DIFF
--- a/src/colmap/estimators/view_graph_calibration.cc
+++ b/src/colmap/estimators/view_graph_calibration.cc
@@ -120,8 +120,9 @@ void CrossValidatePriorFocalLengths(
                                              tvg.F.value(),
                                              camera1.CalibrationMatrix());
       tvg.config = TwoViewGeometry::CALIBRATED;
-      // E now lives in the calibrated frame; solver-estimated intrinsics are
-      // superseded.
+      // E is now built from the K of the cross-validated focal priors, and
+      // consumers calibrate the rays with camera1/camera2 whenever set. Clear
+      // them so the rays use the same K as E, not the solver's own focal.
       tvg.camera1.reset();
       tvg.camera2.reset();
       ++num_upgraded_pairs;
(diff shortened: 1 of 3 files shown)


ahojnnes:
Does it make a difference in terms of AUC on ETH3D etc.?

ahojnnes (approved):

src/colmap/estimators/view_graph_calibration.cc
```diff
@@ -120,6 +120,10 @@ void CrossValidatePriorFocalLengths(
                                              tvg.F.value(),
                                              camera1.CalibrationMatrix());
       tvg.config = TwoViewGeometry::CALIBRATED;
+      // E now lives in the calibrated frame; solver-estimated intrinsics are
+      // superseded.
+      tvg.camera1.reset();
+      tvg.camera2.reset();
```
ahojnnes, line 127: Alternatively could update the focal length?

B1ueber2y: camera1/camera2 are only meant to carry intrinsics that a two-view solver estimated, as a signal that the pair's focal came from the solver rather than from the camera. After the upgrade that's no longer the case in either branch: E is rebuilt from the cross-validated priors in one and from the calibrated focals in the other, and both live in cameras, so downstream consumers should read them from there. Populating camera1/camera2 with those same values would just duplicate the camera and blur what the field means, so resetting seemed cleaner.

I have updated the comments at both sites to spell this out.

B1ueber2y:
> Does it make a difference in terms of AUC on ETH3D etc.?

I have not benchmarked the effects of this PR on the benchmark, but this is a bug fix in my opinion and should anyway be merged. I am going forward to merge it now, as it s blocking the eval of a subsequent one-side focal PR I am going to send in : )

PR #4560, opened 2026-07-18
Restrict projective geometry paths to pinhole based camera models

B1ueber2y:
Fundamental matrix does not natively make sense for perspective Fisheye cameras without focal length prior.

### DIFF
--- a/src/colmap/scene/camera.h
+++ b/src/colmap/scene/camera.h
@@ -132,15 +132,11 @@ struct Camera {
   // panorama), i.e. the EQUIRECTANGULAR model.
   inline bool IsSpherical() const;
 
-  // Whether the camera model is a fisheye model.
-  inline bool IsFisheye() const;
+  // Whether the camera model is perspective and fisheye.
+  inline bool IsPerspectiveFisheye() const;
 
-  // Whether the camera model is built on a pinhole projection
-  // (r = f * tan(theta)), possibly with distortion layered on top. Fisheye
-  // models (r ~ f * theta) and spherical models are not. Required by
-  // estimators whose model is linear in pixel coordinates (F, E, H) or that
-  // recover a focal length from one.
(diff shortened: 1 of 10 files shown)


src/colmap/scene/camera.h
```diff
@@ -135,6 +135,13 @@ struct Camera {
   // Whether the camera model is a fisheye model.
   inline bool IsFisheye() const;
 
+  // Whether the camera model is built on a pinhole projection
+  // (r = f * tan(theta)), possibly with distortion layered on top. Fisheye
+  // models (r ~ f * theta) and spherical models are not. Required by
+  // estimators whose model is linear in pixel coordinates (F, E, H) or that
+  // recover a focal length from one.
+  inline bool HasPinholeProjection() const;
```
ahojnnes: How about:
```
IsPerspective()
IsPerspectivePinhole()
IsPerspectiveFisheye()
IsSpherical()
```
to match our camera model class names. We can still change the IsFisheye, because we introduced it only a few days ago?

ahojnnes: We might even want to rename the underlying functions to CameraIsPerspectivePinhole/CameraIsPerspectiveFisheye.

ahojnnes (approved):

src/colmap/scene/camera.h
B1ueber2y: Good idea. Done.

B1ueber2y:
Made some updates and guard against uncalibrated paths for fisheye images in two view geometry. The estimation of F and H does not make sense in the current case. However, it can be fixed by mapping to a virtual pinhole plane with ``FisheyeToNormal`` first, which I left two TODOs for and leave out of the scope of this PR. PTAL.

ahojnnes (approved):

PR #4561, opened 2026-07-18
Integrate 6pt one-side focal solver in two view geometry estimation.

B1ueber2y:
Integrate 6pt solver (9 solutions) from PoseLib with colmap TinySolver as LO. As a byproduct, this PR resolves pending TODO for two-view geometry between a spherical camera and an uncalibrated perspective camera. 

Tagged as draft for now, as there are two tasks pending:
- [x] fix guided matching with the presence of intrinsic-estimated solver or spherical setup
- [x] benchmarking results

cc: @vlarsson 

## (Edit: July 22, 2026) E2E reconstruction (ETH3D dslr, 10 scenes x 3 seeds, --camera_prior_fraction 0.5)

  `onesided - main`, AUC mean ± std over 3 seeds, averaged over scenes. Positive = branch better.

Here we create a mixed calibration setup and only provide half the cameras intrinsics. 

  | | @0.5 | @1.0 | @5.0 | @10.0 |
  |---|---|---|---|---|
  | **onesided - main** | +0.06 ± 2.07 | +0.08 ± 3.65 | +1.88 ± 3.14 | +1.80 ± 2.95 |

  ## Two-view relative pose AUC (272 mixed-calibration pairs, 4 scenes)

  Error is `max(R_err, t_err)` per pair.

  | | @1° | @3° | @5° | @10° |
  |---|---|---|---|---|
  | main (F) | 0.7 | 3.0 | 6.2 | 12.5 |
  | **one-sided** | **38.3** | **53.2** | **59.8** | **68.6** |

PR #4563, opened 2026-07-19
Tangent sampson error for calibrated two-view geometry

B1ueber2y:
**Highlights:**
* Theoretical improvements (or to be accurate, they are fixes rather than improvements) on calibrated tvg, guided matching, geometric verification, and generalized relative pose on all camera models. 
* Practical improvements on spherical and large FoV models on tvg, guided matching, and geometric verification.
* Analytical Jacobian for tangent sampson error in tiny solver kept the runtime overhead negligible. 

Reference: 
* Section 4.2.6 of the "Multiple View Geometry in Computer Vision", 2000.
* Tangent Sampson Error: Fast Approximate Two-view Reprojection Error for Central Cameras", ICCV 2023.

cc: @vlarsson 

# (Edit: July 21, 2026) Tangent Sampson error for calibrated two-view geometry

## Summary

COLMAP's calibrated two-view geometry scores correspondences by feeding the **Sampson error** unit bearings. The Sampson error is a first-order *pixel* reprojection distance, defined on the image measurements: its denominator is the
residual's gradient with respect to its arguments, so it is a pixel distance only when those arguments are the pixels. Unit bearings are a unit-normalized nonlinear function of the pixels, so plugging them in normalizes by the wrong gradient. This
is not a valid "bearing-space" residual — it is the pixel Sampson error computed incorrectly, and it is wrong for **every** camera, pinhole included. The per-camera threshold conversion (`CamFromImgThreshold`) only patches the leading focal-length scale; the residual still deviates off-axis, and progressively more on wide-FOV and spherical models, where it changes which correspondences are accepted.

This PR switches the calibrated path to the **tangent Sampson error** (Terekhov & Larsson, ICCV 2023): the Sampson error carrying each point's analytic unprojection Jacobian `J = d(ray)/d(pixel)`, so it is the true pixel distance for **any central camera model**. Under a pinhole it reduces to the ordinary Sampson error. One pixel
threshold, no per-camera conversion. It also adds a nonlinear least-squares refine as the LO-RANSAC local optimizer for relative pose, with a hand-derived **analytic Jacobian**.

Net effect: a latent correctness bug is fixed, RANSAC thresholds become a single pixel value with no per-camera conversion, spherical guided-matching recall improves substantially, and relative-pose accuracy roughly doubles — at no runtime premium.

## What changed

- **Essential-matrix RANSAC** scores the pixel-unit tangent Sampson error (`EssentialMatrixTangentSampsonEstimator`) instead of Sampson-on-bearings; the inlier threshold is now a plain pixel value (no `CamFromImgThreshold`).
- **Nonlinear LM refine** added as the LO-RANSAC local optimizer for relative pose, minimizing the same tangent Sampson objective, with an **analytic** 7-parameter Jacobian (validated against autodiff and finite differences).
- **`Camera::CamRayFromImgWithJac`** — analytic unprojection + ray Jacobian for all camera models (reuses the projection Jacobian already computed for undistortion).
- **Generalized relative pose** and **structureless absolute pose** migrated to the same pixel-unit tangent Sampson residual + plain pixel threshold.
- **`RefineRelativePose` / `RefineEssentialMatrix`** (the post-RANSAC Ceres refiner, also `pycolmap.refine_relative_pose`) now minimize tangent Sampson.
- **Guided matching** (`MatchGuided`) uses the tangent (pixel) epipolar filter. Retired the buggy Sampson-on-bearings interfaces (`SampsonError` on unit rays and `...WithCheirality`).
- **Unchanged:** the fundamental-matrix and shared-focal paths, which are already pixel-native.

## Why (correctness)

Sampson error is `num / ‖∇num‖`, the first-order distance to the epipolar variety, normalized by the residual's gradient with respect to its arguments. It is a pixel distance only when those arguments are the pixel measurements. Feeding it unit
bearings silently treats the bearings as the measurements and normalizes by the identity Jacobian `[I₂; 0]`, which is not the bearing-to-pixel Jacobian — so the value is simply wrong.

Tangent Sampson supplies the true `J = d(ray)/d(pixel)`, recovering the pixel distance for any central model. Under a pinhole it equals the ordinary Sampson error (up to the pixel↔normalized-plane focal scale — the PinholeMatchesScaledSampsonExactly` test). Sampson-on-bearings agrees with it only to first order even for a pinhole (`UnitRaysAgreeToFirstOrder`) and diverges outright off-axis and on distorted/spherical models, where it changes inlier selection.

### 1. Relative pose — synthetic accuracy

Median rotation / translation-direction error (degrees), 1000 trials, 1 px noise, three cameras. Columns isolate the two independent changes: correcting the **scoring residual** (Sampson-on-bearings → tangent) and the **LM refine**.

| camera | outliers | main (Sampson on bearings, 5-pt refit) | tangent scoring, 5-pt refit | tangent + LM refine (this PR) |
|---|---|---|---|---|
| SimplePinhole | 0% | 0.045 / 0.039 | 0.043 / 0.039 | 0.023 / 0.020 |
| SimplePinhole | 30% | 0.056 / 0.050 | 0.052 / 0.049 | 0.028 / 0.025 |
| OpenCV (wide + distortion) | 0% | 0.153 / 0.140 | 0.151 / 0.147 | 0.071 / 0.069 |
| OpenCV (wide + distortion) | 30% | 0.177 / 0.163 | 0.175 / 0.171 | 0.092 / 0.084 |
| Equirectangular | 0% | 0.179 / 0.184 | 0.179 / 0.177 | 0.063 / 0.054 |
| Equirectangular | 30% | 0.229 / 0.238 | 0.231 / 0.237 | 0.078 / 0.066 |

- The accuracy gain (~2× on pinhole/OpenCV, ~2.8× on the sphere) comes from the   **LM refine**, not the residual: correcting the scoring alone (columns 1→2) is pose-neutral here. A control that runs the old (incorrect) Sampson-on-bearings residual *inside* the same LM lands within a few percent of the tangent LM, so the win is having a least-squares refine over all inliers. 
- Tangent Sampson's own two-view payoff is therefore the **pixel-consistent threshold** (above) and **guided-matching recall** (§2), not pose accuracy.

### 2. Relative pose — analytic vs. autodiff Jacobian (runtime)

The LM refine ships with an analytic Jacobian. It **reproduces the autodiff LM's pose to 4 decimals** on every row above, at lower cost. Median ms, fixed 100 RANSAC iterations (0% outliers is the slower case here — with all points inliers the LO
fires on more improvements and refines over more inliers, paid equally by every config):

| camera | outliers | main (Sampson on bearings) | tangent, 5-pt refit | LM refine (autodiff) | LM refine (analytic, this PR) |
|---|---|---|---|---|---|
| SimplePinhole | 0% | 2.65ms | 3.51ms | 3.70ms | 3.41ms |
| SimplePinhole | 30% | 1.81ms | 2.14ms | 2.14ms | 1.94ms |
| OpenCV | 0% | 2.55ms | 3.33ms | 3.62ms | 3.26ms |
| OpenCV | 30% | 1.79ms | 2.07ms | 2.15ms | 1.90ms |
| Equirectangular | 0% | 2.25ms | 2.68ms | 2.71ms | 2.45ms |
| Equirectangular | 30% | 1.52ms | 1.63ms | 1.62ms | 1.43ms |

The analytic Jacobian removes the autodiff overhead: the tangent LM now runs at the plain 5-point-refit cost. In isolation the refine step itself is ~3.3× cheaper (≈1.1 → 0.34 ms) than the autodiff version it replaces.

### 3. Guided matching — spherical recall

Equirectangular camera, 300 known correspondences over 40 image pairs, brute-force `MatchGuided` at an 8 px gate. main = Sampson-on-bearings filter; this PR = tangent (pixel) Sampson filter.

| keypoint noise (px) | main recall % | this PR recall % | Δ | precision % |
|---|---|---|---|---|
| 0.5 | 99.92 | 100.00 | +0.08 | 100 |
| 1.0 | 99.42 | 100.00 | +0.58 | 100 |
| 2.0 | 98.44 | 100.00 | +1.56 | 100 |
| 3.0 | 94.70 | 99.26 | +4.56 | 100 |
| 4.0 | 88.22 | 95.65 | +7.43 | 100 |
| 5.0 | 80.85 | 89.46 | **+8.61** | 100 |

The pixel filter is uniform across the sphere where the Sampson-on-bearings filter tightens off-axis, admitting progressively more true correspondences as noise grows, at **100% precision throughout** (no false positives).

### 4. Unprojection Jacobian — cost per call

`CamRayFromImgWithJac`, 100k random pixels:

| camera | `CamRayFromImg` | `+ WithJac` | overhead |
|---|---|---|---|
| SimplePinhole | 6.7 ns | 30.2 ns | +23 ns |
| OpenCV (wide + distortion) | 194.3 ns | 230.8 ns | 1.19× |
| Equirectangular | 41.7 ns | 98.0 ns | 2.35× |

The ray Jacobian is the analytic pseudo-inverse of the projection Jacobian, which is recomputed at the unprojected point (not reused from the undistortion solve). On distorted models that extra evaluation is small next to the iterative undistortion
(1.19×); in absolute terms it is tens of ns per keypoint, paid once during setup (outside the RANSAC loop).

### 5. Reconstruction — ETH3D (no regression)

End-to-end SfM on the ETH3D DSLR **calibrated** set (25 undistorted `PINHOLE` scenes, GT intrinsics as priors), branch vs. main, 5 seeds, matching + two-view geometry recomputed by each binary. Reported is the per-seed **paired** difference (`--threads_per_scene 1`), averaged over scenes and seeds.

Relative-pose AUC delta, **branch − main** (average over scenes, mean ± std over 5 seeds; `+` = branch better):

| mapper | @0.5° | @1.0° | @5.0° | @10.0° |
|---|---|---|---|---|
| incremental | +1.04 ± 1.65 | +0.95 ± 1.46 | +0.43 ± 1.25 | +0.25 ± 1.18 |
| global | −0.40 ± 0.48 | −0.48 ± 1.08 | −0.63 ± 1.71 | −0.76 ± 1.67 |

**No measurable change.** Every delta is within its seed std, and the sign flips between the two mappers — i.e. noise-level, not a real effect. This is the expected result: on a pinhole camera tangent Sampson reduces to the ordinary Sampson error
(up to the focal-length scale), so it neither helps nor hurts. Any two-view-level difference is further absorbed by bundle adjustment. The `overall` (pair-weighted) figures agree but carry larger variance, dominated by a few scenes whose
incremental reconstruction is bistable across seeds — a known property of the mapper's RANSAC, unrelated to this change. This section is a regression gate; tangent Sampson's payoff is on non-pinhole models (§3, guided matching) which
ETH3D DSLR does not exercise.

## Validation

- Solver and estimator unit tests for the tangent Sampson error and the LM refine, including a finite-difference check of the analytic Jacobian and an equivalence check against `ComputeSquaredTangentSampsonError`.
- Generalized-pose and pose-refinement tests updated to pixel-unit thresholds.
- Under a pinhole, tangent Sampson equals the ordinary Sampson error up to the focal-length scale (covered by the `PinholeMatchesScaledSampsonExactly` test).

## Benchmark notes

The relative-pose numbers (§1–§2) are from a synthetic micro-benchmark (fixed seed, identical scenes across all configurations, LO isolated at a fixed iteration budget); absolute milliseconds scale with scene size and iteration count and are meaningful only relatively. §3–§5 are from the earlier guided-matching, Jacobian, and ETH3D measurements. The benchmark tools are temporary and not part of this PR.


src/colmap/geometry/essential_matrix.cc
vlarsson: Should ray1 and ray2 be hnormalized here? @B1ueber2y Or are they already?

vlarsson: hmm, or is the scale absorbed in the quotient maybe

vlarsson: On second thought, I guess it is not, as ray1 and ray2 can have different normalization factors, and the denominator would need different scalars for the four elements then

vlarsson:
I haven't dug into your implementation, but one caveat to keep in mind is that refining TangentSampson over camera parameters requires mixed second derivatives of the projections (as the jacobians are part of the expression). This is a bit annoying, as we only expose jacobians from the camera model. In PoseLib we do some finite diffs of the jacobians to resolve this https://github.com/PoseLib/PoseLib/blob/master/PoseLib/robust/optim/relative.h#L390-L406, but it's not the nicest way.

B1ueber2y:
> I haven't dug into your implementation, but one caveat to keep in mind is that refining TangentSampson over camera parameters requires mixed second derivatives of the projections (as the jacobians are part of the expression). This is a bit annoying, as we only expose jacobians from the camera model. In PoseLib we do some finite diffs of the jacobians to resolve this https://github.com/PoseLib/PoseLib/blob/master/PoseLib/robust/optim/relative.h#L390-L406, but it's not the nicest way.

Thanks! Good point. Until now this PR is only for calibrated setup, so camera parameters are fixed in the estimator. I do not have plans to extend it to uncalibrated one yet, as we always start with estimating focal length assuming pinhole for such camera. One potential case where tangent sampson error on uncalibrated one can start to make sense is the one-side focal estimator https://github.com/colmap/colmap/pull/4561, which I believe is way beyond the scope of this PR.

src/colmap/geometry/essential_matrix.cc
B1ueber2y: The Jacobian was derived with respect to unit bearings, so it should not be hnormalized here. The error is the same sampson error as long as the Jacobian agrees with the convention in tangent sampson error formulation.

vlarsson: But this function is for the normal Sampson error? The regular Samspon error is not the same as Tangent for regular pinhole.

B1ueber2y: I see. Thanks for revealing a big pre-existing issue in the codebase. I will try to send a fix soon.

B1ueber2y:
I updated the PR description and this branch is ready to review from my side. In particular, I edited the highlights in the top of the description : )

src/pycolmap/estimators/pose.cc
```diff
@@ -235,19 +265,26 @@ void BindAbsolutePoseEstimator(py::module& m) {
         "Robust absolute pose estimation with LO-RANSAC "
         "followed by non-linear refinement.");
 
-  m.def("estimate_relative_pose",
```
B1ueber2y, line 238: For clarity, here shows two breaking changes in the following two pycolmap interfaces, which are necessary in this PR.

src/colmap/estimators/generalized_pose.cc
```diff
@@ -456,20 +468,24 @@ bool EstimateStructureLessAbsolutePose(
...
-    query_obs[i].cam_from_rig = Rigid3d();
-    query_obs[i].ray_in_cam = query_camera.CamRayFromImg(query_points2D[i])
-                                  .value_or(Eigen::Vector3d::Zero());
+    const CamRayWithJac world_ray_with_jac =
+        world_cameras[world_camera_idx]
+            .CamRayFromImgWithJac(world_points2D[i])
+            .value_or(CamRayWithJac::Zero());
+    world_obs[i] = {world_cams_from_world[world_camera_idx],
+                    world_ray_with_jac.ray,
+                    world_ray_with_jac.jacobian};
```
ahojnnes: nit: could directly store a CamRayWithJac in the world_obs?

src/colmap/estimators/two_view_geometry.cc
```diff
@@ -54,6 +54,18 @@
...
 namespace {
 
+// Unproject a point to a unit bearing together with the Jacobian of that
+// bearing with respect to the pixel, mapping failures to a zeroed entry.
+//
+// A zero ray and Jacobian yield a zero tangent Sampson numerator *and*
+// denominator, which ComputeSquaredTangentSampsonError reports as an infinite
+// residual. Unprojectable or degenerate points are therefore rejected outright
+// rather than scoring as spuriously good correspondences.
+CamRayWithJac CamRayWithJacOrZero(const Camera& camera,
```
ahojnnes: nit: inline?

src/pycolmap/estimators/pose.cc
```diff
@@ -235,19 +265,26 @@ void BindAbsolutePoseEstimator(py::module& m) {
...
+      "estimate_relative_pose",
+      &PyEstimateRelativePose,
+      "camera1"_a,
+      "points2D1"_a,
+      "camera2"_a,
+      "points2D2"_a,
+      py::arg_v("options", RANSACOptions(), "RANSACOptions()"),
+      "Robustly estimate relative pose from 2D-2D correspondences (given as "
+      "camera + image points) using LO-RANSAC with the tangent Sampson error, "
+      "without non-linear refinement.");
```
ahojnnes: This docstring says that `estimate_relative_pose` runs without non-linear refinement, but the estimator now uses `EssentialMatrixTangentSampsonEstimator` as the LO-RANSAC local estimator. Since that class exposes `Refine`, `loransac.h` invokes the nonlinear tangent-Sampson refinement during local optimization. Could we update the wording to make that behavior explicit?

src/colmap/estimators/cost_functions/sampson_error.h
```diff
@@ -52,17 +56,17 @@ Eigen::Matrix<T, 3, 3> EssentialMatrixFromPoseParams(
...
 template <typename T>
 T SampsonError(const Eigen::Matrix<T, 3, 3>& E,
-               const Eigen::Matrix<T, 3, 1>& cam_ray1,
-               const Eigen::Matrix<T, 3, 1>& cam_ray2) {
-  const Eigen::Matrix<T, 3, 1> epipolar_line1 = E * cam_ray1;
-  const T num = cam_ray2.dot(epipolar_line1);
-  const Eigen::Matrix<T, 4, 1> denom(cam_ray2.dot(E.col(0)),
-                                     cam_ray2.dot(E.col(1)),
+               const Eigen::Matrix<T, 3, 1>& point1,
+               const Eigen::Matrix<T, 3, 1>& point2) {
```
ahojnnes: Should we accept 2d points here then now and use .homogenoues() below?

src/colmap/estimators/cost_functions/sampson_error.h
```diff
@@ -72,31 +76,86 @@ T SampsonError(const Eigen::Matrix<T, 3, 3>& E,
...
+// TangentSampsonErrorCostFunctor, which is pixel-accurate for any central
+// model.
 class SampsonErrorCostFunctor
     : public AutoDiffCostFunctor<SampsonErrorCostFunctor, 1, 7> {
  public:
-  SampsonErrorCostFunctor(const Eigen::Vector3d& cam_ray1,
-                          const Eigen::Vector3d& cam_ray2)
-      : cam_ray1_(cam_ray1), cam_ray2_(cam_ray2) {}
+  SampsonErrorCostFunctor(const Eigen::Vector3d& point1,
+                          const Eigen::Vector3d& point2)
```
ahojnnes: Take 2d instead of 3d points?

src/colmap/estimators/solvers/essential_matrix.cc
```diff
@@ -225,62 +217,57 @@ void EssentialMatrixEightPointEstimator::Estimate(
...
 
-  // Self-seed with the eight-point solver.
-  std::vector<M_t> init_models;
-  EssentialMatrixEightPointEstimator::Estimate(
-      cam_rays1, cam_rays2, &init_models);
-  if (init_models.empty()) {
-    return;
+// Extract the bearings into the contiguous array the five-point solver expects.
+// The Jacobians play no part in estimation. They only affect scoring.
+void UnpackCamRaysWithJac(const std::vector<CamRayWithJac>& cam_rays_with_jac,
```
ahojnnes: nit: Just return the std::vector instead of taking by ref?

src/colmap/feature/sift.cc
```diff
@@ -1568,11 +1606,9 @@ class SiftGPUFeatureMatcher : public FeatureMatcher {
     two_view_geometry->inlier_matches.resize(
         static_cast<size_t>(options_.max_num_matches));
 
-    const float max_residual =
-        use_essential_matrix
-            ? static_cast<float>(ComputeNormalizedGuidedMatchingMaxResidual(
-                  effective_camera1, effective_camera2, max_error))
-            : static_cast<float>(max_error * max_error);
+    // Every config now scores in pixels: the tangent Sampson error for E, the
```
ahojnnes:
```suggestion
    // Every config scores in pixels: the tangent Sampson error for E, the
```

src/colmap/geometry/essential_matrix.h
```diff
@@ -182,28 +177,105 @@ void ComputeSquaredSampsonError(const std::vector<Eigen::Vector3d>& points1,
...
+// @return            Squared tangent Sampson error, in squared pixels.
+double ComputeSquaredTangentSampsonError(const CamRayWithJac& cam_ray1_with_jac,
+                                         const CamRayWithJac& cam_ray2_with_jac,
+                                         const Eigen::Matrix3d& E);
+
+// Overload reading the ray and Jacobian in place, so hot loops (e.g. the CPU
+// guided-matching filter, called O(N1*N2) times) avoid materializing a
+// CamRayWithJac per correspondence.
+double ComputeSquaredTangentSampsonError(const Eigen::Vector3d& cam_ray1,
+                                         const Eigen::Matrix<double, 3, 2>& J1,
```
ahojnnes: Here and elsewhere:
```suggestion
                                         const Eigen::Matrix2x3d& J1,
```

src/colmap/geometry/pose.h
```diff
@@ -43,6 +43,21 @@
 
 namespace colmap {
 
+// A unit bearing and the Jacobian d(ray)/d(pixel) of its unprojection (see
+// Camera::CamRayFromImgWithJac), bundled so RANSAC subsampling keeps them
+// index-aligned.
+struct CamRayWithJac {
+  Eigen::Vector3d ray;
+  Eigen::Matrix<double, 3, 2> jacobian;
```
ahojnnes: Matrix2x3

src/thirdparty/SiftGPU/ProgramCU.cu
```diff
@@ -1565,46 +1565,44 @@ void __global__ MultiplyDescriptorG_Kernel(cudaTextureObject_t texDes1, cudaText
...
+				fx1[0] = F.mat[0][0] * locx + F.mat[0][1] * locy + F.mat[0][2];
+				fx1[1] = F.mat[1][0] * locx + F.mat[1][1] * locy + F.mat[1][2];
+				fx1[2] = F.mat[2][0] * locx + F.mat[2][1] * locy + F.mat[2][2];
 
-				ftx2[0] = F.mat[0][0] * loc2.x + F.mat[1][0] * loc2.y + F.mat[2][0] * loc2.z;
-				ftx2[1] = F.mat[0][1] * loc2.x + F.mat[1][1] * loc2.y + F.mat[2][1] * loc2.z;
-				//ftx2[2] = F.mat[0][2] * loc2.x + F.mat[1][2] * loc2.y + F.mat[2][2] * loc2.z;
+				ftx2[0] = F.mat[0][0] * loc2.x + F.mat[1][0] * loc2.y + F.mat[2][0];
+				ftx2[1] = F.mat[0][1] * loc2.x + F.mat[1][1] * loc2.y + F.mat[2][1];
+				//ftx2[2] = F.mat[0][2] * loc2.x + F.mat[1][2] * loc2.y + F.mat[2][2];
```
ahojnnes: Remove commented line?

ahojnnes (approved):
I didn't check all the details of every single equation but the thorough testing and e2e benchmark give me confidence this is (mostly) all implemented correctly :-) I definitely didn't spend time on all the SiftGPU voodoo...

PR #4565, opened 2026-07-19
Use optional cameras if available at guided matching

B1ueber2y:
This does not affect a lot the behavior for shared focal solver, but would be crucial for the one-side focal case where the calibrated camera can be with any camera model. Without this PR (as in the pre-existing code), the guided matching would wrongly assume pinhole cameras for the already calibrated camera in the one-side focal case.

In preparation for https://github.com/colmap/colmap/pull/4561


ahojnnes (approved):
LGTM, thanks.

Issue #2587, opened 2024-06-04
Incomplete documentation: Convertion to/from OpenCV camera calibration parameters


federkamm:
The comment in the header of `sensor/models.h` clearly states that pixel centers are at half integer coordinates, e.g. top left pixel center is at `(0.5, 0.5)`.  OpenCV uses an image coordinate system with integer pixel centers.  Therefore, if OpenCV returns a camera calibration with calibration matrix

        [750   0 399.5]
    K = [  0 750 299.5]
        [  0   0   1  ]

the difference in origin should be adjusted and the `cameras.txt` should contain a line like

    1 OPENCV 800 600 750 750 400 300 ...

__Can you confirm this is correct?__

Or in other words, is it correct that the feature detector (VLFeat) returns image coordinates with pixel centers at half integer positions (or the returned positions are adjusted accordantly)?  The only reference I could find in the documentation <https://www.vlfeat.org/overview/sift.html#tut.sift.conventions> states that for the MATLAB version, the top left pixel would have coordinates (1,1) but I'm not sure what exactly that means and if it also applies to the C API.

# Suggested possible resolutions:

- State (e.g.) in comment to `OpenCVCameraModel` in `sensor/models.h` (and related classes) that the principal point `(cx, cy)` has to be adjusted by `(0.5, 0.5)` between COLMAP ond OpenCV.
- State (e.g.) at the same place, that no adjustment is necessary despite OpenCV using a different image coordinate system.

specarmi:
I was wondering about this as well. It seems Kalibr uses the same convention as OpenCV according to [this](https://github.com/ethz-asl/kalibr/issues/115). So the question applies to Kalibr calibration results as well.

Note that, in addition to the header file you pointed out, the COLMAP [documentation](https://colmap.github.io/tutorial.html#feature-detection-and-extraction) also states COLMAP's convention:
```
Note that by convention the upper left corner of an image has coordinate (0, 0) and the center of the upper left most pixel has coordinate (0.5, 0.5).
```

I think you are correct that the principal point returned by OpenCV/Kalibr should be increased by 0.5 before the calibration results are used with COLMAP.

Shaked-Q:
Can anyone confirm or deny the need to add 0.5 to the translations of the OpenCV calibration?

PR #4566, opened 2026-07-19
doc: Add FAQ on reusing OpenCV/Kalibr calibration parameters

behnamasadi:
## Summary

Adds a FAQ entry for a recurring question (#2587, also raised for Kalibr) on
reusing externally-calibrated intrinsics in COLMAP:

- **Pixel-coordinate convention:** COLMAP uses a corner origin (the top-left
  pixel center is at `(0.5, 0.5)`, a centered principal point is
  `(width/2, height/2)`), whereas OpenCV/Kalibr use pixel-center integer
  coordinates. The principal point therefore shifts by `+0.5` in both axes
  (`cx_colmap = cx_opencv + 0.5`); focal lengths and distortion coefficients
  transfer unchanged.
- **Distortion parameter order** for the `OPENCV`/`FULL_OPENCV` models.

Verified against `Camera::CreateFromModelId` / `CameraModelInitializeParams`
(principal point is initialized at `width/2, height/2`). Docs build cleanly with
Sphinx.

### DIFF
--- a/doc/faq.rst
+++ b/doc/faq.rst
@@ -91,6 +91,36 @@ with ``SIMPLE_RADIAL`` and inspect the reprojection errors in the model
 statistics.
 
 
+Using calibration from OpenCV, Kalibr, or other tools
+-----------------------------------------------------
+
+If you already calibrated your camera with an external tool such as OpenCV or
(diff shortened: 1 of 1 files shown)


ahojnnes (approved):
Great, thank you!

Issue #4485, opened 2026-06-28
TypeError: pycolmap._core.Database: No constructor defined!


Jacky56:
versions:
```text
pycolmap==4.1.0
```

when you write something like

```python
import pycolmap

pycolmap.Database() # behaviour changed to pycolmap.Database
```

you get:
```text
TypeError: pycolmap._core.Database: No constructor defined!
```

I raised a similar issue to https://github.com/colmap/colmap/issues/4422, is `pycolmap` no longer maintained?

sarlinpe:
As explained in the issue that you linked, the canonical way of opening a pycolmap database has changed to `pycolmap.Database.open(path)`.  We might have failed to appropriately communicate this in a previous release, apologies.

PR #4569, opened 2026-07-19
pycolmap: Document that Database must be opened via Database.open()

behnamasadi:
## Summary

`pycolmap.Database` exposes `Database.open(path)` as the only entry point but has
no constructor and no docstring, so `pycolmap.Database(path)` fails with the
cryptic `TypeError: pycolmap._core.Database: No constructor defined!` (#4485,
#4422 — a maintainer noted the `.open()` workflow was never documented).

This adds:
- a **class docstring** explaining that a `Database` must be opened with
  `Database.open(path)` and can be used as a context manager;
- a **docstring for the `open` static method**.

Docstrings only — no behavior change. Verified `Database::Open` uses
`SQLITE_OPEN_CREATE` (opens the file, creating it if necessary).

### DIFF
--- a/src/pycolmap/scene/database.cc
+++ b/src/pycolmap/scene/database.cc
@@ -453,11 +453,12 @@ void BindDatabase(py::module& m) {
       "the database on exit automatically:\n\n"
       "    with pycolmap.Database.open(path) as db:\n"
       "        ...");
-  PyDatabase.def_static("open",
-                        &Database::Open,
-                        "path"_a,
-                        "Open (creating it if necessary) the database file at "
-                        "the given path and return it.")
+  PyDatabase
+      .def_static("open",
+                  &Database::Open,
+                  "path"_a,
+                  "Open (creating it if necessary) the database file at "
+                  "the given path and return it.")
       .def("close", &Database::Close)
       .def("__enter__", [](Database& self) { return &self; })
       .def("__exit__", [](Database& self, const py::args&) { self.Close(); })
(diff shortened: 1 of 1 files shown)


src/pycolmap/scene/database.cc
```diff
@@ -442,8 +442,22 @@ class PyDatabaseImpl : public Database, py::trampoline_self_life_support {
...
+      m,
+      "Database",
+      "Database storing all extracted and matched information (cameras, "
+      "images, keypoints, descriptors, matches, rigs, and pose priors).\n\n"
+      "A Database cannot be constructed directly; open an existing or new "
+      "database file with ``Database.open(path)`` and release it with "
+      "``close()``. It can also be used as a context manager, which closes "
+      "the database on exit:\n\n"
+      "    with pycolmap.Database.open(path) as db:\n"
+      "        ...");
```
ahojnnes, line 455:
```suggestion
      "Database storing all extracted and matched information (cameras, "
      "images, keypoints, descriptors, matches, rigs, and pose priors).\n\n"
      "A Database cannot be constructed directly; instead open an existing "
      "or new database file with ``Database.open(path)`` and release it with "
      "``close()``. It can also be used with a context manager, which closes "
      "the database on exit automatically:\n\n"
      "    with pycolmap.Database.open(path) as db:\n"
      "        ...");
```

ahojnnes (approved):
Thank you!

Issue #3405, opened 2025-06-14
model_merger from cmd line reduces total number of images and points


YutaoC24:
![Image](https://github.com/user-attachments/assets/2f34243f-1ccd-40c5-b16a-23097b0fd3f0)
I am running model_merger from cmd line with two colmap sparse model inputs. 
For the two sparse models, they both have 90 images in total. They share 20 images with the same name and idx (aligned manually). The final merged result only has 90 images as well as fewer total 3D points. 
Is it possible for the merged result to include all 90 + 90 - 20 = 160 distinct images from both sparse models?

ahojnnes:
This sounds like a bug. Would you be able to share the two models to reproduce the issue?

htchr:
Hello,

I believe I am experiencing a similar error:

```bash
jack@ubuntu:/home/jack/$ colmap model_merger --input_path1 data/56/glo/sparse/0/ --input_path2 data/1/glo/spar
se/0/ --output_path merge/   
I20250915 03:35:11.213244 24262 misc.cc:51] 
Reconstruction 1
----------------
I20250915 03:35:11.213285 24262 model.cc:739] Images: 117
I20250915 03:35:11.213292 24262 model.cc:740] Points: 27491
I20250915 03:35:11.214370 24262 misc.cc:51] 
Reconstruction 2
----------------
I20250915 03:35:11.214376 24262 model.cc:745] Images: 4
I20250915 03:35:11.214378 24262 model.cc:746] Points: 1447
I20250915 03:35:11.214382 24262 misc.cc:51] 
Merging reconstructions
-----------------------
I20250915 03:35:11.239650 24262 model.cc:751] => Merge succeeded
I20250915 03:35:11.239674 24262 misc.cc:51] 
Merged reconstruction
---------------------
I20250915 03:35:11.239677 24262 model.cc:753] Images: 118
I20250915 03:35:11.239686 24262 model.cc:754] Points: 1917
```

I have two models that were generated from the same database.
They overlap by 3 images, so the number of images is correct, but most of the points from the first model are discarded.
The data can be downloaded [here](https://drive.google.com/drive/folders/1XTcU6pZ1x8qOdfWnc9TXJyn7rVrU4X9x?usp=sharing) if you would like to test.

Thank you,

vigneshrk29:
Hi, sorry, has there been any update on this? I am facing similar issues to the above. I have also tried interchanging input_folders1 and 2 as suggested in issue #3175 but I have had no luck. 

[...]

Issue #3218, opened 2025-03-13
colmap model_merger  ID match error occurs


sanqianlouhua:
I'm trying to merge two datasets together, but I'm having a problem with the wrong id, and I don't know how to fix he

Problem Alert:
root@025a23cb81ec:/# colmap model_merger --input_path1 /7904/COLMAP_0_26/sparse/0/ --input_path2 /7904/COLMAP_18_47/sparse/0/ --output_path /7904/47_views/sparse/0/                            
I0313 09:51:46.920598 122950 misc.cc:205] 
Reconstruction 1
----------------
I0313 09:51:46.920643 122950 model.cc:728] Images: 27
I0313 09:51:46.920655 122950 model.cc:729] Points: 7326
I0313 09:51:46.946424 122950 misc.cc:205] 
Reconstruction 2
----------------
I0313 09:51:46.946431 122950 model.cc:734] Images: 30
I0313 09:51:46.946436 122950 model.cc:735] Points: 10951
I0313 09:51:46.946439 122950 misc.cc:205] 
Merging reconstructions
-----------------------
E0313 09:51:46.946496 122950 alignment.cc:76] Check failed: src_images[i]->ImageId() == tgt_images[i]->ImageId() (23 vs. 5) 
terminate called after throwing an instance of 'std::invalid_argument'
  what():  [alignment.cc:76] Check failed: src_images[i]->ImageId() == tgt_images[i]->ImageId() (23 vs. 5)

tsattler:
Best guess (as not enough information is provided): You are running reconstructions with different databases, with some images being part of multiple databases. When you try to merge models reconstructed from different databases, Colmap realizes that some images do not have a consistent `image_id` and then throws an error.

WFram:
As the previous comment says, if you're trying to merge two reconstructions that were built using different databases, you will get mismatches between the indices. When using model merger, what you need to make sure is that both reconstructions have common registered images, having the same image id. And the indices of all the other images should not interfere. I had a similar situation when I couldn't avoid using two separate databases, so I wrote a script to reindex images and points in the source and target reconstructions.

Having two reconstructions built using different databases, I have to do the following:
1. Merging databases into a single database, adding prefixes to the images. For example, for a source reconstruction prefix `1/` will be added, and for a target one, prefix `2/` will be added. Then running an exhaustive matcher on the resulting database (or based on custom candidates).
2. Adding the same image prefixes to the source and target COLMAP reconstructions.
[...]

Issue #3175, opened 2025-02-21
Merged models, but give less points


HungNgoCT:
Hi.,

Thank you for the work.

I scanned a loop model, assume with the path A-B-C-D where A and D is a close loop.

I reconstructed the model A-B-C-D successfully and got 1,667,157 points, call model ABCD.

Then, I (at close loop) extracted 60 images from location A, and 60 images from location B, call segment AD. Reconstructed segment AD, I got 143,784 points

After that, I use: model_merger --input_path1 <segment AD> --input_path2 <ABCD>, I got just 1,653,970 points. It is less than 1,667.157 points. (Image_1)

May I ask: Why I merged more points from segment AD to the model ABCD, but got less points?

But when I revert the other to merge, say: model_merger --input_path1 <ABCD> --input_path2 <segment AB>, I got 1,691,790 points. It is more points than model ABCD (Image_2).

Anyone please help me to understand this. Thank you

![Image](https://github.com/user-attachments/assets/f465614e-7f1e-41ea-908c-1a71cee68703)
Image_1: Less points in merged models

![Image](https://github.com/user-attachments/assets/95eb2670-eb3d-41c1-8686-910862d22d47)
Image_2: more points

cl886699:
same problem

htchr:
Same

Dawars:
I have the same problem.

TL;DR
"Order of inputs matter. The output pcd is cut to the extent of the _input_path2_ pcd."

I used hierarchical_mapper, then manually removed some cameras using image_deleter and merged that model with a snapshot of a full mapper run.

The cameras and points seem fine in the merged model but the left side of the pcd is missing. The missing points are outside of approximately pcd --input_path2. If I change the order of the models, this issue is not noticable/present. This suggests that the output pcd is cut to the extent of the **input_path2** pcd. 

Command: `colmap model_merger --input_path1 colmap/models/0_clean/ --input_path2 colmap/snapshots/1759784355293/ --output_path colmap/models/merged/`

I'm on the latest commit: https://github.com/colmap/colmap/commit/fd25e1a42c2377446330bd7dfd9ee9d5e2f3278e

snapshot:
<img width="1108" height="673" alt="Image" src="https://github.com/user-attachments/assets/9a0b3d89-c7e4-4220-bba7-4e95b5b982c9" />

0_clean:
<img width="1122" height="614" alt="Image" src="https://github.com/user-attachments/assets/d892a859-674b-4984-a29c-a0881312da5d" />

[...]

PR #4570, opened 2026-07-19
Reject reconstruction merges with inconsistent image id<->name mappings

behnamasadi:
## Summary

`MergeReconstructions` classifies each registered source image as "common" or
"missing" purely by image id: if the target already contains that id, the image
is assumed to be the same physical image and is not copied. This is only valid
when both reconstructions share a consistent image id<->name mapping (i.e. were
derived from the same database).

When the two reconstructions come from independent databases that both number
their images `1..N`, distinct physical images collide on id. They are
misclassified as common, never copied, and their tracks are treated as
conflicting — silently dropping images and 3D points from the merged model
(e.g. 90 instead of 160 images). Reported in #3405, #3218, #3175.

The preceding alignment step already matches images by name
(`Reconstruction::FindCommonRegImageIds`), so any merge that reaches the
id-based classification with an inconsistent id<->name mapping produces a
corrupt result. This PR detects that case by comparing the source and target
image names for each colliding id and fails loudly with a clear error message,
instead of silently corrupting the merge.

Full cross-database merging (remapping colliding ids by name) is left as a
follow-up.

## Test

Adds a unit test covering the id-collision case, asserting the merge is rejected
and the target reconstruction is left unmodified.

### DIFF
--- a/src/colmap/estimators/alignment.cc
+++ b/src/colmap/estimators/alignment.cc
@@ -484,32 +484,26 @@ bool MergeReconstructions(const double max_reproj_error,
 
   // Find common and missing images in the two reconstructions. Images are
   // matched by image id, which assumes that both reconstructions share a
-  // consistent image id<->name mapping (i.e. were derived from the same
+  // consistent image_id<->name mapping (i.e. were derived from the same
   // database). If this assumption is violated -- e.g. the reconstructions were
   // built from independent databases that both number their images 1..N -- then
-  // distinct physical images end up with colliding ids. Merging by id would
-  // then silently treat these distinct images as already-present, dropping
-  // images and corrupting tracks (see issues #3405, #3218, #3175). Detect the
-  // inconsistency via the image name and fail loudly instead. Note that the
-  // preceding alignment step already matches images by name (via
-  // Reconstruction::FindCommonRegImageIds), so a merge that reaches this point
-  // with inconsistent id<->name mappings would otherwise produce a corrupt
-  // result.
-  FlatHashSet<image_t> common_image_ids;
-  common_image_ids.reserve(src_reconstruction.NumRegImages());
-  FlatHashSet<image_t> missing_image_ids;
-  missing_image_ids.reserve(src_reconstruction.NumRegImages());
-  for (const image_t image_id : src_reconstruction.RegImageIds()) {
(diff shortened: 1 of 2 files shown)


src/colmap/estimators/alignment.cc
```diff
@@ -482,13 +482,36 @@ bool MergeReconstructions(const double max_reproj_error,
...
+      const std::string& tgt_name = tgt_reconstruction.Image(image_id).Name();
+      if (src_name != tgt_name) {
+        LOG(ERROR)
+            << "Cannot merge reconstructions: image id " << image_id
+            << " refers to \"" << src_name << "\" in one reconstruction but \""
+            << tgt_name << "\" in the other. MergeReconstructions requires both "
+            << "reconstructions to share a consistent image id<->name mapping "
+            << "(i.e. be derived from the same database).";
+        return false;
+      }
```
ahojnnes, line 509:
```suggestion
  // Find common and missing images in the two reconstructions. Images are
  // matched by image id, which assumes that both reconstructions share a
  // consistent image_id<->name mapping (i.e. were derived from the same
  // database). If this assumption is violated -- e.g. the reconstructions were
  // built from independent databases that both number their images 1..N -- then
  // distinct physical images end up with colliding ids. Detect the
  // inconsistency via the image name and fail loudly instead.
FlatHashSet<image_t> common_image_ids;
common_image_ids.reserve(src_reconstruction.NumRegImages());
FlatHashSet<image_t> missing_image_ids;
missing_image_ids.reserve(src_reconstruction.NumRegImages());
for (const image_t image_id : src_reconstruction.RegImageIds()) {
if (tgt_reconstruction.ExistsImage(image_id)) {
      const std::string& src_name = src_reconstruction.Image(image_id).Name();
      const std::string& tgt_name = tgt_reconstruction.Image(image_id).Name();
      if (src_name != tgt_name) {
        LOG(ERROR)
            << "Cannot merge reconstructions: image_id=" << image_id
            << " refers to \"" << src_name << "\" in the source reconstruction but \""
            << tgt_name << "\" in the target. MergeReconstructions requires both "
            << "reconstructions to share a consistent image_id<->name mapping "
            << "(i.e., be derived from the same database).";
        return false;
      }
```

ahojnnes (approved):
Thanks!

PR #4579, opened 2026-07-20
ui: Color camera frustums by reprojection error in the model viewer

behnamasadi:
## Summary

Adds a **"Reprojection error"** option to the *Image colormap* dropdown in the
model viewer, coloring each camera frustum by its mean reprojection error. This
mirrors the existing point-cloud **"Error"** colormap (`PointColormapError`) —
extending the same idea from 3D points to the cameras that observe them — so you
can spot which cameras the solver struggled with directly in the 3D view.

The default colormap is unchanged ("Uniform color"); this is an opt-in choice.
No new dependencies; ~160 lines across `colormaps.{h,cc}` and
`render_options_widget.{h,cc}`.

## How it works

There are really just **three steps**: get a number per camera, squash that
number to 0–1, then look up a color.

### Step 1 — Each camera gets one number: its mean reprojection error (in pixels)

- **Reprojection error** = for a single 3D point seen in a photo, project that 3D
  point back into the image and measure how far (in pixels) it lands from the
  actual detected feature. Small = the 3D model agrees with the photo; large =
  it doesn't.
- A camera sees hundreds of 3D points, each with its own error. **The camera's
  number = the average of those errors.** So each frustum has one value like
  "0.64 px."

For the `south-building` sample model, those per-camera numbers range
**0.47 → 0.79 px** (reported in the log when the colormap is applied).

### Step 2 — Turn that number into a fraction from 0 to 1

This is the *only* place the **max** (red threshold) matters:

```
t = camera_error / max        (then clamp to the 0–1 range)
```

- `t = 0`  → the bottom of the scale
- `t = 1`  → the top of the scale (error is at or above `max`)

### Step 3 — Look up a color for `t` on the "jet" scale

`t` is fed into the existing `JetColormap`:

```
t:     0.0      0.25      0.5       0.75      1.0
color: blue  →  cyan  →  green  →  yellow  →  red
```

That's the whole algorithm. **The error doesn't pick the color directly — it's
error ÷ max that does.** That is why the same camera changes color when you move
the slider.

### Worked example

**At max = 2.0 px** (default):

| camera | t = err ÷ 2.0 | color |
|--------|---------------|-------|
| 0.47 (best)  | 0.24 | blue |
| 0.64 (mean)  | 0.32 | cyan |
| 0.79 (worst) | 0.40 | cyan-green |

→ everything lands in the blue–cyan zone (all `t` are small because 2 px is far
bigger than any real error). That's the "all blue = all good" view.

**At max = 0.8 px:**

| camera | t = err ÷ 0.8 | color |
|--------|---------------|-------|
| 0.47 (best)  | 0.59 | green |
| 0.64 (mean)  | 0.80 | orange |
| 0.79 (worst) | 0.99 | red |

→ now the same 0.47–0.79 range is stretched across green→red, so you see the
differences.

**So:** the colors aren't tied to absolute "good/bad" — they're tied to *where
each camera falls between 0 and the `max` you chose*. Setting `max` near your
worst camera (~0.8) spreads the rainbow across your data; setting it high (2 px)
squashes everyone to blue unless something is genuinely bad.

### Design choice: absolute vs. relative scale

An earlier version normalized colors to each model's own min/max error. That was
misleading: on a well-registered model where every camera is ~0.5 px, it still
painted a full blue→red rainbow, implying some cameras were bad when none were.
The **absolute** scale above keeps colors comparable across models and reserves
red for genuinely high error. The `max` is user-adjustable (with a tooltip), and
applying the colormap logs the model's actual per-image error range so the scale
maps to concrete numbers.

## Screenshots

The same `south-building` model with **"Reproj. error for red [px]"** swept from
a small value to the 2 px default. The *identical* per-camera errors sweep across
the whole colormap — showing that the color reflects `error ÷ max`, not an
absolute good/bad verdict:

**Small max → the whole scale is used up almost immediately, so every camera is red:**

max well below the data — all red

raising max — red/orange/yellow

max near the data range — green/yellow/orange spread

larger max — cyan/green

**Default max = 2 px → all cameras well under tolerance, so all blue/cyan ("all good"):**

default max = 2 px — all blue/cyan

## Notes

- Jet is used for consistency with the existing point "Error" colormap; a
  colorblind-safe coolwarm ramp is an easy swap if preferred.
- Built with `-DGUI_ENABLED=ON` and verified visually. GUI code is not covered by
  the headless unit tests.

### DIFF
--- a/src/colmap/ui/colormaps.cc
+++ b/src/colmap/ui/colormaps.cc
@@ -32,6 +32,11 @@
 #include "colmap/sensor/bitmap.h"
 #include "colmap/util/hash_containers.h"
 
+#include <algorithm>
+#include <unordered_map>
+#include <utility>
+#include <vector>
+
 namespace colmap {
 
 PointColormapBase::PointColormapBase()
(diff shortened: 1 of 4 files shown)


behnamasadi:
@ahojnnes, what do you think of this?

ichsan2895:
I like the idea, it is makes me remember with Lichtfeld Studio that colored the camera frustums with loss/error

behnamasadi:
@ahojnnes Rebased on current main, CI is green, and the diff is contained to four files in `src/colmap/ui/` (`colormaps.{h,cc}`, `render_options_widget.{h,cc}`) — no new dependencies and no change to the default view, since "Uniform color" stays the default and this is an opt-in entry in the existing *Image colormap* dropdown.

The idea is the direct analogue of the point-cloud "Error" colormap that's already there: same jet ramp, same absolute-threshold scaling, just applied to the cameras that observe the points instead of the points themselves. It answers "which images did the solver struggle with?" at a glance, which today means reading per-image statistics out of the log. @ichsan2895 mentioned above that Lichtfeld Studio does the same thing, so the interaction is familiar to people coming from other tools.

One deliberate design note, in case it's the part worth arguing about: the scale is absolute (`error ÷ user-set max`), not normalized to each model's own min/max. Per-model normalization looked better in screenshots but was actively misleading — a well-registered model where every camera sits at ~0.5 px still got painted a full blue→red rainbow. With an absolute threshold, red means genuinely high error and colors stay comparable between models. The threshold is exposed in the UI with a tooltip, and applying the colormap logs the model's actual error range so the number maps to something concrete.

Happy to swap jet for a colorblind-safe coolwarm ramp, or to change the default threshold, if you'd prefer either. If you'd rather not carry this in the GUI at all, just say so and I'll close it — no hard feelings.

PR #4581, opened 2026-07-21
Add browser-based 3D reconstruction viewer

ahojnnes:
## Summary
- add a local-only Three.js viewer for COLMAP sparse binary reconstructions
- support folder selection and drag-and-drop, camera and point inspection, source images, reprojections, and native-style viewer controls
- expose an instance-safe ES module component with scoped CSS and `load`, `clear`, and `dispose` lifecycle APIs
- add the 3D Viewer documentation page, reuse and attribution guidance, frontend build integration, and unit/browser CI coverage
- cancel superseded parser workers and clear stale scene state on failed loads

## Testing
- `./node_modules/.bin/tsc --noEmit`
- `npm test` (14 tests)
- `PYENV_VERSION=colmap make html SPHINXOPTS="-W --keep-going"`


doc/viewer_src/main.ts
```diff
@@ -0,0 +1,563 @@
...
+    inspector.innerHTML = `<div class="viewer-inspector-empty"><h2>Inspector</h2><p>Double-click a point or camera to inspect it.</p></div>`;
+  };
+
+  const displayReconstruction = (parsed: Reconstruction): void => {
+    if (disposed) throw new Error("Cannot load a disposed COLMAP viewer");
+    viewer.setReconstruction(parsed);
+    reconstruction = parsed;
+    drop.hidden = true;
+    setStatus(null);
+    stats.textContent = `${parsed.images.size.toLocaleString()} cameras / ${viewer.visiblePointCount.toLocaleString()} visible points`;
```
B1ueber2y: "cameras" is the wrong word. Here it labels reconstruction.images.size as "cameras". In COLMAP, a camera is an intrinsics record -  reconstruction.cameras is a different, much smaller map. On an official COLMAP page this is actively misleading. Should be "images" (and, with rigs, arguably "frames").

doc/viewer_src/camera_models.ts
```diff
@@ -0,0 +1,182 @@
+import type {Camera, Vec2, Vec3} from "./types";
```
B1ueber2y, line 1: A second, hand-written copy of COLMAP's camera model registry. camera_models.ts:24 (CAMERA_MODEL_PARAM_COUNTS) and project() duplicate sensor/models.h with nothing enforcing sync. Consequences when someone adds or edits a model in C++:

When param count changes → the viewer silently misparses every subsequent record in cameras.bin. parser.ts uses the table to decide how many doubles to consume, so a wrong count desynchronizes the stream. It may still "succeed" (trailing-data check only catches it if the total happens to mismatch) and render garbage.

camera_models.test.ts can't catch this either, as it asserts the TS table against itself. The same drift risk applies to parser.ts vs reconstruction_io_binary.cc for any format change. So a Python test that imports pycolmap and asserts the TS table against pycolmap.CameraModelId / num-params, parsed out of the .ts file. Cheap, and it turns a silent corruption into a red build.

  (The alternative — compiling the C++ reader to WASM — is obviously a much bigger lift and I think choosing against it is reasonable. It's just worth having the trade-off stated in the PR.)

doc/viewer_src/viewer.ts
```diff
@@ -0,0 +1,626 @@
...
+    this.coordinateGrid.visible = false;
+    this.coordinateAxes.visible = false;
+    this.onSelection(null);
+  }
+
+  get visiblePointCount(): number {
+    return this.pointDrawList.length;
+  }
+
+  updateSettings(settings: Partial<ViewerSettings>): void {
```
B1ueber2y, line 319: `updateSettings` has a fast path only for `pointSize`-alone, so every other setting triggers a full `rebuild()` — re-filtering all of `points3D`, reallocating every geometry buffer, and re-emitting the picking scene.

  Camera size is a `<input type="range">` bound to `input`, so a single drag fires this dozens of times per second. Same for the Alt-scroll handler at line 620, and for the min-track / max-error inputs. On anything past a toy model this locks the tab for the duration of the drag.

  `cameraSize` only affects frustum geometry, so it can take a fast path like `pointSize` does. The two filter settings genuinely need a rebuild — those want debouncing (~150ms trailing) rather than per-event.

doc/viewer_src/main.ts
```diff
@@ -0,0 +1,563 @@
...
+    const entries = [...(input.files ?? [])].map((file) => ({path: file.webkitRelativePath || file.name, file}));
+    input.value = "";
+    void acceptEntries(entries).catch(() => undefined);
+  }, listenerOptions);
+  modelSelect.addEventListener("change", () => {
+    const candidate = candidates[Number(modelSelect.value)];
+    if (candidate) void loadCandidate(candidate).catch(() => undefined);
+  }, listenerOptions);
+
+  for (const eventName of ["dragenter", "dragover"] as const) container.addEventListener(eventName, (event) => {
```
B1ueber2y, line 222: Drop overlay gets stuck. Here it sets drop.hidden = false on dragenter/dragover, but the dragleave handler only removes the is-dragging class — it never restores hidden. So: load a model → drag any file over the canvas → drag away without dropping → the "Open a COLMAP reconstruction" panel now permanently covers your reconstruction, with no way to dismiss it short of reloading a model.

doc/viewer_src/viewer.ts
```diff
@@ -0,0 +1,626 @@
...
+  }
+
+  private switchProjection(): void {
+    const previous = this.activeCamera;
+    this.activeCamera = this.settings.projection === "perspective" ? this.perspectiveCamera : this.orthographicCamera;
+    this.activeCamera.position.copy(previous.position);
+    this.activeCamera.quaternion.copy(previous.quaternion);
+    this.activeCamera.up.copy(previous.up);
+    this.controls = this.createControls(this.activeCamera);
+    this.controls.target.set(0, 0, 0);
```
B1ueber2y: Projection toggle loses the orbit pivot. Here it does this.controls.target.set(0, 0, 0) after copying camera position/orientation. Switching perspective↔orthographic snaps the pivot to the scene origin instead of preserving the current target (which resetView correctly sets to viewCenter). The camera stays put but rotation suddenly orbits the wrong point.

doc/viewer_src/viewer.ts
```diff
@@ -0,0 +1,626 @@
...
+    const controls = new OrbitControls(camera, this.canvas);
+    controls.enableDamping = true;
+    controls.dampingFactor = 0.08;
+    controls.screenSpacePanning = true;
+    controls.mouseButtons.LEFT = THREE.MOUSE.ROTATE;
+    controls.mouseButtons.RIGHT = THREE.MOUSE.PAN;
+    return controls;
+  }
+
+  private animate = (): void => {
```
B1ueber2y, line 234: No WebGL context-loss recovery. viewer.ts:188's animate() cancels the RAF and reports on error, and never restarts. No webglcontextlost/webglcontextrestored handlers. On a long-lived docs page (tab backgrounded, GPU reset, driver hiccup) the viewer dies permanently with an error toast.

doc/viewer_src/main.ts
```diff
@@ -0,0 +1,563 @@
...
+  while (true) {
+    const batch = await new Promise<FileSystemEntry[]>((resolve, reject) => reader.readEntries(resolve, reject));
+    if (batch.length === 0) break;
+    children.push(...batch);
+  }
+  const nested = await Promise.all(children.map((child) => readEntry(child, `${path}/${child.name}`)));
+  return nested.flat();
+}
+
+function findImageFile(files: Map<string, File>, name: string): File | null {
```
B1ueber2y: Minor: findImageFile is a linear scan over all dropped files, called once per track element — build a suffix index instead.
  viewer.ts:416's camera.modelId === 17 ? 0.5 : ... is dead code (equirect is routed to sphericalCameraGeometry before cameraGeometry is ever called). imageDrawList is written but never read outside the method that sets it. resize() hardcodes 25 for the FOV instead of reading perspectiveCamera.fov.

doc/Makefile
```diff
@@ -48,18 +51,27 @@ help:
...
+html: viewer-assets
 	$(SPHINXBUILD) -b html $(ALLSPHINXOPTS) $(BUILDDIR)/html
 	@echo
 	@echo "Build finished. The HTML pages are in $(BUILDDIR)/html."
 
-dirhtml:
+serve: html
+	$(PYTHON) -m http.server $(PORT) --directory $(BUILDDIR)/html
+
+dirhtml: viewer-assets
```
B1ueber2y, line 73: make html now hard-requires Node

  What changed. doc/Makefile gained a new target and made the HTML targets depend on it:

  viewer-assets:
        command -v npm >/dev/null 2>&1 || (echo "npm is required to build the interactive viewer"; exit 1)
        test -d node_modules || (echo "Run 'npm ci --prefix doc' from the repository root first"; exit 1)
        npm run build

  html: viewer-assets          # ← and dirhtml, singlehtml
        $(SPHINXBUILD) -b html ...

Make runs prerequisites first, so if either guard trips, make html exits before Sphinx ever runs. You don't get docs-without-viewer; you get no docs.

  Why that's a regression. Before this PR the contract was: pip install -r doc/requirements.txt → make -C doc html. After it, anyone fixing a typo in an .rst file also needs Node + npm installed, plus a manual out-of-band npm ci --prefix doc that pulls the whole tree behind that 1,818-line lockfile from the npm
  registry. The Makefile won't even run it for you — it just tells you to.

And because doc/_static/viewer/ is in .gitignore, there's no prebuilt fallback in the repo. So a source-tarball or offline/air-gapped docs build can't produce any documentation, not just no viewer.

**Suggested fix** — degrade instead of abort:

  viewer-assets:
        if command -v npm >/dev/null 2>&1 && test -d node_modules; then \
          npm run build; \
        else \
          echo "WARNING: skipping 3D viewer build (npm or node_modules not found)"; \
        fi

  The viewer page then renders its <noscript> placeholder and everything else builds normally. If you want CI to stay strict, gate it on a variable (make
  html STRICT_VIEWER=1) rather than on the developer's environment.

  Related inconsistency: doc/make.bat calls bare npm run build under if "%1" == "html" with no guard at all, so Windows users get a raw 'npm' is not recognized + exit /b 1 instead of the Makefile's friendly message. Whatever policy you land on, the two should match.

doc/package.json
```diff
@@ -0,0 +1,22 @@
+{
+  "name": "colmap-documentation-viewer",
+  "private": true,
+  "type": "module",
+  "scripts": {
```
B1ueber2y, line 5: tsc --noEmit isn't actually enforced

  The gap. The PR description lists ./node_modules/.bin/tsc --noEmit under Testing. But:

  - doc/package.json scripts are only build, pretest, test, test:browser — there's no typecheck.
  - The CI step runs npm ci → npm test → playwright install → npm run test:browser. No tsc anywhere.

  Why that means zero type checking. Vite compiles TypeScript with esbuild, which strips types without checking them; Vitest does the same. So type errors
  don't fail the build, don't fail the tests, and don't fail CI. tsc ran once on the author's laptop and that's the entire enforcement.

  That's worth more here than in a typical project, because doc/tsconfig.json sets strict: true and noUncheckedIndexedAccess: true. That second flag is
  exactly why camera_models.ts is full of p[0]! and p[8] ?? 0 — it's the mechanism that catches a wrong index into a params array. Right now nothing checks
  that the assertions are still honest after an edit.

  Suggested fix — two lines:

  "scripts": {
    "typecheck": "tsc --noEmit",
    "build": "vite build",
    ...
  }

        - name: Install and test viewer
          run: |
            npm ci --prefix doc
            npm run typecheck --prefix doc   # ← add
            npm test --prefix doc
            npx --prefix doc playwright install --with-deps chromium
            npm run test:browser --prefix doc

B1ueber2y (approved):
Thank you! The offline demo looks convincing. The PR is so long full of typescripts so I had a pass with my AI and pasted directly those I think look reasonable.

doc/viewer_src/types.ts
```diff
@@ -0,0 +1,61 @@
...
+  camFromWorld: Rigid3d;
+  points2D: Point2D[];
+}
+
+export interface TrackElement {
+  imageId: number;
+  point2DIdx: number;
+}
+
+export interface Point3D {
```
B1ueber2y, line 47: **Memory layout won't survive a full-size reconstruction.**

The concrete issue is in `parser.worker.ts`: `self.postMessage({ok: true, reconstruction})` has no transfer list, so the entire object graph is structured-c>

That would be survivable if the representation were dense, but it's boxed throughout. Against the C++ structs:

| | C++ | JS |
|---|---|---|
| `Point3D` | **64 B** (`Vector3d` 24 + `double` 8 + `Track` 24 + `Vector3ub` 3, padded) | ~600 B (object shell + heap BigInt id + two JSArray shells over `>
| `TrackElement` | **8 B**, contiguous `std::vector` | ~40 B object, pointer-chased through a `FixedArray` |
| `Point2D` | **24 B** (`Vector2d` 16 + `point3D_t` 8), contiguous | ~130 B (object + `xy` JSArray + heap BigInt) |

At mean track length 6 that's 112 B/point native vs. ~600 B in V8, so a 2M-point model is ~250 MB in the GUI and ~1.2 GB here — roughly 2.4 GB at the clone >

Third copy on top: `buildPoints` accumulates `positions`/`colors` as plain `number[]` and then hands them to `Float32BufferAttribute`, which copies into a f>

Suggestions, in descending order of value:

1. **Parse `points3D` into typed arrays and transfer them** — `Float64Array` xyz, `Uint8Array` colors, `Float32Array` errors, `BigUint64Array` ids, and flat>
2. **Same treatment for `points2D`** if (1) isn't enough — that's the other structure that scales with observations rather than images.
3. **Independent of both:** build the geometry into a preallocated `Float32Array` sized to the filtered count instead of growing a `number[]` and copying. C>

If none of that is in scope for this PR, the alternative is to state the intended model size in `viewer.rst` and fail with a clear message above some thresh>

doc/viewer_src/viewer.ts
ahojnnes: Updated the settings paths in 241a9ed84. Point size updates material uniforms, camera size is now a shader uniform with no geometry allocation, projection changes only switch cameras, and connection changes only rebuild their group. The two point filters remain bound to `change`, so their required full rebuild happens on committed input rather than every keystroke/drag event.

doc/viewer_src/main.ts
ahojnnes: Fixed and covered by the browser test. Leaving the host during a drag now removes the dragging state and hides the overlay again when a reconstruction is loaded.

doc/viewer_src/camera_models.ts
ahojnnes: Added `doc/tests/camera_models_registry_test.py`, which parses the TypeScript name/count tables and compares every model ID, name, and parameter count against the freshly built `pycolmap`. The docs CI job now runs this check.

doc/viewer_src/viewer.ts
ahojnnes: Added `webglcontextlost`/`webglcontextrestored` handling. Loss prevents the default teardown, stops the RAF, and shows status; restoration resets renderer state, resizes, clears status, and restarts animation. The lifecycle is also covered in the browser test.

doc/viewer_src/main.ts
ahojnnes: Addressed all four items: image files now use a precomputed suffix index, the equirectangular branch was removed from `cameraGeometry`, `imageDrawList` is used to decode camera pick IDs without a per-item pick table, and orthographic sizing reads `perspectiveCamera.fov`.

doc/Makefile
ahojnnes: Changed both build scripts to degrade cleanly when npm or `doc/node_modules` is unavailable. Unix and Windows now match, `STRICT_VIEWER=1` retains an opt-in hard failure, and the viewer page includes a visible unavailable-build fallback that the JS replaces when assets exist.

doc/viewer_src/viewer.ts
ahojnnes: Fixed. `switchProjection()` now snapshots the current controls target, recreates the controls, restores that target, and updates the controls before resizing.

doc/package.json
ahojnnes: Added the `typecheck` package script and wired `npm run typecheck --prefix doc` into the docs CI job before tests.

doc/viewer_src/main.ts
ahojnnes: Fixed in 241a9ed84. The toolbar now labels `reconstruction.images.size` as images and reports visible points explicitly.

doc/viewer_src/types.ts
ahojnnes: Implemented the compact path in 241a9ed84. Points and observations are stored in typed structure-of-arrays buffers, point IDs remain 64-bit, tracks are flat offset arrays, all buffers are transferred from the worker, and render buffers are preallocated. Point objects are materialized only for an actual selection, and picking no longer retains an object per point. The parser tests also exercise transfer/detachment and unsorted-ID lookup.

PR #4584, opened 2026-07-21
ui: Show camera viewing ray when hovering a pixel in the image viewer

behnamasadi:
## Summary

Adds a rerun-style interaction to the model viewer: **hovering a pixel in a
camera's image draws that pixel's back-projected viewing ray in the 3D view.**

- Double-click a camera frustum → its photo opens in the image window (existing).
- **Hover a pixel in that photo → a ray shoots from the camera center through the
  pixel into the 3D scene**, updating live as the cursor moves and clearing when
  it leaves the image.
- If the hovered pixel is near an observed keypoint, the ray **snaps to and
  highlights the corresponding 3D point** (yellow); otherwise it extends across
  the scene along the pixel's direction.

This makes the 2D image and the 3D reconstruction directly cross-referenceable:
point at anything in a photo and see where it lies in 3D.

## Demo

hover-pixel-ray demo

## How it works

1. **Capture the hovered pixel** — the image window installs an event filter on
   its `QGraphicsView` viewport (mouse tracking on) and maps the cursor to image
   pixel coordinates.
2. **Unproject to a world ray** — `Camera::CamFromImg(pixel)` gives the normalized
   camera-frame direction `(x, y, 1)`, transformed by the image pose
   (`Inverse(image.CamFromWorld())`) into a world-space ray from
   `image.ProjectionCenter()`.
3. **Snap to a 3D point** — the nearest observed `Point2D` with a `point3D_id`
   within ~1% of the image size is found; if present, the ray ends exactly at
   that 3D point and highlights it.
4. **Draw in 3D** — a dedicated `LinePainter` (ray) and `PointPainter` (highlight)
   render in `ModelViewerWidget::paintGL`, using the same
   `model_scale_ * (world + model_origin_)` transform as the rest of the scene.

## Notes

- Files: `model_viewer_widget.{h,cc}` (SetHoverRay/ClearHoverRay + painters),
  `image_viewer_widget.{h,cc}` (event filter + unprojection). `image_viewer_widget.h`
  moves `graphics_scene_`/`graphics_view_` from private to protected so the
  database image viewer can attach the hover filter.
- GUI feature — built with `-DGUI_ENABLED=ON` and verified interactively; not
  covered by the headless unit tests. The ray geometry was checked numerically
  (computed pixel direction matches the camera→3D-point direction).

## Note on the second commit (cherry-picked)

For convenience this branch also carries a second commit,
**"ui: Add reprojection-error image colormap for camera frustums"**, cherry-picked
from #4579, so both GUI viewer features can be built and demoed from a single
branch. That colormap change is reviewed separately in **#4579** — happy to drop
the commit here and keep this PR scoped to the hover-ray change alone if you
prefer.

### DIFF
--- a/src/colmap/ui/colormaps.h
+++ b/src/colmap/ui/colormaps.h
@@ -174,4 +174,29 @@ class ImageColormapNameFilter : public ImageColormapBase {
       image_name_colors_;
 };
 
+// Color images according to their mean reprojection error, on an *absolute*
+// jet scale (0 px = blue, `max_error` px = red). Unlike a relative min/max
+// normalization, this keeps the colors comparable across models: a
+// well-registered model stays blue/cyan, and only genuinely high-error cameras
+// turn red, instead of stretching a tiny error spread over the whole spectrum.
+class ImageColormapReprojectionError : public ImageColormapBase {
+ public:
+  void Prepare(NodeHashMap<camera_t, Camera>& cameras,
+               NodeHashMap<image_t, Image>& images,
+               NodeHashMap<point3D_t, Point3D>& points3D,
+               std::vector<image_t>& reg_image_ids) override;
+
+  void ComputeColor(const Image& image,
+                    Eigen::Vector4f* plane_color,
+                    Eigen::Vector4f* frame_color) override;
+
+  // Errors at or above this value (in pixels) map to the top of the scale
+  // (red). User-adjustable so the sensitivity can be dialed per scene.
+  float max_error = 2.0f;
+
+ private:
+  // Mean reprojection error per image, accumulated from the 3D point tracks.
+  FlatHashMap<image_t, float> image_errors_;
+};
+
 }  // namespace colmap
--- a/src/colmap/ui/image_viewer_widget.cc
+++ b/src/colmap/ui/image_viewer_widget.cc
@@ -29,8 +29,12 @@
 
 #include "colmap/ui/image_viewer_widget.h"
 
+#include "colmap/geometry/rigid3.h"
 #include "colmap/ui/model_viewer_widget.h"
 #include "colmap/util/file.h"
+
+#include <algorithm>
+#include <optional>
 
 namespace colmap {
 
@@ -259,6 +263,12 @@ DatabaseImageViewerWidget::DatabaseImageViewerWidget(
       options_(options) {
   setWindowTitle("Image information");
 
+  // Track the cursor over the image so a viewing ray for the pixel under the
+  // mouse can be drawn in the 3D model viewer.
+  graphics_view_->setMouseTracking(true);
+  graphics_view_->viewport()->setMouseTracking(true);
+  graphics_view_->viewport()->installEventFilter(this);
+
   table_widget_ = new QTableWidget(this);
   table_widget_->setColumnCount(2);
   table_widget_->setRowCount(8);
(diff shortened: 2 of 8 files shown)


behnamasadi:
@ahojnnes what do you say?

src/colmap/ui/colormaps.h
```diff
@@ -174,4 +174,29 @@ class ImageColormapNameFilter : public ImageColormapBase {
       image_name_colors_;
 };
 
+// Color images according to their mean reprojection error, on an *absolute*
```
ahojnnes: This looks like it's an unrelated change to the camera viewing ray functionality, which got accidentally included from one of your other PRs?

src/colmap/ui/image_viewer_widget.cc
```diff
@@ -381,6 +391,95 @@ void DatabaseImageViewerWidget::ShowImageWithId(const image_t image_id) {
...
+    model_viewer_widget_->ClearHoverRay();
+    return;
+  }
+  const Image& image = image_it->second;
+  const auto camera_it = model_viewer_widget_->cameras.find(image.CameraId());
+  if (camera_it == model_viewer_widget_->cameras.end()) {
+    model_viewer_widget_->ClearHoverRay();
+    return;
+  }
+  const Camera& camera = camera_it->second;
```
ahojnnes: This could be extracted once during the ShowImage() call.

src/colmap/ui/image_viewer_widget.cc
```diff
@@ -381,6 +391,95 @@ void DatabaseImageViewerWidget::ShowImageWithId(const image_t image_id) {
...
+
+  // Unproject the pixel to a world-space viewing ray from the camera center.
+  const std::optional<Eigen::Vector2d> cam_point =
+      camera.CamFromImg(Eigen::Vector2d(px, py));
+  if (!cam_point.has_value()) {
+    model_viewer_widget_->ClearHoverRay();
+    return;
+  }
+  Eigen::Vector3d dir_cam(cam_point->x(), cam_point->y(), 1.0);
+  dir_cam.normalize();
```
ahojnnes:
```suggestion
  Eigen::Vector3d cam_ray = cam_point->homogeneous().normalized()
```

src/colmap/ui/image_viewer_widget.cc
```diff
@@ -381,6 +391,95 @@ void DatabaseImageViewerWidget::ShowImageWithId(const image_t image_id) {
...
+  const Rigid3d world_from_cam = Inverse(image.CamFromWorld());
+  const Eigen::Vector3d origin = image.ProjectionCenter();
+  const Eigen::Vector3d dir_world = world_from_cam.rotation() * dir_cam;
```
ahojnnes:
```suggestion
  const Eigen::Vector3d world_ray = image.CamFromWorld().rotation.inverse() * dir_cam;
```

src/colmap/ui/image_viewer_widget.cc
```diff
@@ -381,6 +391,95 @@ void DatabaseImageViewerWidget::ShowImageWithId(const image_t image_id) {
...
+      continue;
+    }
+    const auto p3d_it = model_viewer_widget_->points3D.find(point2D.point3D_id);
+    if (p3d_it == model_viewer_widget_->points3D.end()) {
+      continue;
+    }
+    best_sq_dist = sq_dist;
+    point3D = p3d_it->second.xyz;
+    has_point3D = true;
+  }
```
ahojnnes: Not sure how much value this adds. I would be fine with just showing the ray.

src/colmap/ui/model_viewer_widget.cc
```diff
@@ -867,6 +875,60 @@ void ModelViewerWidget::ShowImageInfo(const image_t image_id) {
...
+  // so the ray endpoints must use the same transform.
+  const Eigen::Vector3f render_origin =
+      (model_scale_ * (origin + model_origin_)).cast<float>();
+  Eigen::Vector3f render_end;
+  if (has_point3D) {
+    render_end = (model_scale_ * (point3D + model_origin_)).cast<float>();
+  } else {
+    // Extend across the scene: uniform scale preserves the world direction.
+    const float ray_length = std::max(2.0f * render_origin.norm(), 1.0f);
+    render_end = render_origin + direction.cast<float>() * ray_length;
```
ahojnnes: If the camera happens to be at the origin, then the ray will be very short. How about extracting the max distance to any 3d point in the image and using that? It could be extracted once during ShowImage().

ahojnnes (commented):
Thanks, it's a cool visualization to have but I am admittedly not very much convinced this adds a lot of practical value. Could you share a bit how you are using this in your workflow? Happy to be convinced if others find it useful.

Apart from these general concerns, it looks like you accidentally entangled a few other changes into this PR.

behnamasadi:
@ahojnnes Thanks for your reviews. I have rebased this on top of my other feature: https://github.com/colmap/colmap/pull/4579
That feature is about setting the camera frustums based on reprojection error so the user gets an intuitive feeling about the reliability of each camera pose. Regarding your question about how I use this feature, it is basically a visual debugging tool to see if 2D/3D correspondence is correct and they land exactly on the 3D scene, Rerun heavily uses it and I get the idea from there. I mean, if you feel like adding this PR, I can make the changes you mentioned. Regards.

src/colmap/ui/colormaps.h
behnamasadi: Correct, that was leaked in from #4579. The branch has been rebased and this PR now touches only the four hover-ray files (`image_viewer_widget.{h,cc}`, `model_viewer_widget.{h,cc}`) — no `colormaps.h` change left.

src/colmap/ui/image_viewer_widget.cc
behnamasadi: Done. All per-image state (camera, projection center, inverse rotation, ray length) is now computed once in `ShowImageWithId()` into a `HoverRayContext` (`image_viewer_widget.cc:395-419`, struct at `image_viewer_widget.h:123`). The mouse-move path only unprojects the pixel.

src/colmap/ui/image_viewer_widget.cc
behnamasadi: Applied — `image_viewer_widget.cc:472`.

src/colmap/ui/image_viewer_widget.cc
behnamasadi: Applied. The inverse rotation is cached per image, so the hover path is just `context.world_from_cam_rotation * cam_ray` (`image_viewer_widget.cc:473`).

src/colmap/ui/image_viewer_widget.cc
behnamasadi: Dropped — only the ray is drawn now.

src/colmap/ui/model_viewer_widget.cc
behnamasadi: Good catch. The length is now the max distance from the projection center to any 3D point observed by this image, computed once in `ShowImageWithId()` (`image_viewer_widget.cc:405-418`), falling back to 1.0 when the image has no triangulated points. That makes it scale-independent and non-degenerate near the origin.

behnamasadi:
@ahojnnes A rebase moved the line anchors, so the review threads above are marked outdated — all six points are addressed in the current head (`44d8463`), and I've replied inline to each. Summary:

- The `colormaps.h` change from #4579 is gone; the diff is now only the four hover-ray files.
- Per-image state is extracted once in `ShowImageWithId()` instead of on every mouse move.
- Both suggestion diffs applied.
- The extra readout is removed — only the ray is drawn.
- Ray length is the max distance to any 3D point observed by the image, so it no longer degenerates when the camera sits near the world origin.

On the value question: I use it as a correspondence sanity check. Hovering a pixel and watching where its ray lands in the 3D view immediately shows whether the 2D observations and the triangulated geometry agree — a ray that misses the surface it visually points at means a bad pose, a wrong camera model, or a mis-registered image, and that is much faster to spot than reading reprojection-error numbers. It's the same idea as Rerun's ray picking. It's a debugging aid, not a workflow feature, and it costs nothing when unused: no state is built unless an image is open and the ray is only computed on hover.

If you'd still rather not carry it in the GUI, I'm fine with closing — but the review comments are all resolved either way, so it's ready if you want it.

behnamasadi:
@ahojnnes Gentle ping on this one. The six review threads are marked outdated after a rebase but all are addressed in `44d8463`; the diff is four files under `src/colmap/ui/`. Happy to close it if GUI debugging aids aren't something you want to carry — just don't want it sitting in your queue undecided.

PR #4586, opened 2026-07-22
doc: Add FAQ on extracting camera poses and camera centers

### LINKED ISSUE #797: How to compose proper transform matrix for image pose from quaternion+translation ?
Hello!

Thanks for your great work!

I'm trying to export images (frames) pose data from existing reconstruction (images.txt) and facing problem converting quaternion data to 4x4 matrix. Here's the code in python: 

`rvar = [float(lvar[0]), float(lvar[1]), float(lvar[2]), float(lvar[2])]
Rotation = transformations.quaternion_matrix(rvar)
Translation = np.eye(4)
Translation[0][3] = float(lvar[4])
Translation[1][3] = float(lvar[5])
Translation[2][3] = float(lvar[6])
Tvar = np.matmul(-Rotation.transpose(),Translation)`

I've tried other combinations of matmul order and transpose (from #682), but the results were absolutely inadequate. Here's screenshots of reconstruction in Colmap and 3d points (camera positions) from transform matrix (coordinates of the points are the fourth row of the matrix):

Colmap:
colmap2
colmap1
colmap3

Results:
Figure_1
Figure_3
Figure_2

Any ideas, how can I get a relevant matrix and what i'm doing wrong?

Thanks in advance!

### LINKED ISSUE #1376: Is it possible to grab 6dof camera poses associated with each image?
When exporting as text I have:

```1 0.014406197300745461 -0.0015940703021185547 0.99712831150560799 0.074330685542238983 0.36718918396685335 -0.5137767828369697 5.1558036640144138 13 device_1_still.png
8 0.99905180602695465 -3.6953248996980325e-05 0.043478689938509597 -0.002256331085761982 1.743857459490727 -0.31915147687515399 6.0468314976936224 14 device_2_still.png
5 -0.17075203506588638 0.0034362913452786418 0.98521817958392233 0.013306879423473651 1.3917064374221519 -0.25369528557724652 4.4701250290315526 15 k4a_still.png
```

And I see from the documentation:
```Here, the first two lines define the information of the first image, and so on. The reconstructed pose of an image is specified as the projection from world to the camera coordinate system of an image using a quaternion (QW, QX, QY, QZ) and a translation vector (TX, TY, TZ). The quaternion is defined using the Hamilton convention, which is, for example, also used by the Eigen library. The coordinates of the projection/camera center are given by -R^t * T, where R^t is the inverse/transpose of the 3x3 rotation matrix composed from the quaternion and T is the translation vector. The local camera coordinate system of an image is defined in a way that the X axis points to the right, the Y axis to the bottom, and the Z axis to the front as seen from the image.
```

In my sparse reconstruction I can see what I am looking for, the 6dof of each image in world space (relatively):
image

**Is there anyway to grab the 6dof of those poses from the colmap GUI?**

### LINKED ISSUE #1484: Colmap-How can I get relative pose between two images?
When I use automatic reconstruction ,I can get images.txt, which contains the absolute pose(qw, qx, qy, qz). But how can I get the relative pose between 2 images?

behnamasadi:
## Summary

Adds a FAQ entry for one of the most frequently repeated questions in the tracker
(#1376, #1484, #797): how to obtain the per-image camera pose and camera center
from a reconstruction.

The entry explains:
- the **world-to-camera** convention of the `(QW, QX, QY, QZ, TX, TY, TZ)` values
  in `images.txt` / `images.bin` (Hamilton quaternion, as also used by Eigen);
- that these values are **not** the camera position — the camera center in world
  coordinates is `C = -R^t * T`, a common point of confusion when comparing
  against GPS/other sensors;
- the pycolmap accessors that return both directly (`image.cam_from_world`,
  `image.cam_from_world.matrix()`, `image.projection_center()`);
- how to compose the **relative pose** between two images.

Cross-links the existing `Output Format` section for the full field layout.

Code example verified against the pycolmap bindings (`cam_from_world`,
`projection_center`, `Rigid3d.matrix()`, `Rigid3d * Rigid3d`, `inverse()`). Docs
build cleanly with Sphinx.

Fixes #1376
Fixes #1484
Fixes #797

### DIFF
--- a/doc/faq.rst
+++ b/doc/faq.rst
@@ -378,6 +378,48 @@ left at the default (greater than 1). So also set that parameter, as below::
         --output_path path/to/dense/workspace/fused.ply
 
 
+.. _faq-extract-poses:
+
+Extracting camera poses and camera centers
+------------------------------------------
+
+Each registered image stores its 6-DoF pose. In the exported ``images.txt`` (and
+the binary ``images.bin``) the pose is given by a quaternion ``(QW, QX, QY, QZ)``
+and a translation ``(TX, TY, TZ)`` that together define the transformation **from
+world to camera coordinates** (Hamilton convention, as also used by Eigen). A
+world point ``X_world`` maps into the camera frame as ``X_cam = R * X_world + T``,
+where ``R`` is the rotation matrix built from the quaternion. See
+:ref:`Output Format <output-format>` for the full field layout.
+
+Note that ``(TX, TY, TZ)`` is **not** the position of the camera in the world. The
+camera center (projection center) in world coordinates is::
+
+    C = -R^t * T
+
+where ``R^t`` is the transpose (inverse) of the rotation matrix. This is a common
+source of confusion when comparing COLMAP poses against GPS or other sensors,
+which usually report the camera center directly.
+
+Using pycolmap, both the world-to-camera pose and the camera center are available
+directly, without manual conversion::
+
+    import pycolmap
+
+    reconstruction = pycolmap.Reconstruction("/path/to/sparse/model")
+    for image_id, image in reconstruction.images.items():
+        cam_from_world = image.cam_from_world        # Rigid3d (world -> camera)
+        world_to_cam = cam_from_world.matrix()       # 3x4 [R | T] matrix
+        center = image.projection_center()           # camera center in world coords
+        print(image.name, center)
+
+The relative pose between two images (transforming points from image ``j``'s
+camera frame into image ``i``'s) can be composed from their world-to-camera poses::
+
+    image_i = reconstruction.images[i]
+    image_j = reconstruction.images[j]
+    i_from_j = image_i.cam_from_world * image_j.cam_from_world.inverse()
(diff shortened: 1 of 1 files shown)


behnamasadi:
Self-review catch: the pycolmap snippet used `image.cam_from_world` as an attribute, but it is bound with `.def(...)`, not `.def_property(...)`, so it is a method and needs parentheses. As written the example raised `AttributeError: 'builtin_function_or_method' object has no attribute 'matrix'`. Fixed and force-pushed — both the pose loop and the relative-pose composition now read `image.cam_from_world()`.

Verified against pycolmap 4.2.0 on a synthetic reconstruction: the corrected snippet runs and returns a `(3, 4)` matrix, a camera center, and a `Rigid3d` relative pose. `image.projection_center()` and `image.data_id` were already correct (the latter is a read-only property, so no parentheses there).

Also added the closing keywords for #1376, #1484 and #797 — the relative-pose paragraph covers #1484 and the quaternion-to-matrix composition covers #797.

PR #4589, opened 2026-07-22
Add multi-component support to global mapper

lpanaf:
## Summary

Global SfM keeps only the largest connected component of the view graph, so everything else in a disconnected capture is discarded. This PR reconstructs every component instead, producing one model per component.

The process has three steps:

1. Extract the initial connected components of the view graph.
2. Run rotation averaging independently on each of them, on an isolated copy of the base reconstruction so that rig calibration cannot leak across components. This is what settles the final components: filtering edges by relative rotation error breaks apart clusters that were only held together by outlier edges, so one input component can yield several genuine ones. Doing the split here rather than on the raw pose graph matters because no later stage discards images.
3. Run the full global mapper on each final component, over its own `DatabaseCache` restricted to that component's images.

The resulting models are sorted by registered-frame count, and those with fewer than `min_num_frames` frames are dropped.

## Options

- `GlobalMapper.reconstruct_all_components` (default `true`): reconstruct every component versus only the largest one.
- `GlobalMapper.min_num_frames` (default `3`): discard models with fewer registered frames.

Both are exposed through `OptionManager` and pycolmap's `GlobalMapperOptions`.

## Behavior changes

- Multi-component reconstruction is on by default, so runs on disconnected datasets now emit several models instead of one.
- `min_num_frames` also applies to the single-component path, where any successful model was previously kept.
- The reconstruction is no longer added to the `ReconstructionManager` before the mapper starts, so the GUI no longer renders intermediate states during global mapping; models appear once complete.

## Benchmark: IMC2025, all 13 scenes

No covisibility filtering, same feature/match databases, `--random_seed 0 --num_threads 1`, measured at `b647561c`. `global (main)` and `global (single)` (`--reconstruct_all_components 0`) come out bit-identical, confirming the largest-component path is unchanged.

Pooled over all scenes:

| method         | AUC@0.5° | AUC@1° | AUC@5° | AUC@10° | reg imgs / 1823 | #comp |
|----------------|----------|--------|--------|---------|-----------------|-------|
| incremental    | 8.52     | 17.01  | 36.25  | 41.67   | 1345            | 34    |
| global (main)  | 7.45     | 11.53  | 19.22  | 21.14   | 634             | 12    |
| global (multi) | 13.85    | 23.15  | 40.33  | 44.33   | 1544            | 68    |

Per-scene average (mAA):

| method         | AUC@0.5° | AUC@1° | AUC@5° | AUC@10° |
|----------------|----------|--------|--------|---------|
| incremental    | 10.80    | 21.24  | 43.00  | 48.55   |
| global (main)  | 7.71     | 12.59  | 22.78  | 25.04   |
| global (multi) | 14.60    | 25.21  | 44.74  | 48.91   |

Registered images more than double (634 to 1544 of 1823) and overall AUC@10° goes from 21.14 to 44.33, slightly ahead of the incremental mapper. Full per-scene tables are in [this comment](https://github.com/colmap/colmap/pull/4589#issuecomment-5107407840).

## Test plan

New end-to-end tests in `controllers/global_pipeline_test.cc`:

- `MultiComponents`: two rigs with all cross-rig matches removed yield one reconstruction per component, each matching its ground-truth subset.
- `MultiComponentsWithOutlierEdges`: two clusters bridged only by outlier edges are still recovered separately, since rotation averaging leaves large residuals on the bridges and filters them out.
- `MultiComponentsWithOutlierEdgesUsingGravity`: even when every cross-cluster edge is an outlier, which the gravity-free solver cannot resolve because the inter-cluster gauge is free, gravity priors make the bridges detectable.
- `MultiComponentsEmptyViewGraph`: a database with no matches produces no reconstructions.

New tests in `scene/pose_graph_test.cc` cover the component ordering and the edgeless case.

- [x] `ctest -R "controllers/global_pipeline_test|scene/pose_graph_test"`

### DIFF
--- a/src/colmap/controllers/global_pipeline.cc
+++ b/src/colmap/controllers/global_pipeline.cc
@@ -219,7 +219,8 @@ void GlobalPipeline::RunMultiComponents(
     return;
   }
 
-  LOG(INFO) << "Found " << component_manager.Size() << " connected component(s)";
+  LOG(INFO) << "Found " << component_manager.Size()
+            << " connected component(s)";
 
   // Run the full pipeline independently per component, restricted to that
   // component's images.
--- a/src/colmap/controllers/global_pipeline.cc
+++ b/src/colmap/controllers/global_pipeline.cc
@@ -30,6 +30,7 @@
 #include "colmap/controllers/global_pipeline.h"
 
 #include "colmap/estimators/alignment.h"
+#include "colmap/estimators/rotation_averaging.h"
 #include "colmap/estimators/two_view_geometry.h"
 #include "colmap/scene/database_cache.h"
 #include "colmap/scene/pose_graph.h"
@@ -39,7 +40,6 @@
 #include "colmap/util/timer.h"
 
 #include <algorithm>
-#include <deque>
 
 namespace colmap {
 namespace {
@@ -76,6 +76,66 @@ size_t NumFramesForImages(const Reconstruction& reconstruction,
     frame_ids.insert(reconstruction.Image(image_id).FrameId());
   }
   return frame_ids.size();
+}
+
+struct ComponentDecomposition {
+  std::vector<FlatHashSet<image_t>> components;
+  size_t num_failed = 0;
+};
+
+// Split every input view-graph component once using the same rotation
+// averaging and relative-rotation filtering as the global mapper. Components
+// that are already too small are passed through so the caller can account for
+// them without running rotation averaging.
+ComponentDecomposition ComputeComponentsByRotationAveraging(
+    const RotationEstimatorOptions& options,
+    const PoseGraph& pose_graph,
+    const Reconstruction& base,
+    const std::vector<PosePrior>& pose_priors,
+    int min_num_frames) {
+  ComponentDecomposition result;
+  const std::vector<FlatHashSet<image_t>> input_components =
+      pose_graph.ComputeConnectedComponentImageIds(
+          base, /*filter_unregistered=*/false);
+
+  for (const auto& input_component : input_components) {
+    if (static_cast<int>(NumFramesForImages(base, input_component)) <
+        min_num_frames) {
+      result.components.push_back(input_component);
+      continue;
+    }
+
+    Reconstruction reconstruction = base;
+    PoseGraph component_pose_graph = pose_graph;
+    component_pose_graph.InvalidatePairsOutsideActiveImageIds(input_component);
+
+    RotationEstimatorOptions decomposition_options = options;
+    decomposition_options.filter_unregistered = false;
+    if (!RunRotationAveragingOnComponent(decomposition_options,
+                                         component_pose_graph,
+                                         input_component,
+                                         reconstruction,
+                                         pose_priors)) {
+      ++result.num_failed;
+      continue;
+    }
+
+    if (decomposition_options.max_rotation_error_deg > 0) {
+      FilterEdgesByRelativeRotation(
+          component_pose_graph,
+          reconstruction,
+          decomposition_options.max_rotation_error_deg);
+    }
+
+    std::vector<FlatHashSet<image_t>> sub_components =
+        component_pose_graph.ComputeConnectedComponentImageIds(
+            reconstruction, /*filter_unregistered=*/true);
+    for (auto& sub_component : sub_components) {
+      result.components.push_back(std::move(sub_component));
+    }
+  }
+
+  return result;
 }
 
 }  // namespace
@@ -123,10 +183,8 @@ GlobalPipeline::ReconstructionResult GlobalPipeline::ReconstructSingleComponent(
             << " seconds";
 
   // A stop requested through the callback is reported as success, so false
-  // only denotes a genuine mapping failure. Return the failed reconstruction
-  // temporarily so multi-component orchestration can skip its registered
-  // frames while retrying the remainder; the caller removes it from the output
-  // manager.
+  // only denotes a genuine mapping failure. The caller removes failed
+  // reconstructions from the output manager.
   if (!success) {
     LOG(ERROR) << "Global mapping failed";
     return {std::move(reconstruction), false};
@@ -216,42 +274,39 @@ GlobalPipeline::ReconstructionStats GlobalPipeline::ReconstructMultiComponents(
     const GlobalMapperOptions& mapper_options) {
   ReconstructionStats stats;
 
-  // Build the base reconstruction and pose graph from the cache.
+  // Build the base reconstruction, pose graph, and pose priors from the cache.
   Reconstruction base;
   base.Load(*database_cache_);
   PoseGraph pose_graph;
   pose_graph.Load(*database_cache_->CorrespondenceGraph());
+  const std::vector<PosePrior>& pose_priors = database_cache_->PosePriors();
   if (pose_graph.Empty()) {
     LOG(ERROR) << "Cannot continue with empty pose graph";
     return stats;
   }
 
-  // Partition the unfiltered input view graph. The full mapper may split these
-  // components further when it filters relative rotations below.
-  std::vector<FlatHashSet<image_t>> components =
-      pose_graph.ComputeConnectedComponentImageIds(
-          base, /*filter_unregistered=*/false);
+  // Decompose the view graph once after rotation filtering. The full mapper is
+  // then run at most once per resulting component; any additional fragments
+  // rejected by its refinement pass are not recursively retried.
+  ComponentDecomposition decomposition =
+      ComputeComponentsByRotationAveraging(mapper_options.RotationAveraging(),
+                                           pose_graph,
+                                           base,
+                                           pose_priors,
+                                           options_.min_num_frames);
+  stats.num_failed += decomposition.num_failed;
+  std::vector<FlatHashSet<image_t>>& components = decomposition.components;
 
-  LOG(INFO) << "Found " << components.size() << " connected component(s)";
+  LOG(INFO) << "Found " << components.size()
+            << " connected component(s) after rotation filtering";
 
-  // Run the full pipeline independently per input component. The mapper keeps
-  // the largest connected component after rotation filtering. Remove its
-  // registered frames, decompose the remainder, and enqueue every resulting
-  // component. This uses the mapper's actual two-pass rotation averaging and
-  // does not require a disposable preliminary solve.
-  std::deque<FlatHashSet<image_t>> pending_components;
-  for (auto& component : components) {
-    pending_components.push_back(std::move(component));
-  }
-  size_t component_idx = 0;
-  while (!pending_components.empty()) {
+  for (size_t component_idx = 0; component_idx < components.size();
+       ++component_idx) {
     if (CheckIfStopped()) {
       return stats;
     }
 
-    FlatHashSet<image_t> image_ids = std::move(pending_components.front());
-    pending_components.pop_front();
-    ++component_idx;
+    const FlatHashSet<image_t>& image_ids = components[component_idx];
 
     if (static_cast<int>(NumFramesForImages(base, image_ids)) <
         options_.min_num_frames) {
@@ -259,11 +314,10 @@ GlobalPipeline::ReconstructionStats GlobalPipeline::ReconstructMultiComponents(
       continue;
     }
 
-    LOG_HEADING1(
-        StringPrintf("Reconstructing component %d with %d images (%d pending)",
-                     static_cast<int>(component_idx),
-                     static_cast<int>(image_ids.size()),
-                     static_cast<int>(pending_components.size())));
+    LOG_HEADING1(StringPrintf("Reconstructing component %d / %d with %d images",
+                              static_cast<int>(component_idx + 1),
+                              static_cast<int>(components.size()),
+                              static_cast<int>(image_ids.size())));
 
     DatabaseCache::Options cache_options;
     cache_options.image_names.reserve(image_ids.size());
@@ -278,24 +332,6 @@ GlobalPipeline::ReconstructionStats GlobalPipeline::ReconstructMultiComponents(
     const std::shared_ptr<Reconstruction>& reconstruction =
         result.reconstruction;
 
-    const std::vector<frame_t> reg_frame_ids = reconstruction->RegFrameIds();
-    const FlatHashSet<frame_t> registered_frame_ids(reg_frame_ids.begin(),
-                                                    reg_frame_ids.end());
-    FlatHashSet<image_t> remaining_image_ids;
-    remaining_image_ids.reserve(image_ids.size());
-    for (const image_t image_id : image_ids) {
-      if (registered_frame_ids.count(base.Image(image_id).FrameId()) == 0) {
-        remaining_image_ids.insert(image_id);
-      }
-    }
-
-    if (remaining_image_ids.size() == image_ids.size()) {
-      LOG(ERROR) << "Global mapping made no registration progress";
-      reconstruction_manager_->Delete(reconstruction_manager_->Size() - 1);
-      ++stats.num_failed;
-      continue;
-    }
-
     if (!result.success) {
       reconstruction_manager_->Delete(reconstruction_manager_->Size() - 1);
       ++stats.num_failed;
@@ -303,26 +339,6 @@ GlobalPipeline::ReconstructionStats GlobalPipeline::ReconstructMultiComponents(
                options_.min_num_frames) {
       reconstruction_manager_->Delete(reconstruction_manager_->Size() - 1);
       ++stats.num_too_small;
-    }
-
-    if (CheckIfStopped()) {
-      return stats;
-    }
-
-    if (!remaining_image_ids.empty()) {
-      PoseGraph remaining_pose_graph = pose_graph;
-      remaining_pose_graph.InvalidatePairsOutsideActiveImageIds(
-          remaining_image_ids);
-      std::vector<FlatHashSet<image_t>> remaining_components =
-          remaining_pose_graph.ComputeConnectedComponentImageIds(
-              base, /*filter_unregistered=*/false);
-      if (remaining_components.empty()) {
-        ++stats.num_too_small;
-      } else {
-        for (auto& component : remaining_components) {
-          pending_components.push_back(std::move(component));
-        }
-      }
     }
   }
 
--- a/src/colmap/controllers/global_pipeline.cc
+++ b/src/colmap/controllers/global_pipeline.cc
@@ -81,12 +81,12 @@ size_t NumFramesForImages(const Reconstruction& reconstruction,
 struct ComponentDecomposition {
   std::vector<FlatHashSet<image_t>> components;
   size_t num_failed = 0;
+  size_t num_too_small = 0;
 };
 
 // Split every input view-graph component once using the same rotation
 // averaging and relative-rotation filtering as the global mapper. Components
-// that are already too small are passed through so the caller can account for
-// them without running rotation averaging.
+// that are already too small are discarded without running rotation averaging.
 ComponentDecomposition ComputeComponentsByRotationAveraging(
     const RotationEstimatorOptions& options,
     const PoseGraph& pose_graph,
@@ -101,7 +101,7 @@ ComponentDecomposition ComputeComponentsByRotationAveraging(
   for (const auto& input_component : input_components) {
     if (static_cast<int>(NumFramesForImages(base, input_component)) <
         min_model_size) {
-      result.components.push_back(input_component);
+      ++result.num_too_small;
       continue;
     }
 
@@ -295,6 +295,7 @@ GlobalPipeline::ReconstructionStats GlobalPipeline::ReconstructMultiComponents(
                                            pose_priors,
                                            options_.min_model_size);
   stats.num_failed += decomposition.num_failed;
+  stats.num_too_small += decomposition.num_too_small;
   std::vector<FlatHashSet<image_t>>& components = decomposition.components;
 
   LOG(INFO) << "Found " << components.size()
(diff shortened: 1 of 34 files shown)


lpanaf:
# IMC2025 global reconstruction — incremental vs global (main / single / multi)

All 13 IMC2025 scenes, no covisibility filtering. Same feature/match databases.
`global (main)` = commit 4d1d5932 default `global_mapper` (largest component).
`global (single)` = branch b647561c `--reconstruct_all_components 0`.
`global (multi)`  = branch b647561c `--reconstruct_all_components 1`.
All at `--random_seed 0 --num_threads 1`. `global (main)` and `global (single)` come out
bit-identical, confirming the branch does not change the largest-component path.

## Overall (pooled over all 13 scenes)

| method          | AUC@0.5° | AUC@1° | AUC@5° | AUC@10° | reg imgs | total | #comp |
|-----------------|----------|--------|--------|---------|----------|-------|-------|
| incremental     | 8.52     | 17.01  | 36.25  | 41.67   | 1345     | 1823  | 34    |
| global (main)   | 7.45     | 11.53  | 19.22  | 21.14   | 634      | 1823  | 12    |
| global (single) | 7.45     | 11.53  | 19.22  | 21.14   | 634      | 1823  | 12    |
| global (multi)  | 13.85    | 23.15  | 40.33  | 44.33   | 1544     | 1823  | 68    |

## Per-scene average (mAA)

| method          | AUC@0.5° | AUC@1° | AUC@5° | AUC@10° |
|-----------------|----------|--------|--------|---------|
| incremental     | 10.80    | 21.24  | 43.00  | 48.55   |
| global (main)   | 7.71     | 12.59  | 22.78  | 25.04   |
| global (single) | 7.71     | 12.59  | 22.78  | 25.04   |
| global (multi)  | 14.60    | 25.21  | 44.74  | 48.91   |

## Per-scene AUC@10°

| scene                             | incremental | global (main) | global (single) | global (multi) |
|-----------------------------------|-------------|---------------|-----------------|----------------|
| ETs                               | 93.16       | 52.02         | 52.02           | 94.31          |
| amy_gardens                       | 1.26        | 0.34          | 0.34            | 2.23           |
| fbk_vineyard                      | 0.34        | 0.10          | 0.10            | 1.20           |
| imc2023_haiper                    | 63.82       | 28.75         | 28.75           | 63.18          |
| imc2023_heritage                  | 66.53       | 31.82         | 31.82           | 61.63          |
| imc2023_theather_imc2024_church   | 37.49       | 22.54         | 22.54           | 32.99          |
| imc2024_dioscuri_baalshamin       | 57.06       | 31.35         | 31.35           | 37.88          |
| imc2024_lizard_pond               | 3.34        | 2.79          | 2.79            | 7.18           |
| pt_brandenburg_british_buckingham | 70.37       | 27.37         | 27.37           | 84.33          |
| pt_piazzasanmarco_grandplace      | 81.75       | 54.99         | 54.99           | 83.00          |
| pt_sacrecoeur_trevi_tajmahal      | 77.81       | 27.96         | 27.96           | 80.87          |
| pt_stpeters_stpauls               | 78.26       | 45.45         | 45.45           | 87.05          |
| stairs                            | 0.00        | 0.00          | 0.00            | 0.00           |
| **overall**                       | 41.67       | 21.14         | 21.14           | 44.33          |

## Per-scene AUC@5°

| scene                             | incremental | global (main) | global (single) | global (multi) |
|-----------------------------------|-------------|---------------|-----------------|----------------|
| ETs                               | 86.17       | 48.48         | 48.48           | 88.63          |
| amy_gardens                       | 1.07        | 0.23          | 0.23            | 1.76           |
| fbk_vineyard                      | 0.26        | 0.05          | 0.05            | 0.49           |
| imc2023_haiper                    | 61.05       | 26.21         | 26.21           | 59.84          |
| imc2023_heritage                  | 62.00       | 30.03         | 30.03           | 57.11          |
| imc2023_theather_imc2024_church   | 28.73       | 18.58         | 18.58           | 27.54          |
| imc2024_dioscuri_baalshamin       | 51.03       | 28.94         | 28.94           | 34.67          |
| imc2024_lizard_pond               | 3.01        | 2.13          | 2.13            | 6.17           |
| pt_brandenburg_british_buckingham | 55.33       | 23.22         | 23.22           | 74.42          |
| pt_piazzasanmarco_grandplace      | 70.66       | 48.62         | 48.62           | 73.99          |
| pt_sacrecoeur_trevi_tajmahal      | 71.52       | 26.96         | 26.96           | 75.56          |
| pt_stpeters_stpauls               | 68.11       | 42.62         | 42.62           | 81.42          |
| stairs                            | 0.00        | 0.00          | 0.00            | 0.00           |
| **overall**                       | 36.25       | 19.22         | 19.22           | 40.33          |

## Per-scene registered images

| scene                             | incremental | global (main) | global (single) | global (multi) | total |
|-----------------------------------|-------------|---------------|-----------------|----------------|-------|
| ETs                               | 19          | 10            | 10              | 19             | 19    |
| amy_gardens                       | 80          | 41            | 41              | 147            | 200   |
| fbk_vineyard                      | 7           | 13            | 13              | 111            | 163   |
| imc2023_haiper                    | 54          | 31            | 31              | 54             | 54    |
| imc2023_heritage                  | 141         | 57            | 57              | 142            | 148   |
| imc2023_theather_imc2024_church   | 71          | 47            | 47              | 69             | 76    |
| imc2024_dioscuri_baalshamin       | 95          | 52            | 52              | 92             | 114   |
| imc2024_lizard_pond               | 59          | 33            | 33              | 90             | 180   |
| pt_brandenburg_british_buckingham | 224         | 75            | 75              | 224            | 225   |
| pt_piazzasanmarco_grandplace      | 168         | 100           | 100             | 168            | 168   |
| pt_sacrecoeur_trevi_tajmahal      | 225         | 75            | 75              | 225            | 225   |
| pt_stpeters_stpauls               | 200         | 100           | 100             | 200            | 200   |
| stairs                            | 2           | 0             | 0               | 3              | 51    |
| **overall**                       | 1345        | 634           | 634             | 1544           | 1823  |

## Per-scene number of components

| scene                             | incremental | global (main) | global (single) | global (multi) |
|-----------------------------------|-------------|---------------|-----------------|----------------|
| ETs                               | 2           | 1             | 1               | 2              |
| amy_gardens                       | 5           | 1             | 1               | 14             |
| fbk_vineyard                      | 1           | 1             | 1               | 15             |
| imc2023_haiper                    | 2           | 1             | 1               | 2              |
| imc2023_heritage                  | 4           | 1             | 1               | 5              |
| imc2023_theather_imc2024_church   | 2           | 1             | 1               | 2              |
| imc2024_dioscuri_baalshamin       | 2           | 1             | 1               | 7              |
| imc2024_lizard_pond               | 5           | 1             | 1               | 10             |
| pt_brandenburg_british_buckingham | 3           | 1             | 1               | 3              |
| pt_piazzasanmarco_grandplace      | 2           | 1             | 1               | 2              |
| pt_sacrecoeur_trevi_tajmahal      | 3           | 1             | 1               | 3              |
| pt_stpeters_stpauls               | 2           | 1             | 1               | 2              |
| stairs                            | 1           | 0             | 0               | 1              |
| **total**                         | 34          | 12            | 12              | 68

src/colmap/controllers/global_pipeline.h
```diff
@@ -61,6 +61,15 @@ struct GlobalPipelineOptions {
   // Whether to decompose relative poses from two-view geometries.
   bool decompose_relative_pose = true;
 
+  // If true, reconstruct every connected component of the view graph (one
+  // model per component). If false (default), reconstruct only the largest
+  // connected component.
+  bool reconstruct_all_components = false;
```
ahojnnes: I think this should default to true?

src/colmap/controllers/global_pipeline.h
```diff
@@ -81,6 +90,19 @@ class GlobalPipeline : public BaseController {
   void Run() override;
 
  private:
+  // Run the full global SfM pipeline on the given database cache and return
+  // the resulting reconstruction.
+  std::shared_ptr<Reconstruction> RunSingleReconstruction(
```
ahojnnes: nit: ReconstructSingleComponent

src/colmap/controllers/global_pipeline.h
```diff
@@ -81,6 +90,19 @@ class GlobalPipeline : public BaseController {
...
+      const std::shared_ptr<const DatabaseCache>& database_cache,
+      const GlobalMapperOptions& mapper_options);
+
+  // Partition the view graph into connected components and run the full
+  // pipeline per component. Appends the resulting reconstructions to the
+  // given vector.
+  void RunMultiComponents(
```
ahojnnes: nit: ReconstructMultiComponents and return vector of shared_ptrs ?

src/colmap/controllers/global_pipeline.cc
```diff
@@ -127,25 +117,139 @@ void GlobalPipeline::Run() {
...
+
+  if (pose_graph.Empty()) {
+    LOG(ERROR) << "Cannot continue with empty pose graph";
+    return;
+  }
+
+  // Partition the view graph into connected components via rotation averaging.
+  ReconstructionManager component_manager;
+  *component_manager.Get(component_manager.Add()) = *base;
+  if (!RunRotationAveragingMultiComponents(mapper_options.RotationAveraging(),
```
ahojnnes: Maybe I am missing something but it feels odd that rotation averaging gets a special treatment and is multi-component, as it doesn't really do anything special with the multiple components. Instead, I suggest to extract the decomposition of the pose graph into multiple components and then perform all the reconstruction steps (including rotation averaging) on a single component.

ahojnnes (commented):
Thanks, I think it's great to have multi-component support for the global mapper but I don't like the current design choice of entangling the decomposition of connected components into the rotation averaging step. I can see how that makes sense when rotation averaging logic is needed to do the splitting, but it seems like the multi-component splitting is really just happening on the input pose graph?

lpanaf:
> Thanks, I think it's great to have multi-component support for the global mapper but I don't like the current design choice of entangling the decomposition of connected components into the rotation averaging step. I can see how that makes sense when rotation averaging logic is needed to do the splitting, but it seems like the multi-component splitting is really just happening on the input pose graph?

Thanks for the review! I also don't think splitting after rotation averaging is elegant but this was a design choice I settled with. The reason of doing it in this stage is that rotation averaging is a stage that can still further split the input. For example, if there are two clusters connected by several outliers edges, then the error on these several edges would very likely be high simultaneously, so they would be pruned in the filtering stage, resulting in disconnected components. If we only perform clustering in the input stage, some of the connected components would be pruned and will never be recovered.

On the other hand, later stages do not discard images.

What do you think?

src/colmap/estimators/rotation_averaging.cc
```diff
@@ -602,89 +765,10 @@ bool RunRotationAveraging(const RotationEstimatorOptions& options,
...
-
-    RotationEstimatorOptions options_ra = options;
-    options_ra.skip_initialization = true;
-    options_ra.use_stratified = false;
-    RotationEstimator rotation_estimator(options_ra);
-    if (!rotation_estimator.EstimateRotations(
-            pose_graph, pose_priors, active_image_ids, reconstruction)) {
-      return false;
-    }
+  if (!RunRotationAveragingOnComponent(
```
B1ueber2y, line 679: The comment here does not align with the code logic. Maybe extract active_image_ids of the largest component ids outside and add an argument?

src/colmap/estimators/rotation_averaging.cc
```diff
@@ -110,6 +111,70 @@ FlatHashSet<image_t> ComputeLargestConnectedComponentImageIds(
   return image_ids;
 }
 
+// Computes all connected components and returns their image ids, sorted by
```
B1ueber2y: Since this logic is relatively straightforward and does not concern any rotations, why not move it to upper logic (or just inlined completely in pose graph) and make rotation averaging one of the steps in the global pipeline. 

I feel that we dont really need RotationAveragingMultiComponents after this refactorization?

B1ueber2y:
Thank you! Multi-component support for global mapping would be really useful.

src/colmap/estimators/rotation_averaging.cc
```diff
@@ -274,6 +339,104 @@ void FilterEdgesByRelativeRotation(PoseGraph& pose_graph,
...
+    const RotationEstimatorOptions& options,
+    PoseGraph& pose_graph,
+    Reconstruction& reconstruction,
+    const std::vector<PosePrior>& pose_priors) {
+  FlatHashSet<image_t> active_image_ids;
+
+  // Step 1: Solve rotation averaging on the largest connected component.
+  if (!HasUnknownCamsFromRig(reconstruction)) {
+    // All cam_from_rig are known, solve directly.
+    active_image_ids = ComputeLargestConnectedComponentImageIds(
```
ahojnnes: Since this is called already with a connected component, do we still need this logic here? It seems duplicated with the one below that splits the components?

src/colmap/estimators/rotation_averaging.cc
```diff
@@ -721,4 +806,62 @@ bool RunRotationAveraging(const RotationEstimatorOptions& options,
...
+bool RunRotationAveragingMultiComponents(
+    const RotationEstimatorOptions& options,
+    PoseGraph& pose_graph,
+    ReconstructionManager& reconstruction_manager,
+    const std::vector<PosePrior>& pose_priors) {
+  THROW_CHECK_GT(reconstruction_manager.Size(), 0)
+      << "Reconstruction manager must contain a base reconstruction at index 0";
+
+  // The base reconstruction provides the frames, cameras, and rigs template.
+  const Reconstruction base = *reconstruction_manager.Get(0);
```
ahojnnes: This logic is a bit weird. Shouldn't we extract the connected components based on the images in the respective reconstructions?

src/colmap/estimators/rotation_averaging.h
```diff
@@ -156,4 +157,16 @@ bool RunRotationAveraging(const RotationEstimatorOptions& options,
                           Reconstruction& reconstruction,
                           const std::vector<PosePrior>& pose_priors);
 
+// Multi-component rotation averaging. Reads the base reconstruction (frames,
+// cameras, rigs) from `reconstruction_manager.Get(0)`, solves rotation
+// averaging independently on each initial connected component of the view
+// graph, filters outlier pairs by relative rotation (without de-registering),
+// and repopulates the manager with one reconstruction per disjoint component of
+// the resulting filtered view graph.
+bool RunRotationAveragingMultiComponents(
```
ahojnnes: I suggest we only expose a single interface for rotation averaging, which takes a reconstruction manager.

src/colmap/scene/pose_graph.cc
```diff
@@ -49,10 +49,32 @@ FlatHashSet<frame_t> PoseGraph::ComputeLargestConnectedFrameComponent(
...
+
+FlatHashSet<frame_t> PoseGraph::ComputeLargestConnectedFrameComponent(
+    const Reconstruction& reconstruction, bool filter_unregistered) const {
+  std::vector<FlatHashSet<frame_t>> components =
+      ComputeConnectedFrameComponents(reconstruction, filter_unregistered);
+  if (components.empty()) {
+    return {};
+  }
+  // Components are sorted by descending size, so the first is the largest.
+  return std::move(components.front());
```
ahojnnes: Not sure if the std::move is really necessary here.

src/colmap/controllers/global_pipeline.cc
```diff
@@ -126,25 +189,132 @@ void GlobalPipeline::Run() {
...
+
+  // Sort reconstructions by the number of registered frames (descending) and
+  // discard those that failed (null) or are too small. Null entries sort last.
+  std::sort(reconstructions.begin(),
+            reconstructions.end(),
+            [](const std::shared_ptr<Reconstruction>& lhs,
+               const std::shared_ptr<Reconstruction>& rhs) {
+              const size_t lhs_num = lhs ? lhs->NumRegFrames() : 0;
+              const size_t rhs_num = rhs ? rhs->NumRegFrames() : 0;
+              return lhs_num > rhs_num;
```
ahojnnes: I think this can just turn into a THROW_CHECK_NOTNULL on lhs and rhs, because the reconstructions are never null?

src/colmap/controllers/global_pipeline.cc
```diff
@@ -126,25 +189,132 @@ void GlobalPipeline::Run() {
...
+        static_cast<int>(reconstruction->NumRegFrames()) <
+            options_.min_num_frames) {
+      ++num_discarded;
+      continue;
+    }
+
+    // Output the reconstruction.
+    Reconstruction& output_reconstruction =
+        *reconstruction_manager_->Get(reconstruction_manager_->Add());
+    output_reconstruction = *reconstruction;
```
ahojnnes: Here, we should be able to std::move the reconstruction.

src/colmap/controllers/global_pipeline.cc
```diff
@@ -126,25 +189,132 @@ void GlobalPipeline::Run() {
...
+  const std::vector<FlatHashSet<image_t>> components =
+      ComputeComponentsByRotationAveraging(
+          mapper_options.RotationAveraging(), pose_graph, base, pose_priors);
+
+  LOG(INFO) << "Found " << components.size() << " connected component(s)";
+
+  // Run the full pipeline independently per component, restricted to that
+  // component's images.
+  for (size_t i = 0; i < components.size(); ++i) {
+    FlatHashSet<std::string> image_names;
```
ahojnnes: Instead of moving them later into the cache_options, we could just directly insert them into the options?

src/colmap/controllers/global_pipeline.cc
```diff
@@ -62,6 +68,78 @@ void WarnInsufficientPriorFocalLengths() {
...
+      reconstruction, /*filter_unregistered=*/true);
+}
+
+// Partitions the view graph into components using rotation averaging: computes
+// the initial connected components of `pose_graph`, refines each on an isolated
+// copy of `base` (to avoid cross-component rig calibration leakage), filters
+// outlier pairs by relative rotation, and returns the image ids of every
+// resulting disjoint component. Outlier edge filtering during rotation
+// averaging may further split a weakly connected component into several.
+std::vector<FlatHashSet<image_t>> ComputeComponentsByRotationAveraging(
```
ahojnnes: Does this mean we run rotation averaging twice now? Once to partition and then again to solve for the rotations?

src/colmap/estimators/rotation_averaging.h
```diff
@@ -156,4 +156,24 @@ bool RunRotationAveraging(const RotationEstimatorOptions& options,
...
+// PoseGraph::ComputeConnectedComponentImageIds) and registers its frames.
+// Handles rigs with unknown cam_from_rig by first solving on an expanded
+// reconstruction. Does NOT perform outlier filtering or de-registration;
+// callers compose these primitives to implement the desired post-processing.
+bool RunRotationAveragingOnComponent(
+    const RotationEstimatorOptions& options,
+    PoseGraph& pose_graph,
+    const FlatHashSet<image_t>& active_image_ids,
+    Reconstruction& reconstruction,
+    const std::vector<PosePrior>& pose_priors);
```
ahojnnes, line 169: nit: Reorder to have output arguments at the end. Same for function below.

lpanaf (commented):
Thanks for the update! New changes LGTM.

src/colmap/controllers/global_pipeline.h
```diff
@@ -81,6 +90,29 @@ class GlobalPipeline : public BaseController {
   void Run() override;
 
  private:
+  struct ReconstructionStats {
+    size_t num_failed = 0;
```
B1ueber2y, line 96: Lets add some comments here?

src/colmap/scene/pose_graph.cc
```diff
@@ -69,27 +124,9 @@ void PoseGraph::InvalidatePairsOutsideActiveImageIds(
 int PoseGraph::MarkConnectedComponents(const Reconstruction& reconstruction,
```
B1ueber2y: This method seems to be stale for a while. Lets drop it here given that it anyway looks a bit ugly?

src/colmap/scene/pose_graph.h
```diff
@@ -69,6 +71,19 @@ class PoseGraph {
...
+  // Compute all connected components of frames, returned sorted by descending
+  // size. If filter_unregistered is true, only considers frames with HasPose().
+  std::vector<FlatHashSet<frame_t>> ComputeConnectedFrameComponents(
+      const Reconstruction& reconstruction,
+      bool filter_unregistered = true) const;
+
+  // Compute all connected components as sets of image ids, returned sorted by
+  // descending frame count. If filter_unregistered is true, only considers
+  // frames with HasPose().
+  std::vector<FlatHashSet<image_t>> ComputeConnectedComponentImageIds(
```
B1ueber2y: The naming and comment (especially "descending frame count") of this method is a bit confusing to me. How about ComputeConnectedImageIdsForFrameComponents? Meanwhile, to be less verbose, how about we drop Compute in all methods here in this class?

src/colmap/controllers/global_pipeline.h
```diff
@@ -61,6 +61,15 @@ struct GlobalPipelineOptions {
   // Whether to decompose relative poses from two-view geometries.
   bool decompose_relative_pose = true;
 
+  // If true (default), reconstruct every connected component of the view graph
+  // (one model per component). If false, reconstruct only the largest connected
+  // component.
+  bool reconstruct_all_components = true;
```
B1ueber2y: How about "multiple_models" to be consistent with the incremental pipeline?

ahojnnes (approved):

src/colmap/controllers/global_pipeline.h
```diff
@@ -81,6 +90,29 @@ class GlobalPipeline : public BaseController {
   void Run() override;
 
  private:
+  struct ReconstructionStats {
+    size_t num_failed = 0;
+    size_t num_too_small = 0;
+  };
+
+  struct ReconstructionResult {
```
B1ueber2y: This struct seems to be unnecessary. We can use std::optional?

src/colmap/controllers/global_pipeline.h
B1ueber2y: Funny that I am one minute late for posting this and we had the same thoughts : ) @ahojnnes

PR #4590, opened 2026-07-26
Add TartanAir panorama SfM reconstruction benchmark

ahojnnes:
## Summary

- add a curated, release-backed TartanAir V2 benchmark with 100 diverse scenes and 5,000 panoramas
- evaluate both native spherical cameras and overlapping perspective rigs through the panorama SfM pipeline
- package quality-97, 4:4:4 progressive JPEG panoramas and 512x256 16-bit PNG ground-truth depth maps
- derive ground-truth covisibility from depth and remove verified pairs with zero shared samples before mapping
- add deterministic seed, thread, CPU/GPU, and mapper controls to the panorama example
- include the dataset manifest, checksums, packaging utility, license, and focused tests

Dataset release: https://github.com/colmap/colmap/releases/tag/datasets-tartanair-v2-v1

## Dataset artifacts

The deterministic 13-shard release contains exactly 100 scenes, 5,000 RGB images, and 5,000 depth maps (6.183 GB total). Relative to packaging PNG RGB with the same 16-bit depth maps (14.442 GB), JPEG reduces the archive size by 57.2%.

On a representative image sample, quality-97 JPEG with 4:4:4 chroma sampling measured 44.52 dB mean PSNR (42.90 dB p10) and 0.9946 mean SSIM (0.9911 p10). Depth is min-pooled from 2048x1024 to 512x256, quantized at 0.125 m per unit, and stored losslessly as 16-bit PNG; zero represents invalid depth.

Covisibility samples depth every four pixels, projects samples through the ground-truth poses, and accepts range agreement within `max(0.25 m, 1%)` up to 256 m. The resulting directed overlap matrix is cached locally in `covisibility.npz`. Only pairs with zero shared samples are removed. A threshold of five was rejected because it removed a valid low-overlap pair in `Ruins` and reduced perspective-global AUC@10 by 0.513 points.

## Original end-to-end benchmark

CPU-only on an Apple M3 Pro, seed 0, four parallel scenes, and one thread per scene:

| Pipeline | AUC@0.5/1/5/10 | Registered images | Wall time |
|---|---|---:|---:|
| Perspective rig | 75.35 / 82.60 / 88.78 / 89.72 | 4,772 / 5,000 | 4:28:43 |
| Native spherical | 78.19 / 86.12 / 92.68 / 93.57 | 4,800 / 5,000 | 1:19:18 |

The native spherical path is 3.39x faster and improves AUC@5 by 3.90 points.

## Mapper-only four-way comparison

[...]

PR #4591, opened 2026-07-26
Avoid cross-sensor sequential rig pairs

ahojnnes:
## Summary

- require sequential seed pairs to use the same camera when rig expansion is enabled
- retain same-frame rig pairs and expand valid temporal pairs across all rig sensors
- add linear and quadratic rig pairing regression coverage

## Motivation

Folder-major image ordering can place the last image of one rig sensor next to the first image of another sensor. Sequential pairing interpreted those boundaries as temporal neighbors and expanded them into complete frame-pair graphs. On the TartanAir panorama rig this added geometrically unrelated frame pairs.

## Benchmark

Using 11 TartanAir scenes from #4590 with fixed seed 0, CPU SIFT, and one thread per scene:

- attempted image pairs per scene: 59,724 -> 34,128 (-42.9%)
- summed matching time: 4,322.5s -> 2,435.5s (-43.7%)
- pooled relative-pose AUC at 0.5/1/5/10 degrees: 44.28/46.82/48.89/49.20 -> 51.38/54.89/58.72/59.81
- mapping time: 394.7s -> 313.7s

Rendered images and descriptors were held fixed; only pairing, verification, and mapping were rerun.

## Tests

- `ctest -R controllers/pairing_test --output-on-failure`

### DIFF
--- a/src/colmap/controllers/pairing.cc
+++ b/src/colmap/controllers/pairing.cc
@@ -492,8 +492,7 @@ void SequentialPairGenerator::MaybeExpandRigImages(image_t image_id1,
 
 bool SequentialPairGenerator::IsValidSequentialNeighbor(
     image_t image_id1, image_t image_id2) const {
-  if (!options_.expand_rig_images ||
-      !image_to_frame_id_.contains(image_id1) ||
+  if (!options_.expand_rig_images || !image_to_frame_id_.contains(image_id1) ||
       !image_to_frame_id_.contains(image_id2)) {
     return true;
   }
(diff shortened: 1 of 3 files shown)


B1ueber2y (approved):
Thank you!

src/colmap/controllers/pairing.cc
```diff
@@ -515,11 +515,37 @@ std::vector<std::pair<image_t, image_t>> SequentialPairGenerator::Next() {
     }
   };
 
+  auto IsValidSequentialNeighbor = [this, image_id1](image_t image_id2) {
```
sarlinpe: Does this need to be a lambda? can't this be a method with arguments image_id1 and image_id2? Similarly for MaybeExpandRigImages.

sarlinpe (approved):

PR #4592, opened 2026-07-26
Initialize low-texture rigs from aggregate frame matches

ahojnnes:
## Summary

- fall back to frame-level correspondence support when every constituent rig image pair is below the initialization threshold
- estimate and validate generalized relative pose directly for that fallback
- preserve the existing individual-pair ranking and stability gate whenever a normal pair is available
- add a synthetic pipeline test where aggregate rig support succeeds above every individual pair

## Motivation

Rig initialization already estimates generalized relative pose from all matched sensor pairs, but it first required one constituent image pair to meet `init_min_num_inliers` and the individual two-view stability checks. A low-texture multi-camera frame can have sufficient aggregate evidence while every individual sensor pair remains below the threshold.

The fallback is deliberately narrow. An earlier implementation replaced normal rig ranking and regressed a successful TartanAir control by selecting a distant frame; this version leaves the established path unchanged and activates only when no constituent pair anywhere in the source frame qualifies.

## Benchmark

On 11 TartanAir scenes from #4590, fixed seed 0 and one thread per scene, this conservative fallback is neutral: pooled relative-pose AUC remains 44.28/46.82/48.89/49.20 at 0.5/1/5/10 degrees, with 388/550 registered panoramas. None of those real scenes has both sufficient aggregate support and no qualifying constituent pair. In combination with #4591, the result matches #4591 at 51.38/54.89/58.72/59.81. All four successful controls remain fully registered, with at most 0.03 AUC@10 change.

## Tests

- `ctest -R sfm/incremental_mapper_test --output-on-failure`
- `ctest -R controllers/incremental_pipeline_test --output-on-failure`

### DIFF
--- a/src/colmap/controllers/incremental_pipeline_test.cc
+++ b/src/colmap/controllers/incremental_pipeline_test.cc
@@ -136,6 +136,37 @@ TEST(IncrementalPipeline, WithoutNoiseAndWithNonTrivialFrames) {
   }
 }
 
+TEST(IncrementalPipeline, InitializesRigFromAggregatedFrameMatches) {
+  const auto database_path = CreateTestDir() / "database.db";
+
+  auto database = Database::Open(database_path);
+  Reconstruction gt_reconstruction;
+  SyntheticDatasetOptions synthetic_dataset_options;
+  synthetic_dataset_options.num_rigs = 1;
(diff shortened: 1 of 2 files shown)


B1ueber2y (approved):

ahojnnes:
Closing this as it didn't materially change the final metrics.

PR #4593, opened 2026-07-27
Fix pose-prior BA gating to count frames, not images (#4368)

### LINKED ISSUE #4368: Unable to align a sparse reconstruction from a sensor rig with pose priors.
I've been trying to get a valid reconstruction with a camera rig utilizing pose priors.
My setup consists of a rig with multiple camera sensors, for which I have accurate pose priors.
The final reconstruction itself is valid, but not aligned with the pose priors. Although the reconstruction is not positioned around the origin, it is still several hundred meters away from where it should be.
After the registration of the initial image pair, I get the two warnings:
[alignment.cc:272] Not enough valid pose priors for alignment
[bundle_adjustment_ceres.cc:1050] Alignment w.r.t. prior positions failed

I'm using pycolmap, built from source (version 4.0.4), and my workflow is the following:
- Populate the database manually.
  - add cameras
  - add a single rig, set one camera as reference, and add the other sensors with a transformation from the reference to this camera. 
  - add frames, ensure data_t will match the images.
  - add images
- add pose priors for the reference camera sensors to each frame, position is in world coordinates, coordinate system is cartesian, covariance is the identity matrix.
- feature extraction using Aliked.
- sequential feature matching.
- incremental_mapping, with use_prior_position=True

If I run the same pipeline with a single camera sensor in the rig, everything works fine, no warnings, and the alignment is correct.

It does feel like I'm just overseeing something, but I'm very much stuck, so any help is greatly appreciated. Thanks!

kunjikakgp:
Fixes #4368.

AdjustGlobalBundle() decided whether to enable pose-prior bundle adjustment using `ba_config.NumImages() > 2`. This counts every image across every sensor, but a multi-camera rig registers several images per frame/position. E.g. a 2-camera rig with only 2 registered frames already has 4 images, so the ">2" check passed with just 2 distinct positions. AlignReconstructionToPosePriors() requires >= 3 correspondences to fit a similarity transform, so it then failed, leaving the reconstruction unaligned with the pose priors (see #4368, diagnosed by @hayamdk).

This changes the check to use `reconstruction_->NumRegFrames()`, which counts distinct registered frames/positions instead of raw images.

Testing: this is a targeted, minimal condition change (swap `ba_config.NumImages()` for `reconstruction_->NumRegFrames()`, both already used elsewhere in this file/function) reviewed by reading `incremental_mapper.cc`, `alignment.cc`, and `bundle_adjustment.h`. I was not able to build the full COLMAP/Ceres toolchain in my environment to run `incremental_mapper_test`, so I'd appreciate a maintainer double-checking against CI and, ideally, against the rig database @MarwinSc attached to #4368.

### DIFF
--- a/src/colmap/sfm/incremental_mapper.cc
+++ b/src/colmap/sfm/incremental_mapper.cc
@@ -1172,9 +1172,14 @@ bool IncrementalMapper::AdjustGlobalBundle(
     }
   }
 
-  // Only use prior pose if at least 3 images have been registered.
+  // Only use prior pose if at least 3 frames (not images) have been
+// registered. Count frames, not images: a multi-camera rig registers
+// several images per frame/position, so NumImages() overstates the
+// number of distinct positions, which could enable pose-prior BA with
+// fewer than 3 usable positions and make AlignReconstructionToPosePriors()
+// fail (see #4368).
   const bool use_prior_position =
-      options.use_prior_position && ba_config.NumImages() > 2;
+      options.use_prior_position && reconstruction_->NumRegFrames() > 2;
 
   std::unique_ptr<BundleAdjuster> bundle_adjuster;
   if (!use_prior_position) {


ahojnnes:
Thank you but I don't think this is the right fix. `AlignReconstructionToPosePriors()` requires at least 3 registered images, which `ba_config.NumImages()` should be equivalent to. There can still be a degeneracy, where the 3 images' poses are not distinct or colinear. Is this issue still confirmed to be present in latest colmap versions?

ahojnnes:
PTAL: https://github.com/colmap/colmap/pull/4594

PR #4597, opened 2026-07-28
Agent/bundle adjuster pose priors

jizongFox:


jizongFox:
accidential PR, please ignore:)

PR #4598, opened 2026-07-29
Fix missing gflags::gflags alias for downstream CMake consumers

StonerLing:
## Description

When consuming COLMAP as a CMake package with:

```cmake
find_package(colmap CONFIG REQUIRED)
target_link_libraries(app PRIVATE colmap::colmap)
```

the downstream project may fail during linking/configuration if the transitive gflags target namespace does not match the one expected by the exported dependencies.

```
LINK : fatal error LNK1181: cannot open input file 'gflags.lib'
```

## Root cause

The exported glog::glog target references gflags::gflags as an interface dependency. However, gflags only defines this namespaced target when `GFLAGS_USE_TARGET_NAMESPACE=ON`; otherwise, it only provides the plain gflags target.

COLMAP relies on the Ceres → glog → gflags dependency chain, so this target namespace mismatch can propagate to downstream consumers linking against `colmap::colmap`.

The current pycolmap CI does not reproduce this issue because it builds with the vcpkg `CMAKE_TOOLCHAIN_FILE`, which keeps the dependency configuration consistent. However, other consumers using the same installed packages without that toolchain file (or using a different packaging setup) may encounter this issue.

### DIFF
--- a/python/CMakeLists.txt
+++ b/python/CMakeLists.txt
@@ -32,38 +32,6 @@ endif()
 
 find_package(colmap REQUIRED)
 
-# On Windows with MSVC, the linker receives bare "gflags.lib" (no path) because
-# CMake's multi-config generators fail to propagate IMPORTED_IMPLIB_<CONFIG>
-# from the gflags imported target to the linker. Both glog::glog and Ceres::ceres
-# reference gflags in their INTERFACE_LINK_LIBRARIES, so replace those references
-# with the absolute path to gflags.lib to bypass CMake target resolution.
-if(MSVC)
-    if(TARGET gflags::gflags_shared)
-        set(_gflags_target gflags::gflags_shared)
-    elseif(TARGET gflags_shared)
-        set(_gflags_target gflags_shared)
-    endif()
-    if(_gflags_target)
-        get_target_property(_gflags_implib ${_gflags_target} IMPORTED_IMPLIB_RELEASE)
-        if(_gflags_implib)
-            if(TARGET glog::glog)
-                get_target_property(_glog_iface glog::glog INTERFACE_LINK_LIBRARIES)
-                string(REPLACE "gflags::gflags" "${_gflags_implib}" _glog_iface_fixed "${_glog_iface}")
-                set_target_properties(glog::glog PROPERTIES
-                    INTERFACE_LINK_LIBRARIES "${_glog_iface_fixed}")
-            endif()
-            if(TARGET Ceres::ceres)
-                get_target_property(_ceres_iface Ceres::ceres INTERFACE_LINK_LIBRARIES)
-                string(REPLACE ";gflags;" ";${_gflags_implib};" _ceres_iface_fixed "${_ceres_iface}")
-                set_target_properties(Ceres::ceres PROPERTIES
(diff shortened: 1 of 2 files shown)


ahojnnes:
Thank you @StonerLing. Do you see the same issue also for the hosted pycolmap binaries on PyPI?

StonerLing:
No. The pycolmap package from PyPI works fine when running `benchmark/reconstruction/evaluate.py`. 

This PR addresses issues encountered when building pycolmap locally with MSVC and vcpkg. However, there is another platform compatibility issue that I attempted to resolve in #4600.

python/CMakeLists.txt
```diff
@@ -32,6 +32,38 @@ endif()
...
+            if(TARGET Ceres::ceres)
+                get_target_property(_ceres_iface Ceres::ceres INTERFACE_LINK_LIBRARIES)
+                string(REPLACE ";gflags;" ";${_gflags_implib};" _ceres_iface_fixed "${_ceres_iface}")
+                set_target_properties(Ceres::ceres PROPERTIES
+                    INTERFACE_LINK_LIBRARIES "${_ceres_iface_fixed}")
+            endif()
+        endif()
+    endif()
+    unset(_gflags_target)
+    unset(_gflags_implib)
```
sarlinpe: I suggest extracting the new code into a function such that this cleanup is unnecessary.

sarlinpe (commented):
Thank you!
1) How important is it to run in editable mode on MSVC, and is it worth the extra patching code?
2) Could this be fixed in VCPKG instead of COLMAP?

StonerLing:
@sarlinpe Thanks for your question! It helped me trace the root cause, which is actually not specific to `pycolmap` — it comes from an inconsistency in the transitive CMake dependency chain exposed by `find_package(colmap)`, **and therefore can affect any downstream project consuming `colmap` through CMake.**

---

**Root cause**

The exported `glog::glog` target references `gflags::gflags` as an interface dependency, while `gflags` only provides this namespaced target when `GFLAGS_USE_TARGET_NAMESPACE=ON`. Otherwise, only the plain `gflags` target exists.

Since COLMAP relies on the Ceres → glog → gflags dependency chain, this namespace mismatch can propagate to downstream consumers linking against `colmap::colmap`.

**Fix**

I added a small compatibility fix in the latest commit (`cmake/FindDependencies.cmake`) to bridge this namespace mismatch when necessary. PTAL!

This makes the COLMAP package configuration self-contained: `find_package(colmap)` + `target_link_libraries(... colmap::colmap)` works without requiring consumers to know how COLMAP's dependencies were built. Existing working configurations are unaffected.

---

Regarding your questions:

1. Editable mode is mainly useful for developers who modify the source code; it is unrelated to this issue.
    
2. Strictly speaking, using the vcpkg install tree without `CMAKE_TOOLCHAIN_FILE` is not the officially supported workflow. However, this is not vcpkg-specific: the exported `glog::glog` target assumes the existence of `gflags::gflags`, which may not hold across different packaging setups. Since COLMAP cannot control downstream dependency resolution, fixing the exported config is the more robust solution.

ahojnnes (approved):
Great, LGTM. Thank you!

PR #4600, opened 2026-07-30
Fix preexec_fn crash on Windows in benchmark evaluation script

StonerLing:
`subprocess.Popen` on Windows does not support the `preexec_fn` parameter. Passing it unconditionally in `_run_with_log` and the `database_cleaner` call raises `ValueError`, preventing the benchmark from running on Windows.

Guard these two calls with `platform.system() == "Linux"` so the script works cross-platform. On non-Linux systems child-process cleanup already relies on `multiprocessing.Pool.terminate()`, so `preexec_fn` is merely a best-effort addition on Linux.


sarlinpe (approved):

ahojnnes:
Thank you. I think we also need this on Mac and not just Linux?

StonerLing:
I came across some information suggesting that `prctl(PR_SET_PDEATHSIG)` is Linux-specific. Could you confirm if that's correct? I don't have access to a macOS machine to verify it myself.

StonerLing:
https://github.com/colmap/colmap/blob/954de4871784f5370ae9198696e895c49bc7490f/benchmark/reconstruction/evaluation/utils.py#L65-L69

Alright, I just noticed this is already guarded by platform checks. On macOS, this feature is simply skipped...

More precisely, I guess the platform differences are: prctl(PR_SET_PDEATHSIG) is only supported on Linux; macOS supports preexec_fn but not prctl(PR_SET_PDEATHSIG); Windows supports neither.

ahojnnes (approved):
Thank you!

ahojnnes:
Needs a code-format. Thanks.

PR #4607, opened 2026-08-04
Optimize image warping for small resizes

ahojnnes:
## Summary

- add `WarpImageOptions::direct_warp_min_scale`, defaulting to `0.5`, to warp directly at the target resolution for equal-size outputs and reductions of up to 2x per dimension
- retain the source-resolution warp plus OpenImageIO resize when either target/source dimension ratio is below the threshold
- propagate the options through camera undistortion and stereo rectification
- expose `WarpImageOptions` and `UndistortCameraOptions.warp_options` in pycolmap

## Performance

Release benchmarks at 1920x1080 on the development host:

| Camera model | Legacy | Direct warp (default) | Speedup |
|---|---:|---:|---:|
| OPENCV | 230.6 ms | 34.0 ms | 6.8x |
| OPENCV_FISHEYE | 296.2 ms | 71.3 ms | 4.2x |

The improvement comes from avoiding both the higher source-resolution pixel loop and the final generic OpenImageIO resize. These benchmark outputs qualify for direct warping with the new `0.5` default.

## Accuracy tradeoff

For a pure, aligned 2x reduction, COLMAP's pixel-center convention makes bilinear interpolation average the corresponding 2x2 source pixels. This makes `0.5` a practical cutoff for ordinary imagery rather than limiting direct warping to very small reductions.

This is not an unconditional antialiasing guarantee: undistortion and homography mappings can produce rotated, stretched, or spatially varying source footprints, and a 2x2 box average is not an ideal low-pass filter. Reductions larger than 2x retain the legacy source-resolution warp and OpenImageIO resize. Callers can set the threshold to `1.0` to retain that path for every downscale, or lower it to favor performance.

## Tests

- `ctest --output-on-failure -R '(image/(warp|undistortion)|controllers/undistorters)_test'`
- pycolmap undistortion and image pipeline tests (21 passed)
- `git diff --check`

### DIFF
--- a/src/colmap/image/undistortion_test.cc
+++ b/src/colmap/image/undistortion_test.cc
@@ -34,6 +34,10 @@
 #include "colmap/sensor/bitmap.h"
 #include "colmap/util/eigen_matchers.h"
 
+#include <algorithm>
+#include <cstdint>
+#include <cstdlib>
+
 #include <gtest/gtest.h>
 
 namespace colmap {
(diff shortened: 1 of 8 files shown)


src/colmap/image/undistortion_test.cc
```diff
@@ -225,6 +225,41 @@ TEST(UndistortCamera, NoBlankPixels) {
   }
 }
```
lpanaf, line 264: Does it make sense to add a new TEST comparing the results with direct warp and rescaled warp to ensure that the difference is not large when it satisfy the direct warp condition?

lpanaf (approved):

src/colmap/image/undistortion_test.cc
ahojnnes: Added in 8d70f877e. The test now uses the default direct-warp threshold, verifies that the output scale satisfies it, forces the legacy resize path for comparison, and checks that a smooth RGB image differs by less than 0.5 intensity levels on average and at most 1 level per channel.

PR #4608, opened 2026-08-04
Expose interpolation mode for image warping

ahojnnes:
## Summary

- add `WarpImageOptions::Interpolation` with nearest-neighbor and bilinear sampling modes
- keep bilinear interpolation as the default to preserve existing behavior
- dispatch once per image to interpolation-specialized warp implementations, avoiding a switch in the pixel loops
- apply the selected interpolation to both option-aware camera warp functions
- expose the enum as `pycolmap.WarpImageOptions.Interpolation` and the selected mode as `WarpImageOptions.interpolation`

Python usage:

```python
options = pycolmap.WarpImageOptions()
options.interpolation = (
    pycolmap.WarpImageOptions.Interpolation.NEAREST_NEIGHBOR
)
```

## Performance

In a randomized 1920x1080 benchmark, selecting interpolation inside the pixel loop added 7.4% median CPU time relative to an otherwise equivalent fixed-bilinear loop. Dispatching once to specialized implementations reduced the measured gap to 3.3%; the generated code contains no per-pixel interpolation dispatch.

## Tests

- `ctest --output-on-failure -R '(image/(warp|undistortion)|controllers/undistorters)_test'`
- pycolmap undistortion and image pipeline tests (23 passed)
- `ruff check src/pycolmap/image/undistortion_test.py`
- `git diff --check`

### DIFF
--- a/src/colmap/image/warp_test.cc
+++ b/src/colmap/image/warp_test.cc
@@ -183,6 +183,33 @@ TEST(Warp, DirectWarpMinScale) {
                                            &default_image));
 }
 
+TEST(Warp, Interpolation) {
+  const Camera source_camera =
+      Camera::CreateFromModelId(1, CameraModelId::kPinhole, 4, 4, 4);
+  Camera target_camera = source_camera;
+  target_camera.SetPrincipalPointX(1.5);
+
+  Bitmap source_image(4, 4, /*as_rgb=*/false);
+  for (int y = 0; y < source_image.Height(); ++y) {
+    source_image.SetPixel(0, y, BitmapColor<uint8_t>(0));
+    source_image.SetPixel(1, y, BitmapColor<uint8_t>(100));
+  }
+
+  WarpImageOptions options;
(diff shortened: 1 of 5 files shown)


src/colmap/image/warp_test.cc
```diff
@@ -320,6 +352,51 @@ TEST(Warp, HomographyDirectWarpMinScale) {
                                             &default_image));
 }
```
lpanaf, line 354: Potentially can add another test which take a source camera and a shifted camera (shift left by 1.1px, for example) as the target camera. Then, with the nearest neighbor strategy, it should expect the images to be exactly the same as shifting left by 1px?

lpanaf (changes requested):

lpanaf (approved):

src/colmap/image/warp_test.cc
ahojnnes: Thanks, I think the existing test that checks for a single value is sufficient.

PR #4609, opened 2026-08-04
Inline bitmap interpolation methods

ahojnnes:
## Summary

- Move `Bitmap::InterpolateNearestNeighbor` and `Bitmap::InterpolateBilinear` into `bitmap.h` so callers can inline them.
- Preserve the existing interpolation algorithms, API, and numerical results.
- Enable the optimization for all call sites, including image warping, MVS fusion and texture mapping, and reconstruction color extraction.

## Performance

Measured with a 1920x1080 RGB camera warp in a Release build with IPO on an Apple M3 Pro. Results are median CPU time over 30 randomized repetitions:

| Interpolation | Out of line | Inlined | Improvement |
| --- | ---: | ---: | ---: |
| Bilinear | 34.3 ms | 31.1 ms | 9.3% |
| Nearest neighbor | 22.9 ms | 22.7 ms | 0.9% |

Bilinear benefits more because inlining lets the compiler optimize interpolation together with the subsequent `BitmapColor<float>` to `BitmapColor<uint8_t>` conversion and eliminate the intermediate `std::optional` boundary.

I also evaluated changing bilinear arithmetic to `float`. Float accumulation was only 0.4% faster in the complete warp and full-float coordinates were 0.4% slower, both within noise, so this PR retains double-precision behavior.

## Testing

- `ctest --output-on-failure -R "^(sensor/bitmap_test|image/warp_test)$"`

### DIFF
--- a/src/colmap/sensor/bitmap.cc
+++ b/src/colmap/sensor/bitmap.cc
@@ -231,68 +231,6 @@ void Bitmap::Fill(const BitmapColor<uint8_t>& color) {
   }
 }
 
-std::optional<BitmapColor<uint8_t>> Bitmap::InterpolateNearestNeighbor(
-    const double x, const double y) const {
-  const int xx = static_cast<int>(std::round(x));
-  const int yy = static_cast<int>(std::round(y));
-  return GetPixel(xx, yy);
-}
-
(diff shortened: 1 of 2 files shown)


lpanaf (approved):
Seems like a straightforward refactoring, so I did not check in details and trust the tests results.

PR #4610, opened 2026-08-04
Support sampson refinement of fundamental matrix with tiny solver

B1ueber2y:
Refinement on sampson error with tiny solver. Enabled by default. Benchmark numbers posted below.


B1ueber2y:
## Two-view geometry — 11,468 ETH3D pairs

Every cached image pair from the 25 uncalibrated ETH3D dslr workspaces. Both arms run in one process on identical correspondences with identical RANSAC seeds, so the comparison is exactly paired. The estimated `F` is converted to
`E` with the ground-truth intrinsics and decomposed to a relative pose, scored against ground truth by `max(rotation, translation)` error. AUC is a C++ port of `compute_auc()` from `benchmark/reconstruction/evaluation/utils.py`, verified to match it to 10 decimals.

| | AUC@0.5 | AUC@1 | AUC@5 | AUC@10 | med rot | med trans | inliers | ms/call |
|---|---:|---:|---:|---:|---:|---:|---:|---:|
| `off` | 20.22 | 29.81 | 49.76 | 56.14 | 0.426° | 1.519° | 396.8 | 8.77 |
| `on` | 21.83 | 31.00 | 50.02 | 56.21 | 0.397° | 1.492° | 397.0 | 9.58 |
| **delta** | **+1.60** | **+1.20** | **+0.27** | **+0.07** | −0.029° | −0.027° | +0.2 | **1.09×** |

## End-to-end reconstruction (incremental) — ETH3D, 25 scenes × 5 seeds

`benchmark/reconstruction/evaluate.py`, ETH3D dslr, `--uncalibrated`, incremental mapper, `--quality high`, 5 seeds per arm, 125 paired (scene, seed) draws. Each leg re-verifies the cached raw matches (`--overwrite_two_view_geometries`) and re-runs the mapper. Mean ± std over seeds; the delta is paired per seed.

| | AUC@0.5 | AUC@1 | AUC@5 | AUC@10 |
|---|---:|---:|---:|---:|
| `off` | 36.31 ± 0.71 | 58.13 ± 1.04 | 87.16 ± 1.64 | 92.04 ± 1.71 |
| `on` | 36.52 ± 0.83 | 58.27 ± 1.20 | 87.11 ± 1.71 | 92.00 ± 1.73 |
| **delta** | **+0.21 ± 1.05** | **+0.14 ± 1.42** | **−0.06 ± 2.46** | **−0.04 ± 2.56** |

## Summary

The refinement measurably improves two-view geometry on real data (+1.60 AUC@0.5, p ≈ 1e-14) for 1.09× the estimation cost, and that improvement does not reach reconstruction accuracy: it does not change much the inlier counts, so the mapper receives the same correspondences, and bundle adjustment re-estimates the geometry downstream.

B1ueber2y:
### Global mapper

| | AUC@0.5 | AUC@1 | AUC@5 | AUC@10 |
|---|---:|---:|---:|---:|
| `off` | 35.01 ± 0.26 | 56.90 ± 0.35 | 86.30 ± 0.51 | 91.36 ± 0.59 |
| `on` | 34.91 ± 0.54 | 56.65 ± 1.15 | 85.91 ± 1.82 | 91.12 ± 1.83 |
| **delta** | **−0.10 ± 0.66** | **−0.25 ± 1.13** | **−0.39 ± 1.39** | **−0.24 ± 1.29** |

ahojnnes (commented):
Two findings from my review:

1. **The independent point normalizations change the objective being minimized.** In `RefineFundamentalMatrixSampson`, each view receives its own Hartley scale before the Sampson residual is evaluated. Sampson error is invariant to a common coordinate scale, but not to two different scales: the two gradient terms in the denominator are then weighted differently. As a result, this is not generally minimizing the pixel-space Sampson error used by RANSAC. In a deterministic synthetic reproduction, a 2x scale difference between the two views let the refinement increase the original pixel-space mean squared Sampson error by as much as 5.22x. Please use a common normalization scale for both views, or compensate the residual/Jacobian so the optimized metric remains the original pixel-space metric.

2. **The new opt-out is not wired through the normal user-facing entry points.** `TwoViewGeometryOptions::use_sampson_refinement` defaults to true, but `OptionManager::AddTwoViewGeometryOptions()` does not register `TwoViewGeometry.use_sampson_refinement`, so feature-matching CLI/config users cannot disable it. In addition, `pycolmap.estimate_fundamental_matrix` selects `FundamentalMatrixSampsonEstimator` unconditionally and only accepts `RANSACOptions`, leaving that API without an opt-out. Please expose the option in both paths.

B1ueber2y:
> Two findings from my review:
> 
Agreed with both. Addressed: 40ae486e7b10d63955d0e299566a30d1d3b183fa

ahojnnes (approved):
Thank you, LGTM.

PR #4611, opened 2026-08-04
Add OPENCV camera model support to Caspar GPU bundle adjustment

bjoernellens1:
Adds native OPENCV camera-model support to Caspar's GPU bundle adjustment (CUDA only).

- `bundle_adjustment_caspar.cc`/`caspar_model_adapter.h`: dispatch for `OPENCV`-model cameras through Caspar (previously PINHOLE/SIMPLE_RADIAL only).
- `Symforce-Caspar/generated/f32/`: kernel tree extended with `opencv_core`/`opencv_split_core` node types (4 merged + 11 split variants), cross-checked against `OpenCVCameraModel::Distortion`/`ImgFromCam` for an exact param match.

Per @ahojnnes's request, this PR is back to its original CUDA-only OpenCV scope. HIP support for Caspar is now a separate PR: #4618. No dependency between them.

Verification unchanged from original submission (CUDA build, real distorted-lens data).


xiemeilong:
Are you planning to add support for OPENCV_FISHEYE?

bjoernellens1:
> Are you planning to add support for OPENCV_FISHEYE?

Not yet, but it's a natural next step and should follow the same recipe as this PR's OPENCV support — happy to take a stab at it if there's interest, or if someone else wants to pick it up this PR's caspar_generate.py changes are a reasonable template to follow.

ahojnnes:
Thank you. It will be great, if we can disentangle the two changes and have separate PRs for (1) adding HIP support for Caspar and (2) adding OPENCV as a supported camera model to Caspar.

bjoernellens1:
Split per your request:

- This PR (#4611): back to original scope, OPENCV camera model for Caspar, CUDA-only.
- #4618: ROCm/HIP compilation support for Caspar (PINHOLE/SIMPLE_RADIAL only), stacked on #4420 like #4614.

Independently mergeable, no dependency either direction. (The GlobalMapper fixed-rotation fix that was also in here got merged separately as #4612 already.)

bjoernellens1:
Pushed a fix for a real correctness bug found while validating this on real GPU hardware: the two OpenCV split-variant score kernels (`kernel_opencv_split_fixed_principal_point_score.cu` and its `_fixed_pose_` sibling) read their per-thread score accumulator unconditionally in `SumStore()`, but only ever assign it inside a `problem_size` guard. For any thread beyond `problem_size` (the common case — any block that isn't an exact multiple of 1024 factors), that's an uninitialized-variable read, which reliably decoded as NaN for these two kernels specifically (larger, more register-pressured than SimpleRadial's/Pinhole's equivalents) and corrupted the whole reduction from iteration 0.

Fix: explicitly zero the accumulator for out-of-range threads before the reduction, removing the undefined-behavior read.

Verified on a 613-image/54458-point real reconstruction: went from a false 3-iteration convergence exit with unchanged output to a genuine 70-90-iteration optimization (score `4.5487e5 -> ~4.4298e5`), consistent across repeated runs. PINHOLE/SIMPLE_RADIAL/OpenCV-with-zeroed-distortion controls still converge correctly with no regression.

PR #4612, opened 2026-08-05
Fix Caspar BA option handling and GlobalMapper fallback

bjoernellens1:
## Summary

- Use Ceres for the fixed-rotation stage of GlobalMapper iterative bundle adjustment when Caspar is selected.
- Make the Caspar backend honor `min_track_length` and `refine_points3D`.
- Report that `constant_rig_from_world_rotation` is unsupported by Caspar instead of silently ignoring it.
- Add backend diagnostics and shared Ceres/Caspar tests for the supported options, including external observations.

## Motivation

GlobalMapper performs iterative bundle adjustment in two stages: it first holds rig rotations constant to stabilize positions, then runs full joint optimization. Caspar represents each pose as one retracted `Pose3`, so it cannot keep rotation fixed while refining translation. The fixed-rotation option was therefore ignored and the first stage became another full joint-optimization pass.

A downstream three-scene accuracy sweep found internally inconsistent scale regimes in two non-trivial reconstructions when `GlobalMapper.ba_backend=CASPAR` was used. The standalone Caspar bundle adjuster also did not consistently apply the generic minimum-track-length and fixed-point options.

## Fix

GlobalMapper now uses Ceres only for its fixed-rotation sub-stage and retains Caspar for the more expensive full joint-optimization stage. Caspar now filters short tracks, keeps points fixed when point refinement is disabled, and emits an explicit diagnostic for its remaining constant-rotation limitation.

## Verification

- All 17 tests in the local Ceres-enabled `bundle_adjustment_test` binary pass.
- Shared Ceres tests cover `min_track_length` and `refine_points3D`.
- Dedicated Caspar tests cover `min_track_length`, including external observations, and `refine_points3D`.
- Both previously regressed real scenes reconstruct without inconsistent scale regimes after the GlobalMapper fallback (previously 2 and 5 regimes). The internal BA loop retains a substantial speedup over CPU/Ceres (approximately 299 s versus 884 s on a 1,599-frame scene).

Related: #4484, #4611.

### DIFF
--- a/src/colmap/sfm/global_mapper.cc
+++ b/src/colmap/sfm/global_mapper.cc
@@ -355,16 +355,7 @@ bool GlobalMapper::IterativeBundleAdjustment(
       // Caspar's pose node is a single retracted Pose3 (rotation+
       // translation together) with no mechanism to hold rotation
       // constant while translation is free -- constant_rig_from_world_
-      // rotation is silently ignored when backend == CASPAR (confirmed:
-      // zero references to it anywhere in bundle_adjustment_caspar.cc),
-      // so this stage would otherwise silently degrade into a second full
-      // joint optimization pass instead of a stabilizing partial one.
-      // Live 3-scene accuracy sweep confirmed this causes real
-      // scale-regime-split regressions on non-trivial scenes (floor2:
-      // 0->2, trolley_femto: 0->5; only the smallest/simplest scene
-      // stayed clean). Force Ceres for just this sub-stage -- it's the
-      // cheap half of the loop anyway; the joint-optimization stage below
-      // keeps the requested backend (Caspar's actual speed win).
+      // rotation is silently ignored when backend == CASPAR.
       if (opts_position_only.backend == BundleAdjustmentBackend::CASPAR) {
         opts_position_only.backend = BundleAdjustmentBackend::CERES;
       }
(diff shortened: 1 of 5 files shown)


src/colmap/sfm/global_mapper.cc
```diff
@@ -352,6 +352,22 @@ bool GlobalMapper::IterativeBundleAdjustment(
...
+      // rotation is silently ignored when backend == CASPAR (confirmed:
+      // zero references to it anywhere in bundle_adjustment_caspar.cc),
+      // so this stage would otherwise silently degrade into a second full
+      // joint optimization pass instead of a stabilizing partial one.
+      // Live 3-scene accuracy sweep confirmed this causes real
+      // scale-regime-split regressions on non-trivial scenes (floor2:
+      // 0->2, trolley_femto: 0->5; only the smallest/simplest scene
+      // stayed clean). Force Ceres for just this sub-stage -- it's the
+      // cheap half of the loop anyway; the joint-optimization stage below
+      // keeps the requested backend (Caspar's actual speed win).
```
ahojnnes:
```suggestion
      // Caspar's pose node is a single retracted Pose3 (rotation+
      // translation together) with no mechanism to hold rotation
      // constant while translation is free -- constant_rig_from_world_
      // rotation is silently ignored when backend == CASPAR.
```

ahojnnes (approved):
Thanks, I added a few more tweaks.

PR #4614, opened 2026-08-05
Add ROCm/HIP support for GPU SIFT (feature extraction + matching)

bjoernellens1:
## Summary

Adds ROCm/HIP support for GPU SIFT (feature extraction + matching via the bundled `SiftGPU` third-party library), extending the dense-reconstruction HIP port in #4420 to feature extraction/matching. Includes a real, well-diagnosed bug fix for an out-of-bounds texture read that only manifests on HIP (not CUDA).

**This PR is stacked on #4420** ("Add ROCm/HIP support for patch_match_stereo (AMD GPU)", by @iShengnan) and cannot be reviewed/merged independently of it — GPU SIFT under HIP needs the `HIP_ENABLED` CMake option, `enable_language(HIP)` plumbing, and `cuda_to_hip.h` compat header that PR introduces. The branch is `ishengnan/rocm-support` rebased onto current `upstream/main`, with the SIFT-specific commits added on top; only the last 4 commits (see below) are new content for this PR — everything else is #4420's own, unmodified.

### What's added

- `src/thirdparty/SiftGPU/CMakeLists.txt`: parallel `if(HIP_ENABLED)` block alongside the existing `if(CUDA_ENABLED)` block, both defining `SIFTGPU_CUDA_ENABLED` (the third-party's GPU-backend-on gate).
- Two CUDA-tolerated-but-HIP-fatal bugs fixed so GPU SIFT actually runs on AMD hardware:
  1. `CuTexImage::CuTexObj` lacked rule-of-five — the compiler-synthesized copy constructor copied the raw texture handle, so any copy-initialization path (e.g. NRVO not eliding a copy) triggered a double-`cudaDestroyTextureObject`/`hipDestroyTextureObject` call.
  2. `ComputeOrientation`/`ComputeDescriptor` bound their pitched buffer as `cudaResourceTypePitch2D` and read it with `tex2D`. AMD's `hipResourceTypePitch2D` requires the pitch to be a multiple of `texturePitchAlignment` (256 bytes on gfx90a/gfx1151), which a tightly packed `CuTexImage` buffer can't guarantee for arbitrary pyramid levels — fixed by routing `tex2D` reads through linear texture binding on HIP instead.
- **The interesting one:** `ListGen_Kernel`'s tail-thread texture read was unguarded. The kernel's launch grid is rounded up to a multiple of `LISTGEN_BLOCK_DIM`, so `idx1` can run past `list_len` for the tail block. The only bounds check in the original (pre-HIP) code guards the *write* (`if (idx1 < list_len) d_list[idx1] = pos;`), but the *read* above it, `tex1Dfetch<int4>(texDataList, idx1)`, is unconditional. CUDA's `tex1Dfetch` on a linear texture silently returns zero for an out-of-range index, so on CUDA the tail threads' garbage `pos` is simply discarded by the write guard and nothing bad happens — this bug has been latent in upstream SIFT for as long as the OOB-tolerant CUDA behavior masked it. HIP does not provide that clamp-to-zero fallback: an out-of-range `tex1Dfetch` can read from unmapped device memory and faults with a "Page not present" signature. Root-caused with `AMD_LOG_LEVEL=3` + `HIP_LAUNCH_BLOCKING=1` tracing (the fault is synchronous with the `ListGen_Kernel` launch inside `GenerateFeatureList`'s GPU list-generation path), confirmed by ruling out two more plausible-looking suspects (the two commits above) with real instrumented builds before finding the actual cause. Fixed by clamping the tail-thread read index to `list_len - 1` before the `tex1Dfetch`, gated under `COLMAP_HIP_ENABLED` so the CUDA path stays byte-identical to upstream.

### Verification

Built and run on real gfx1151 hardware (AMD Strix Halo APU) inside a `rocm/pytorch:rocm7.2.4_ubuntu24.04_py3.12_pytorch_release_2.10.0` container, `-DCUDA_ENABLED=OFF -DHIP_ENABLED=ON -DCMAKE_HIP_ARCHITECTURES=gfx1151`. GPU SIFT extraction + matching completes end-to-end on real datasets:

- 3 fresh-process runs each on two different image sets (1280x720 and 640x480, 20 images each, chosen to exercise two different feature-list buffer growth trajectories) — 6/6 runs complete all 20 images with correct nonzero keypoint counts on every image, zero faults.

This isolated branch (this PR's commits on top of a fresh `upstream/main` rebase of #4420) was also test-built end-to-end in a clean container to confirm it compiles standalone with only #4420 + these commits (no dependency on any other unrelated work).


jeffdaily:
Not that I needed to have my ego stroked by submitting my own PR for GPU SIFT support since it was my agent that largely did the work, this current PR isn't clean.  I'm working on the rebase/report of the GPU SIFT work now.  Wanted to go through the entire moat porting process again to be 100% sure it works.

bjoernellens1:
Makes sense — closing this in favor of your from-scratch redo. One thing worth carrying over if useful: a real bug found and fixed in `ListGen_Kernel` (`ProgramCU.cu`) — an unguarded tail-thread `tex1Dfetch` read past `list_len` that's silently tolerated by CUDA's out-of-range clamp-to-zero but faults on HIP. Fix is a one-line clamp, commit `44d9b147a4` on this branch if you want to pull it in.

bjoernellens1:
@jeffdaily are you still working on that? contact me if you need help

jeffdaily:
@bjoernellens1 see https://github.com/colmap/colmap/pull/4635.

PR #4616, opened 2026-08-05
Minor: Expose optional cheirality check in ImgFromCam

B1ueber2y:
Also include some small cosmetic consistency improvements on the macro naming in models.h 

**No change of behaviors**

### DIFF
--- a/src/colmap/estimators/pose.cc
+++ b/src/colmap/estimators/pose.cc
@@ -91,7 +91,9 @@ bool EstimateAbsolutePose(const AbsolutePoseEstimationOptions& options,
     }
 
     ImgFromCamFunc img_from_cam_func =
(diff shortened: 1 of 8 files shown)


sarlinpe (approved):
1) I think that splitting this PR into the new argument vs the renaming of the macros would make it easier to parse.
2) Why would we need to disable the chirality check? I fail to see what justifies the added complexity.

B1ueber2y:
Thank you!

> 1. I think that splitting this PR into the new argument vs the renaming of the macros would make it easier to parse.
> 
I consider this a minor PR (as in the title) so I broke a bit the proper practice for fast dev.

> 2. Why would we need to disable the chirality check? I fail to see what justifies the added complexity.

It can be helpful for debugging for non-cheirality relevant cases.

PR #4621, opened 2026-08-06
Add pycolmap bindings for the hierarchical mapper

ahojnnes:
Exposes the hierarchical mapping pipeline to Python, and brings the global mapper bindings up to the same level of completeness.

## Hierarchical mapper (new)

- `pycolmap.SceneClusteringOptions` and `pycolmap.HierarchicalPipelineOptions` dataclasses
- `pycolmap.HierarchicalPipeline` class with `run()`
- `pycolmap.hierarchical_mapping(database_path, image_path, output_path, options)`, mirroring `incremental_mapping()` and `global_mapping()`

```python
options = pycolmap.HierarchicalPipelineOptions()
options.clustering_options.leaf_max_num_images = 500
reconstructions = pycolmap.hierarchical_mapping(
    database_path, image_path, output_path, options
)
```

To share one code path between the CLI and Python, `RunHierarchicalMapperImpl()` is factored out of `RunHierarchicalMapper()`, mirroring the existing `RunGlobalMapperImpl()`.

`add_callback`/`callback` are deliberately not bound on `HierarchicalPipeline` — unlike the other two controllers it registers no callbacks, so they would be dead API.

## Global mapper

While adding the above it became apparent that the global mapper bindings were the least complete of the three:

- `GlobalPipeline` was not bound as a class at all, only the free `global_mapping()` function. It is now bound along with its `MODEL_UPDATE_CALLBACK` (new `pycolmap.GlobalPipelineCallback` enum), so the in-progress reconstruction can be rendered from Python.
- `GlobalPipelineOptions.image_path`, `GlobalMapperOptions.image_path` and `GlobalMapperOptions.ba_gpu_index` were never bound. Not broken in practice, since `global_mapping()` sets `image_path` from its argument, but unreachable when driving the options directly.
- The `RotationAveraging()`, `GlobalPositioning()`, `BundleAdjustment()` and `Retriangulation()` accessors — the analogues of `IncrementalPipelineOptions.get_mapper()` — were not bound. They are now exposed as `get_*`.

`GlobalMapperOptions`/`GlobalPipelineOptions` are moved from `pipeline/sfm.cc` into the new `sfm/global_mapper.cc`, so all three mappers follow the same layout: controller and options in `pycolmap/sfm/<name>_mapper.cc`, free pipeline function in `pycolmap/pipeline/sfm.cc`. This requires binding the global mapper after the incremental triangulator (`GlobalMapperOptions.retriangulation` is an `IncrementalTriangulatorOptions`), which is noted in `sfm/bindings.cc`. Happy to drop this move if it is more churn than wanted here.

## Tests

Previously no pycolmap test ran a mapping pipeline to completion — `sfm_test.py` only asserted the functions were `callable`, and the only end-to-end coverage lived in `python/examples/custom_incremental_pipeline_test.py`, which drives a Python reimplementation of the loop rather than `IncrementalPipeline.run()`.

Added end-to-end tests on synthetic datasets, mirroring the corresponding C++ controller tests and checking recovered poses against ground truth via `compare_reconstructions`:

- `IncrementalPipeline.run()`, `GlobalPipeline.run()`, `HierarchicalPipeline.run()`
- `incremental_mapping()`, `global_mapping()`, `hierarchical_mapping()`

The global pipeline test also asserts that `MODEL_UPDATE_CALLBACK` actually fires, and gives `calibrate_view_graph()` its first real coverage. Each new test runs in well under a second.

### DIFF
--- a/src/colmap/exe/sfm.cc
+++ b/src/colmap/exe/sfm.cc
@@ -422,6 +422,28 @@ int RunGlobalMapper(int argc, char** argv) {
   return EXIT_SUCCESS;
 }
 
+bool RunHierarchicalMapperImpl(
+    const std::filesystem::path& database_path,
+    const std::filesystem::path& image_path,
+    const std::filesystem::path& output_path,
+    const std::shared_ptr<HierarchicalPipelineOptions>& mapper_options,
+    std::shared_ptr<ReconstructionManager>& reconstruction_manager) {
+  HierarchicalPipelineOptions options = *mapper_options;
+  options.image_path = image_path;
+
+  HierarchicalPipeline hierarchical_mapper(
+      options, Database::Open(database_path), reconstruction_manager);
+  hierarchical_mapper.Run();
+
+  if (reconstruction_manager->Size() == 0) {
+    LOG(ERROR) << "Failed to create sparse model";
+    return false;
+  }
+
+  reconstruction_manager->Write(output_path);
+  return true;
+}
+
 int RunHierarchicalMapper(int argc, char** argv) {
   std::filesystem::path output_path;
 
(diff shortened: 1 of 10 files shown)


sarlinpe (approved):
Thank you!

PR #4617, opened 2026-08-05
Estimate homography on rays for distorted cameras and spherical cases

B1ueber2y:
The existing homography estimation is not geometrically correct for distorted cameras. 

This PR includes the following improvements:

* Homography estimation on rays rather than pixels when one or both of the cameras have distortions. The homography estimation in practice is gated in practice, and remain the same for two non-distorted cameras (as the ray homography is 1.5 - 6x slower compared to the pixel homography). Scoring is kept the same on the pixel space. 
* Therefore able to add homography path for spherical cameras, and thereby support detection of planar/panoramic mode. 

Plane-dominated two-view scenes, 500 points, 1 px noise, 300 problems per cell, 4.0 px threshold.

AUC@1/5/10 deg on max(rot, trans) error, and mean ms per problem.

| camera             | outl | pixel          | pixel normalized     | ray            | ms px/pxnorm/ray    |
|--------------------|------|----------------|----------------|----------------|-------------------|
| PINHOLE            |   0% | 49.0 74.3 81.1 | 79.8 90.6 92.1 | 79.4 90.5 92.1 |  0.21  0.31  0.29 |
| PINHOLE            |  40% | 46.4 73.9 81.3 | 75.5 88.4 90.2 | 74.6 88.0 89.8 |  0.29  0.37  0.75 |
| OPENCV k1=0.05     |   0% |  0.3  4.8 14.4 |  0.2  4.7 13.9 | 81.1 92.5 94.1 |  0.43  0.58  0.31 |
| OPENCV k1=0.05     |  40% |  0.3  4.5 13.6 |  0.3  4.7 13.6 | 75.1 88.8 90.7 |  1.38  1.32  0.84 |
| OPENCV k1=0.30     |   0% |  0.0  0.7  1.7 |  0.0  0.7  1.5 | 69.3 85.8 88.7 |  5.83  5.43  0.29 |
| OPENCV k1=0.30     |  40% |  0.0  0.4  1.0 |  0.0  0.3  1.1 | 67.4 85.2 88.2 | 34.80 33.79  0.89 |
| OPENCV_FISHEYE     |   0% |  0.0  0.2  1.3 |  0.0  0.2  1.2 | 71.0 80.6 82.0 | 11.06 10.50  0.34 |
| OPENCV_FISHEYE     |  40% |  0.0  0.0  1.1 |  0.0  0.0  1.0 | 68.5 80.9 82.6 | 56.09 56.07  1.42 |
[...]

PR #4623, opened 2026-08-06
Budget the homography ransac with minimum inlier ratio

B1ueber2y:
Splitted from https://github.com/colmap/colmap/pull/4617 and also cover shared focal and one sided focal. 

| ETH3D undistorted PINHOLE, 2068 pairs | AUC@1/5/10 | ms | 
  |---|---|---|
  | baseline | 89.5 97.6 98.7 | 14.20 |
  | + budget | 89.6 97.6 98.7 | 9.47 |

  | ETH3D thinprism, 992 pairs | AUC@1/5/10 | ms |
  |---|---|---|
  | baseline | 88.4 97.2 98.5 | 14.70 |
  | + budget | 88.5 97.3 98.6 | 9.92 |

### DIFF
--- a/src/colmap/estimators/two_view_geometry.cc
+++ b/src/colmap/estimators/two_view_geometry.cc
@@ -158,8 +158,13 @@ TwoViewGeometry EstimateCalibratedHomography(
 
   // Estimate planar or panoramic model.
 
+  auto ransac_options = options.ransac_options;
+  if (options.min_inlier_ratio > 0) {
+    ransac_options.min_inlier_ratio = options.min_inlier_ratio;
+  }
+
   LORANSAC<HomographyMatrixEstimator, HomographyMatrixEstimator> H_ransac(
-      options.ransac_options);
+      ransac_options);
   const auto H_report =
       H_ransac.Estimate(matched_img_points1, matched_img_points2);
   geometry.H = H_report.model;
@@ -224,8 +229,15 @@ TwoViewGeometry EstimateUncalibratedTwoViewGeometry(
 
   // Estimate planar or panoramic model.
 
+  // Budget the search for the ratio the homography must beat to be selected
+  // below, since a weaker one is discarded anyway.
+  auto H_ransac_options = options.ransac_options;
+  H_ransac_options.min_inlier_ratio =
+      std::max(options.ransac_options.min_inlier_ratio,
+               options.max_H_inlier_ratio * F_report.support.num_inliers /
+                   matches.size());
   LORANSAC<HomographyMatrixEstimator, HomographyMatrixEstimator> H_ransac(
-      options.ransac_options);
+      H_ransac_options);
   const auto H_report =
       H_ransac.Estimate(matched_img_points1, matched_img_points2);
   geometry.H = H_report.model;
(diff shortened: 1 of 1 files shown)


src/colmap/estimators/two_view_geometry.cc
```diff
@@ -224,8 +229,15 @@ TwoViewGeometry EstimateUncalibratedTwoViewGeometry(
 
   // Estimate planar or panoramic model.
 
+  // Budget the search for the ratio the homography must beat to be selected
+  // below, since a weaker one is discarded anyway.
```
sarlinpe:
```suggestion
  // Budget the search for the inlier ratio that the homography must reach
  // to be selected below, since a weaker one is discarded anyway.
```

src/colmap/estimators/two_view_geometry.cc
```diff
@@ -916,8 +928,16 @@ TwoViewGeometry EstimateCalibratedTwoViewGeometry(
 
   // Estimate planar or panoramic model.
 
+  // As above. The competing model depends on the branch taken below, so the
+  // smaller inlier count is the lowest bar the homography might have to clear.
```
sarlinpe:
```suggestion
  // Budget the estimation as above. The competing model depends on the branch 
  // taken below so the homography must reach the smallest count of the two.
```

src/colmap/estimators/two_view_geometry.cc
```diff
@@ -1266,8 +1292,14 @@ TwoViewGeometry EstimateOneSidedFocalTwoViewGeometry(
   // spherical one it is skipped, as in the spherical path. A default report has
   // no inliers, so the checks below then behave as a failed estimate.
   const bool has_image_plane = !camera2.IsSpherical();
+  // As in EstimateUncalibratedTwoViewGeometry.
```
sarlinpe:
```suggestion
  // Budget the estimation as in EstimateUncalibratedTwoViewGeometry.
```

src/colmap/estimators/two_view_geometry.cc
```diff
@@ -224,8 +229,15 @@ TwoViewGeometry EstimateUncalibratedTwoViewGeometry(
 
   // Estimate planar or panoramic model.
 
+  // Budget the search for the ratio the homography must beat to be selected
+  // below, since a weaker one is discarded anyway.
+  auto H_ransac_options = options.ransac_options;
+  H_ransac_options.min_inlier_ratio =
+      std::max(options.ransac_options.min_inlier_ratio,
+               options.max_H_inlier_ratio * F_report.support.num_inliers /
+                   matches.size());
```
sarlinpe, line 238: Could you define a helper function for this, since it's duplicated 4 times in this file?

ahojnnes:
Does the same logic apply to the EF inlier ratio logic?

src/colmap/estimators/two_view_geometry.cc
B1ueber2y: Thanks. However, the logic in each of the four times is different in terms of the support it compares with, and it s only a small block. I think a helper function would be unnecessary here. I would keep it as it is.

B1ueber2y:
> Does the same logic apply to the EF inlier ratio logic?

Thanks. It can make sense from the first look by switching the order of E and F, and make E budgeted based on F. However, it would weaken the H budget and lead to behavior change, as we are currently doing max_H_inlier_ratio * min(E_inliers, F_inliers). So I suggest to keep the change to be only on H.

ahojnnes (approved):

PR #4628, opened 2026-08-09
Fix loading of database cache images in image registrator

### LINKED ISSUE #4602: Add --Mapper.load_all_images option for image_registrator
### Motivation

Currently, `image_registrator` uses `DatabaseCache::Options::load_all_images`, which defaults to `false`. Therefore, during `DatabaseCache::Load()`, only images that have valid two-view geometries and satisfy `min_num_matches` are loaded into `DatabaseCache::images_`.

This behavior works well for the standard incremental SfM pipeline where the reconstruction is generated by `mapper`. However, it causes problems for reconstructions generated by other pipelines.

For example, a sparse model can be created using:

```
Images
  ↓
Feature extraction
  ↓
Feature matching
  ↓
Triangulation using known/prior camera poses
  ↓
Bundle Adjustment
  ↓
Sparse model
```

In this case, images may already exist in the reconstruction with valid poses and 3D observations, but they may not have corresponding `two_view_geometry` entries or may have fewer inlier matches than `Mapper.min_num_matches`.

When running `image_registrator` with such a model:

1. The reconstruction contains registered images.
2. `DatabaseCache::Load()` skips some of these images because `load_all_images=false`.
3. `CorrespondenceGraph` is built only from the loaded images.
4. Later, `ObservationManager` calls:

```cpp
CorrespondenceGraph::FindCorrespondences(image_id, point2D_idx)
```

The first line:

```cpp
const Image& image = images_.at(image_id);
```

can fail because the `image_id` exists in the reconstruction but does not exist in the correspondence graph.

---

### Proposed Solution

Expose `DatabaseCache::Options::load_all_images` as a user-controllable option for `image_registrator`.

Add a new command-line option:

```
--Mapper.load_all_images
```

When enabled:

* All images are loaded into `DatabaseCache`.
* Images are not filtered only based on `two_view_geometry` and `min_num_matches`.
* This allows `image_registrator` to work with externally generated models (e.g., prior-pose triangulation + BA).

---

#### 1. `option_manager.cc`

In:

```cpp
OptionManager::AddMapperOptions()
```

Add:

```cpp
AddDefaultOption("Mapper.load_all_images",
                 &mapper->load_all_images);
```

---

#### 2. `image.cc`

In:

```cpp
RunImageRegistrator()
```

Add:

```cpp
database_cache_options.load_all_images =
    options.mapper->load_all_images;
```

---

### Expected Behavior

Default behavior remains unchanged:

```
--Mapper.load_all_images 0
```

preserves the current SfM behavior.

For externally generated models:

```
--Mapper.load_all_images 1
```

allows `image_registrator` to load all images and avoids invalid assumptions about the relationship between reconstruction images and the correspondence graph.

---

### Use Case

This option is useful for workflows where the reconstruction is generated from external information, such as:

* Known camera poses
* Robot/vehicle trajectories
* LiDAR-camera calibration pipelines
* Prior-pose triangulation
* Custom SfM pipelines

These workflows can produce valid COLMAP sparse models without relying on incremental mapper registration.

Adding this option would make `image_registrator` more general and robust for non-standard reconstruction pipelines.

tianrking:
## Summary

- expose the existing `Mapper.load_all_images` setting through mapper CLI/config options
- forward the setting to the database cache used by `image_registrator`
- cover command-line parsing and project-file round trips for the option

This keeps the default behavior unchanged while allowing externally registered images without two-view geometry to be loaded when requested.

Fixes #4602

## Testing

- `scripts/format/c++.sh` with the required clang-format 22.1.5 (3 changed files)
- configured a CPU-only release build with tests enabled
- built `colmap_controllers_option_manager_test` and `colmap_main` (371/371 build steps)
- ran `option_manager_test` (14/14 passed)
- ran `colmap image_registrator --help` and verified `--Mapper.load_all_images arg (=0)` is exposed

## AI assistance

AI tools assisted with repository analysis and implementation. I reviewed the resulting diff, traced the existing database-cache option flow, and ran the tests and executable checks listed above.

### DIFF
--- a/src/colmap/exe/image.cc
+++ b/src/colmap/exe/image.cc
@@ -273,6 +273,9 @@ int RunImageRegistrator(int argc, char** argv) {
     return EXIT_FAILURE;
   }
 
+  auto reconstruction = std::make_shared<Reconstruction>();
+  reconstruction->Read(input_path);
+
   LOG_HEADING1("Loading database");
 
   std::shared_ptr<DatabaseCache> database_cache;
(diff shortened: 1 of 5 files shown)


ahojnnes:
Thanks for addressing the issue. Could a more principled fix be to always load the registered images in the reconstruction?

src/colmap/exe/image.cc
```diff
@@ -287,6 +287,7 @@ int RunImageRegistrator(int argc, char** argv) {
         options.mapper->ignore_watermarks;
     database_cache_options.image_names = {options.mapper->image_names.begin(),
                                           options.mapper->image_names.end()};
+    database_cache_options.load_all_images = options.mapper->load_all_images;
```
sarlinpe: Rather than exposing load_all_images, should we instead set it to true here? We already do so for the point_triangulator https://github.com/colmap/colmap/blob/5d968cf38aa2574fd284b0dc86c3d58d124940b3/src/colmap/exe/sfm.cc#L658

sarlinpe (commented):
Thank you!

src/colmap/exe/image.cc
ahojnnes: ^^ This or augmenting "options.mapper->image_names" with the already registered images?

cometicray:
Thanks for addressing this issue. My use case is building the reconstruction from known camera poses + triangulation + BA, so registered images without two-view geometry are still valid reconstruction images. I agree that handling this internally in image_registrator would be a cleaner solution.

src/colmap/exe/image.cc
ahojnnes: Done.

ahojnnes:
Took a stab at this, also added a test, and directly pushed to your branch. Please let me know if this fixes the issue for you.

sarlinpe (approved):

cometicray:
Thanks for the fix! This looks like it addresses my use case.
I will test it with my workflow (known camera poses + triangulation + BA) and report back if I run into any issues.

Issue #4630, opened 2026-08-10
Spatial matcher fails when some images have no pose prior


hotplot:
**Describe the bug**

The spatial matcher applies an offset to center pose priors around a local origin for improved numerical precision. When a subset of images do not have pose priors, the offset is calculated incorrectly and can result in a failure to match sets of images.

The relevant code is here: https://github.com/colmap/colmap/blob/85350623e67c5a074dd36808de4aefadd4e805f2/src/colmap/controllers/pairing.cc#L718

I think this is a straight-forward fix and I'm happy to propose a PR, unless you're opposed to this.

**To Reproduce**

1. Create a database with image pose priors in a coordinate system that has large coordinate values (e.g. NZTM/EPSG:2193, where coordinate values are on the order of 1e6)
2. Remove pose priors for 20% of images
3. Run `spatial_matcher`

**Expected behavior**
Successful pose-based matching of features in all images with poses, and exhaustive matching of features in the remaining images without poses.

**Actual behaviour**
Partial matching of features for some images, but complete failure to match features for most images in dataset.

**Environment:**
 - OS: Ubuntu 24.04
 - COLMAP Version 4.2.0

PushpakAg:
I'd take a closer look at how the offset is being calculated when some images lack pose priors. It seems like that's where things might be going wrong.

hotplot:
This has been fixed by #4631

PR #4631, opened 2026-08-11
Fix calculation of local origin during spatial matching

hotplot:
Bugfix for #4630 

**Summary**

`SpatialPairGenerator::ReadPositionPriorData` calculates the mean coordinate of the image position priors and subtracts this from the positions matrix to improve numerical precision. In the case where some images do not have a pose prior, their corresponding zero-valued rows were still included in calculation of the mean position, and this could significantly bias the calculated mean position. 

For pose priors using projected Cartesian coordinates in particular, where coordinate values might be on the order of 1e6, this could result in large numerical errors and cause subsequent spatial matching to fail.

**Changes**

- `SpatialPairGenerator::ReadPositionPriorData` has been updated to trim unpopulated position rows prior to calculation of mean coordinate
- Debug logging has been added that prints the calculated offset at log level 1
- A test `CentersLargeCoordinatesWithMissingPosePrior` has been added to test for regressions

Hopefully this PR is welcome and follows general expectations; let me know if there are issues.

### DIFF
--- a/src/colmap/controllers/pairing.cc
+++ b/src/colmap/controllers/pairing.cc
@@ -747,11 +747,25 @@ Eigen::RowMajorMatrixXf SpatialPairGenerator::ReadPositionPriorData(
     }
   }
 
+  // Trim unused rows for images without a valid position before calculating
+  // the mean coordinate below.
+  const size_t num_populated_rows = position_idxs_.size();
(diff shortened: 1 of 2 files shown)


ahojnnes (approved):
Thanks for the fix, looks great.

PR #4640, opened 2026-08-18
Update conditional compilation for CUDA and HIP for dense reconstruct

baranelitez:
https://github.com/colmap/colmap/pull/4420 adds ROCm, HIP support to colmap. But on an attempt of starting the stereo step, the dense reconstruction gui gives an error by only checking against the presence of CUDA, even if the user has built it with HIP.

This commit fixes that. If user has HIP enabled, but not CUDA, the process can start.


ahojnnes:
Thank you, LGTM.

ahojnnes (approved):

ahojnnes:
Needed to fix a bunch of unrelated issues to appease CI failures.

PR #4641, opened 2026-08-18
Add changelog for 4.2.0 release

ahojnnes:
## Summary

- add draft release notes for COLMAP 4.2.0
- highlight multi-component global mapping, LoMa, ROCm/HIP PatchMatch, the browser viewer, and mapper API expansion
- document performance improvements, bug fixes, and migration-relevant breaking changes

## Testing

- `git diff --check`
- `PYENV_VERSION=colmap make -C doc html SPHINXOPTS="-W --keep-going"`

PR #4643, opened 2026-08-24
Restore pycolmap relative pose ray overloads

ahojnnes:
Follow-up to review feedback in #4641.

## Summary
- restore the camera-ray overloads of pycolmap.estimate_relative_pose and pycolmap.refine_relative_pose
- attach identity Jacobians to preserve the legacy Sampson-error normalization
- add regression coverage for both compatibility overloads

## Testing
- PYENV_VERSION=colmap scripts/format/c++.sh
- PYENV_VERSION=colmap scripts/format/python.sh
- PYENV_VERSION=colmap python -m pytest -q src/pycolmap/estimators/pose_test.py

### DIFF
--- a/src/pycolmap/estimators/pose.cc
+++ b/src/pycolmap/estimators/pose.cc
@@ -115,6 +115,41 @@ py::typing::Optional<py::dict> PyEstimateAndRefineAbsolutePose(
   return result;
 }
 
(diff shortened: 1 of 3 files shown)


B1ueber2y (approved):
Thank you!

Issue #3798, opened 2025-12-04
Accessing items in pycolmap `RigMap` causes crash on Windows


StonerLing:
**Describe the bug**
Accessing or printing a `RigMap` with items on Windows triggers the crash.

**To Reproduce**
Minimal code to reproduce the behavior:

```python
import pycolmap

rig_map = pycolmap.RigMap()
rig_map[1] = pycolmap.Rig()
print(rig_map)
```
I met the crash when running benchmark scripts at:

```python
if sparse is not None:
    largest_component = max(largest_component, sparse.num_images())
    for image in sparse.images.values():
        if image.image_id in sparse_merged.images:
            continue
        if image.camera_id not in sparse_merged.cameras:
            sparse_merged.add_camera(image.camera)
        if image.frame_id not in sparse_merged.frames:
            if image.frame.rig_id not in sparse_merged.rigs:
                sparse_merged.add_rig(image.frame.rig)
```
Crash info:
```
*** Aborted at 1764843252 (unix time) try "date -d @1764843252" if you are using GNU date ***
    @     0x7ffdfbb40ef5 _seh_filter_exe
    @     0x7ff794c81f5c OPENSSL_Applink
    @     0x7ffde790f47f __C_specific_handler
    @     0x7ffdfdc1247f __chkstk
    @     0x7ffdfdbc14f4 RtlRaiseException
    @     0x7ffdfdc10f8e KiUserExceptionDispatcher
    @     0x7ffd67d77ddf public: struct faiss::InterruptCallback & __ptr64 __cdecl faiss::InterruptCallback::operator=(struct faiss::InterruptCallback const & __ptr64) __ptr64
    @     0x7ffd68177993 PyInit__core
    @     0x7ffd68177b14 PyInit__core
    @     0x7ffd67d73cf2 (unknown)
    @     0x7ffd7142142d PyCFunction_GetFlags
    @     0x7ffd713dc2c8 _PyObject_MakeTpCall
    @     0x7ffd713deb44 PyMethod_Self
    @     0x7ffd7143bba7 _PyTuple_Resize
    @     0x7ffd71449d46 PyType_Ready
    @     0x7ffd71424849 PyObject_Str
    @     0x7ffd713fa3c9 PyFile_WriteObject
    @     0x7ffd714e6b3e _PyWarnings_Init
    @     0x7ffd714210cb PyCFunction_GetFlags
    @     0x7ffd714e88e1 _PyOS_URandomNonblock
[...]

PR #4648, opened 2026-08-25
Add RigMap repr and iteration regression coverage

tandede:
## Summary

Add regression coverage for the non-empty `RigMap` access paths reported in #3798:

- representing a directly constructed `RigMap`;
- representing `Reconstruction.rigs.values()`;
- iterating over `Reconstruction.rigs.values()`;
- iterating over `Reconstruction.rigs.items()`.

## Investigation

Issue #3798 reports a Windows crash when accessing or printing a non-empty `pycolmap.RigMap`, including the `Reconstruction.rigs.values()` path.

On the current `main` branch, the original reproducer no longer crashes in a Windows/vcpkg-style Boost container configuration after the recent hash-container backend changes. I therefore did not add a speculative production-code change or claim a root-cause fix that could not be reproduced against the current implementation.

The existing tests only checked `RigMap` length and indexed access. They did not exercise the representation and view-iteration paths from the report, leaving those paths unprotected against regression.

## Implementation

The existing direct `RigMap` test now verifies that representing a non-empty map includes the stored rig.

A second test constructs a `Reconstruction`, adds a camera with a trivial rig, and exercises:

- the values-view representation;
- values iteration;
- items iteration.

The assertions validate the retained rig identifiers instead of depending on container ordering beyond this single-entry fixture.

## Validation

- `python -m pytest -q src/pycolmap/sensor` — 46 passed
- full pycolmap test suite — 814 passed, 33 skipped
- `ruff check src/pycolmap/sensor/rig_test.py`
- `ruff format --check src/pycolmap/sensor/rig_test.py`
- `git diff --check`

The skipped full-suite tests require MVS or CUDA support that is not enabled in the local build.

Related: #3798

### DIFF
--- a/src/pycolmap/sensor/rig_test.py
+++ b/src/pycolmap/sensor/rig_test.py
@@ -145,3 +145,22 @@ def test_rig_map_insert_and_access():
     rig_map[1] = rig
     assert len(rig_map) == 1
     assert rig_map[1].rig_id == 1
+    assert "RigMap{1: Rig(rig_id=1" in repr(rig_map)
+
+
+def test_reconstruction_rig_map_views_and_iteration():
+    reconstruction = pycolmap.Reconstruction()
+    camera = pycolmap.Camera.create_from_model_id(
+        camera_id=1,
+        model=pycolmap.CameraModelId.SIMPLE_PINHOLE,
+        focal_length=1.0,
+        width=1,
+        height=1,
+    )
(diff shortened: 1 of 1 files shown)


ahojnnes:
Thanks, I think the issue was already covered by existing tests but the reason CI didn't catch the issue was that we didn't run pycolmap tests under Windows. This is fixed now here: #4651.

PR #4651, opened 2026-08-25
Consolidate bundle adjustment backend tests

ahojnnes:
## Summary

Consolidates backend-independent bundle adjustment tests into the shared interface test suite.

- Parameterizes shared tests across Ceres and Caspar when Caspar is enabled
- Moves common coverage for camera/point constraints, partial tracks, track-length filtering, intrinsics, and ignored points
- Removes duplicated tests from the Ceres and Caspar implementation suites
- Keeps backend-specific behavior and solver-internal tests in their respective suites
- Fixes coverage where the previous Caspar constant-camera test accidentally constructed a Ceres adjuster

## Verification

- `estimators/bundle_adjustment_test`
- `estimators/bundle_adjustment_ceres_test`
- Caspar-enabled compilation of all three affected test translation units
- Formatting and `git diff --check`


sarlinpe (approved):
Thank you!

PR #4653, opened 2026-08-26
Fix compiler caching in PyCOLMAP CI

ahojnnes:
## Summary

- isolate CUDA and non-CUDA compiler caches
- make ccache reusable across cibuildwheel temporary build directories
- compile the macOS prerequisite build without sudo so it uses the configured cache
- cap cache size and report fresh per-run statistics after wheel compilation

## Motivation

Recent main builds restored compiler caches, but macOS recorded no hits and the Linux CUDA job could not save its updates because it shared a key with the non-CUDA job. The existing statistics also ran before the wheel builds and therefore did not show whether pycolmap compilation itself benefited.

## Validation

- parsed build-pycolmap.yml with PyYAML
- ran bash syntax checks on the modified Linux and macOS scripts
- verified the new ccache environment settings with ccache --show-config
- ran git diff --check


ahojnnes:
Verified compiler-cache behavior with an exact rerun of PyCOLMAP run [32962134797](https://github.com/colmap/colmap/actions/runs/32962134797) (attempt 2), restoring the primary cache keys created by attempt 1:

| Job | COLMAP snapshot | Final snapshot | PyCOLMAP binding delta |
|---|---:|---:|---:|
| macOS arm64 | 366/366 hits | 524/524 hits | 158/158 hits |
| Linux CPU | 371/371 hits | 529/529 hits | 158/158 hits |
| Linux CUDA | 390/390 hits | 548/548 hits | 158/158 hits |
| Windows | 363/363 hits | 521/521 hits | 158/158 hits |

All four jobs had zero compiler-cache misses. The final cache sizes stay within the 1.5 GB bound (macOS 0.4 GB, Linux CPU 1.4 GB, Linux CUDA 1.3 GB, Windows 1.1 GB).

This confirms the caches cover both the prerequisite COLMAP C/C++ build and the compiled pycolmap bindings. Pure Python sources are not compiler-cache inputs.

lpanaf (approved):

PR #4654, opened 2026-08-26
Keep CI compiler caches within the GitHub Actions quota

ahojnnes:
## Summary

- replace per-run compiler cache keys with weekly immutable generations
- restore caches in every build, but save only from trusted main/release runs
- centralize weekly restore/save behavior in a reusable `compiler-cache` composite action
- cap each configuration according to its observed working set
- prune stale PyCOLMAP ccache entries and recompress retained entries at zstd level 6 before the weekly save
- remove cross-run FetchContent caches, which consumed roughly 4 GB of the repository quota by themselves
- preserve stable ccache path/compiler hashing and per-run statistics

## Motivation

The repository had 11.06 GB of active GitHub Actions caches, above the standard 10 GiB limit. Run-specific compiler archives and PR-scoped copies were evicting recently saved main caches. The latest main build restored only 1 of 10 native compiler caches even though available caches produced approximately 99.3–99.9% hits.

Pull requests now restore shared main caches read-only. One trusted build seeds each configuration for the current UTC week; subsequent builds restore the exact weekly key without producing more archives. At the next weekly epoch, the previous generation is used as a fallback and the refreshed generation is saved once.

Before saving a new PyCOLMAP generation, ccache entries not used during the current full wheel build are evicted. The remaining entries are recompressed from ccache's default zstd level 1 to level 6. This work runs only on the first successful trusted build that creates a weekly cache.

## Cache budget

Configured compiler-cache limits are based on the observed working sets:

- native Docker/macOS: 100 MB each
- native Ubuntu: 100–150 MB per configuration
- native Windows: 200–300 MB per configuration
- PyCOLMAP: 350 MB–1 GB per configuration

One complete generation is capped at approximately 4.4 GB. During the seven-day overlap between weekly generations, the projected peak before save-time compaction is approximately 8.8–9.1 GB including cached compiler tools, leaving limited headroom below the 10 GiB repository quota. Existing FetchContent entries will age out because the workflows no longer restore them.

## Validation

- parsed all workflow and composite-action YAML
- verified only the shared composite action invokes `actions/cache/restore` and `actions/cache/save`
- verified pull requests and manual PyCOLMAP runs cannot save compiler caches
- ran an isolated save/restore smoke test: https://github.com/colmap/colmap/actions/runs/33055245598
  - the reusable action evicted the deliberately stale entry
  - the retained entry survived recompression and archive restore as a cache hit
  - the stale entry produced the expected cache miss
- `git diff --check`

CI exercises restore behavior on all hosted runner platforms. The first trusted main run will seed the weekly keys; a subsequent run will demonstrate warm-cache hit rates.


lpanaf (approved):

lpanaf:
Nothing ever deletes blobs. GitHub evicted unused entries after 7 days; Azure keeps them until told otherwise. Worth adding a lifecycle rule on the colmap-ci-cache/ prefix, since otherwise every future COMPILER_CACHE_VERSION bump silently orphans a full per-configuration set. This comment seems to be relevant, for your reference.

ahojnnes:
> Nothing ever deletes blobs. GitHub evicted unused entries after 7 days; Azure keeps them until told otherwise. Worth adding a lifecycle rule on the colmap-ci-cache/ prefix, since otherwise every future COMPILER_CACHE_VERSION bump silently orphans a full per-configuration set. This comment seems to be relevant, for your reference.

Thanks but the azure blob storage container is already configured like that but with a 30d ttl.

PR #4660, opened 2026-08-27
Restrict sequential loop detection by image distance

ahojnnes:
## Summary

- add `SequentialMatching.loop_detection_min_image_distance` to exclude nearby images from loop retrieval
- apply candidate filtering before top-N selection so nearby frames do not consume the loop-detection budget
- expose the option through the CLI, GUI, and PyCOLMAP and document its behavior

The option defaults to zero to preserve existing behavior.

## Verification

- `ctest --output-on-failure -R "controllers/pairing_test|retrieval/visual_index_test"`
- built `colmap_ui` and `colmap_main`
- built the PyCOLMAP `_core` extension and verified the new property


doc/tutorial.rst
```diff
@@ -330,8 +330,12 @@ matching modes, which are intended for different input scenarios:
   other. This matching mode has built-in loop detection based on a vocabulary
   tree, where every N-th image (``--SequentialMatching.loop_detection_period``)
   is matched against its visually most similar images
-  (``--SequentialMatching.loop_detection_num_images``). Note that image
-  file names must be ordered sequentially (e.g., ``image0001.jpg``,
+  (``--SequentialMatching.loop_detection_num_images``). Retrieved images can
+  be restricted to those that are sufficiently far from the query in the
+  sequence (``--SequentialMatching.loop_detection_min_image_distance``), so
```
B1ueber2y: I think image_distance is a bit confusing. It can be some hamming, ssim or psnr etc. How about index distance?

src/colmap/controllers/pairing.cc
```diff
@@ -518,6 +538,17 @@ bool SequentialPairGenerator::IsValidSequentialNeighbor(
          cache_->GetImage(image_id2).CameraId();
 }
 
+bool SequentialPairGenerator::IsValidLoopDetectionPair(
+    const image_t image_id1, const image_t image_id2) const {
+  const size_t image_idx1 = image_id_to_idx_.at(image_id1);
+  const size_t image_idx2 = image_id_to_idx_.at(image_id2);
+  const size_t image_distance = image_idx1 > image_idx2
```
B1ueber2y: Similarly, this should be image_idx_distance?

B1ueber2y (approved):

PR #4662, opened 2026-08-28
Add graceful shutdown support for processing pipelines

ahojnnes:
## Summary

- Add process-wide cooperative handling for SIGINT and SIGTERM, with a second signal forcing immediate termination.
- Propagate cancellation through feature extraction/matching, incremental mapping, bundle adjustment, point triangulation, image registration, image undistortion/rectification, patch-match stereo, and stereo fusion.
- Support graceful shutdown in `automatic_reconstructor` when using its incremental mapper; global and hierarchical mapper modes remain excluded.
- Preserve and write usable in-progress reconstruction or fusion results before returning.
- Expose a reusable `pycolmap.CancellationToken` across supported Python pipeline entry points and raise `InterruptedError` only after cleanup.
- Keep guided geometric-verification database updates consistent across cancellation.
- Document CLI and Python shutdown behavior.

Global mapping is intentionally not included because its intermediate state cannot currently be resumed safely.

## Testing

- `PYENV_VERSION=colmap scripts/format/python.sh`
- Targeted C++ build for the CLI and affected test targets.
- Seven affected C++ test suites pass:
  - `controllers/bundle_adjustment_test`
  - `controllers/undistorters_test`
  - `estimators/global_positioning_test`
  - `exe/sfm_test`
  - `mvs/fusion_test`
  - `sfm/global_mapper_test`
  - `util/cancellation_test`
- pycolmap shared-module build and stub generation.
- `src/pycolmap/util/cancellation_test.py`: 7 passed.

The full `install` build remains blocked by the pre-existing `util/cache_test` linker error for `testing::internal::GetWithoutMatchers()`; the successfully built libraries were installed directly for pycolmap verification.

The implementation and follow-up fixes were independently reviewed with Claude Code.

### DIFF
--- a/doc/cli.rst
+++ b/doc/cli.rst
@@ -122,12 +122,14 @@ and can be continued with ``--input_path``::
         --input_path $DATASET_PATH/sparse/0 \
         --output_path $DATASET_PATH/sparse/0
 
-Point triangulation, image registration, bundle adjustment, and stereo fusion
-write their usable partial results before exiting. Image undistortion can be
-resumed by rerunning the same command. Ceres optimizations stop between
-iterations. The Caspar bundle-adjustment backend can only stop after its current
-solver invocation. The global mapper does not support graceful shutdown because
-its intermediate reconstructions cannot currently be resumed.
+Point triangulation, image registration, and bundle adjustment write their
+usable partial results before exiting. Stereo fusion writes non-empty partial
+results to a sibling path containing ``.partial`` so that an existing completed
+result is not overwritten. Interrupted image undistortion can be restarted by
+rerunning the same command. Ceres optimizations stop between iterations. The
+Caspar bundle-adjustment backend can only stop after its current solver
+invocation. The global mapper does not support graceful shutdown because its
+intermediate reconstructions cannot currently be resumed.
 
 The equivalent pycolmap functions accept an optional
 ``cancellation_token``. Cancellation finishes cleanup and then raises
--- a/doc/cli.rst
+++ b/doc/cli.rst
@@ -97,6 +97,50 @@ of commands as an alternative to the automatic reconstruction command::
         --input_path $DATASET_PATH/dense/meshed-poisson.ply \
         --output_path $DATASET_PATH/dense/textured
 
+Graceful shutdown and resuming
+------------------------------
+
+The feature extraction and matching commands, ``mapper``,
+``pose_prior_mapper``, ``global_mapper``, and ``patch_match_stereo`` handle
+``SIGINT`` and ``SIGTERM`` cooperatively. The first signal stops work at a safe
+point and writes any usable in-progress results. The process then exits with
+status 130 for ``SIGINT`` or 143 for ``SIGTERM``. A second signal terminates
+immediately and may interrupt an output write.
+
+Feature extraction and matching can be resumed by rerunning the same command
+against the same database. PatchMatch can likewise be rerun against the same
+workspace; complete depth and normal maps are skipped. An interrupted
+incremental reconstruction is written to its normal numbered model directory
+and can be continued with ``--input_path``::
+
+    $ colmap mapper \
+        --database_path $DATASET_PATH/database.db \
+        --image_path $DATASET_PATH/images \
+        --input_path $DATASET_PATH/sparse/0 \
+        --output_path $DATASET_PATH/sparse/0
+
+The global mapper also writes a coherent partial reconstruction when possible,
+but automatic continuation of the global pipeline from that model is not
+currently supported. Ceres optimizations stop between iterations. The Caspar
+bundle-adjustment backend can only stop after its current solver invocation.
+
+The equivalent pycolmap functions accept an optional
+``cancellation_token``. Cancellation finishes cleanup and then raises
+``InterruptedError``. Applications can connect cloud-preemption signals to the
+token without pycolmap replacing the application's signal handlers::
+
+    import signal
+    import pycolmap
+
+    token = pycolmap.CancellationToken()
+    signal.signal(signal.SIGTERM, lambda *_: token.cancel())
+    pycolmap.incremental_mapping(
+        database_path,
+        image_path,
+        output_path,
+        cancellation_token=token,
+    )
+
 To use the global SfM pipeline instead of the incremental mapper, replace the
 ``mapper`` step with ``global_mapper``. The global mapper depends on good focal
 length priors, so if reliable intrinsics are not available (e.g., from EXIF or
--- a/doc/cli.rst
+++ b/doc/cli.rst
@@ -131,10 +131,12 @@ Caspar bundle-adjustment backend can only stop after its current solver
 invocation. The global mapper does not support graceful shutdown because its
 intermediate reconstructions cannot currently be resumed.
 
-The equivalent pycolmap functions accept an optional
-``cancellation_token``. Cancellation finishes cleanup and then raises
-``InterruptedError``. Applications can connect cloud-preemption signals to the
-token without pycolmap replacing the application's signal handlers::
+The equivalent pycolmap functions automatically handle ``SIGINT`` (Ctrl-C) by
+performing a graceful shutdown and raising ``KeyboardInterrupt``. For other
+termination events, they accept an optional ``cancellation_token``. Explicit
+cancellation finishes cleanup and then raises ``InterruptedError``. This allows
+applications to connect cloud-preemption signals (like ``SIGTERM``) to the token
+without pycolmap replacing the host application's signal handlers::
 
     import signal
     import pycolmap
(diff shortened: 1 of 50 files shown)


doc/cli.rst
```diff
@@ -97,6 +97,50 @@ of commands as an alternative to the automatic reconstruction command::
...
+
+The global mapper also writes a coherent partial reconstruction when possible,
+but automatic continuation of the global pipeline from that model is not
+currently supported. Ceres optimizations stop between iterations. The Caspar
+bundle-adjustment backend can only stop after its current solver invocation.
+
+The equivalent pycolmap functions accept an optional
+``cancellation_token``. Cancellation finishes cleanup and then raises
+``InterruptedError``. Applications can connect cloud-preemption signals to the
+token without pycolmap replacing the application's signal handlers::
```
sarlinpe:
```suggestion
The equivalent pycolmap functions automatically handle ``SIGINT`` (Ctrl-C) 
by performing a graceful shutdown and raising ``KeyboardInterrupt``. 
For other termination events, they accept an optional ``cancellation_token``. 
Explicit cancellation finishes cleanup and then raises ``InterruptedError``. 
This allows applications to connect cloud-preemption signals (like ``SIGTERM``) 
to the token without pycolmap replacing the host application's signal handlers::
```

src/pycolmap/pipeline/mvs.cc
```diff
@@ -15,13 +16,79 @@
...
 namespace py = pybind11;
 
+void PatchMatchStereo(
+    const std::filesystem::path& workspace_path,
+    std::string workspace_format,
+    const std::string& pmvs_option_name,
+    const mvs::PatchMatchOptions& options,
+    const std::filesystem::path& config_path,
+    const std::shared_ptr<CancellationToken>& cancellation_token) {
+#if defined(COLMAP_CUDA_ENABLED)
```
sarlinpe:
```suggestion
#if defined(COLMAP_CUDA_ENABLED) || defined(COLMAP_HIP_ENABLED)
```

src/pycolmap/pipeline/mvs.cc
```diff
@@ -15,13 +16,79 @@
...
+
+  auto controller = std::make_shared<mvs::PatchMatchController>(
+      options, workspace_path, workspace_format, pmvs_option_name, config_path);
+  ControllerThread<mvs::PatchMatchController> thread(std::move(controller));
+  thread.Start();
+  PyWait(&thread, cancellation_token);
+#else
+  static_cast<void>(cancellation_token);
+  RunPatchMatchStereoImpl(
+      workspace_path, workspace_format, pmvs_option_name, options, config_path);
```
sarlinpe:
```suggestion
  LOG(FATAL_THROW) << "Dense stereo reconstruction requires CUDA or HIP, "
                      "neither of which is available on your system.";
```

src/pycolmap/pipeline/sfm.cc
```diff
@@ -44,19 +44,34 @@ std::shared_ptr<Reconstruction> TriangulatePoints(
...
   CreateDirIfNotExists(output_path);
 
   py::gil_scoped_release release;
+  PyInterrupt py_interrupt(1.0);
+  bool python_interrupt_raised = false;
+  const auto check_if_stopped = [&]() {
+    python_interrupt_raised = py_interrupt.Raised();
+    return python_interrupt_raised ||
+           (cancellation_token && cancellation_token->IsCancelled());
+  };
```
sarlinpe: To avoid the boilerplate code duplicated across all functions in `pipeline/`, consider adding the following helper:
```c++
#include <atomic>
#include <thread>

// ...

class PyInterruptChecker {
 public:
  explicit PyInterruptChecker(
      std::shared_ptr<CancellationToken> cancellation_token = nullptr,
      double gap = 1.0)
      : py_interrupt_(gap),
        cancellation_token_(std::move(cancellation_token)),
        calling_thread_id_(std::this_thread::get_id()) {}

  // Returns a callback that safely checks for interrupts, even if called 
  // from C++ worker threads.
  std::function<bool()> Callback() {
    return [this]() {
      if (std::this_thread::get_id() == calling_thread_id_) {
        if (py_interrupt_.Raised()) {
          python_interrupt_raised_.store(true, std::memory_order_relaxed);
        }
      }
      return python_interrupt_raised_.load(std::memory_order_relaxed) ||
             (cancellation_token_ && cancellation_token_->IsCancelled());
    };
  }

  // Checks the flags and throws the appropriate Python exception.
  void CheckAndThrow() {
    if (python_interrupt_raised_.load(std::memory_order_relaxed)) {
      ThrowPythonError();
    }
    if (cancellation_token_ && cancellation_token_->IsCancelled()) {
      ThrowCancelled();
    }
  }

 private:
  PyInterrupt py_interrupt_;
  std::shared_ptr<CancellationToken> cancellation_token_;
  std::thread::id calling_thread_id_;
  std::atomic<bool> python_interrupt_raised_{false};
};
```
Then we'd have:
```c++
  PyInterruptChecker checker(cancellation_token);
  const bool success = RunGlobalMapperImpl(database_path,
                                           image_path,
                                           output_path,
                                           options_,
                                           reconstruction_manager,
                                           checker.Callback());
  checker.CheckAndThrow();
```

sarlinpe (approved):

src/colmap/exe/colmap.cc
```diff
@@ -189,7 +222,18 @@ int main(int argc, char** argv) {
...
+      std::unique_ptr<colmap::ScopedSignalHandler> signal_handler;
+      if (SupportsGracefulShutdown(command)) {
+        signal_handler = std::make_unique<colmap::ScopedSignalHandler>();
+      }
+
+      const int exit_code = matched_command_func(command_argc, command_argv);
+      if (signal_handler && signal_handler->ReceivedSignal() != 0) {
+        LOG(INFO) << "Graceful shutdown completed after receiving signal "
+                  << signal_handler->ReceivedSignal();
+        return 128 + signal_handler->ReceivedSignal();
```
sarlinpe: Consider adding method GetExitCode to ScopedSignalHandler to avoid repeating this magic number 128.

doc/cli.rst
ahojnnes: Applied in fd8c2afb6. The documentation now distinguishes automatic SIGINT/KeyboardInterrupt handling from explicit cancellation tokens used for signals such as SIGTERM.

src/pycolmap/pipeline/mvs.cc
ahojnnes: Applied in fd8c2afb6. The pycolmap PatchMatch path is now enabled for either CUDA or HIP builds.

src/pycolmap/pipeline/sfm.cc
ahojnnes: Applied in fd8c2afb6. Added PyInterruptChecker and refactored triangulation, incremental mapping, bundle adjustment, undistortion, guided verification, and stereo fusion to use it.

src/pycolmap/pipeline/mvs.cc
ahojnnes: Applied in fd8c2afb6. Backend-less builds now raise the explicit CUDA-or-HIP error directly.

src/colmap/exe/colmap.cc
ahojnnes: Applied in fd8c2afb6. ScopedSignalHandler::GetExitCode() now centralizes signal-to-exit-code conversion and is covered by the cancellation test.

fallenmi (dismissed):
`triangulate_points()` can destroy the caller's existing model before honoring an already-cancelled token. With the default `clear_points=True`, `RunPointTriangulatorImpl()` calls `DeleteAllPoints2DAndPoints3D()` and transcribes image IDs before the supplied stop callback is installed; the first check happens later inside `TriangulateReconstruction()`. The Python wrapper only raises `InterruptedError` after that implementation returns.

I reproduced this with the current merge build `6f0d40d` (exact head `fd8c2afb`): a synthetic reconstruction containing 50 points, passed with a token cancelled before invocation, raised `InterruptedError` but left the caller's in-memory reconstruction with 0 points and wrote a 0-point model to the output path. The relevant current-merge files are identical to the exact head.

Please short-circuit a pre-existing cancellation before the clear/transcribe mutation and before writing a cleared model, and add a regression test showing that an already-cancelled default call preserves the caller's reconstruction. The exact-head CI matrix is otherwise green, and `git diff --check` is clean.

Reviewed with OpenAI Codex assistance; I independently traced the exact base/head/current-merge control flow and reproduced the behavior with the repository's exact current-merge macOS wheel.

ahojnnes:
Thanks, I don't think this is a real issue, if I understand correctly. Cancellation is a best effort. If you immediately cancel before/after calling the function, then I don't think there should be an expectation that nothing is modified in the input data.

PR #4664, opened 2026-08-30
Add scaled generalized absolute pose estimation (GP4PS)

nushakrishnan:
## Summary
Adding joint estimation of the pose of a generalized camera and the scale of its internal rig geometry from 2D-3D correspondences. Used for localizing a VIO trajectory against a metric reference map. The trajectory is treated as one generalized camera whose cams_from_rig are rigid but of unknown global scale, and the estimated model is a Sim3d rig_from_world.

- Robust estimation of pose and scale with LO-RANSAC, built on PoseLib's 4-point GP4PS minimal solver. Because no non-minimal solver exists for this problem, local optimization refines pose and scale over the inlier set inside the loop.
- Degenerate configurations are handled explicitly - panoramic rigs (all cameras sharing a projection center), where scale is unobservable.

## Breaking changes
`refine_generalized_absolute_pose` and `estimate_and_refine_generalized_absolute_pose` now return the refined cameras under the `"cameras"` key of the result dict. Previously the refined cameras were silently discarded (pybind11 copies the input list into a temporary, so in-place refinement never reached the caller); read them from the return value instead.

## Testing
Please download the data and script [here](https://github.com/user-attachments/files/31617876/gp4ps_example.zip).

Output:
```
59097 matches over 256 keyframes
num_inliers (unique 3D points): 2000
scale (query -> reference): 1.043718
```


ahojnnes:
Review findings:

1. **Invalid projections become perfect fits during refinement.** In both `TinyScaledRigReprojCostFunctor` (`generalized_absolute_pose.cc`, lines 87–96) and `ScaledRigReprojErrorCostFunctor` (`reprojection_error.h`, lines 447–457), points behind the camera or otherwise not projectable receive a zero residual. This lets refinement accept an invalid pose as a perfect solution. I reproduced both `GP4PSEstimator::Refine` and `RefineScaledGeneralizedAbsolutePose` returning `true` unchanged when every point was behind the cameras. Invalid trial projections should reject the evaluation or receive a penalty rather than zero error.

2. **The public refinement API accepts scale-unobservable inputs.** `RefineScaledGeneralizedAbsolutePose` does not apply the panoramic-rig check to the observations selected by `inlier_mask`. With coincident projection centers—or an inlier mask that selects observations from only one center—the scale is unobservable, but the function reports success and returns the arbitrary initial scale. I reproduced a perfect central-rig fit returning `true` while preserving an initial scale of 2. The active inlier camera centers should be checked before solving.

3. **Empty inputs are dereferenced before rejection.** `EstimateScaledGeneralizedAbsolutePose` calls `ThrowCheckCameras` before its empty-input check; that helper dereferences the result of `std::minmax_element` on an empty `camera_idxs`. `RefineScaledGeneralizedAbsolutePose` similarly dereferences `std::min_element`/`std::max_element` without checking for an empty input. Python callers can therefore hit undefined behavior instead of receiving `None`/failure.

The two targeted C++ test binaries build and pass. `git diff --check` also reports trailing whitespace in `src/colmap/estimators/solvers/generalized_absolute_pose.h:134`.

nushakrishnan:
hello @ahojnnes, your comments have been addressed. I also added the pybind functions to the python test. please let me know what you think!

ahojnnes:
Thanks for addressing the earlier comments. I reviewed the updated implementation and built commit `8131b0919`; both targeted C++ test binaries pass and `git diff --check` is clean. I still have the following concerns before approval:

1. The behind-camera fix is too coarse. Both refinement cost functors return `false` when any single observation cannot project. This aborts an initial solve, or rejects an entire trial step, rather than treating only that observation as invalid; it also bypasses the Cauchy loss and makes refinement brittle to one stale inlier. GP3P and GP4PS RANSAC scoring are already consistent because both assign `max()` to the individual invalid residual. The inconsistency is specifically in nonlinear refinement. I would prefer a finite cheirality-aware penalty/angular residual, or explicit initial validation plus clearly documented hard-domain behavior.

2. Scale observability remains incomplete. `RefineScaledGeneralizedAbsolutePose` now checks the centers selected by `inlier_mask`, but `GP4PSEstimator::Refine` does not, and `EstimateScaledGeneralizedAbsolutePose` checks all input observations rather than the final RANSAC consensus. A dataset containing multiple centers can therefore still report an arbitrary scale when its final inliers originate from only one center. Please check the LO-refinement inputs and final inlier set as well.

3. The full Ceres refinement does not preserve positive scale. The tiny refiner uses log-scale, while `RefineScaledGeneralizedAbsolutePose` optimizes the raw scale with an unconstrained Euclidean manifold. It can cross zero and only gets rejected afterward, potentially leaving the caller's transform modified. Please use a positive parameterization/bound and preferably commit the result only after a successful solve.

4. The Python tests only verify that the new functions are callable. They do not exercise estimation/refinement behavior, empty inputs, degeneracy, invalid projections, covariance, or returned values. Also, refined camera intrinsics are not returned; automatic STL conversion gives the binding a temporary `std::vector<Camera>`, so camera updates are not observable by Python callers.

The empty-input UB and whitespace findings are fixed, and the public refinement panoramic/empty-inlier check is fixed at its immediate call site. The remaining points above make the overall fix incomplete.

nushakrishnan:
Hey @ahojnnes, I have addressed your comments and made the following changes:

1. Refinement step now checks at the initial estimate which inliers project in front of their camera and drops the rest. The cost is documented as undefined behind the camera.
2. Scale observability is checked in three places. The minimal solver rejects samples from a single center, the local refinement rejects inputs from a single center, and the final RANSAC consensus set is rejected if all its inliers come from one center. The Ceres refinement optimises the scale in log space so it stays positive.
3. The Python bindings return the refined cameras. 
4. The pycolmap tests cover estimation, refinement, covariance, empty inputs, degenerate rigs, stale inliers, points behind cameras and invalid scale.

Let me know what you think, thanks!

ahojnnes:
@nushakrishnan I have a few suggested changes: https://github.com/nushakrishnan/colmap/pull/1. Feel free to merge, if it makes sense to you.

nushakrishnan:
@ahojnnes Hello! It looked good to me, thank you.

ahojnnes (approved):

PR #4665, opened 2026-08-31
Enforce Python function type annotations

ahojnnes:
## Summary

- require complete function annotations and check all typed function bodies with mypy
- annotate every tracked Python function, including tests, benchmarks, docs, examples, and the Caspar generator
- include documentation and the Caspar generator in CI type checking
- exempt generated pybind11 stubs from source annotation enforcement and narrowly suppress generated-binding diagnostics

## Testing

- `PYENV_VERSION=colmap scripts/format/python.sh`
- exact CI mypy commands against an installed wheel
- `PYENV_VERSION=colmap pytest` (955 passed)
- `PYENV_VERSION=colmap make -C doc html SPHINXOPTS="-W --keep-going"`
- AST audit confirming no executable behavior changes beyond annotations

### DIFF
--- a/benchmark/reconstruction/evaluation/blended_mvs.py
+++ b/benchmark/reconstruction/evaluation/blended_mvs.py
@@ -151,7 +151,9 @@ class DatasetBlendedMVS(Dataset):
             frame = pycolmap.Frame(frame_id=i)
             frame.rig_id = i
             frame.add_data_id(image.data_id)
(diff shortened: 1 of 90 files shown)


sarlinpe (approved):
Thank you! At some point we'll need to find a way to fix the numpy shape annotations in the stubs, the inlined ignores are pretty annoying...

PR #4669, opened 2026-09-02
Centralize Ceres loss and stock utility seams

Zador-Pataki:
## Motivation

Bundle adjustment currently owns COLMAP's standard Ceres loss construction. Moving it to a shared utility lets other estimators reuse the same validation and adds an optional positive loss weight through `ceres::ScaledLoss`.

Separately, triangulation-angle filtering currently combines classification with deletion. Exposing the classification step allows callers to inspect or schedule filtering without mutating the reconstruction.

## Changes

- Add the shared `CeresLossFunctionType`, `IsValidCeresLossFunction()`, and `CreateCeresLossFunction()` APIs, including optional positive loss weighting.
- Migrate the existing bundle-adjustment loss call sites to the shared API.
- Add `ObservationManager::FindPoints3DWithSmallTriangulationAngle()`.
- Implement the existing mutating triangulation-angle filter through that query.
- Expose the query and loss weight through PyCOLMAP.

## API note

The C++ loss enum moves from `CeresBundleAdjustmentOptions::LossFunctionType` to `colmap::CeresLossFunctionType`, and loss construction moves from `CeresBundleAdjustmentOptions::CreateLossFunction()` to the shared free function.

PyCOLMAP retains the existing `pycolmap.LossFunctionType` name, adds `loss_function_weight`, and removes `CeresBundleAdjustmentOptions.create_loss_function()`.

## Validation

- `bundle_adjustment_ceres_test`: 21/21 passed
- `observation_manager_test`: 14/14 passed


src/colmap/sfm/observation_manager.cc
```diff
@@ -432,18 +432,18 @@ size_t ObservationManager::FilterObservationsWithNegativeDepth() {
...
 
-size_t ObservationManager::FilterPoints3DWithSmallTriangulationAngle(
-    const double min_tri_angle, const FlatHashSet<point3D_t>& point3D_ids) {
-  // Number of filtered observations.
-  size_t num_filtered_observations = 0;
-
+std::vector<point3D_t>
+ObservationManager::FindPoints3DWithSmallTriangulationAngle(
+    const double min_tri_angle,
+    const std::vector<point3D_t>& point3D_ids) const {
```
ahojnnes: If this function takes a flag hash set, then the FilterPoints3DWithSmallTriangulationAngle doesn't need to copy from the input hash set into a vector?

src/colmap/sfm/observation_manager.cc
```diff
@@ -432,18 +432,18 @@ size_t ObservationManager::FilterObservationsWithNegativeDepth() {
   return num_filtered;
 }
 
-size_t ObservationManager::FilterPoints3DWithSmallTriangulationAngle(
-    const double min_tri_angle, const FlatHashSet<point3D_t>& point3D_ids) {
-  // Number of filtered observations.
-  size_t num_filtered_observations = 0;
-
+std::vector<point3D_t>
+ObservationManager::FindPoints3DWithSmallTriangulationAngle(
```
ahojnnes, line 436: Suggest adding a minimal unit test to cover this new function.

src/pycolmap/estimators/bundle_adjustment.cc
```diff
@@ -181,17 +181,16 @@ void BindBundleAdjuster(py::module& m) {
...
+          .value("TRIVIAL", CeresLossFunctionType::TRIVIAL)
+          .value("SOFT_L1", CeresLossFunctionType::SOFT_L1)
+          .value("CAUCHY", CeresLossFunctionType::CAUCHY)
+          .value("HUBER", CeresLossFunctionType::HUBER);
   AddStringToEnumConstructor(PyCeresLossFunctionType);
 
   auto PyCeresBundleAdjustmentOptions =
       py::classh<CeresBAOpts>(m, "CeresBundleAdjustmentOptions")
           .def(py::init<>())
-          .def("create_loss_function", &CeresBAOpts::CreateLossFunction)
```
ahojnnes, line 194: We should expose the new free standing CreateCeresLossFunction function in the pycolmap bindings?

ahojnnes (approved):
Thanks, a few follow-up comments and, on a more high-level note, the PR entangles a few unrelated changes into one PR. It would be easier to review if the point filtering changes were separate from Ceres changes. Not urgent to change now for this PR but a note for the future. Cheers!

src/colmap/estimators/ceres_loss.h
```diff
@@ -0,0 +1,25 @@
...
+
+#include <memory>
+
+#include <ceres/loss_function.h>
+
+namespace colmap {
+
+enum class CeresLossFunctionType { TRIVIAL, SOFT_L1, CAUCHY, HUBER };
+
+// Standard construction accepts non-negative scale (including positive
```
B1ueber2y: Since this becomes a stand alone function, let's document again here what the scale and weight means. Alternatively, as we are already here, how about robust_scale?

src/colmap/estimators/ceres_loss.h
```diff
@@ -0,0 +1,25 @@
...
+// infinity for compatibility) and finite positive weight.
```
B1ueber2y: I suggest not to say infinity explicitly in the comment, as this does not seem to be a common way to use the robust loss in ceres.

B1ueber2y (approved):
Thank you!

src/colmap/sfm/observation_manager.cc
```diff
@@ -432,18 +432,18 @@ size_t ObservationManager::FilterObservationsWithNegativeDepth() {
...
+    const double min_tri_angle,
+    const FlatHashSet<point3D_t>& point3D_ids) const {
   // Minimum triangulation angle in radians.
   const double min_tri_angle_rad = DegToRad(min_tri_angle);
 
   // Cache for image projection centers.
   FlatHashMap<image_t, Eigen::Vector3d> proj_centers;
+  std::vector<point3D_t> small_angle_point3D_ids;
```
sarlinpe, line 444: To avoid making FilterPoints3DWithSmallTriangulationAngle slower (non-preallocated vector small_angle_point3D_ids), would it make sense to abstract the loop body into a new helper HasPoint3DSufficientTriangulationAngle?
```c++
private:
  // Helper to check if a single 3D point has a sufficient triangulation angle.
  bool HasPoint3DSufficientTriangulationAngle(
      point3D_t point3D_id, 
      double min_tri_angle_rad, 
      FlatHashMap<image_t, Eigen::Vector3d>& proj_centers) const;
```

Zador-Pataki: Done, thanks! However, `FindPoints3DWithSmallTriangulationAngle` is no longer used internally by COLMAP. Is that okay? Its Python binding is still very useful.

sarlinpe: Thanks! It's fine to have some extra code if it's tested and used externally.

Zador-Pataki:
In the previous CI run, all checks passed except Clang-Tidy, which flagged the intentionally ignored return value of `loss.release()`. I looked into it and do not see a safety issue.  I added a targeted NOLINT suppression in [fd7fdf27](https://github.com/Zador-Pataki/colmap/commit/fd7fdf27bb257a2ca1d0db9701a3282687aef55d). Thoughts?

src/colmap/estimators/ceres_loss.cc
```diff
@@ -0,0 +1,61 @@
...
+      loss = std::make_unique<ceres::CauchyLoss>(robust_scale);
+      break;
+    case CeresLossFunctionType::HUBER:
+      loss = std::make_unique<ceres::HuberLoss>(robust_scale);
+      break;
+  }
+  if (weight == 1.0) return loss;
+  auto scaled_loss = std::make_unique<ceres::ScaledLoss>(
+      loss.get(), weight, ceres::TAKE_OWNERSHIP);
+  loss.release();  // NOLINT(bugprone-unused-return-value)
```
ahojnnes: This should be the right way, no?
```suggestion
  auto scaled_loss = std::make_unique<ceres::ScaledLoss>(
      loss.release(), weight, ceres::TAKE_OWNERSHIP);
```

src/colmap/estimators/ceres_loss_function.h
```diff
@@ -0,0 +1,26 @@
+// Copyright (c), ETH Zurich and UNC Chapel Hill.
```
ahojnnes, line 1: Suggest adding full license header for consistency. Here and below. We can decide separately, if we want to shorten it consistently across the codebase.

ahojnnes (approved):
Thank you!

src/colmap/estimators/ceres_loss_function.h
```diff
@@ -0,0 +1,26 @@
...
+// All rights reserved.
+
+#pragma once
+
+#include <memory>
+
+#include <ceres/loss_function.h>
```
B1ueber2y, line 34: Sorry for another nit. Should we rename the two files in this PR to ceres_loss_function.h to align with the naming.

Zador-Pataki: no worries. done:)

ahojnnes (approved):
Thank you!

PR #4670, opened 2026-09-02
Expose reusable global positioning

Zador-Pataki:
## Motivation

Adding constraints to COLMAP's global-positioning Ceres problem currently requires hacking its implementation because `RunGlobalPositioning(...)` constructs, solves, and publishes the problem in one call.

Following `CreateDefaultBundleAdjuster(...)`, this PR adds `GlobalPositioner::CreateDefault(...)`, which returns the prepared standard problem so callers can extend it before solving. `RunGlobalPositioning(...)` remains the stock one-shot path with unchanged behavior.

## Changes

- Add `GlobalPositioner::CreateDefault(...)` and a prepared `GlobalPositioner` lifecycle.
- Expose the prepared Ceres problem, temporary frame centers, and parameter-ordering extension.
- Add an optional caller-provided loss.
- Expose the construction and lifecycle workflow through PyCOLMAP.

## Why global positioning needs additional seams

The bundle-adjustment workflow can expose a prepared problem through parameter blocks that are already externally addressable. Global positioning also owns temporary centers and observation scales internally, so its prepared problem needs additional seams.

<details>
<summary>[Click to expand for details]</summary>

- **Temporary frame centers.** The solver does not optimize the pose translations stored in `Reconstruction` directly. It copies poses into temporary world-frame centers and converts the optimized centers back during finalization. `FrameCenterParameterBlock(frame_id)` exposes this temporary parameter to code adding constraints to the prepared Ceres problem. `Finalize()` performs the required publication back to `Reconstruction`.
- **Internal observation scales.** BATA observation scales are stored inside `GlobalPositioner` rather than `Reconstruction`.
- **Late parameter blocks.** Stock parameter ordering is created during construction. `ExtendParameterBlockOrdering()` adds caller-created blocks without rebuilding existing assignments and supports explicit group overrides.

</details>

<details>
<summary>[Click to expand for a possible future refactor]</summary>

A deeper refactor could eliminate the two global-positioning-specific state seams:

- Optimize the translation blocks in reconstruction-owned `Rigid3d` poses instead of copied frame centers.
- Make observation scales explicit input and output. `GlobalPositioner::CreateDefault(...)` could consume a scale map keyed by observation, allowing callers to provide initial values and inspect the optimized results instead of relying on hidden owner storage.

</details>

### DIFF
--- a/src/colmap/estimators/cost_functions/motion_averaging.h
+++ b/src/colmap/estimators/cost_functions/motion_averaging.h
@@ -1,6 +1,8 @@
 
 #pragma once
 
+#include "colmap/estimators/cost_functions/utils.h"
+
 #include <Eigen/Core>
 #include <ceres/ceres.h>
 #include <ceres/rotation.h>
@@ -13,7 +15,8 @@ namespace colmap {
 // 3D point.
 // Reference: Zhuang et al., "Baseline Desensitizing In Translation Averaging",
 // CVPR 2018.
-struct BATAPairwiseDirectionCostFunctor {
+struct BATAPairwiseDirectionCostFunctor
+    : public AutoDiffCostFunctor<BATAPairwiseDirectionCostFunctor, 3, 3, 3, 1> {
   explicit BATAPairwiseDirectionCostFunctor(
       const Eigen::Vector3d& pos2_from_pos1_dir)
       : pos2_from_pos1_dir_(pos2_from_pos1_dir) {}
@@ -30,59 +33,18 @@ struct BATAPairwiseDirectionCostFunctor {
     return true;
   }
 
-  static ceres::CostFunction* Create(
-      const Eigen::Vector3d& pos2_from_pos1_dir) {
-    return (
-        new ceres::
-            AutoDiffCostFunction<BATAPairwiseDirectionCostFunctor, 3, 3, 3, 1>(
-                new BATAPairwiseDirectionCostFunctor(pos2_from_pos1_dir)));
-  }
-
   const Eigen::Vector3d pos2_from_pos1_dir_;
-};
-
-// Covariance-weighted BATA residual with fixed camera rotation
-// (W = Sigma^(-1/2) * R_cam_from_world).
-struct WeightedBATAPairwiseDirectionCostFunctor {
-  WeightedBATAPairwiseDirectionCostFunctor(
-      const Eigen::Vector3d& pos2_from_pos1_dir,
-      const Eigen::Matrix3d& weight_matrix)
-      : pos2_from_pos1_dir_(pos2_from_pos1_dir),
-        weight_matrix_(weight_matrix) {}
-
-  template <typename T>
-  bool operator()(const T* pos1,
-                  const T* pos2,
-                  const T* scale,
-                  T* residuals) const {
-    using Vec3T = Eigen::Matrix<T, 3, 1>;
-    const Vec3T r_world = pos2_from_pos1_dir_.cast<T>() -
-                          scale[0] * (Eigen::Map<const Vec3T>(pos2) -
-                                      Eigen::Map<const Vec3T>(pos1));
-    Eigen::Map<Vec3T> residual_map(residuals);
-    residual_map = weight_matrix_.cast<T>() * r_world;
-    return true;
-  }
-
-  static ceres::CostFunction* Create(const Eigen::Vector3d& pos2_from_pos1_dir,
-                                     const Eigen::Matrix3d& weight_matrix) {
-    return new ceres::AutoDiffCostFunction<
-        WeightedBATAPairwiseDirectionCostFunctor,
-        3,
-        3,
-        3,
-        1>(new WeightedBATAPairwiseDirectionCostFunctor(pos2_from_pos1_dir,
-                                                        weight_matrix));
-  }
-
-  const Eigen::Vector3d pos2_from_pos1_dir_;
-  const Eigen::Matrix3d weight_matrix_;
 };
 
 // Computes the error between a translation direction and the direction formed
 // from a camera (c) and 3D point (p) with constant rig extrinsics, such that:
 // t_ij - scale * (p - c + t_rig) is minimized.
-struct RigBATAPairwiseDirectionConstantRigCostFunctor {
+struct RigBATAPairwiseDirectionConstantRigCostFunctor
+    : public AutoDiffCostFunctor<RigBATAPairwiseDirectionConstantRigCostFunctor,
+                                 3,
+                                 3,
+                                 3,
+                                 1> {
   RigBATAPairwiseDirectionConstantRigCostFunctor(
       const Eigen::Vector3d& cam_from_point3D_dir,
       const Eigen::Vector3d& cam_from_rig_translation)
@@ -103,18 +65,6 @@ struct RigBATAPairwiseDirectionConstantRigCostFunctor {
     return true;
   }
 
-  static ceres::CostFunction* Create(
-      const Eigen::Vector3d& cam_from_point3D_dir,
-      const Eigen::Vector3d& cam_from_rig_translation) {
-    return (new ceres::AutoDiffCostFunction<
-            RigBATAPairwiseDirectionConstantRigCostFunctor,
-            3,
-            3,
-            3,
-            1>(new RigBATAPairwiseDirectionConstantRigCostFunctor(
-        cam_from_point3D_dir, cam_from_rig_translation)));
-  }
-
   const Eigen::Vector3d cam_from_point3D_dir_;
   const Eigen::Vector3d cam_from_rig_translation_;
 };
@@ -122,7 +72,13 @@ struct RigBATAPairwiseDirectionConstantRigCostFunctor {
 // Computes the error between a translation direction and the direction formed
 // from a camera (c) and 3D point (p) with variable rig extrinsics, such that:
 // t_ij - scale * (p - c + t_rig) is minimized.
-struct RigBATAPairwiseDirectionCostFunctor {
+struct RigBATAPairwiseDirectionCostFunctor
+    : public AutoDiffCostFunctor<RigBATAPairwiseDirectionCostFunctor,
+                                 3,
+                                 3,
+                                 3,
+                                 3,
+                                 1> {
   RigBATAPairwiseDirectionCostFunctor(
       const Eigen::Vector3d& cam_from_point3D_dir,
       const Eigen::Quaterniond& rig_from_world_rot)
@@ -146,19 +102,6 @@ struct RigBATAPairwiseDirectionCostFunctor {
                     Eigen::Map<const Eigen::Matrix<T, 3, 1>>(rig_in_world) -
                     cam_from_rig_translation);
     return true;
-  }
-
-  static ceres::CostFunction* Create(
-      const Eigen::Vector3d& cam_from_point3D_dir,
-      const Eigen::Quaterniond& rig_from_world_rot) {
-    return (new ceres::AutoDiffCostFunction<RigBATAPairwiseDirectionCostFunctor,
-                                            3,
-                                            3,
-                                            3,
-                                            3,
-                                            1>(
-        new RigBATAPairwiseDirectionCostFunctor(cam_from_point3D_dir,
-                                                rig_from_world_rot)));
   }
 
   const Eigen::Vector3d cam_from_point3D_dir_;
--- a/src/colmap/estimators/cost_functions/motion_averaging.h
+++ b/src/colmap/estimators/cost_functions/motion_averaging.h
@@ -41,6 +41,44 @@ struct BATAPairwiseDirectionCostFunctor {
   const Eigen::Vector3d pos2_from_pos1_dir_;
 };
 
+// Covariance-weighted BATA residual with fixed camera rotation
+// (W = Sigma^(-1/2) * R_cam_from_world).
+struct WeightedBATAPairwiseDirectionCostFunctor {
+  WeightedBATAPairwiseDirectionCostFunctor(
+      const Eigen::Vector3d& pos2_from_pos1_dir,
+      const Eigen::Matrix3d& weight_matrix)
+      : pos2_from_pos1_dir_(pos2_from_pos1_dir),
+        weight_matrix_(weight_matrix) {}
+
+  template <typename T>
+  bool operator()(const T* pos1,
+                  const T* pos2,
+                  const T* scale,
+                  T* residuals) const {
+    using Vec3T = Eigen::Matrix<T, 3, 1>;
+    const Vec3T r_world = pos2_from_pos1_dir_.cast<T>() -
+                          scale[0] * (Eigen::Map<const Vec3T>(pos2) -
+                                      Eigen::Map<const Vec3T>(pos1));
+    Eigen::Map<Vec3T> residual_map(residuals);
+    residual_map = weight_matrix_.cast<T>() * r_world;
+    return true;
+  }
+
+  static ceres::CostFunction* Create(const Eigen::Vector3d& pos2_from_pos1_dir,
+                                     const Eigen::Matrix3d& weight_matrix) {
+    return new ceres::AutoDiffCostFunction<
+        WeightedBATAPairwiseDirectionCostFunctor,
+        3,
+        3,
+        3,
+        1>(new WeightedBATAPairwiseDirectionCostFunctor(pos2_from_pos1_dir,
+                                                        weight_matrix));
+  }
+
+  const Eigen::Vector3d pos2_from_pos1_dir_;
+  const Eigen::Matrix3d weight_matrix_;
+};
+
 // Computes the error between a translation direction and the direction formed
 // from a camera (c) and 3D point (p) with constant rig extrinsics, such that:
 // t_ij - scale * (p - c + t_rig) is minimized.
--- a/src/colmap/estimators/cost_functions/motion_averaging.h
+++ b/src/colmap/estimators/cost_functions/motion_averaging.h
@@ -1,8 +1,6 @@
 
 #pragma once
 
-#include "colmap/estimators/cost_functions/utils.h"
-
 #include <Eigen/Core>
 #include <ceres/ceres.h>
 #include <ceres/rotation.h>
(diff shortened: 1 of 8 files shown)


src/colmap/estimators/cost_functions/motion_averaging.h
```diff
@@ -41,6 +41,44 @@ struct BATAPairwiseDirectionCostFunctor {
   const Eigen::Vector3d pos2_from_pos1_dir_;
 };
 
+// Covariance-weighted BATA residual with fixed camera rotation
+// (W = Sigma^(-1/2) * R_cam_from_world).
+struct WeightedBATAPairwiseDirectionCostFunctor {
```
ahojnnes: Can we just use our CovarianceWeightedCostFunctor on the existing BATA cost function?

src/colmap/estimators/global_positioning.cc
```diff
@@ -440,11 +448,102 @@ void GlobalPositioner::ConvertBackResults(Reconstruction& reconstruction) {
...
+std::unique_ptr<GlobalPositioner> CreateDefaultGlobalPositioner(
+    const GlobalPositionerOptions& options,
+    const PoseGraph& pose_graph,
+    Reconstruction* reconstruction,
+    const ObservationWhiteningMap& observation_whitening,
+    std::shared_ptr<ceres::LossFunction> loss_function) {
+  if (reconstruction == nullptr) {
+    throw std::invalid_argument("reconstruction must not be null");
+  }
+  std::unique_ptr<GlobalPositioner> positioner(new GlobalPositioner(options));
```
ahojnnes: make_unique

src/colmap/estimators/global_positioning.cc
```diff
@@ -440,11 +448,102 @@ void GlobalPositioner::ConvertBackResults(Reconstruction& reconstruction) {
...
+  options_.solver_options.linear_solver_ordering = std::move(ordering);
+}
+
+std::unique_ptr<GlobalPositioner> CreateDefaultGlobalPositioner(
+    const GlobalPositionerOptions& options,
+    const PoseGraph& pose_graph,
+    Reconstruction* reconstruction,
+    const ObservationWhiteningMap& observation_whitening,
+    std::shared_ptr<ceres::LossFunction> loss_function) {
+  if (reconstruction == nullptr) {
```
ahojnnes: Let's just make the argument a reference, so we can skip this nullptr check?

src/colmap/estimators/global_positioning.cc
```diff
@@ -440,11 +448,102 @@ void GlobalPositioner::ConvertBackResults(Reconstruction& reconstruction) {
...
-  GlobalPositioner positioner(options);
-  return positioner.Solve(pose_graph, reconstruction);
+  if (reconstruction.NumImages() == 0) {
+    LOG(ERROR) << "Number of images = " << reconstruction.NumImages();
+    return false;
+  }
+  if (reconstruction.NumPoints3D() == 0) {
+    LOG(ERROR) << "Number of tracks = " << reconstruction.NumPoints3D();
+    return false;
+  }
```
ahojnnes:
```suggestion
  if (reconstruction.NumImages() == 0 || reconstruction.NumPoints3D() == 0) {
    LOG(ERROR) << "Failed to run global positioning for empty incomplete reconstruction: " << reconstruction;
    return false;
  }
```

src/colmap/estimators/global_positioning.h
```diff
@@ -115,8 +148,28 @@ class GlobalPositioner {
   // Temporary storage for camera-in-rig positions when cam_from_rig is unknown
   // and needs to be estimated.
   NodeHashMap<sensor_t, Eigen::Vector3d> cams_in_rig_;
+
+  // Retained for late parameter ordering and Finalize().
+  Reconstruction* reconstruction_ = nullptr;
+  std::unique_ptr<ceres::Problem> problem_;
+
+ private:
+  friend std::unique_ptr<GlobalPositioner> CreateDefaultGlobalPositioner(
```
ahojnnes: This is a bit of an awkward design. Let's instead add a static CreateDefaul() method to the global positioner and remove the friend declaration.

src/colmap/scene/track.h
```diff
@@ -49,6 +50,9 @@ struct TrackElement {
   inline bool operator!=(const TrackElement& other) const;
 };
 
+// Identifies one observation in a Point3D track for per-observation lookup.
+using Point3DTrackElementKey = std::pair<point3D_t, uint64_t>;
```
ahojnnes, line 54: Not sure I understand what this is? Is it point3D_id x track_el_idx ?

src/colmap/estimators/global_positioning.cc
```diff
@@ -25,61 +25,67 @@ GlobalPositioner::GlobalPositioner(const GlobalPositionerOptions& options)
...
+
+void GlobalPositioner::Prepare(
+    const PoseGraph& pose_graph,
+    Reconstruction& reconstruction,
+    const ObservationWhiteningMap& observation_whitening,
+    std::shared_ptr<ceres::LossFunction> loss_function) {
   if (reconstruction.NumImages() == 0) {
     LOG(ERROR) << "Number of images = " << reconstruction.NumImages();
-    return false;
+    throw std::runtime_error("global positioning requires images");
```
ahojnnes, line 86: Why did we change this to an exception? We could return a boolean and then avoid the redundant empty check in the "CreateDefault..." ?

src/colmap/estimators/global_positioning.h
```diff
@@ -22,6 +29,12 @@ struct GlobalPositionerOptions {
   bool optimize_positions = true;
   bool optimize_points = true;
   bool optimize_scales = true;
+  // Whether to fix the first active observation scale to remove the gauge
+  // ambiguity when optimizing scales.
+  bool fix_observation_scale_gauge = true;
```
B1ueber2y: These new options seem to be irrelevant to what the PR title describes. Consider splitting.

src/colmap/estimators/cost_functions/motion_averaging.h
B1ueber2y: In COLMAP we no longer pass the covariance matrix into the cost function. Instead, we have a unified covariance reweighted wrapper to achieve this.

B1ueber2y (commented):
Thanks. Quickly skimmed through the PR on the train. Will have a detailed pass later.

src/pycolmap/estimators/motion_averaging.cc
```diff
@@ -35,6 +37,16 @@ void BindGlobalPositioner(py::module& m) {
           .def_readwrite("optimize_scales",
                          &GlobalPositionerOptions::optimize_scales,
                          "Whether to optimize scales.")
+          .def_readwrite(
+              "fix_observation_scale_gauge",
+              &GlobalPositionerOptions::fix_observation_scale_gauge,
+              "Whether to fix the first active observation scale to remove "
+              "the scale gauge.")
+          .def_readwrite(
+              "downweight_uncalibrated_observations",
```
lpanaf: also this one, seems to be not directly related to the current pr

src/colmap/estimators/global_positioning.cc
```diff
@@ -136,14 +142,18 @@ void GlobalPositioner::InitializeRandomPositions(
 }
 
 void GlobalPositioner::AddPointToCameraConstraints(
-    Reconstruction& reconstruction) {
+    Reconstruction& reconstruction,
+    const ObservationWhiteningMap& observation_whitening) {
```
lpanaf: maybe introduce the whitening in a different pr?

ahojnnes (commented):
@lpanaf Can you take a look as well?

src/colmap/estimators/global_positioning.cc
```diff
@@ -185,15 +200,30 @@ void GlobalPositioner::AddPoint3DToProblem(point3D_t point3D_id,
...
+        throw std::invalid_argument("observation weights must be finite");
+      }
+      whitening_matrix = &whitening_it->second;
+      if (!image.IsRefInFrame()) {
+        throw std::invalid_argument(
+            "weighted BATA does not support non-reference rig images");
+      }
+    }
 
-    if (!options_.generate_scales && random_initialization) {
```
lpanaf, line 196: why is this random_initialization condition got removed?

src/colmap/estimators/global_positioning.cc
```diff
@@ -266,48 +294,27 @@ void GlobalPositioner::AddPoint3DToProblem(point3D_t point3D_id,
...
   }
+}
 
-  // Add the cam_in_rig to be estimated into the parameter group
-  for (auto& [sensor_id, center] : cams_in_rig_) {
-    if (problem_->HasParameterBlock(center.data())) {
-      parameter_ordering->AddElementToGroup(center.data(), group_id);
-    }
-  }
+void GlobalPositioner::AddObservationResidual(
```
lpanaf: The design is a bit imbalanced, since this only sets up the residuale for the non-rig / ref images.

src/colmap/estimators/global_positioning.cc
```diff
@@ -325,9 +332,9 @@ void GlobalPositioner::ParameterizeVariables(Reconstruction& reconstruction) {
 
   // If not optimizing positions, set frame centers to be constant.
   if (!options_.optimize_positions) {
-    for (auto& [frame_id, center] : frame_centers_) {
-      if (problem_->HasParameterBlock(center.data())) {
-        problem_->SetParameterBlockConstant(center.data());
+    for (const auto& [frame_id, _] : frame_centers_) {
+      if (double* center = FrameCenterParameterBlock(frame_id)) {
+        problem_->SetParameterBlockConstant(center);
```
lpanaf: Since frame_id comes from iterating frame_centers, it would waste the extra search?

src/colmap/estimators/global_positioning.h
Zador-Pataki: Agreed that these are behavioral changes beyond the core prepared-problem API. @lpanaf also mentioned this and additionally raised the related scale-initialization case, so I am grouping them here. I would prefer to address them in a follow-up PR, but their clean interfaces depend on the factory introduced here, so I am unsure how to split them without stacking the follow-up. Suggestions are welcome. 

The motivation is to replace hard-coded behavior and heuristics that restrict downstream research: 
<details>
<summary>[Scale gauge]</summary> 
GP unconditionally fixes an internal observation scale that callers cannot access or later unfix. This is necessary for stock GP’s scale-ambiguous objective, but becomes a harmful additional hard constraint when metric measurements, such as depth or IMU residuals, already determine scale. 
</details>
<details>
<summary>[Calibration weighting]</summary> The fixed `0.5` multiplier is a stock heuristic. In a problem extended with other residuals or modalities, it silently downweights uncalibrated BATA observations relative to every additional residual family and also wraps any caller-provided loss. A cleaner API could expose explicit calibrated and uncalibrated weights and default `CreateDefaultGlobalPositioner` to equal weighting (`1.0` for both). `RunGlobalPositioning` could explicitly retain `0.5` for uncalibrated observations to preserve existing behavior. 
</details>
<details>
<summary>[Scale initialization]</summary>  This appears to be a bug fix? GP exposes `generate_random_points = false` to preserve user-provided 3D point positions and `generate_scales = false` to derive observation scales from those positions and the initialized camera centers. However, the scale computation is additionally gated on random point initialization, so starting from user-defined 3D points still leaves every observation scale at `1`. Removing that condition makes the two options work together and enables continuation from an existing reconstruction or checkpoint. Please correct me if I am missing an intended reason for the existing condition. 
</details>

Not addressing these points would limit the usefulness of the global positioner. However, I am completely open to different plumbing that preserves its legacy behavior by default while allowing users to configure these choices explicitly.

B1ueber2y: Makes sense. Its perfectly fine to either stack PRs (you can select the target to merge into, but this requires developing in a non-forked repo), or wait for this PR to be merged until you create another one. Alternatively, you can also create a draft PR and make it ready after the upstream is merged. Given that we are not rushing in a time-sensitive condition, either of the three ways should work fine : )

src/colmap/estimators/cost_functions/motion_averaging.h
Zador-Pataki: Thanks, updated in [541f7d26](https://github.com/colmap/colmap/pull/4670/commits/541f7d2679aff182f73b79127915c8b263ff5e6b). This required a small refactor to make the existing BATA functors compatible with `AutoDiffCostFunctor`; most of the diff is the resulting removal of the custom weighted functor and redundant construction code.

@B1ueber2y you also had mentioned this

Zador-Pataki: Re [comment](https://github.com/colmap/colmap/pull/4670#discussion_r3912173335), shall I make this a follow up PR?

src/colmap/estimators/global_positioning.cc
Zador-Pataki: [Done](https://github.com/colmap/colmap/pull/4670/changes/dc961d5ce3b5a4a35cc4f947fca2a2326072509f). `std::make_unique` could not use the protected constructor, so I followed `CreateDefaultCeresBundleAdjuster` and added an internal `DefaultGlobalPositioner` with a public constructor. Does this look right?

src/colmap/scene/track.h
Zador-Pataki: Yes, exactly. I updated the comment accordingly. Longer term, I think `GlobalPositioner::CreateDefault` should optionally accept externally owned observation scales, allowing custom initialization and inspection after optimization. That would require the same indexing mechanism.

src/colmap/estimators/global_positioning.cc
Zador-Pataki: With the new factory design, `Prepare()` runs inside the `DefaultGlobalPositioner` constructor, so it cannot return `false`. Mirroring `CreateDefaultBundleAdjuster`, the factory can assume valid input because `RunGlobalPositioning()` already performs this check and returns `false`. I therefore propose removing this check.

src/colmap/estimators/global_positioning.cc
Zador-Pataki: nice thanks

src/colmap/estimators/global_positioning.h
Zador-Pataki: okay thanks I removed them from this PR and will make a follow up

ahojnnes:
I reviewed the current merge result against `main` and found the following:

- **P1 — The new Python test unconditionally imports optional `pyceres`.** [`motion_averaging_test.py:3`](https://github.com/colmap/colmap/blob/921b4a0e7ad6c1021b152afdb18d0e82a290e21e/src/pycolmap/estimators/motion_averaging_test.py#L3) imports `pyceres`, but it is absent from the project and cibuildwheel test dependencies. A clean `pytest` environment will fail during collection with `ModuleNotFoundError`. Please add it to the relevant test dependencies or skip this test when the optional package is unavailable.

- **P2 — The temporary frame-center parameter is not exposed to Python.** The [`GlobalPositioner` binding](https://github.com/colmap/colmap/blob/921b4a0e7ad6c1021b152afdb18d0e82a290e21e/src/pycolmap/estimators/motion_averaging.cc#L69-L79) omits `FrameCenterParameterBlock`. Python callers therefore cannot associate a frame ID with the temporary center optimized by the prepared problem, so they cannot add the position constraints that motivate this API.

- **P2 — Late ordering does not cover arbitrary extension parameters.** [`SetParameterBlockOrdering()`](https://github.com/colmap/colmap/blob/921b4a0e7ad6c1021b152afdb18d0e82a290e21e/src/colmap/estimators/global_positioning.cc#L503-L513) only adds unknown parameter blocks when their size is 1. Any caller-added vector block remains uncovered and the final count check throws. I reproduced this by adding a 2D parameter block to the composable-problem test. Unknown extension blocks should receive a safe final group; only known BATA scale addresses should be classified into group 0.

- **P2 — The covariance-map argument cannot be populated from Python.** [`ObservationCovarianceMap`](https://github.com/colmap/colmap/blob/921b4a0e7ad6c1021b152afdb18d0e82a290e21e/src/pycolmap/estimators/motion_averaging.cc#L81-L89) is made opaque and registered without a constructor or mapping operations. Thus Python cannot create a nonempty value for the exposed `observation_covariances` argument.

I also checked all existing review threads. The covariance-wrapper, `make_unique`, reference argument, track-key clarification, rig handling, random-initialization, and redundant-lookup comments are addressed in the current head. The request to move covariance whitening into a separate PR remains substantively unaddressed and still seems worth considering, particularly given the incomplete Python covariance API. The exception-design thread remains unchanged, but throwing from the direct factory seems defensible and is not a blocker.

The current merge result builds, and all five focused C++ global-positioning tests pass. GitHub currently reports no checks for the PR branch.

src/pycolmap/estimators/motion_averaging_test.py
```diff
@@ -72,6 +75,30 @@ def test_global_positioner_options_default_init() -> None:
...
+    options.random_seed = 42
+    loss = pyceres.CauchyLoss(0.1)
+    owner = pycolmap.create_default_global_positioner(
+        options,
+        pycolmap.PoseGraph(),
+        reconstruction,
+        loss_function=loss,
+    )
+    del loss
+    gc.collect()
```
ahojnnes, line 97: Why do we need this?

Zador-Pataki:
@ahojnnes Thanks for the review. I’ve addressed the remaining points:

  - PyCeres-dependent tests now skip when it is unavailable.
  - Frame-center parameter blocks are exposed as writable NumPy arrays that keep the positioner alive.
  - Restored stock construction-time parameter ordering instead of rebuilding it afterward. `ExtendParameterBlockOrdering()` now only extends that ordering. This addresses uncovered
  vector blocks: new blocks default to group 3, while independent single-residual scalars can use group 0. Callers can explicitly override assignments.
  - Covariance weighting has been removed from this PR and preserved for a follow-up.
  - Although throwing from the factory is defensible, I removed the duplicate checks to follow default BA more closely, as [proposed earlier](https://github.com/colmap/colmap/pull/4670#discussion_r3919840543). The factory assumes valid input, while `RunGlobalPositioning()` retains the checks and returns `false`. Happy to restore the exceptions if you
  prefer.
-  I also updated the PR description to reflect these changes.

ahojnnes (approved):

PR #4671, opened 2026-09-02
Add composable Ceres rotation averaging

Zador-Pataki:
## Motivation

Adding constraints to COLMAP's rotation-averaging problem currently requires hacking its implementation. The existing estimator uses L1-IRLS rather than Ceres.

This PR adds a separate Ceres rotation averager, defaulting to a Huber loss. Following `CreateDefaultBundleAdjuster(...)`, `CreateDefaultCeresRotationAverager(...)` returns the prepared problem so callers can extend it before solving. The existing rotation estimator retains its current behavior.

## Changes

- Add `CeresRotationAveragerOptions`, `CeresRotationAverager`, and `CreateDefaultCeresRotationAverager(...)`.
- Build a Ceres problem with parameter blocks referencing reconstruction-owned frame and sensor rotations, with configurable loss and solver options.
- Expose the prepared problem, solver options, `Solve()`, and insertion of relative-rotation residuals with explicit losses.
- Extract the existing `InitializeFromMaximumSpanningTree(...)` for reuse by both rotation estimators.
- Expose the construction and solve workflow through PyCOLMAP.

### DIFF
--- a/src/colmap/estimators/rotation_averaging_ceres.cc
+++ b/src/colmap/estimators/rotation_averaging_ceres.cc
@@ -1,0 +1,194 @@
+#include "colmap/estimators/rotation_averaging_ceres.h"
+
+#include "colmap/estimators/cost_functions/manifold.h"
+#include "colmap/estimators/cost_functions/quaternion_utils.h"
+#include "colmap/estimators/rotation_averaging.h"
+#include "colmap/scene/pose_graph.h"
+#include "colmap/scene/reconstruction.h"
+
+#include <algorithm>
+#include <limits>
+#include <stdexcept>
+#include <vector>
+
+#include <ceres/rotation.h>
+
+namespace colmap {
+namespace {
+
+struct RelativeRotationError {
+  explicit RelativeRotationError(const Eigen::Quaterniond& cam2_from_cam1)
+      : cam2_from_cam1(cam2_from_cam1) {}
+
+  template <typename T>
+  bool operator()(const T* const rotation1,
+                  const T* const rotation2,
+                  T* residuals) const {
+    const Eigen::Matrix<T, 3, 3> error =
+        EigenQuaternionMap<T>(rotation2).toRotationMatrix().transpose() *
+        cam2_from_cam1.cast<T>().toRotationMatrix() *
+        EigenQuaternionMap<T>(rotation1).toRotationMatrix();
+    ceres::RotationMatrixToAngleAxis(error.data(), residuals);
+    return true;
+  }
+
+  static ceres::CostFunction* Create(const Eigen::Quaterniond& cam2_from_cam1) {
+    return new ceres::AutoDiffCostFunction<RelativeRotationError, 3, 4, 4>(
+        new RelativeRotationError(cam2_from_cam1));
+  }
+
+  const Eigen::Quaterniond cam2_from_cam1;
+};
+
+}  // namespace
+
+CeresRotationAveragerOptions::CeresRotationAveragerOptions() {
+  solver_options.linear_solver_type = ceres::SPARSE_NORMAL_CHOLESKY;
+}
+
+CeresRotationAverager::CeresRotationAverager(
+    std::unique_ptr<ceres::Problem> problem,
+    ceres::Solver::Options solver_options,
+    Reconstruction& reconstruction)
+    : solver_options_(std::move(solver_options)),
+      reconstruction_(reconstruction),
+      problem_(std::move(problem)) {}
+
+ceres::Solver::Summary CeresRotationAverager::Solve() {
+  ceres::Solver::Summary summary;
+  ceres::Solve(solver_options_, problem_.get(), &summary);
+  if (summary.IsSolutionUsable()) {
+    for (const auto& [frame_id, frame] : reconstruction_.Frames()) {
+      if (frame.HasPose() &&
+          problem_->HasParameterBlock(
+              frame.RigFromWorld().rotation().coeffs().data())) {
+        reconstruction_.RegisterFrame(frame_id);
+      }
+    }
+  }
+  return summary;
+}
+
+ceres::Problem& CeresRotationAverager::Problem() { return *problem_; }
+const ceres::Problem& CeresRotationAverager::Problem() const {
+  return *problem_;
+}
+const ceres::Solver::Options& CeresRotationAverager::SolverOptions() const {
+  return solver_options_;
+}
+
+void CeresRotationAverager::AddRelativeRotationResidual(
+    const image_t image_id1,
+    const image_t image_id2,
+    const Eigen::Quaterniond& cam2_from_cam1,
+    std::shared_ptr<ceres::LossFunction> loss_function) {
+  Frame& frame1 =
+      reconstruction_.Frame(reconstruction_.Image(image_id1).FrameId());
+  Frame& frame2 =
+      reconstruction_.Frame(reconstruction_.Image(image_id2).FrameId());
+  double* rotation1 = frame1.RigFromWorld().rotation().coeffs().data();
+  double* rotation2 = frame2.RigFromWorld().rotation().coeffs().data();
+  if (!problem_->HasParameterBlock(rotation1) ||
+      !problem_->HasParameterBlock(rotation2)) {
+    throw std::invalid_argument(
+        "relative rotation residual references an image outside the Ceres "
+        "problem");
+  }
+  ceres::LossFunction* loss = loss_function.get();
+  if (loss != nullptr) {
+    losses_.try_emplace(loss, std::move(loss_function));
+  }
+  problem_->AddResidualBlock(RelativeRotationError::Create(cam2_from_cam1),
+                             loss,
+                             rotation1,
+                             rotation2);
+}
+
+std::unique_ptr<CeresRotationAverager> CreateDefaultCeresRotationAverager(
+    const CeresRotationAveragerOptions& options,
+    const PoseGraph& pose_graph,
+    Reconstruction& reconstruction) {
+  std::shared_ptr<ceres::LossFunction> loss(CreateCeresLossFunction(
+      options.loss_function_type, options.loss_function_scale));
+  FlatHashSet<image_t> image_ids;
+  std::vector<image_pair_t> pair_ids;
+  for (const auto& [pair_id, edge] : pose_graph.ValidEdges()) {
+    const auto [image_id1, image_id2] = PairIdToImagePair(pair_id);
+    if (!reconstruction.ExistsImage(image_id1) ||
+        !reconstruction.ExistsImage(image_id2)) {
+      throw std::invalid_argument("rotation edge references unknown image");
+    }
+    const Image& image1 = reconstruction.Image(image_id1);
+    const Image& image2 = reconstruction.Image(image_id2);
+    if (image1.FrameId() == image2.FrameId() || !image1.IsRefInFrame() ||
+        !image2.IsRefInFrame() || image1.FramePtr()->NumDataIds() != 1 ||
+        image2.FramePtr()->NumDataIds() != 1) {
+      throw std::invalid_argument(
+          "Ceres rotation averaging does not support multi-camera rigs");
+    }
+    image_ids.insert(image_id1);
+    image_ids.insert(image_id2);
+    pair_ids.push_back(pair_id);
+  }
+  if (pair_ids.empty()) {
+    throw std::invalid_argument("Ceres rotation averaging requires edges");
+  }
+  if (pose_graph
+          .ConnectedFrameComponents(reconstruction,
+                                    /*filter_unregistered=*/false)
+          .size() != 1) {
+    throw std::invalid_argument(
+        "Ceres rotation averaging requires a connected pose graph");
+  }
+  std::sort(pair_ids.begin(), pair_ids.end());
+
+  if (!options.skip_initialization) {
+    for (const auto& [image_id, cam_from_world] :
+         ComputeImageRotationsFromMaximumSpanningTree(pose_graph, image_ids)) {
+      Frame& frame =
+          reconstruction.Frame(reconstruction.Image(image_id).FrameId());
+      if (!frame.HasPose()) {
+        Rigid3d rig_from_world;
+        rig_from_world.translation().setConstant(
+            std::numeric_limits<double>::quiet_NaN());
+        frame.SetRigFromWorld(rig_from_world);
+      }
+      frame.RigFromWorld().rotation() = cam_from_world.rotation();
+    }
+  }
+
+  ceres::Problem::Options problem_options;
+  problem_options.loss_function_ownership = ceres::DO_NOT_TAKE_OWNERSHIP;
+  auto problem = std::make_unique<ceres::Problem>(problem_options);
+  for (const image_t image_id : image_ids) {
+    Frame& frame =
+        reconstruction.Frame(reconstruction.Image(image_id).FrameId());
+    if (!frame.HasPose()) {
+      throw std::invalid_argument(
+          "rotation averaging frame has no initial pose");
+    }
+    double* rotation = frame.RigFromWorld().rotation().coeffs().data();
+    problem->AddParameterBlock(
+        rotation, 4, CreateEigenQuaternionManifold().release());
+  }
+  const image_t root_image_id =
+      *std::min_element(image_ids.begin(), image_ids.end());
+  Frame& root_frame =
+      reconstruction.Frame(reconstruction.Image(root_image_id).FrameId());
+  problem->SetParameterBlockConstant(
+      root_frame.RigFromWorld().rotation().coeffs().data());
+
+  std::unique_ptr<CeresRotationAverager> owner(new CeresRotationAverager(
+      std::move(problem), options.solver_options, reconstruction));
+  for (const image_pair_t pair_id : pair_ids) {
+    const auto [image_id1, image_id2] = PairIdToImagePair(pair_id);
+    owner->AddRelativeRotationResidual(
+        image_id1,
+        image_id2,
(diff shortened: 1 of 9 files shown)


B1ueber2y:
Is there a reason why this only supports non-rig setup?

Zador-Pataki:
@B1ueber2y No fundamental limitation; this was mainly to limit the initial implementation effort. Rig support could be added afterward. Is this scope acceptable? Or is it important to already support rig?

Zador-Pataki:
@B1ueber2y I added the rig support

src/colmap/estimators/rotation_averaging_ceres.cc
```diff
@@ -0,0 +1,224 @@
...
+
+ceres::Problem& CeresRotationAverager::Problem() { return *problem_; }
+const ceres::Problem& CeresRotationAverager::Problem() const {
+  return *problem_;
+}
+const ceres::Solver::Options& CeresRotationAverager::SolverOptions() const {
+  return solver_options_;
+}
+
+void CeresRotationAverager::AddRelativeRotationResidual(
```
ahojnnes, line 165: The header documents that frame/sensor rotations "must be configured in Problem()", but nothing enforces it here. If a caller passes image ids whose blocks were never added (e.g. images outside the pose graph), `AddResidualBlock` will implicitly create plain Euclidean parameter blocks without the quaternion manifold, and the rotation will be optimized off-manifold with no error — silently wrong results. Suggest `THROW_CHECK(problem_->HasParameterBlock(...))` for `rotation1`/`rotation2` and the non-null sensor blocks (mirroring how BA config validates its setup), or at minimum spelling out the consequence in the doc comment.

src/colmap/estimators/rotation_averaging_ceres.cc
```diff
@@ -0,0 +1,224 @@
...
+    throw std::invalid_argument("Ceres rotation averaging requires edges");
+  }
+  const auto frame_components = pose_graph.ConnectedFrameComponents(
+      reconstruction, /*filter_unregistered=*/false);
+  if (frame_components.size() != 1) {
+    throw std::invalid_argument(
+        "Ceres rotation averaging requires a connected pose graph");
+  }
+
+  if (!options.skip_initialization) {
```
ahojnnes, line 91: MST initialization mutates `reconstruction` before the sensor-rotation validation below, so when construction throws `invalid_argument` for a missing sensor rotation, the input is left partially initialized. Consider validating sensor presence first so a throw leaves the reconstruction untouched. Related: with `refine_sensor_from_rig=false` the missing-sensor case throws via `THROW_CHECK` inside init instead of `invalid_argument` — inconsistent exception contract depending on the flag.

src/colmap/estimators/rotation_averaging_ceres.cc
```diff
@@ -0,0 +1,224 @@
...
+    }
+  }
+  const image_t root_image_id =
+      *std::min_element(image_ids.begin(), image_ids.end());
+  Frame& root_frame =
+      reconstruction.Frame(reconstruction.Image(root_image_id).FrameId());
+  problem_->SetParameterBlockConstant(
+      root_frame.RigFromWorld().rotation().coeffs().data());
+
+  for (const auto& [pair_id, edge] : pose_graph.ValidEdges()) {
```
ahojnnes, line 135: Note that `Edge::num_matches` is ignored here — every edge gets the same loss/scale, unlike the L1-IRLS estimator's `INLIER_MATCH_COUNT` reweighting. For graphs with widely varying match counts this is a quality gap vs. the existing estimator. Worth either documenting the limitation or (as follow-up) a per-edge `ScaledLoss`.

src/colmap/estimators/rotation_averaging_ceres.cc
```diff
@@ -0,0 +1,224 @@
...
+    losses_.try_emplace(loss, std::move(loss_function));
+  }
+  if (sensor1 == nullptr && sensor2 == nullptr && !same_frame) {
+    problem_->AddResidualBlock(RelativeRotationError::Create(cam2_from_cam1),
+                               loss,
+                               rotation1,
+                               rotation2);
+    return;
+  }
+  std::vector<double*> blocks;
```
ahojnnes, line 197: Degenerate case: for a self-loop edge (same image twice, or two ref images in one frame), `blocks` stays empty here and the cost function ends up with zero parameter blocks. The L1 implementation explicitly skips such self-loops — suggest skipping or throwing on empty `blocks`.

src/colmap/estimators/rotation_averaging_ceres.h
```diff
@@ -0,0 +1,72 @@
...
+#include <ceres/ceres.h>
+
+namespace colmap {
+
+class PoseGraph;
+class Reconstruction;
+
+struct CeresRotationAveragerOptions {
+  CeresLossFunctionType loss_function_type = CeresLossFunctionType::HUBER;
+  double loss_function_scale = DegToRad(5.0);
```
ahojnnes, line 21: Please document the units here (radians — residuals are angle-axis vectors). The sibling option `irls_loss_parameter_sigma` is documented "in degrees", so a user setting `loss_function_scale = 5.0` by analogy would get a ~286° threshold and an effectively non-robust loss. Same for the Python docstring.

src/colmap/estimators/rotation_averaging_ceres.h
```diff
@@ -0,0 +1,72 @@
...
+      image_t image_id2,
+      const Eigen::Quaterniond& cam2_from_cam1,
+      std::shared_ptr<ceres::LossFunction> loss_function);
+
+ private:
+  ceres::Solver::Options solver_options_;
+  Reconstruction& reconstruction_;
+  FlatHashMap<ceres::LossFunction*, std::shared_ptr<ceres::LossFunction>>
+      losses_;
+  // Keep losses alive until the problem is destroyed.
```
ahojnnes, line 61: Nit: this comment belongs above `losses_` rather than `problem_`.

src/pycolmap/estimators/rotation_averaging_ceres.cc
```diff
@@ -0,0 +1,58 @@
...
+#include "pycolmap/helpers.h"
+#include "pycolmap/pybind11_extension.h"
+
+#include <pybind11/pybind11.h>
+
+using namespace colmap;
+using namespace pybind11::literals;
+namespace py = pybind11;
+
+void BindCeresRotationAverager(py::module& m) {
```
ahojnnes, line 17: Two nits: (1) please call `IsPyceresAvailable()` here like `BindBundleAdjuster`/`BindCostFunctions` do, so the `ceres::Problem`/`LossFunction` cross-module casters resolve without the user importing pyceres first — otherwise `averager.problem` raises a conversion error unless pyceres happens to be imported. (2) Neighboring estimator bindings document all members, while this one has no docstrings (notably `loss_function_scale` units, see C++ comment).

src/pycolmap/estimators/rotation_averaging_ceres.cc
```diff
@@ -0,0 +1,58 @@
...
+           &CeresRotationAverager::AddRelativeRotationResidual,
+           "image_id1"_a,
+           "image_id2"_a,
+           "cam2_from_cam1"_a,
+           "loss"_a)
+      .def_property_readonly(
+          "problem",
+          py::overload_cast<>(&CeresRotationAverager::Problem),
+          py::return_value_policy::reference_internal)
+      .def_property_readonly("solver_options",
```
ahojnnes, line 48: `SolverOptions()` is exposed with `copy`, so `averager.solver_options.max_num_iterations = N` in Python mutates a temporary and is silently discarded. For a composable tune-then-solve API, `reference_internal` seems intended; otherwise please document the snapshot semantics.

ahojnnes (commented):
Reviewed the full diff plus the surrounding code (existing L1-IRLS estimator, `PoseGraph`, `Frame`/`Rig`, pycolmap/pyceres interop). The residual math checks out — I verified the error composition against `Frame::SensorFromWorld`, including the same-frame and shared-sensor cases — the MST-init extraction is behavior-preserving, gauge fixing is now deterministic (min image id), and test coverage is solid. Findings are inline: no correctness bugs in the formulation, mainly an unenforced precondition, undocumented units, and small Python-binding issues.

One general note: unlike `RunRotationAveraging`, this is a primitive with no gravity mode, pose priors, outlier filtering, or de-registration — worth one doc sentence on intended use.

Zador-Pataki:
@ahojnnes Addressed the comments. Regarding missing sensor rotations (https://github.com/colmap/colmap/pull/4671#discussion_r4049740795), I made this match the existing rotation averager: missing sensor rotations now start at identity when refinement is enabled. When refinement is disabled, missing sensor rotations explicitly throw std::invalid_argument before modifying the reconstruction.

src/pycolmap/estimators/rotation_averaging_ceres.cc
```diff
@@ -0,0 +1,78 @@
...
+           "cam2_from_cam1"_a,
+           "loss"_a,
+           "Add a relative-rotation residual between images whose frame and "
+           "sensor rotations are configured in the problem.")
+      .def_property_readonly(
+          "problem",
+          py::overload_cast<>(&CeresRotationAverager::Problem),
+          py::return_value_policy::reference_internal,
+          "Ceres problem for inspecting and modifying the optimization.")
+      .def_property_readonly("solver_options",
```
ahojnnes: Removing the explicit `copy` here is a no-op: pybind11's `automatic` policy resolves lvalue-reference returns to `copy` (see `cast.h`), and `GlobalPositioner` binds this exact accessor shape with explicit `copy`. So this property is a snapshot — either restore explicit `copy` and drop post-construction tuning from the test, or add a non-const C++ accessor and bind with `reference_internal`.

ahojnnes (commented):
Reviewed the latest push (`fc1b67c4`). The residual math, gauge fixing, and ownership/lifetime design look good, and most earlier findings are addressed and covered by tests. Two must-fix items (the `solver_options` copy semantics + failing assertion, and the ruff violation that breaks the format gate), plus minor notes — all inline.

src/pycolmap/estimators/rotation_averaging_ceres_test.py
```diff
@@ -0,0 +1,65 @@
...
+    options = pycolmap.CeresRotationAveragerOptions()
+    assert options.reweighting == pycolmap.RotationAveragingReweighting.UNIFORM
+    options.reweighting = (
+        pycolmap.RotationAveragingReweighting.INLIER_MATCH_COUNT
+    )
+    options.solver_options.num_threads = 1
+    averager = pycolmap.create_default_ceres_rotation_averager(
+        options, pose_graph, reconstruction
+    )
+    averager.solver_options.num_threads = 2
```
ahojnnes, line 41: This mutates a temporary copy (see the `solver_options` binding comment), so `summary.num_threads_given` will be 1 and the assertion below fails. Please set `num_threads` via options before construction instead (or make the binding live).

src/pycolmap/estimators/rotation_averaging_ceres_test.py
```diff
@@ -0,0 +1,65 @@
+# SPDX-License-Identifier: BSD-3-Clause
+
+import gc
+import subprocess
+import sys
+
+import pytest
```
ahojnnes, line 5: `ruff check` flags this import block (`I001`), which fails the CI format gate (`scripts/format/python.sh` runs `ruff check --fix` plus `git diff --exit-code`). Please run the format script.

src/colmap/estimators/rotation_averaging_ceres.cc
```diff
@@ -0,0 +1,260 @@
...
+    add_rotation(frame.RigFromWorld().rotation());
+    if (!image.IsRefInFrame()) {
+      auto& pose =
+          frame.RigPtr()->MaybeSensorFromRig(image.CameraPtr()->SensorId());
+      if (!pose.has_value()) {
+        pose = Rigid3d(Eigen::Quaterniond::Identity(),
+                       Eigen::Vector3d::Constant(
+                           std::numeric_limits<double>::quiet_NaN()));
+      }
+      if (!pose->rotation().coeffs().allFinite()) {
```
ahojnnes: This throw happens after earlier loop iterations already posed frames and created sensor entries, so the reconstruction is left partially mutated (unlike the pre-init validation above). Consider hoisting the finiteness check before init. Also inconsistent: with refinement enabled a missing sensor starts at identity, but a NaN-rotation sensor throws.

src/colmap/estimators/rotation_averaging_ceres.cc
```diff
@@ -0,0 +1,260 @@
...
+    throw std::invalid_argument("Ceres rotation averaging requires edges");
+  }
+  const auto frame_components = pose_graph.ConnectedFrameComponents(
+      reconstruction, /*filter_unregistered=*/false);
+  if (frame_components.size() != 1) {
+    throw std::invalid_argument(
+        "Ceres rotation averaging requires a connected pose graph");
+  }
+
+  if (!options.refine_sensor_from_rig) {
```
ahojnnes: Note the shared MST init below validates *all* rigs in the reconstruction when `refine_sensor_from_rig=false`, while this pre-check covers only graph images — averaging a subgraph can throw for unrelated rigs' sensors. Worth a doc line (or scoping).

src/colmap/estimators/rotation_averaging_ceres.h
```diff
@@ -0,0 +1,76 @@
...
+  bool skip_initialization = false;
+
+  // When false, treat each non-ref sensor's cam_from_rig rotation as a
+  // pre-calibrated constant
+  bool refine_sensor_from_rig = true;
+
+  CeresRotationAveragerOptions();
+};
+
+// Optimizes rotations directly in the reconstruction, which must outlive
```
ahojnnes: Worth one sentence on intended use: unlike `RunRotationAveraging`, this is a primitive with no gravity mode, pose priors, outlier filtering, or de-registration.

src/colmap/estimators/rotation_averaging_ceres.h
```diff
@@ -0,0 +1,76 @@
...
+      RotationAveragingReweighting::UNIFORM;
+
+  ceres::Solver::Options solver_options;
+
+  // Flag to skip maximum spanning tree initialization.
+  bool skip_initialization = false;
+
+  // When false, treat each non-ref sensor's cam_from_rig rotation as a
+  // pre-calibrated constant
+  bool refine_sensor_from_rig = true;
```
ahojnnes, line 35: Please clarify that `true` only estimates *uncalibrated* (NaN-translation) sensors — fully calibrated sensor rotations stay constant unless the user flips them via `Problem()`.

src/pycolmap/estimators/rotation_averaging_ceres.cc
```diff
@@ -0,0 +1,78 @@
...
+              "refine_sensor_from_rig",
+              &CeresRotationAveragerOptions::refine_sensor_from_rig,
+              "When False, treat each non-ref sensor's cam_from_rig rotation "
+              "as a pre-calibrated constant.");
+  MakeDataclass(PyOptions);
+
+  py::classh<CeresRotationAverager>(m, "CeresRotationAverager")
+      .def("solve",
+           &CeresRotationAverager::Solve,
+           py::call_guard<py::gil_scoped_release>(),
```
ahojnnes, line 52: `solve` releases the GIL, unlike the sibling `solve` bindings (`BundleAdjuster`, `GlobalPositioner` hold it; only free-function wrappers release). Likely safe here, but please confirm it's intentional.

src/pycolmap/estimators/rotation_averaging_ceres_test.py
```diff
@@ -0,0 +1,65 @@
...
+
+import pycolmap
+
+
+def test_ceres_rotation_averager_adds_residual_and_solves() -> None:
+    pytest.importorskip("pyceres")
+    subprocess.run([sys.executable, __file__], check=True, timeout=30)
```
ahojnnes: The subprocess pattern is novel for pycolmap estimator tests — in-process plus `importorskip` (as in `motion_averaging_test.py`) would match conventions and give better failure output. Minor.

src/pycolmap/estimators/rotation_averaging_ceres.cc
```diff
@@ -0,0 +1,78 @@
...
+                         &CeresRotationAveragerOptions::loss_function_type,
+                         "Loss function for relative-rotation residuals.")
+          .def_readwrite("loss_function_scale",
+                         &CeresRotationAveragerOptions::loss_function_scale,
+                         "Loss function scale in radians.")
+          .def_readwrite("reweighting",
+                         &CeresRotationAveragerOptions::reweighting,
+                         "Reweighting scheme for relative-rotation "
+                         "constraints: UNIFORM or INLIER_MATCH_COUNT.")
+          .def_readwrite("solver_options",
```
ahojnnes, line 34: Nit: the BA binding's `solver_options` docstring notes it requires PyCeres — worth the same note here.

Zador-Pataki:
@ahojnnes updated, thanks!
- [Solver options](https://github.com/colmap/colmap/pull/4671#discussion_r4057393007): I overlooked BA’s simpler direct-member pattern, so I switched to that instead of adding another accessor.
- [Initialization](https://github.com/colmap/colmap/pull/4671#discussion_r4057393015): unlike default RA, which can return false for NaN starting rotations, Ceres aborts when adding them to the problem. I now chose to treat refinable NaNs as missing rotations and initialize them.
- [GIL](https://github.com/colmap/colmap/pull/4671#discussion_r4057393029): releasing the GIL was intentional, following the existing `run_rotation_averaging()` binding. Here, releasing the GIL also lets Ceres worker threads call Python loss functions without deadlocking.

PR #4673, opened 2026-09-03
Stop deriving the hash map backend from the build machine

### LINKED ISSUE #4672: Hash map backend selection makes COLMAP ABI depend on the Boost version of the host
## Summary

Starting from COLMAP 4.2.0, `FlatHashMap`/`FlatHashSet` resolve to either `boost::unordered_flat_*` or `std::unordered_*` depending on a build option that is **auto-detected from the Boost version found on the build machine**. These aliases appear as data members in public headers, so two builds from identical source, with identical CMake arguments, are ABI-incompatible if the two machines have different Boost versions. The difference is header-only, so it produces no link error — only memory corruption at runtime.

## Details

- `cmake/FindDependencies.cmake:45` selects the backend when `COLMAP_HASH_MAP_BACKEND` is unset: Boost >= 1.84 (`:38`) gives `BOOST`, anything older gives `STD`. 
- `src/colmap/util/hash_containers.h:67` switches the aliases on the resulting `COLMAP_HASH_BOOST` macro.
- The aliases are used as data members in public headers, e.g.  `CorrespondenceGraph::image_pairs_`, `BundleAdjustmentConfig::image_ids_` and the other six sets, `ObservationManager`, `IncrementalMapper`,   `IncrementalTriangulator`, `Covariance`.

Consequently `sizeof` and the member layout of those classes differ between an Ubuntu 24.04 build (Boost 1.83 -> `STD`) and a build on any newer distribution or a vcpkg-based build (-> `BOOST`).

Within a single CMake project this is consistent: `cmake/CMakeHelper.cmake:103` and `:116` propagate the macro as `PUBLIC`/`INTERFACE`, so `find_package(COLMAP)` consumers inherit the same choice.

The problem is the **prebuilt pycolmap wheels**, which are built with `BOOST`. Any downstream project that links its own COLMAP and also loads pycolmap into the same interpreter — a pybind11 extension that accepts `pycolmap` objects, or one whose bound classes derive from COLMAP types registered by pycolmap — gets
two COLMAPs with different layouts for the classes above. Nothing in the wheel records or exposes which backend it was built with, so the downstream cannot detect the mismatch, at build time or at run time.

## Reproduction

1. `pip install pycolmap==4.2.0` (built with `BOOST`).
2. On Ubuntu 24.04 (Boost 1.83, so `STD` is selected), build a pybind11 extension against COLMAP 4.2.0.
3. Pass a `pycolmap`-constructed object of an affected type into that extension, or register a bound class deriving from one of them.

Result: segfault. In our case the crash was inside `slot_tp_repr` during module import, because pybind11 evaluates `repr()` of default arguments at registration time; other access patterns corrupt memory more quietly.

## Environment

- COLMAP 4.2.0 / pycolmap 4.2.0 (PyPI wheels, manylinux and macOS arm64)
- Ubuntu 24.04, Boost 1.83, GCC 13

B1ueber2y:
Fixes https://github.com/colmap/colmap/issues/4672

FlatHashMap/FlatHashSet and their node variants are data members of classes in public headers, so COLMAP_HASH_MAP_BACKEND determines the layout of Reconstruction, CorrespondenceGraph, BundleAdjustmentConfig and others: sizeof(Reconstruction) is 320 bytes with STD and 280 with BOOST. Both backends are header-only, so a mismatch changes no mangled name and produces no link error, only memory corruption.

The backend was auto-selected from the Boost found at configure time, which made that layout a property of the build machine rather than of the source and the CMake arguments. The published manylinux pycolmap wheels are built through vcpkg and got BOOST; an apt-based Ubuntu 24.04 build of the same commit gets STD. An extension linking its own COLMAP and sharing bound types with pycolmap in one interpreter then reads those objects at the wrong offsets, since pybind11 keys registration on type_index, which is identical for both layouts.

Default to STD instead, and keep the old behavior available as an explicit AUTO that warns. BOOST stays an opt-in that has to be applied to everything in the process. colmap-config.cmake now pre-sets the value COLMAP was built with, so find_package() consumers reproduce that choice instead of re-deriving it from their own Boost, turning a mismatch into a configure-time error. The choice is also recorded in the binary as colmap::kHashMapBackend, reported by GetBuildInfo(), and exposed as pycolmap.__hash_map_backend__ for downstreams to assert on.

### DIFF
--- a/.github/workflows/build-mac.yml
+++ b/.github/workflows/build-mac.yml
@@ -77,8 +77,8 @@ jobs:
           cmake --version
           mkdir -p build
           cd build
-          # Only CI coverage of the non-default BOOST hash map backend. The
-          # bundle ships no headers or libraries, so nothing links against it.
+          # The app bundle ships no headers or libraries, so nothing links
+          # against it and it can use the faster non-default hash map backend.
           cmake .. \
             -GNinja \
             -DCMAKE_BUILD_TYPE=${{ matrix.config.cmakeBuildType }} \
--- a/.github/workflows/build-mac.yml
+++ b/.github/workflows/build-mac.yml
@@ -77,10 +77,8 @@ jobs:
           cmake --version
           mkdir -p build
           cd build
-          # The only job covering the non-default BOOST hash map backend: the
-          # Ubuntu jobs' apt Boost is too old, and the Windows job exports a
-          # linkable install tree, which must keep the default ABI. This job
-          # publishes a self-contained app bundle, so the backend stays private.
+          # Only CI coverage of the non-default BOOST hash map backend. The
+          # bundle ships no headers or libraries, so nothing links against it.
           cmake .. \
             -GNinja \
             -DCMAKE_BUILD_TYPE=${{ matrix.config.cmakeBuildType }} \
--- a/.github/workflows/build-mac.yml
+++ b/.github/workflows/build-mac.yml
@@ -77,11 +77,16 @@ jobs:
           cmake --version
           mkdir -p build
           cd build
+          # The only job covering the non-default BOOST hash map backend: the
+          # Ubuntu jobs' apt Boost is too old, and the Windows job exports a
+          # linkable install tree, which must keep the default ABI. This job
+          # publishes a self-contained app bundle, so the backend stays private.
           cmake .. \
             -GNinja \
             -DCMAKE_BUILD_TYPE=${{ matrix.config.cmakeBuildType }} \
             -DTESTS_ENABLED=ON \
             -DWERROR_ENABLED=ON \
+            -DCOLMAP_HASH_MAP_BACKEND=BOOST \
             -DQt6_DIR="$(brew --prefix qt)/lib/cmake/Qt6"
           ninja
 
--- a/.github/workflows/build-ubuntu.yml
+++ b/.github/workflows/build-ubuntu.yml
@@ -169,9 +169,9 @@ jobs:
           fi
 
           # Ubuntu's apt Boost (1.74 on 22.04, 1.83 on 24.04) predates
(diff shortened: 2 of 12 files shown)


.github/workflows/build-mac.yml
```diff
@@ -77,11 +77,16 @@ jobs:
           cmake --version
           mkdir -p build
           cd build
+          # The only job covering the non-default BOOST hash map backend: the
```
ahojnnes: Is this true? Ubuntu 26.06 afaik supports it?

.github/workflows/build-mac.yml
```diff
@@ -77,11 +77,16 @@ jobs:
...
+          # Ubuntu jobs' apt Boost is too old, and the Windows job exports a
+          # linkable install tree, which must keep the default ABI. This job
```
ahojnnes: Not sure I understand the concern for Windows? Why can it not use boost? The ABI issue is not boost specific but a C++ issue and applies also to the libc++/libstdc++ version.

ahojnnes (approved):
Thank you! Enforcing a consistent boost version is a good safety check during configure time. I left two questions where I am not sure I understand the issue.

.github/workflows/build-mac.yml
B1ueber2y: Yes. 26.04 has Boost 1.90. The Ubuntu jobs share one configure line across 22.04/24.04/26.04, and the older two are below 1.84. But since STD is now just the default, I dropped that comment instead.

.github/workflows/build-mac.yml
B1ueber2y: Good point. Nothing blocks it on windows. I just updated it and now boost is pinned on both windows and macos.

PR #4674, opened 2026-09-03
Vendor Boost so the hash map backend does not depend on the host

### LINKED ISSUE #4672: Hash map backend selection makes COLMAP ABI depend on the Boost version of the host

B1ueber2y:
Fixes https://github.com/colmap/colmap/pull/4672

An alternative to https://github.com/colmap/colmap/pull/4673. Comparing with the other one, this PR always uses boost as backend which keeps the speed up at the sacrifice of vendoring boost when the system boost version is lower than 1.84. 

The hash map backend is auto-selected from the Boost found at configure time, so `sizeof(Reconstruction)` is 280 bytes in the vcpkg-built pycolmap wheels and 320 in an apt-based Ubuntu build of the same commit. Both backends are header-only, so mixing them in one process corrupts memory instead of failing to link.

This fixes the default to `BOOST` and adds `FETCH_BOOST=AUTO`, which checks the system Boost first and fetches a pinned 1.88 only when it is too old for the selected backend. The backend is also exposed as `colmap::kHashMapBackend` / `pycolmap.__hash_map_backend__` so downstreams can assert on it.

### DIFF
--- a/.github/workflows/build-ubuntu.yml
+++ b/.github/workflows/build-ubuntu.yml
@@ -25,6 +25,8 @@ jobs:
             os: ubuntu-26.04,
             qtVersion: 6,
             cmakeBuildType: Release,
+            # The only job building the non-default STD hash map backend.
+            hashMapBackend: STD,
             asanEnabled: false,
             guiEnabled: true,
             cudaEnabled: false,
@@ -169,9 +171,9 @@ jobs:
           fi
 
           # Ubuntu's apt Boost (1.74 on 22.04, 1.83 on 24.04) predates
-          # boost::unordered_node_map (Boost >= 1.84), so COLMAP auto-selects the
-          # STD hash map backend here. The BOOST backend is exercised on the
-          # macOS (brew) and Windows (vcpkg) jobs. See COLMAP_HASH_MAP_BACKEND.
+          # boost::unordered_node_map (Boost >= 1.84), so FETCH_BOOST=AUTO fetches
+          # and builds Boost on these jobs. See FETCH_BOOST and, for the hash map
+          # backend it makes available, COLMAP_HASH_MAP_BACKEND.
           sudo apt-get update && sudo apt-get install -y \
             build-essential \
             cmake \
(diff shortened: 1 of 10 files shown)


B1ueber2y:
This alternative does not work, as we swapped Boost out from under CGAL, and CGAL doesn't survive the swap. Closing.

PR #4676, opened 2026-09-04
Prototype: make boost-unordered the only hash container backend

ahojnnes:
Prototype for the direction discussed on #4675 / #4672: instead of detecting a hash-container ABI mismatch, remove the possibility of one.

## Why

`FlatHashMap`/`NodeHashMap` are data members of classes in public headers, so `sizeof(Reconstruction)` is 320 bytes with `STD` and 280 with `BOOST`, and two COLMAP builds that disagree corrupt memory in one process rather than failing to link. #4673 addressed that by pinning the default to `STD`, which costs the faster containers everywhere — including the published `pycolmap` wheels, which had been getting `BOOST` from vcpkg and silently dropped to `STD`.

With one backend there is nothing to mismatch, no ABI tag needed, and no default to argue about.

## What

- `hash_containers.h` is unconditionally `boost::unordered`. `COLMAP_HASH_MAP_BACKEND`, `AUTO`, `colmap::kHashMapBackend` and `pycolmap.__hash_map_backend__` are gone, along with the explicit `-DCOLMAP_HASH_MAP_BACKEND=BOOST` in the macOS and Windows jobs.
- `boost::unordered_node_map` requires Boost >= 1.84, newer than apt Boost on Ubuntu 24.04 (1.83) and 22.04 (1.74). `FETCH_BOOST_UNORDERED` (ON by default, following `FETCH_POSELIB`/`FETCH_FAISS`/`FETCH_ONNX`) vendors a pinned copy when the available Boost is too old.
- The vendored set is the nine header-only modules in the include closure of `unordered_{flat,node}_{map,set}.hpp` at the pinned tag `boost-1.92.0`, determined empirically: `unordered`, `assert`, `config`, `container_hash`, `core`, `describe`, `mp11`, `predef`, `throw_exception`. 4.8 MB of tarballs, 3.2 MB of headers installed. They are only downloaded, never `add_subdirectory()`'d, so the per-module Boost CMake files can't collide with the `Boost::` targets from `find_package(Boost)`.
- Headers install to `include/colmap/thirdparty/boost-unordered` and are exported as an include directory of `colmap_boost_unordered`, so `find_package(colmap)` consumers compile COLMAP's public headers against the same copy.
- `colmap-config.cmake` records whether COLMAP used the system Boost. If it did, a consumer with an older Boost gets a configure-time error instead of a different container layout.

## Container layout is stable across Boost versions

I checked the concern that vendoring only shifts the problem from backend choice to Boost version. `sizeof` of all four containers is 48 bytes on 1.84, 1.86, 1.88, 1.90 and 1.92 (x86-64/arm64, `std::hash` as the hash), against 56 for `std::unordered_map`. Not a proof of identical layout, but no drift in the range that matters. Pinning the fetched copy makes the question moot for the vendored path either way.

## vcpkg

`vcpkg.json` already declares `boost-unordered`, and vcpkg's Boost is far newer than 1.84, so `COLMAP_BOOST_UNORDERED_FROM_SYSTEM` is TRUE and the fetch path never runs under vcpkg — the vcpkg builds (Windows CI, the pycolmap wheels) just get one Boost, as before. That is also the configuration this change is meant to restore: those wheels were on `BOOST` until #4673.

## Verified locally (macOS, arm64)

- System-Boost path (1.92): configures, builds.
- Forced-fetch path: configures, builds all 391 targets, installs, and `doc/sample-project` builds and runs against the install tree with `-isystem .../colmap/thirdparty/boost-unordered` on the consumer's command line.

## Open question this PR exists to answer

Ubuntu 22.04, where vendored 1.92 `unordered` headers land on the include path ahead of the system Boost 1.74 headers that Boost.Graph and Boost.ProgramOptions come from, in the same translation units (`math/graph_cut.h` includes both). macOS 1.92-vs-1.92 proves nothing about that gap. If 22.04 breaks, the fallback is vendoring more of Boost, or dropping 22.04 support.

Draft until CI reports on that.

### DIFF
--- a/.github/workflows/build-mac.yml
+++ b/.github/workflows/build-mac.yml
@@ -77,14 +77,11 @@ jobs:
           cmake --version
           mkdir -p build
           cd build
-          # The app bundle ships no headers or libraries, so nothing links
-          # against it and it can use the faster non-default hash map backend.
           cmake .. \
             -GNinja \
             -DCMAKE_BUILD_TYPE=${{ matrix.config.cmakeBuildType }} \
             -DTESTS_ENABLED=ON \
             -DWERROR_ENABLED=ON \
-            -DCOLMAP_HASH_MAP_BACKEND=BOOST \
             -DQt6_DIR="$(brew --prefix qt)/lib/cmake/Qt6"
           ninja
 
--- a/.github/workflows/build-windows.yml
+++ b/.github/workflows/build-windows.yml
@@ -149,11 +149,8 @@ jobs:
             # during CUDA compiler detection.
             $env:CUDAFLAGS = "-allow-unsupported-compiler"
           }
-          # The package ships no headers or libraries, so nothing links against
-          # it and it can use the faster non-default hash map backend.
           cmake .. `
             -GNinja `
-            -DCOLMAP_HASH_MAP_BACKEND=BOOST `
             -DCMAKE_MAKE_PROGRAM=ninja `
             -DCMAKE_BUILD_TYPE=Release `
             -DTESTS_ENABLED=${{ matrix.config.testsEnabled }} `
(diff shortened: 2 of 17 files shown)


B1ueber2y:
Thanks! This looks indeed much better than the discarded https://github.com/colmap/colmap/pull/4674.

ahojnnes:
@B1ueber2y I think we should go with #4677.

PR #4677, opened 2026-09-05
Build a pinned Boost from source instead of vendoring boost-unordered

ahojnnes:
Alternative to #4676 — **only one of the two should be merged.**

Both PRs make `boost::unordered` the only hash container backend (see #4676 for that rationale, which is unchanged here). They differ only in how a new enough boost-unordered is guaranteed on distros whose Boost is older than 1.84.

## Why an alternative

#4676 vendors just the nine header-only modules in boost-unordered's include closure. To win, those headers have to sit ahead of the system Boost on the include path — which also shadows the support modules they bring with them (`core`, `config`, `mp11`, `container_hash`, ...). The rest of the system Boost then compiles against them.

That is already visible: on Ubuntu 24.04, system Boost 1.83's `property_tree` calls `boost::swap`, which vendored 1.92's `boost/core` has deprecated. #4676 needs `BOOST_ALLOW_DEPRECATED` to build. The deprecation is benign, but the shadowing it reveals is not bounded to warnings: **`Boost::graph` and `Boost::program_options` are compiled, dynamically linked system libraries**, and their headers are now being compiled against newer support headers than the `.so` was built from.

Extending the vendored set to cover `property_tree` makes this worse, not better: its closure adds `mpl`, `type_traits`, `iterator`, `range`, `concept_check`, `preprocessor` — 20 modules, ~23 MB of headers — which is most of what Boost.Graph's headers are built out of.

## What this PR does instead

Take all of Boost from one place, so headers and compiled libraries always match.

A system Boost >= 1.84 is always preferred and used as it is — that is the case on Ubuntu 26.04 (1.90), homebrew (1.92) and vcpkg, where nothing is downloaded and nothing extra is installed.

- Only where no such Boost is found does `FETCH_BOOST` kick in: it downloads the pinned `boost-1.92.0-cmake.tar.xz` release archive and builds it as a subproject. It is ON by default, but unlike `FETCH_POSELIB`/`FETCH_FAISS`/`FETCH_ONNX` that does not mean Boost gets fetched — it is a fallback switch, and turning it OFF makes a too-old system Boost a configure-time error instead.
- `BOOST_INCLUDE_LIBRARIES` limits configuration to the ten libraries COLMAP uses plus their dependencies; only `graph`, `program_options` and their compiled deps actually build.
- `BOOST_SKIP_INSTALL_RULES=OFF` — Boost defaults to skipping its install rules as a subproject, which would both leave the headers out of COLMAP's install tree and keep its targets out of any export set.
- Boost installs into COLMAP's prefix, and `colmap-config.cmake` points consumers at that copy rather than re-fetching or falling back to their own Boost.
- No `BOOST_ALLOW_DEPRECATED`, no `SYSTEM BEFORE` include-path juggling, no `colmap_boost_unordered` target.

`Boost::boost`, `Boost::graph` and `Boost::program_options` are all provided by Boost's own CMake build, so every existing link line in COLMAP is unchanged.

One ordering change was required: `colmap-config.cmake` now includes `FindDependencies.cmake` **before** `colmap-targets.cmake`. A vendored Boost is exported from its own export set rather than imported, so CMake generates an existence check for those targets in the targets file.

## Cost, measured

| | #4676 | this PR |
|---|---|---|
| download | 4.8 MB, always | 102 MB (`.tar.xz`), only when the system Boost is too old |
| extracted | — | 692 MB |
| Boost configure | — | ~4 s |
| Boost build | — | **~8 s** (88 targets, 10 cores) |
| headers installed | 3.2 MB, always | 110 MB, only when fetched |
| shadows system Boost | yes, 9 modules | no |

The build time is negligible. The real cost is the 102 MB download and a larger install tree, and it is only paid on distros whose Boost is older than 1.84 (Ubuntu 24.04 and earlier).

## Verified locally (macOS, arm64)

- System-Boost path (1.92): configures, `Using system Boost 1.92.0`.
- Forced-fetch path: configures, builds all 479 targets, installs (292 MB total), and `doc/sample-project` configures, builds and **runs** against the install tree — including Boost.ProgramOptions parsing, i.e. the vendored compiled library is the one being linked.

## Open question for CI

COLMAP is not Boost's only consumer in its own link line — CGAL in particular is built on Boost headers and, on a distro build, was packaged against the system Boost. This PR does not shadow Boost for COLMAP's own translation units any more, but a distro CGAL still gets compiled against the vendored headers. That boundary is what CI needs to answer here, the same way Ubuntu 22.04 was the open question for #4676.

### DIFF
--- a/.github/workflows/build-mac.yml
+++ b/.github/workflows/build-mac.yml
@@ -77,14 +77,11 @@ jobs:
           cmake --version
           mkdir -p build
           cd build
-          # The app bundle ships no headers or libraries, so nothing links
-          # against it and it can use the faster non-default hash map backend.
           cmake .. \
             -GNinja \
             -DCMAKE_BUILD_TYPE=${{ matrix.config.cmakeBuildType }} \
             -DTESTS_ENABLED=ON \
             -DWERROR_ENABLED=ON \
-            -DCOLMAP_HASH_MAP_BACKEND=BOOST \
             -DQt6_DIR="$(brew --prefix qt)/lib/cmake/Qt6"
           ninja
 
--- a/.github/workflows/build-windows.yml
+++ b/.github/workflows/build-windows.yml
@@ -149,11 +149,8 @@ jobs:
             # during CUDA compiler detection.
             $env:CUDAFLAGS = "-allow-unsupported-compiler"
           }
-          # The package ships no headers or libraries, so nothing links against
-          # it and it can use the faster non-default hash map backend.
           cmake .. `
             -GNinja `
-            -DCOLMAP_HASH_MAP_BACKEND=BOOST `
             -DCMAKE_MAKE_PROGRAM=ninja `
             -DCMAKE_BUILD_TYPE=Release `
             -DTESTS_ENABLED=${{ matrix.config.testsEnabled }} `
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -70,7 +70,9 @@ option(BENCHMARK_ENABLED "Whether to enable runtime benchmarking support" OFF)
 option(FETCH_POSELIB "Whether to consume PoseLib using FetchContent or find_package" ON)
 option(FETCH_FAISS "Whether to consume faiss using FetchContent or find_package" ON)
 option(FETCH_ONNX "Whether to consume ONNX using FetchContent or find_package" ON)
-option(FETCH_BOOST "Whether to build a pinned Boost using FetchContent when the available Boost is older than the required version" ON)
+# Unlike the FETCH_* options above, this one is only a fallback: a system Boost
+# that is new enough is always preferred and nothing is downloaded then.
+option(FETCH_BOOST "Whether to fall back to building a pinned Boost using FetchContent if no system Boost of the required version is found" ON)
 option(BUILD_SHARED_LIBS "Whether to build shared libraries (faster linktime, slower runtime)" OFF)
 option(ALL_SOURCE_TARGET "Whether to create a target for all source files (for Visual Studio / XCode development)" OFF)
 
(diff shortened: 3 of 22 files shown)


B1ueber2y (approved):
Thanks a lot! Agreed that we should go with this one. 

This installs Boost into the prefix root, so a default install puts Boost 1.92 at /usr/local/include/boost, which precedes the default search path and so shadows the system Boost for anything else built on that machine. Should we install it under a COLMAP-private directory the way https://github.com/colmap/colmap/pull/4676 does and point colmap-config.cmake there?

ahojnnes:
Good catch — fixed in a81b8c725.

Boost keys all of its install rules off `CMAKE_INSTALL_INCLUDEDIR`, `CMAKE_INSTALL_LIBDIR` and `BOOST_INSTALL_CMAKEDIR`, so those are now pointed at a private subdirectory for the subproject and restored afterwards. The vendored copy lands in:

```
include/colmap/thirdparty/boost/boost/...
lib/colmap/thirdparty/boost/libboost_*.a
lib/colmap/thirdparty/boost/cmake/Boost-1.92.0/
```

Nothing Boost-related is left at the prefix root — verified on a full fetch+install: no `include/boost`, no `lib/libboost_*`, no `lib/cmake/Boost*`. `colmap-config.cmake` already derived the consumer path from `COLMAP_BOOST_INSTALL_CMAKEDIR`, so it follows automatically, and `doc/sample-project` configures, builds and runs against the install tree, reporting `Using Boost vendored by COLMAP at <prefix>/lib/colmap/thirdparty/boost/cmake/Boost-1.92.0`.

The same commit fixes a second bug this surfaced. `CMAKE_BUILD_TYPE` was defaulted to Release *after* `FindDependencies.cmake` ran, so a configure without an explicit `-DCMAKE_BUILD_TYPE` built Boost with an empty configuration. Boost then installed no `boost_<lib>-targets-<config>.cmake`, and consumers got `IMPORTED_LOCATION not set for imported target "Boost::graph"`. The build type is now settled before dependencies are configured.

PR #4678, opened 2026-09-05
Fix LO-RANSAC refinement below sample minimum

ahojnnes:
## Summary

- stop recursive local optimization when an updated inlier set falls below the local estimator minimum
- prevent MSAC improvements with fewer inliers from invoking the five-point essential-matrix refiner with only four rays
- add a regression test reproducing a 6-to-4 inlier-set shrink

## Testing

- `cmake --build build --target colmap_optim_loransac_test -j2`
- `ctest --test-dir build -R "^optim/loransac_test$" --output-on-failure`


ahojnnes:
This was crashing on me on a dataset.

B1ueber2y (approved):

PR #4681, opened 2026-09-06
Remove pycolmap deprecations predating 4.2.0

ahojnnes:
Follow-up to #4081, which kept these entries as too new at the time. All of them shipped deprecated in 4.2.0, so they can go now:

- `feature/extraction.cc`: remove deprecated `Sift` class in favor of `FeatureExtractor.create()`, plus its test and unused local typedefs.
- `scene/camera.cc`: remove `Camera.create` alias of `create_from_model_id`.
- `geometry/pose_prior.cc`: remove `PosePrior.is_valid` / `is_covariance_valid` aliases of `has_position` / `has_position_cov`.
- `scene/correspondence_graph.cc`: remove `num_correspondences_between_images`, `num_correspondences_between_all_images`, and `find_correspondences_between_images` aliases.

Deliberately left alone:
- `DefDeprecation` helper in `helpers.h` (generic infra for future deprecations).
- `is_deprecated_image_prior` in the `Database` bindings, which mirrors the C++ core interface and needs a core-side decision first.

Test plan:
- All 4 edited translation units recompile cleanly (full `_core` link is blocked by a stale `install/lib` in this checkout — pre-existing, unrelated).
- `pytest src/pycolmap/feature/extraction_test.py src/pycolmap/geometry/pose_prior_test.py src/pycolmap/scene/camera_test.py src/pycolmap/scene/correspondence_graph_test.py`: 92 passed.
- `ruff check` + `ruff format --check` on the edited test file pass.

### DIFF
--- a/python/README.md
+++ b/python/README.md
@@ -304,23 +304,35 @@ import numpy as np
 import pycolmap
 from PIL import Image, ImageOps
 
-# Input should be grayscale image with range [0, 1].
+# Input should be a grayscale bitmap.
 img = Image.open('image.jpg').convert('RGB')
 img = ImageOps.grayscale(img)
-img = np.array(img).astype(np.float) / 255.
+bitmap = pycolmap.Bitmap.from_array(np.array(img))
 
 # Optional parameters:
-# - options: dict or pycolmap.SiftExtractionOptions
+# - options: pycolmap.FeatureExtractionOptions (default: SIFT)
 # - device: default pycolmap.Device.auto uses the GPU if available
-sift = pycolmap.Sift()
+extractor = pycolmap.FeatureExtractor.create()
 
 # Parameters:
-# - image: HxW float array
-keypoints, descriptors = sift.extract(img)
+# - bitmap: pycolmap.Bitmap (grayscale for SIFT)
+keypoints, descriptors = extractor.extract(bitmap)
 # Returns:
-# - keypoints: Nx4 array; format: x (j), y (i), scale, orientation
-# - descriptors: Nx128 array; L2-normalized descriptors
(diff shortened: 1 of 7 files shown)


B1ueber2y (approved):
Thanks. From a quick pass of AI:

1.	python/examples/convert_legacy_rotation_averaging_format.py:180 calls pycolmap.Camera.create(camera_id=..., model=..., focal_length=..., width=..., height=...). Kwargs map 1:1 to create_from_model_id, so it’s a one-line rename, but as it stands this PR breaks the example at runtime. CI won’t catch it — testpaths only collects test_*, and mypy runs on --package pycolmap, not python/examples/.

2.	python/README.md:300-323 still has the whole “SIFT Feature Extraction” section built on sift = pycolmap.Sift() / sift.extract(img). That’s not just a rename: FeatureExtractor.extract_from_float32_array returns FeatureKeypoints/FeatureDescriptors objects, not the documented Nx4 array plus L2-normalized floats, and — unlike the removed Sift — it does not downscale to EffMaxImageSize() or rescale keypoints back (that Thumbnail call lives in controllers/feature_extraction.cc:123). Worth saying so explicitly, since it’s a silent behavior change for anyone migrating on large images, and on the GPU path sift.cc:687 will THROW_CHECK_EQ on an oversized bitmap.

Nit: src/pycolmap/feature/extraction.cc:32, static std::map<int, std::unique_ptr<std::mutex>> sift_gpu_mutexes is unused (the real one is sift_gpu_mutexes_ in colmap/feature/sift.cc). Already dead before this PR, but the PR removes the typedefs right next to it, so it may as well go.

PR #4684, opened 2026-09-07
Add ScaleWeightedCostFunctor for stddev-based residual whitening

B1ueber2y:
Adds `ScaleWeightedCostFunctor`, a sibling of `CovarianceWeightedCostFunctor` that
whitens residuals with standard deviations instead of a full covariance, and
drops a `cast<T>()` from the existing covariance wrapper.

## Motivation

Weighting a cost functor by an isotropic or diagonal uncertainty currently means
constructing `stddev^2 * I` and paying, per residual block, a matrix inverse and
a Cholesky factorization at setup plus a dense `N x N` product on every
evaluation. That is the common case in tree: `AbsolutePoseRefinementOptions::
position_prior_covariance` defaults to identity, and the bundle adjustment pose
prior falls back to `prior_position_fallback_stddev^2 * I`.

`ceres::ScaledLoss` is not a substitute when a robust loss is attached. It forms
`a * rho(s)`, whereas whitening forms `rho(s / sigma^2)`, so the inlier/outlier
knee sits at `delta` regardless of `a` rather than scaling with `sigma`. It also
requires one heap loss object per distinct weight, since every problem here sets
`loss_function_ownership = DO_NOT_TAKE_OWNERSHIP`.

## Changes

- `ScaleWeightedCostFunctor<CostFunctor>`, taking either a per-residual vector
  of standard deviations or a single one broadcast over all residuals.
  `Create(stddev, ...)` produces exactly the same residuals and jacobians as
  `CovarianceWeightedCostFunctor::Create(stddev * stddev * I, ...)`. The
  reciprocal is taken once in the constructor, so the hot path multiplies.
- `CovarianceWeightedCostFunctor` now applies `left_sqrt_info_` as a `double`
  matrix. Ceres defines mixed `Jet`/scalar operator traits, so the previous
  `cast<T>()` was materializing an `N x N` matrix of Jets and multiplying their
  all-zero derivative parts. Results are bit-identical, no call site changes.
- pycolmap bindings for both forms on every functor that already has a
  covariance overload: `ReprojErrorCost`, the constant-pose and
  constant-3D-point variants, `RigReprojErrorCost` and its constant-rig
  variant, `AbsolutePosePriorCost`, `AbsolutePosePositionPriorCost`,
  `RelativePosePriorCost` and `Point3DAlignmentCost`. A float selects the
  isotropic overload and an array of per-residual standard deviations the
  diagonal one.
- Tests covering the standard deviation vector, the broadcast scalar, and
  equivalence with the covariance wrapper on both residuals and jacobians.

## Runtime

Release, `-O3 -march=native`, pinned to one core, best of 5 runs of 2M
`Evaluate()` calls with jacobians.

| cost functor | unweighted | Covariance (before) | Covariance (after) | Scale |
|---|---|---|---|---|
| `ReprojErrorCostFunctor<Pinhole>` (N=2, 14 params) | 345.2 ns | 400.3 ns | 401.8 ns | 351.1 ns |
| `AbsolutePosePositionPriorCostFunctor` (N=3, 7 params) | 353.5 ns | 474.5 ns | 450.2 ns | 363.2 ns |
| `AbsolutePosePriorCostFunctor` (N=6, 7 params) | 453.0 ns | 864.2 ns | 756.7 ns | 468.2 ns |

**Whitening overhead drops from +16% / +34% / +91% to +2% / +3% / +3%, and
removing the cast alone recovers 5% at N=3 and 12% at N=6 for callers that need
a full covariance.** The cast makes no difference at N=2, where the temporary is
a single 2x2 block.

### DIFF
--- a/src/colmap/estimators/bundle_adjustment_ceres.cc
+++ b/src/colmap/estimators/bundle_adjustment_ceres.cc
@@ -1029,32 +1029,43 @@ class PosePriorBundleAdjuster : public CeresBundleAdjuster {
 
     const Eigen::Vector3d normalized_position =
         normalized_from_metric_ * pose_prior.position;
-    const Eigen::Matrix3d normalized_from_metric_scaled_rotation =
+    // The fallback is isotropic and stays isotropic under the similarity
+    // transform, so it only needs a scale rather than a whitening matrix.
+    const double normalized_position_stddev =
         normalized_from_metric_.scale() *
-        normalized_from_metric_.rotation().toRotationMatrix();
-    const Eigen::Matrix3d position_cov =
-        pose_prior.HasPositionCov()
-            ? pose_prior.position_covariance
-            : (prior_options_.prior_position_fallback_stddev *
-               prior_options_.prior_position_fallback_stddev *
-               Eigen::Matrix3d::Identity());
-    const Eigen::Matrix3d normalized_position_cov =
-        normalized_from_metric_scaled_rotation * position_cov *
-        normalized_from_metric_scaled_rotation.transpose();
+        prior_options_.prior_position_fallback_stddev;
+    Eigen::Matrix3d normalized_position_cov;
+    if (pose_prior.HasPositionCov()) {
+      const Eigen::Matrix3d normalized_from_metric_scaled_rotation =
+          normalized_from_metric_.scale() *
+          normalized_from_metric_.rotation().toRotationMatrix();
+      normalized_position_cov =
+          normalized_from_metric_scaled_rotation *
+          pose_prior.position_covariance *
+          normalized_from_metric_scaled_rotation.transpose();
+    }
 
     if (image.IsRefInFrame()) {
       problem.AddResidualBlock(
-          CovarianceWeightedCostFunctor<AbsolutePosePositionPriorCostFunctor>::
-              Create(normalized_position_cov, normalized_position),
+          pose_prior.HasPositionCov()
+              ? CovarianceWeightedCostFunctor<
+                    AbsolutePosePositionPriorCostFunctor>::
+                    Create(normalized_position_cov, normalized_position)
+              : ScaleWeightedCostFunctor<AbsolutePosePositionPriorCostFunctor>::
+                    Create(normalized_position_stddev, normalized_position),
           prior_loss_function_.get(),
           rig_from_world.params.data());
     } else {
       Rigid3d& cam_from_rig =
           frame.RigPtr()->SensorFromRig(image.CameraPtr()->SensorId());
       problem.AddResidualBlock(
-          CovarianceWeightedCostFunctor<
-              AbsoluteRigPosePositionPriorCostFunctor>::
-              Create(normalized_position_cov, normalized_position),
+          pose_prior.HasPositionCov()
+              ? CovarianceWeightedCostFunctor<
+                    AbsoluteRigPosePositionPriorCostFunctor>::
+                    Create(normalized_position_cov, normalized_position)
+              : ScaleWeightedCostFunctor<
+                    AbsoluteRigPosePositionPriorCostFunctor>::
+                    Create(normalized_position_stddev, normalized_position),
           prior_loss_function_.get(),
           cam_from_rig.params.data(),
           rig_from_world.params.data());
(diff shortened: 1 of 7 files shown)


src/colmap/estimators/cost_functions/utils.h
```diff
@@ -164,4 +166,63 @@ class CovarianceWeightedCostFunctor {
...
+
+  template <typename... Args>
+  static ceres::CostFunction* Create(const StddevVec& stddevs, Args&&... args) {
+    return CreateAutoDiffCostFunction(new ScaleWeightedCostFunctor<CostFunctor>(
+        stddevs, std::forward<Args>(args)...));
+  }
+
+  template <typename... Args>
+  static ceres::CostFunction* Create(double stddev, Args&&... args) {
+    return CreateAutoDiffCostFunction(new ScaleWeightedCostFunctor<CostFunctor>(
```
ahojnnes, line 318: As a further optimization, it should be trivial to derive manual Jacobians for this? Same for Covariance...CostFunctor?

ahojnnes (approved):
Thanks, looks great. One suggestion left inline.

src/colmap/estimators/cost_functions/utils.h
B1ueber2y: Good point. Thanks! ns per `Evaluate()` with jacobians, measured interleaved in one process, min over 200 rounds, outputs bit-identical:

| | autodiff | manual |
|---|---|---|
| covariance, `ReprojErrorCostFunctor<Pinhole>` (N=2, blocks 3+7+4) | 307.2 | 274.3 |
| covariance, `AbsolutePosePriorCostFunctor` (N=6, block 7) | 440.0 | 397.2 |
| scale, `ReprojErrorCostFunctor<Pinhole>` (N=2, blocks 3+7+4) | 272.9 | 275.3 |
| scale, `AbsolutePosePriorCostFunctor` (N=6, block 7) | 378.9 | 385.5 |

Therefore, I switched the covariance wrapper to manual jacobians and kept autodiff for the scale one.

ahojnnes: Great, thanks.

src/colmap/estimators/cost_functions/utils.h
```diff
@@ -131,15 +196,54 @@ class CovarianceWeightedCostFunctor {
   using CovMat = Eigen::Matrix<double, kNumResiduals, kNumResiduals>;
 
   template <typename... Args>
-  explicit CovarianceWeightedCostFunctor(const CovMat& cov, Args&&... args)
-      : left_sqrt_info_(LeftSqrtInformation(cov)),
+  static ceres::CostFunction* Create(const CovMat& cov, Args&&... args) {
+    return new CovarianceWeightedCostFunction<CostFunctor, kParameterDims>(
+        cov,
+        CreateAutoDiffCostFunction(
```
ahojnnes, line 209: How does this work when I want to combine the covariance weighting with a cost function that already has manual Jacobians? Is this possible?

src/colmap/estimators/bundle_adjustment_ceres.cc
```diff
@@ -1014,32 +1014,43 @@ class PosePriorBundleAdjuster : public CeresBundleAdjuster {
...
+        prior_options_.prior_position_fallback_stddev;
+    Eigen::Matrix3d normalized_position_cov;
+    if (pose_prior.HasPositionCov()) {
+      const Eigen::Matrix3d normalized_from_metric_scaled_rotation =
+          normalized_from_metric_.scale() *
+          normalized_from_metric_.rotation().toRotationMatrix();
+      normalized_position_cov =
+          normalized_from_metric_scaled_rotation *
+          pose_prior.position_covariance *
+          normalized_from_metric_scaled_rotation.transpose();
```
sarlinpe: Can this be factored into a new method called in the `image.IsRefInFrame()` branch such that we don't instantiate `normalized_position_cov` if `HasPositionCov` is false?

sarlinpe (approved):
Thank you!

src/colmap/estimators/bundle_adjustment_ceres.cc
B1ueber2y: Done.

src/colmap/estimators/cost_functions/utils.h
B1ueber2y: It is now possible with the newest commit. PTAL on the new test.

ahojnnes (approved):

PR #4685, opened 2026-09-07
Refactor `BundleAdjustmentController` to take `BundleAdjustmentConfig` directly

StonerLing:
## Summary

- Refactor `BundleAdjustmentController` to take `(BundleAdjustmentOptions, BundleAdjustmentConfig, ...)` instead of `OptionManager`; callers now express the BA problem (images, constants, gauge, point selection) directly.
- Add a pose-prior-based overload `(..., PosePriorBundleAdjustmentOptions, pose_priors, ...)`.
- CLI `bundle_adjuster`: prior flags and redundant-point two-stage refinement.
- pycolmap `bundle_adjustment`: backward-compatible extension (`ba_config`, `pose_priors`, `use_prior_position`, `prior_options`).
- GUI widget builds the standard config locally.
- Add shared `RunBundleAdjustmentImpl()` used by CLI and pycolmap.

## Testing

- Extended `controllers/bundle_adjustment_test.cc` (fallback, pose-prior reconstruction).

### DIFF
--- a/src/colmap/controllers/bundle_adjustment.cc
+++ b/src/colmap/controllers/bundle_adjustment.cc
@@ -29,16 +29,37 @@
 
 #include "colmap/controllers/bundle_adjustment.h"
 
-#include "colmap/estimators/bundle_adjustment_ceres.h"
 #include "colmap/sfm/observation_manager.h"
+#include "colmap/util/logging.h"
 #include "colmap/util/misc.h"
 #include "colmap/util/timer.h"
 
 namespace colmap {
+
 BundleAdjustmentController::BundleAdjustmentController(
-    const OptionManager& options,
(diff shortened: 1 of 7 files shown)


ahojnnes:
Could you share the motivation for this change? It's not immediately obvious why we want this change.

StonerLing:
This started when I wanted to add an optional BA step after the `HierarchicalPipeline` and found it required duplicating a lot of setup code.

The old constructor took `OptionManager`, which tied the controller to global CLI state, callers could not express which images, constants, gauge, or points to optimize, and the controller hard-coded "all images". Features available elsewhere (pose priors, constant rigs/cameras, point selection/pruning) were unreachable from the standalone BA path.

PR #4686, opened 2026-09-08
Compute pitched row offsets in size_t in the rotate/transpose/flip kernels (fixes #3207)

### LINKED ISSUE #3207: CUDA Illegal Memory Access in patch_match_stereo with Large Images (8000x6000)
### **Description**
When running `colmap patch_match_stereo` on high-resolution images (8000x6000), I encounter a CUDA illegal memory access error during execution. However, when I limit the `--PatchMatchStereo.max_image_size` to **7700**, the process runs successfully.

### **Error Output**
```bash
CUDA error at /home/felix/workspace/colmap/src/colmap/mvs/gpu_mat.h:374 - an illegal memory access was encountered
This error is likely caused by the graphics card timeout detection mechanism of your operating system.
```

### **Steps to Reproduce**
1. Run the following command on an **8000x6000** image dataset:
   ```bash
    mkdir -p data/reconstruction
    colmap feature_extractor \
        --database_path data/reconstruction/database.db \
        --image_path data/images

    # Match features
    colmap exhaustive_matcher \
        --database_path data/reconstruction/database.db

    # Create sparse model
    mkdir -p data/reconstruction/sparse
    colmap mapper \
        --database_path data/reconstruction/database.db \
        --image_path data/images \
        --output_path data/reconstruction/sparse

    # Create dense model
    mkdir -p data/reconstruction/dense/0
    colmap image_undistorter \
        --image_path data/images \
        --input_path data/reconstruction/sparse/0 \
        --output_path data/reconstruction/dense/0 \
        --output_type COLMAP

    # Patch match stereo
   colmap patch_match_stereo \
       --workspace_path data/reconstruction/dense/0 \
       --workspace_format COLMAP \
       --PatchMatchStereo.geom_consistency true   ```
2. Observe the CUDA error in `gpu_mat.h:374`.
3. Add `--PatchMatchStereo.max_image_size 7700` and rerun. It works fine.

### **System Information**
- **COLMAP Commit Hash**: a586e7cb223cc86c609105246ecd3a10e0c55131
- **OS**: Ubuntu 22.04
- **GPU**: NVIDIA RTX 4090, 24GB VRAM
- **CUDA Version**: CUDA 12.2
- **NVIDIA Driver Version**: 11.5/12.2

### **Additional Notes**
- The issue occurs in the function **`GpuMat<T>::Rotate`** in `gpu_mat.h:374`.
- It might be related to **out-of-bounds memory access** when handling large images.
- Reducing the image size (`max_image_size`) avoids the crash, suggesting a potential **memory management issue** in PatchMatchStereo.


Would appreciate any insights from the maintainers on whether this is a known issue or if there's a potential fix!

maxfurni:
Fixes #3207 (CUDA illegal memory access in `patch_match_stereo` with large images).

### Cause

`CudaRotateKernel`, `CudaTransposeKernel` and `CudaFlipHorizontalKernel` take the pitch and the row index as `int` and compute the byte offset as `row * pitch` in 32-bit arithmetic. For wide element types that product exceeds `INT_MAX` on large images. The patch-match random-state map holds 48-byte `curandState` elements, so above roughly 44.7 MP the offsets of the last few hundred rows wrap negative and the kernels touch memory before the allocation. The slice base pointers in `GpuMat::Rotate` / `Transpose` / `FlipHorizontal` are already computed in `size_t`; only the per-row products inside the kernels were narrow.

The arithmetic matches the reports exactly:

| frame | rows whose offset overflows (input / rotated) | observed |
|---|---|---|
| 7700×5775 (the issue's `max_image_size 7700` workaround) | 0 / 0 | runs |
| 8192×5464 | 2 / 15 | appears to run — the wrapped writes land inside a neighbouring allocation (silent corruption, not a fault) |
| 8000×6000 (the issue) | 407 / 550 | faults |
| 7980×5981 (ours) | 381 / 503 | faults on an L4 and on an L40S (48 GB) alike — not a memory-capacity problem |

### Change

Compute the byte offsets in `size_t` in the three kernels. No behavioural change below the overflow line.

### Test

`GpuMat.RotateLargeRowOffsets` rotates an 8000×6000 `GpuMatPRNG` (two 2.3 GB buffers) and checks the device reports no error; it skips itself when the GPU has less than ~5 GB free, so CI without a large GPU is unaffected.

Verified by static analysis of every known data point; a full `patch_match_stereo` run on the 7980×5981 set with this patch is in progress on an L40S and I will report the result here.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/src/colmap/mvs/cuda_flip.h
+++ b/src/colmap/mvs/cuda_flip.h
@@ -72,8 +72,8 @@ __global__ void CudaFlipHorizontalKernel(T* output_data,
   for (int i = 0; i < TILE_DIM_FLIP; i += BLOCK_ROWS_FLIP) {
     const int x = min(x_index, width - 1);
     const int y = min(y_index, height - i - 1);
-    tile[tile_y + i][tile_x] =
-        *((T*)((char*)input_data + y * input_pitch + i * input_pitch) + x);
+    tile[tile_y + i][tile_x] = *(
+        (T*)((char*)input_data + static_cast<size_t>(y + i) * input_pitch) + x);
   }
 
   __syncthreads();
@@ -82,7 +82,8 @@ __global__ void CudaFlipHorizontalKernel(T* output_data,
   if (x_index < width) {
     for (int i = 0; i < TILE_DIM_FLIP; i += BLOCK_ROWS_FLIP) {
       if (y_index + i < height) {
-        *((T*)((char*)output_data + y_index * output_pitch + i * output_pitch) +
+        *((T*)((char*)output_data +
+               static_cast<size_t>(y_index + i) * output_pitch) +
           x_index) = tile[threadIdx.y + i][threadIdx.x];
       }
     }
--- a/src/colmap/mvs/cuda_rotate.h
+++ b/src/colmap/mvs/cuda_rotate.h
@@ -70,11 +70,6 @@ __global__ void CudaRotateKernel(T* output_data,
   int output_x = input_y;
   int output_y = width - 1 - input_x;
 
-  // Byte offsets must be computed in size_t: for wide element types
-  // (curandState is 48 bytes) a row offset of `y * pitch` exceeds
-  // INT_MAX above ~44.7 MP, wrapping negative and touching memory before
-  // the allocation (colmap/colmap#3207: illegal memory access on
-  // 8000x6000 images).
   *((T*)((char*)output_data + static_cast<size_t>(output_y) * output_pitch) +
     output_x) =
       *((T*)((char*)input_data + static_cast<size_t>(input_y) * input_pitch) +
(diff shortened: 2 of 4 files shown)


maxfurni:
Verified end to end on real data.

A full photometric `patch_match_stereo` pass over 25 reference views of 7980×5981 undistorted frames (8064×6048 iPhone captures, `--PatchMatchStereo.max_image_size -1`, 20 source images per reference image) on an NVIDIA L40S, built from `4bbab7a` plus this patch:

- unpatched: every view aborts at once with `CUDA error at gpu_mat.h:185 - an illegal memory access was encountered` (reproduced twice on an L4 and once on an L40S, so not a memory-capacity effect);
- patched: all 25 views complete, ~68 s each, 5 iterations × 4 sweeps per view (each sweep exercising `Rotate` on the 48-byte `curandState` map), exit 0, checkpoint written.

Same container, same inputs, only the three kernels changed.

ahojnnes:
Thanks, the fix looks good to me. I suggest removing the regression test, as it will fail run OOM on smaller GPUs.

ahojnnes (approved):

PR #4687, opened 2026-09-09
Refactor camera models into self-contained sensor/models/ headers

ahojnnes:
Decomposes the 3145-line `sensor/models.h` into an umbrella header plus a new `sensor/models/` folder, and removes the associated boilerplate. No behavior change.

**Split (mechanical, verified lossless):**
- `base.h`: model id enum, definition/registration macros, CRTP bases
- One header per model family (`pinhole`, `radial`, `opencv`, `fov`, `division`, `eucm`, `fisheye`, `fisheye_radial`, `thin_prism`, `spherical`)
- `runtime.h`: `CameraModel*` runtime dispatch (aggregates the family headers)
- `jacobian.h`: moved from `models_jacobian.h` (rename only)

**Follow-up simplifications:**
- All model implementations defined inline in the class bodies; definition macros slimmed to constants + static members
- Parameter-group index arrays derived from the macro group sizes via a `constexpr` helper; `params_info` passed through the macros
- `model_name`/`params_info`/idx members initialized directly (`constexpr`/inline); all `Initialize*()` methods and their `models.cc` definitions removed
- Name/id lookup: jump-table switch for id->name, lazily-built O(1) map for name->id (drops the old builders and their typo)
- Fisheye predicate derived from the base class instead of a hand-maintained model list
- Rename `CamRayFromImgJacobian` to `CamRayFromImgJac`

**Verification:**
- Script-checked that every moved function body/signature matches the original modulo formatting
- All index arrays, params info strings, and name/id round trips checked against the previous implementation
- Full build green; `sensor/*`, `scene/camera_test`, `scene/projection_test`, `estimators/cost_functions` pass
- Formatted with the pinned clang-format 22.1.5 / ruff 0.15.20


sarlinpe (approved):
Did not review every line because it's impossible to differentiate code moved vs changed.

ahojnnes:
I intentionally kept the tests unchanged and I think we have pretty thorough coverage on the models, so this should be a safe change.

PR #4689, opened 2026-09-09
Fix race condition in DownloadAndCacheFile directory creation

sarlinpe:
When running tests or downloads concurrently across multiple processes (such as during parallel test execution with `ctest -j`), multiple callers can race to create the download cache directory.

Previously, `DownloadAndCacheFile` used a check-then-act pattern:
```cpp
    if (!std::filesystem::exists(download_cache_dir)) {
      THROW_CHECK(std::filesystem::create_directories(download_cache_dir));
    }
```

  Because std::filesystem::create_directories returns false when the target directory already exists (e.g. when created by another racing process between the check and creation), THROW_CHECK(false) treated this normal success condition as a fatal error and threw an exception, causing flaky test failures in feature/onnx_matchers_test and feature/loma_test.


src/colmap/util/file.cc
```diff
@@ -490,7 +490,8 @@ std::string DownloadAndCacheFile(const std::string& uri) {
 
   if (!std::filesystem::exists(download_cache_dir)) {
     VLOG(2) << "Creating download cache directory: " << download_cache_dir;
-    THROW_CHECK(std::filesystem::create_directories(download_cache_dir));
+    THROW_CHECK(std::filesystem::create_directories(download_cache_dir) ||
```
ahojnnes, line 493: Should we reverse the order? First check for existence and then create?

ahojnnes (approved):
There may be another type of race condition when downloading the files in different processes. It would probably need some filesystem locking but it's a bit complicated.

src/colmap/util/file.cc
sarlinpe: Reversing the order would reintroduce the race condition we're trying to fix.

If we did `is_directory(...) || create_directories(...)`:
1. Thread A checks `is_directory` -> `false` (it does not exist yet).
2. Racing Thread B creates the directory in the meantime.
3. Thread A proceeds to `create_directories(...)`. Because the directory now already exists, `create_directories` returns `false` (per C++ spec, it only returns `true` if a new directory was created).
4. `false || false` evaluates to `false`, causing `THROW_CHECK` to fail.

With `create_directories(...) || is_directory(...)`:
- We attempt creation first. If it succeeds, it returns `true` and short-circuits.
- If it returns `false` because another racing thread/process already created it, we fall back to `is_directory` to confirm that a valid directory is now present.

PR #4690, opened 2026-09-09
Share tiny-solver refinement with GP3P; use LO-RANSAC for generalized absolute pose

ahojnnes:
Stacked on #4664 (base `pr-4664-base` = #4664 head; will rebase onto main after #4664 merges).

Shares the TinySolver rig refinement introduced for GP4PS with the rigid GP3P estimator and switches `EstimateGeneralizedAbsolutePose` to LO-RANSAC:
- `TinyRigCostFunctor<kScaled>` + `RefineRigPoseWithTinySolver` shared helper (pure refactor of the #4664 code, no behavior change)
- New `GP3PEstimator::Refine` honoring the estimator's residual type (reprojection error, or sine residual for cosine-distance scoring)
- `EstimateGeneralizedAbsolutePose` now runs LO-RANSAC with GP3P as its own local estimator

Measured on 500 paired-seed synthetic problems per config (60 points, 3-camera rig, threshold 4e-3): 3-6x lower rotation/translation error than plain RANSAC across inlier ratios 0.3-0.8 and noise levels 1e-4 to 1e-3, essentially all GT inliers recovered, 0-8% fewer trials (35% fewer at 3e-3 noise), at ~4 extra 10us refinements per estimate.

Test plan: `generalized_absolute_pose_test` 16/16 (incl. 3 new GP3P Refine tests), `generalized_pose_test` 16/16, clang-format clean.


src/colmap/estimators/solvers/generalized_absolute_pose_test.cc
```diff
@@ -142,6 +142,120 @@ INSTANTIATE_TEST_SUITE_P(GP3PEstimatorTests,
                                            std::make_pair(1, true),
                                            std::make_pair(2, true)));
 
+struct GP3PSProblem {
```
B1ueber2y, line 145: I assume this name is wrong and the letter S should not be there?

B1ueber2y (commented):
Nice improvement. Thank you!

src/colmap/estimators/solvers/generalized_absolute_pose_test.cc
ahojnnes: Thanks, fixed.

B1ueber2y (approved):

PR #4694, opened 2026-09-10
Unify OptionManager once-guards with RegisterOptionGroupOnce

ahojnnes:
Replace the 27 per-group `added_*_` bool flags in `BaseOptionManager` and `OptionManager` with a single shared `FlatHashSet<string>` of added option group names, guarded through the new `BaseOptionManager::RegisterOptionGroupOnce()`. `Check()` queries membership via the new `HasOptionGroup()`, and `Reset()` clears the set instead of resetting individual flags.

Also fixes `Reset()` never resetting `added_hierarchical_mapper_options_`, which previously made re-adding hierarchical mapper options after `Reset()` a silent no-op.

## Test plan

- `ninja colmap option_manager_test base_option_manager_test` builds clean (Release, Ninja)
- `ctest -R controllers/` — 16/16 suites pass, including `AddOptionsIdempotent`, `Reset`, `AddAllOptions`, `ReRead`
- CLI smoke: `colmap mapper --help` exits 0 with full listing; `colmap mapper` (missing required args) exits 1
- `clang-format --dry-run --Werror` clean on all 4 touched files

### DIFF
--- a/src/colmap/controllers/base_option_manager.cc
+++ b/src/colmap/controllers/base_option_manager.cc
@@ -57,20 +57,27 @@ BaseOptionManager::BaseOptionManager(bool add_project_options) {
   AddLogOptions();
 }
 
+bool BaseOptionManager::RegisterOptionGroupOnce(
+    const std::string& option_group) {
+  return added_option_groups_.insert(option_group).second;
+}
+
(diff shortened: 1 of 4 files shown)


lpanaf (approved):
Thanks! Everything looks a lot simpler now

PR #4695, opened 2026-09-10
Deduplicate two-view geometry estimation helpers

ahojnnes:
Extract shared file-local helpers in `two_view_geometry.cc` for the blocks repeated across the calibrated/spherical/uncalibrated/focal estimation paths:

- `ExtractMatchedImagePoints` / `ExtractMatchedCamRaysWithJac` (returned as moved pairs, bound with `const auto [...]`); `ExtractInlierImagePoints` converted to the same pair-return style
- `CheckMinMatchesOrDegenerate` / `CheckMinInlierRatioOrDegenerate` guards
- `RansacOptionsWithMinInlierRatio` override and `HomographyRansacOptions` search budget, inlined at the RANSAC construction sites
- `MaybeMarkWatermark` epilogue
- `FitModelFromInliers<Estimator>` for the E/F/H fallback fits in `MaybeFitMissingTwoViewGeometryMatrix`

Per-path model-selection logic is intentionally untouched (it differs in ratios, configs, and side effects per path). This PR is behavior-preserving: the uncalibrated path keeps ignoring `TwoViewGeometryOptions::min_inlier_ratio`, exactly as before. Making the ratio consistent across paths is split into a stacked follow-up PR.

## Test plan

- `ninja` full build clean (Release, Ninja)
- `ctest -R estimators` — 35/35 pass, incl. `two_view_geometry_test`
- Downstream geometry consumers (`feature_matching_test`, `pairing_test`, pipelines, sfm, feature, retrieval) — all pass
- `clang-format --dry-run --Werror` clean
- Note: `scene/reconstruction_test` (`AddFrame`/`AddImage`) fails identically on pristine main — pre-existing, unrelated (that binary does not link the estimators lib)

### DIFF
--- a/src/colmap/estimators/two_view_geometry.cc
+++ b/src/colmap/estimators/two_view_geometry.cc
@@ -192,6 +192,132 @@ FeatureMatches ExtractOutlierMatches(const FeatureMatches& matches,
   return outlier_matches;
 }
 
+std::pair<std::vector<Eigen::Vector2d>, std::vector<Eigen::Vector2d>>
+ExtractMatchedImagePoints(const std::vector<Eigen::Vector2d>& points1,
+                          const std::vector<Eigen::Vector2d>& points2,
+                          const FeatureMatches& matches) {
+  std::vector<Eigen::Vector2d> matched_points1(matches.size());
+  std::vector<Eigen::Vector2d> matched_points2(matches.size());
+  for (size_t i = 0; i < matches.size(); ++i) {
+    matched_points1[i] = points1[matches[i].point2D_idx1];
+    matched_points2[i] = points2[matches[i].point2D_idx2];
+  }
+  return std::make_pair(std::move(matched_points1), std::move(matched_points2));
+}
+
+std::pair<std::vector<CamRayWithJac>, std::vector<CamRayWithJac>>
+ExtractMatchedCamRaysWithJac(const Camera& camera1,
+                             const std::vector<Eigen::Vector2d>& points1,
+                             const Camera& camera2,
+                             const std::vector<Eigen::Vector2d>& points2,
+                             const FeatureMatches& matches) {
+  std::vector<CamRayWithJac> matched_rays1(matches.size());
+  std::vector<CamRayWithJac> matched_rays2(matches.size());
+  for (size_t i = 0; i < matches.size(); ++i) {
+    const point2D_t idx1 = matches[i].point2D_idx1;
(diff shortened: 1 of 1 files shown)


src/colmap/estimators/two_view_geometry.cc
```diff
@@ -1001,72 +1084,65 @@ TwoViewGeometry EstimateCalibratedTwoViewGeometry(
...
   // The tangent Sampson residual is in pixels, matching the fundamental matrix
   // and homography paths below, so all three share the same unscaled pixel
   // threshold. This also removes the former CamFromImgThreshold conversion,
   // whose single per-camera focal length is only exact at the principal point.
   LORANSAC<EssentialMatrixTangentSampsonEstimator,
            EssentialMatrixTangentSampsonEstimator,
            MEstimatorSupportMeasurer>
-      E_ransac(ransac_options);
+      E_ransac(RansacOptionsWithMinInlierRatio(options));
   const auto E_report =
```
B1ueber2y, line 1108: RansacOptionsWithMinInlierRatio(options) is called four times, and HomographyRansacOptions(...) is duplicated in both arms of the ternary. The old local ransac_options / H_ransac_options read better. Hoisting a const RANSACOptions H_ransac_options above the ternary fixes it.

src/colmap/estimators/two_view_geometry.cc
```diff
@@ -1184,56 +1255,44 @@ TwoViewGeometry EstimateSharedFocalTwoViewGeometry(
...
-  if (options.min_inlier_ratio > 0) {
-    ransac_options.min_inlier_ratio = options.min_inlier_ratio;
+    matched_centered_points1[i] = matched_img_points1[i] - principal_point;
+    matched_centered_points2[i] = matched_img_points2[i] - principal_point;
   }
 
   // Shared-focal relative pose. Residuals are pixel-space squared Sampson
-  // error, so the pixel threshold in `ransac_options` is used unscaled (unlike
-  // the calibrated essential-matrix path, which rescales it into ray space).
+  // error, so the pixel threshold is used unscaled, unlike the calibrated
```
B1ueber2y: Here the comment is actually stale (not from this PR). Can we directly fix here as well?

Specifically, it still says the calibrated E path "rescales it into ray space". The calibrated path no longer does that (see line 1101, "removes the former CamFromImgThreshold conversion"). The PR rewrote this comment, so it should fix the claim too.

B1ueber2y (approved):

PR #4697, opened 2026-09-10
Deduplicate UI painters with shared PainterBase

ahojnnes:
Point, line, triangle, and mesh painters shared near-identical OpenGL setup/upload/render boilerplate. This moves the common logic (shader setup, VAO/VBO management, vertex attribute binding, render bind/release) into a new `PainterBase` class. Public `Setup`/`Upload`/`Render` APIs are unchanged.

Stack 1/3 (no dependencies).

Test plan: `cmake --build build --target colmap_ui` passes; full `colmap` executable links.


lpanaf (approved):
UI changes. Did not check in details since as long as it renders, it should be good.

PR #4698, opened 2026-09-10
Deduplicate UI option-widget row builders

ahojnnes:
Shares label/row creation (`CreateRowLabel`), file/dir path rows (`AddOptionPath`), and row visibility (`SetRowVisible`) in `OptionsWidget`, and reuses them in the feature extraction/matching, dense reconstruction, and automatic reconstruction widgets via shared option blocks, matcher runners, and button/tab helpers. Qt ownership, signals, and layout are unchanged.

Stack 2/3, based on #4697 (mechanical stacking only; no code dependency).

Test plan: `cmake --build build --target colmap_ui` passes.


lpanaf (approved):
UI changes only, approved with trust.

PR #4704, opened 2026-09-11
Point AGENTS.md at optional AGENTS.override.md

ahojnnes:
Some agents don't respect AGENTS.overrid.md and thus loading AGENTS.md currently never see repo-local personal instructions. This one-liner tells them to also read AGENTS.override.md when present (kept out of git via .gitignore), with its instructions taking precedence over the shared guide. Useful for instructing the model about user-specific Python environments, etc.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -6,6 +6,7 @@ CMakeUserPresets.json
 .clangd
 compile_commands.json
 .claude/settings.local.json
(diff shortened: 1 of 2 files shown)


B1ueber2y (approved):

sarlinpe (approved):
Handy, thank you!

PR #4711, opened 2026-09-12
Keep at least one panorama rendering worker

trueoneplusone:
## Summary

When `os.cpu_count()` returns 1, panorama perspective rendering subtracts one reserved CPU and passes `max_workers=0` to `ThreadPoolExecutor`. This raises `ValueError: max_workers must be greater than 0` before any panorama is processed.

Keep at least one rendering worker. The existing unknown-CPU fallback, reservation of one CPU on multicore hosts, and 32-worker cap remain unchanged.

Add a regression test for reported CPU counts of 1, `None`, 4, and 64. It uses the real executor and verifies that every submitted panorama is processed, replacing only the image-processing work to avoid optional dependencies.

## Validation

- Before the fix: the single-CPU regression raises the reported exception; the three comparison cases pass.
- After the fix: all six tests in `python/pycolmap/panorama_test.py` pass.
- A separate smoke test exercised actual rendering with `cpu_count=1`, two synthetic 16x8 panoramas, Pillow/OpenCV, and the native PyCOLMAP bitmap writer. It produced two rendered images and two masks, all 4x4.
- Ruff check, Ruff format check, and `git diff --check` pass.

The Python tests used this checkout's source with the native PyCOLMAP 4.2.0 Windows wheel (Python 3.13.5). COLMAP's C++ code and the full reconstruction pipeline were not built or tested.

AI assistance: OpenAI Codex identified the edge case, implemented the fix, and ran the regression tests.

### DIFF
--- a/python/pycolmap/panorama.py
+++ b/python/pycolmap/panorama.py
@@ -482,7 +482,7 @@ def render_perspective_images(
     )
 
     num_panos = len(pano_image_names)
-    max_workers = min(32, (os.cpu_count() or 2) - 1)
+    max_workers = max(1, min(32, (os.cpu_count() or 2) - 1))
 
     pbar = None
     if show_progress:
--- a/python/pycolmap/panorama_test.py
+++ b/python/pycolmap/panorama_test.py
@@ -1,12 +1,16 @@
 from pathlib import Path
 
 import numpy as np
+import pytest
 
 import pycolmap
 
 from .panorama import (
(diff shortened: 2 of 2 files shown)


ahojnnes (approved):

ahojnnes:
Thank you! Removed the test, as I am not sure it adds a lot of value and covering all edge cases/branches is not possible anyway.

PR #4718, opened 2026-09-13
Update pybind11 to 3.1.0

ahojnnes:
Bump to the latest pybind11 release (3.1.0) and adopt two of its new features.

Version bump:
- `pyproject.toml`: build pin `3.0.4` -> `3.1.0`
- `python/CMakeLists.txt`: `find_package` minimum `3.0.2` -> `3.1.0`

Notable pybind11 3.1.0 changes vs 3.0.x: Python 3.8 and MSVC 2017 support dropped (no impact, pycolmap requires >=3.10), PEP 484 strict-mode numeric casts relaxed, plus `stl_bind` slice-deletion, `enum_` shutdown-crash, and `scoped_ostream_redirect` fixes.

Feature adoption:
- Custom `__str__` on `enum_` (pybind/pybind11#6078): `str()` on enums using `AddStringToEnumConstructor` now returns the plain member name (e.g. `str(Device.auto) == 'auto'`) and round-trips through the string constructor. `repr()` keeps the qualified `Type.MEMBER` form via its own function object (pybind11 chains overloads by mutating the shared record, so the old `__repr__ = __str__` alias would have picked up the custom `__str__` too).
- Explicit `py::mod_gil_used()` module declaration (new 3.1.0 spelling for the default): documents that `_core` requires the GIL; no behavior change.

Verified locally: full C++ rebuild + install, pycolmap incremental rebuild against pybind11 3.1.0 (incl. stub generation), and the full pytest suite (962 passed, incl. 2 new enum str/repr tests).


B1ueber2y (approved):
Thanks! Two minor things from AI pass:

1. CreateRepresentationFromAttributes (src/pycolmap/helpers.h:249) and CreateSummary (helpers.h:213) both use py::str(attribute), not py::repr. So repr(IncrementalMapperOptions()) now prints image_selection_method=MAX_VISIBLE_POINTS_NUM instead of ImageSelectionMethod.MAX_VISIBLE_POINTS_NUM — which contradicts the PR's stated goal of keeping repr qualified. Switching that to py::repr fixes it and lets the manual py::isinstance<py::str> quoting branch go away.

2.	Nit: the __repr__ lambda captures enm by value, keeping a py::object alive in a function record for the module lifetime. py::type::of(self).attr("__name__") avoids the capture entirely.

ahojnnes:
Thanks, both addressed in 4bb5262dd:

1. `CreateRepresentationFromAttributes` and `CreateSummary` now use `py::repr`, so dataclass reprs keep qualified enums (`image_selection_method=ImageSelectionMethod.MIN_UNCERTAINTY`); the manual string-quoting branch is gone. Also applied the same to `AddDefaultsToDocstrings` so docstring defaults stay qualified/quoted.
2. `__repr__` lambda now uses `py::type::of(self)` with no capture.

Added 5 regression tests in `dataclass_test.py`; full pytest suite passes (967 passed).

PR #4723, opened 2026-09-13
Extend solvers/utils with ray helpers and share pose packing

ahojnnes:
Extend solvers/utils with `CalibratedRays` (replacing two verbatim copies), plus `PoseParamsFromRigid3d` and `Rigid3dFromPoseParams` for the 7-parameter TinySolver packing used by the essential and focal refine drivers.

Part 4/5 of the solver-simplify stack (stacked on #4722).

Test Plan:
- estimators_solvers/utils_test (extended)
- estimators_cost_functions/sampson_error_test (incl. new round-trip test)
- estimators_solvers/essential_matrix_test
- estimators_solvers/relpose_one_sided_focal_test
- estimators_solvers/relpose_shared_focal_test

PR #4724, opened 2026-09-13
Replace hand-rolled Kabsch with Eigen::umeyama in EPnP and GR8P

ahojnnes:
Use Umeyama's method (as in SimilarityTransformEstimator) instead of the hand centroid+SVD alignment in EPNPEstimator::EstimateRT and the GR8P rotation initialization, preserving the inverted rotation direction the caller transposes back.

Part 5/6 of the solver-simplify stack (stacked on #4723).

Test Plan:
- estimators_solvers/absolute_pose_test
- estimators_solvers/generalized_relative_pose_test


ahojnnes:
Dropping this from the stack. The refactor is numerically exact (verified bit-identical outputs), but `Eigen::umeyama` heap-allocates two 3xN demeaned copies per call, whereas the hand-rolled Kabsch accumulated the 3x3 covariance in place. `EstimateRT` runs 3x per `EPNPEstimator::ComputePose`, and EPnP is the local-optimization estimator in `LORANSAC<P3PEstimator, EPNPEstimator>`, so this is a hot path during image registration.

Measured on `EPNPEstimator::Estimate` (Release, arm64, min of 5 runs x 1500 reps):

| n | before | after | delta |
|---|--------|-------|-------|
| 100 | 33.4 us | 35.8 us | +7% |
| 500 | 59.6 us | 63.7 us | +7% |
| 2000 | 150.6 us | 179.7 us | +19% |

Not worth the line savings. #4725 has been restacked directly onto #4723.

PR #4727, opened 2026-09-13
Share reprojection error epilogue between autodiff reprojection functors

ahojnnes:
Add `ComputeImgReprojError` and use it in the four autodiff reprojection functors, replacing four verbatim copies of the project / subtract-observation / seam-wrap / zero-on-failure epilogue. Adds an analytical behind-camera zeroing test.

Part 5/5 of the solver-simplify stack (stacked on #4723).

Test Plan:
- estimators_cost_functions/reprojection_error_test


src/colmap/estimators/cost_functions/reprojection_error_test.cc
```diff
@@ -47,6 +47,45 @@ TEST(ReprojErrorCostFunctor, Nominal) {
...
+
+  AnalyticalReprojErrorCostFunction<CameraModel> cost_function(kPoint2D);
+  EXPECT_TRUE(cost_function.Evaluate(parameters, residuals, jacobians));
+  EXPECT_EQ(residuals[0], 0);
+  EXPECT_EQ(residuals[1], 0);
+  for (double v : j_point) EXPECT_EQ(v, 0);
+  for (double v : j_pose) EXPECT_EQ(v, 0);
+  for (double v : j_params) EXPECT_EQ(v, 0);
+
+  for (double& v : j_point) v = 1;
```
B1ueber2y, line 76: Should we reset residuals to 1 here as well?

B1ueber2y (approved):

PR #4730, opened 2026-09-14
Expose selectable MVS estimators

ahojnnes:
Stack 5/5. Depends on #4732. Adds the generic CLI, GUI and automatic reconstruction selection, pycolmap bindings, namespaced fusion inputs, tests, and documentation. Defaults to MVSFORMER_PP in ONNX-enabled builds.


ahojnnes:
ETH3D e2e calibration showed that MVSFormer++ benefits from the upstream-style two-view fusion minimum. StereoFusion.min_num_pixels now defaults to -1 (automatic): MVSFormer++ depth-map suffixes resolve to 2, while PatchMatch and other inputs retain 5. Any explicit non-negative value overrides the automatic choice.

PR #4750, opened 2026-09-16
Reuse shared image feature cache

ahojnnes:
Stack 2/2. Depends on #4745. Introduces ImageFeatureCache in feature/utils, a per-image cache retaining the two most recently used entries, and uses it in the SIFT CPU matcher and the ONNX brute-force/LightGlue matchers, replacing hand-rolled per-image caching. Also drops the unused descriptor index prefetch in guided SIFT matching. No behavior change.

Test plan: `ctest -R 'feature/(utils_test|sift_test|onnx_utils_test|onnx_matchers_test|aliked_test|loma_test)'` passes.


src/colmap/feature/utils.h
```diff
@@ -46,4 +47,38 @@ std::vector<float> HWCToCHW(const uint8_t* data,
...
+// a third image evicts the least recently used entry, so matching consecutive
+// image pairs (A, B) then (B, C) only creates features for C. Returned
+// references stay valid until the entry is evicted. Entries with invalid
+// image ids are never reused. Images must expose an image_id member.
+template <typename T>
+class ImageFeatureCache {
+ public:
+  template <typename Image, typename CreateFn>
+  T& GetOrCreate(const Image& image, CreateFn&& create) {
+    for (int i = 0; i < 2; ++i) {
```
sarlinpe: Should the value 2 be defined by a constexpr to be reused below?

sarlinpe (approved):

src/colmap/feature/utils.h
ahojnnes: Thx, done.

Issue #4752, opened 2026-09-17
[Bug] 4.2.0: images.txt poses from a rig reconstruction do not match image.CamFromWorld() (stale per-image poses)


adisonshadow:
### Description

When reconstructing a multi-camera rig (frames with multiple sensors, constant per-sensor `sensor_from_rig` poses), the per-image poses written to the exported `images.txt` do **not** match `image.CamFromWorld()` of the in-memory reconstruction. `frames.txt` + `rigs.txt`, however, **are** self-consistent: composing `cam_from_world = cam_from_rig ∘ rig_from_world` from them reproduces `image.CamFromWorld()` exactly (0.0000° for all images and all sensors).

In other words, after a rig-constrained reconstruction the model in memory is perfectly rigid (per-timestamp baseline norms between all sensors are constant, as expected for a physical rig), but the exported TXT carries per-image poses that violate this rigidity (baseline norms varying by ~±50% over time; per-image rotation errors up to 63° relative to the composition of the exported `frames.txt`/`rigs.txt`).

This makes any downstream consumer of the standard TXT format (undistortion, training pipelines, further SfM tools) see non-rigid cameras even though the rig-constrained solution itself is rigid.

### Steps to reproduce

1. Build a rig dataset per doc/rigs.html (two-pass workflow, unknown sensor poses):
   - Pass 1: plain `colmap mapper` without rigs (per-camera trivial rigs), partial model is fine.
   - `colmap rig_configurator --database_path db --input_path <pass1 model> --rig_config_path rig.json` (JSON omits `cam_from_rig_*` so poses are averaged from the pass-1 model; 448 multi-sensor frames written).
   - Pass 2: `colmap mapper --Mapper.ba_refine_sensor_from_rig 0` (also reproduced with refinement enabled and with `bundle_adjuster --BundleAdjustment.refine_sensor_from_rig 1`).
2. Load the final model and compare, for every image, `image.CamFromWorld()` against the exported `images.txt`:

```python
import pycolmap, numpy as np

rec = pycolmap.Reconstruction("sparse/0")   # final rig model (binary)
Rig = rec.rigs[next(iter(rec.rigs))]

def R_of(r3d):
    return np.asarray(r3d.rotation.matrix())

def center(r3d):
    R = R_of(r3d); t = np.asarray(r3d.translation)
    return -R.T @ t

max_ang, n_mismatch, n = 0.0, 0, 0
for img_id, img in rec.images.items():
    cfw = img.cam_from_world()
[...]

PR #4753, opened 2026-09-17
Prefer reference sensor poses when converting to rig frames

ahojnnes:
Related to #4752.

When `rig_configurator` converts an unconstrained reconstruction to rig frames (`UpdateRigsAndFramesFromDatabase`), each frame pose was set from every registered image in arbitrary iteration order (last wins). For frames with inconsistent per-image poses, as produced by the unconstrained first pass of the two-pass rig workflow, a non-reference image could overwrite the reference-derived pose — even though non-reference poses depend on the averaged `sensor_from_rig` calibration and are less reliable.

This change prefers reference-sensor poses and only falls back to the first available non-reference pose when no reference pose exists.

Test plan:
- New `ApplyRigConfig.PrefersRefSensorForFramePoses` test fails before the fix (converted frame poses diverge from reference images) and passes after.
- `ctest -R "scene/(rig_test|reconstruction_io_test|reconstruction_test)" --output-on-failure`: 3/3 pass.
- Synthetic end-to-end checks (direct rig mapper, two-pass mapper, converter round-trip) show exported `images.txt` consistent with `image.CamFromWorld()`.


src/colmap/scene/rig.cc
```diff
@@ -200,13 +200,21 @@ void UpdateRigsAndFramesFromDatabase(const Database& database,
...
     reconstruction_frame.AddDataId(reconstruction_image.DataId());
     if (reconstruction_image.HasPose()) {
-      if (database_rig.IsRefSensor(database_sensor_id)) {
-        reconstruction_frame.SetRigFromWorld(
-            reconstruction_image.CamFromWorld());
-      } else {
-        reconstruction_frame.SetRigFromWorld(
-            Inverse(database_rig.SensorFromRig(database_sensor_id)) *
-            reconstruction_image.CamFromWorld());
+      const bool is_ref = database_rig.IsRefSensor(database_sensor_id);
```
B1ueber2y: nit: a small improvement with equivalent logic:
```
if (database_rig.IsRefSensor(database_sensor_id)) {
  reconstruction_frame.SetRigFromWorld(reconstruction_image.CamFromWorld());
} else if (!reconstruction_frame.HasPose()) {
  reconstruction_frame.SetRigFromWorld(
      Inverse(database_rig.SensorFromRig(database_sensor_id)) *
      reconstruction_image.CamFromWorld());
}
```

B1ueber2y (approved):

PR #4756, opened 2026-09-20
Expand tiny solver test coverage

ahojnnes:
Extends `optim/tiny_solver_test` to cover the production configurations and failure handling of `colmap::TinySolver`:

- Dynamically-sized residual path and solver instance reuse.
- All termination statuses (`GRADIENT_TOO_SMALL`, `COST_TOO_SMALL`, `RELATIVE_STEP_SIZE_TOO_SMALL`, `HIT_MAX_ITERATIONS`, `COST_CHANGE_TOO_SMALL`).
- Cost-function failure at the initial point, at trial points, and on post-accept re-evaluation; linear-solver factorization failure via an injected stub.

Test plan: `ctest -R optim/tiny_solver_test` passes (14 tests).


src/colmap/optim/tiny_solver_test.cc
```diff
@@ -115,5 +118,262 @@ TEST(TinySolver, ReportsCostFunctionFailure) {
...
+      residuals[i] = parameters[0] - values[i];
+      if (jacobian != nullptr) {
+        jacobian[i] = 1.0;
+      }
+    }
+    return true;
+  }
+};
+
+TEST(TinySolver, DynamicResidualsConvergeAndResizeOnReuse) {
```
B1ueber2y, line 144: There is no "ResizeOnReuse" behavior in this test?

src/colmap/optim/tiny_solver_test.cc
```diff
@@ -4,6 +4,9 @@
 
 #include "colmap/estimators/cost_functions/tiny_manifold.h"
 
+#include <stdexcept>
```
B1ueber2y: Unused?

B1ueber2y (approved):

src/colmap/optim/tiny_solver_test.cc
ahojnnes: Done — the test now reuses the same solver instance with 3, then 5, then 2 residuals, covering buffer growth and shrinkage.

src/colmap/optim/tiny_solver_test.cc
ahojnnes: Done — removed the unused include.

PR #4757, opened 2026-09-20
Handle non-finite values in tiny solver

ahojnnes:
Stacked on #4756.

Makes `colmap::TinySolver` robust against cost functions that report success with garbage output, instead of spinning all iterations on NaNs:

- New `NUMERICAL_FAILURE` status for non-finite costs, gradients, steps, and initial parameters.
- Non-finite retraction outputs are rejected like failed trial evaluations, shrinking the trust region.
- The final cost is refreshed from the accepted point's trial residuals (including when its Jacobian re-evaluation fails), so `final_cost` always describes the returned parameters. The cost-change termination check still runs before the re-evaluation, so converging solves skip the extra Jacobian evaluation.
- The estimator drivers return false on `NUMERICAL_FAILURE` instead of post-solve `allFinite` checks; the log-space pre-checks now also reject non-finite focals/scales.

Test plan: `optim/tiny_solver_test` (21 tests) plus the dependent `tiny_manifold`, `tiny_sampson_error`, `essential_matrix`, `fundamental_matrix`, `generalized_absolute_pose`, `relpose_one_sided_focal`, and `relpose_shared_focal` suites all pass.


src/colmap/optim/tiny_solver.h
```diff
@@ -302,19 +336,28 @@ class TinySolver {
...
         x = x_new_;
 
-        if (std::abs(cost_change) < options.function_tolerance) {
+        // Refresh the linearization at the accepted point before checking
+        // termination, so the summary always describes the returned x. The
+        // cost function already evaluated successfully at x_new_ == x above,
+        // so re-evaluating (now also for the Jacobian) is not expected to
+        // fail; guard against it regardless, reporting the cost of the
+        // accepted point, whose residuals are known and finite.
+        if (!Update(function, x)) {
```
B1ueber2y, line 349: Hoisting Update() above the COST_CHANGE_TOO_SMALL check adds a full autodiff Jacobian eval to the normal convergence path, inside RANSAC. The old code already did cost_ = f_x_new_.squaredNorm() / 2; immediately before the break, so final_cost already described the returned x. The reorder buys a fresher gradient_max_norm in the summary and nothing else.

src/colmap/estimators/solvers/relpose_shared_focal.cc
```diff
@@ -195,15 +195,9 @@ bool RelativePoseSharedFocalEstimator::Refine(const std::vector<X_t>& points1,
   x[7] = std::log(model->focal);
   solver.Solve(f, &x, options);
```
B1ueber2y: I am confused here. I assume something like the following was missed?

if (solver.Solve(f, &x, options).status == NUMERICAL_FAILURE) return false;

src/colmap/optim/tiny_solver.h
ahojnnes: Done — moved the COST_CHANGE_TOO_SMALL check back above Update(), restoring the old order (cost refreshed directly from the trial residuals). The cost refresh on Update failure / non-finite Jacobian is kept.

src/colmap/estimators/solvers/relpose_shared_focal.cc
ahojnnes: Done — added the NUMERICAL_FAILURE check as suggested, applied consistently to all five Refine drivers in this PR (essential, fundamental, GP3P/GP4P, one-sided and shared focal).

B1ueber2y (approved):
Thank you!

PR #4762, opened 2026-09-20
Fix pycolmap link failure against COLMAP built with pinned Boost

### LINKED ISSUE #4761: pycolmap fails to link against an installed COLMAP since the pinned Boost
**Describe the bug**
Since #4677, building pycolmap from source with `python -m pip install .` against an installed COLMAP fails at the final link step on Ubuntu 24.04.

```
/usr/bin/ld: /usr/local/lib/colmap/thirdparty/boost/libboost_thread.a(thread.cpp.o):(.data.rel.ro._ZTVN5boost12bad_weak_ptrE[_ZTVN5boost12bad_weak_ptrE]+0x20): undefined reference to `boost::bad_weak_ptr::what() const'
  /usr/bin/ld: /usr/local/lib/colmap/thirdparty/boost/libboost_thread.a(thread.cpp.o):(.data.rel.ro._ZTVN5boost10wrapexceptINS_12bad_weak_ptrEEE[_ZTVN5boost10wrapexceptINS_12bad_weak_ptrEEE]+0x50): undefined reference to `boost::bad_weak_ptr::what() const'
  collect2: error: ld returned 1 exit status
  ninja: build stopped: subcommand failed.

  *** CMake build failed
  error: subprocess-exited-with-error
  
  × Building wheel for pycolmap (pyproject.toml) did not run successfully.
  │ exit code: 1
  ╰─> No available output.
 ```

**To Reproduce**
  ```bash
  git checkout b3f4c100   # or current main
  cmake -S . -B build -GNinja -DCMAKE_BUILD_TYPE=Release -DCMAKE_CUDA_ARCHITECTURES=native
  ninja -C build
  sudo ninja -C build install
  python -m pip install -v .
  ```

**Environment:**
 - OS: Ubuntu 24.04.5, x86_64

nushakrishnan:
Fixes #4761.

pybind11 enables LTO for the `_core` module by default, but the Boost that COLMAP builds from source (when the system Boost is older than 1.84, e.g. Ubuntu 24.04) is installed as non-LTO static archives, and with GCC that mix fails at link time with `undefined reference to boost::bad_weak_ptr::what() const` from `libboost_thread.a`. This sets `CMAKE_INTERPROCEDURAL_OPTIMIZATION` to `OFF` for pycolmap only when compiling with GCC against a COLMAP that uses the pinned Boost, which matches the root build.

Tested on Ubuntu 24.04 with `python -m pip install .` against an installed COLMAP.


ahojnnes (approved):
Thanks, pushed directly to your branch with a small improvement.

PR #4763, opened 2026-09-20
Make global positioning scale and weighting policies configurable

Zador-Pataki:
## Motivation

Global positioning hard-codes scale-gauge fixing and uncalibrated-camera weighting, and couples scale initialization to point randomization. These choices restrict downstream research, for example when extending global positioning with additional residuals.

## Changes

- Remove the dependency of scale initialization on point randomization and optimization, and rename and invert `generate_scales` to `initialize_scales_from_geometry`.
- Replace the hard-coded uncalibrated-observation loss multiplier with `uncalibrated_observation_weight` (default `0.5`). `1.0` gives calibrated and uncalibrated observations equal weight.
- Allow callers to disable the previously unconditional fixing of the first observation scale with `fix_observation_scale_gauge` (default `true`).


src/colmap/estimators/global_positioning.cc
B1ueber2y: Stale comment?

src/colmap/estimators/global_positioning.cc
```diff
@@ -39,6 +40,8 @@ class DefaultGlobalPositioner final : public GlobalPositioner {
 
 GlobalPositioner::GlobalPositioner(const GlobalPositionerOptions& options)
     : options_(options) {
+  THROW_CHECK(std::isfinite(options_.uncalibrated_observation_weight));
+  THROW_CHECK_GE(options_.uncalibrated_observation_weight, 0.0);
```
B1ueber2y: Should be THROW_CHECK_GT here?

src/colmap/estimators/global_positioning.h
```diff
@@ -19,14 +19,17 @@ struct GlobalPositionerOptions {
...
-  bool generate_scales = true;
+  // Derive scales from camera and 3D-point positions; otherwise use 1.
+  bool initialize_scales_from_geometry = false;
 
   // Flags for which parameters to optimize
   bool optimize_positions = true;
   bool optimize_points = true;
   bool optimize_scales = true;
+  // Fix the first active observation scale when optimizing scales.
+  bool fix_observation_scale_gauge = true;
```
B1ueber2y: Just fix_first_scale?

src/colmap/estimators/global_positioning_test.cc
```diff
@@ -173,6 +173,86 @@ TEST(GlobalPositioning, ComposableProblem) {
...
+      problem.GetParameterBlocks(&blocks);
+      int num_scales = 0;
+      int num_constant = 0;
+      for (double* block : blocks) {
+        if (problem.ParameterBlockSize(block) != 1) continue;
+        ++num_scales;
+        num_constant += problem.IsParameterBlockConstant(block);
+      }
+      ASSERT_GT(num_scales, 1);
+      EXPECT_EQ(num_constant, optimize ? int(fix_gauge) : num_scales);
```
B1ueber2y: nit: int(fix_gauge) -> static_cast<int>()?

src/colmap/estimators/global_positioning_test.cc
```diff
@@ -173,6 +173,86 @@ TEST(GlobalPositioning, ComposableProblem) {
...
+      options.generate_random_points = false;
+      options.optimize_points = optimize_points;
+      options.initialize_scales_from_geometry = from_geometry;
+      auto positioner =
+          GlobalPositioner::CreateDefault(options, PoseGraph(), reconstruction);
+      std::vector<ceres::ResidualBlockId> residuals;
+      positioner->Problem().GetResidualBlocks(&residuals);
+      ASSERT_FALSE(residuals.empty());
+      std::vector<double*> blocks;
+      positioner->Problem().GetParameterBlocksForResidualBlock(
```
B1ueber2y, line 246: nit: ASSERT_EQ(blocks.size(), 3)?

src/colmap/estimators/global_positioning.cc
```diff
@@ -211,7 +216,7 @@ void GlobalPositioner::AddPoint3DToProblem(point3D_t point3D_id,
 
     double& scale = scales_.emplace_back(1);
 
-    if (!options_.generate_scales && random_initialization) {
+    if (options_.initialize_scales_from_geometry) {
```
B1ueber2y, line 219: Here there seems to be a minor behavior change, with && random_initialization being dropped?

Zador-Pataki: Yes, intentional. This removes the dependency on point randomization, allowing scales to be initialized from supplied camera positions and 3D points when resuming optimization. I separately realized I had also removed the requirement that point optimization be enabled. I added this back to keep the change focused on removing the randomization dependency.

Zador-Pataki:
@B1ueber2y thanks! I addressed everything and clarified the remaining open question.

src/colmap/estimators/global_positioning.h
```diff
@@ -19,14 +19,17 @@ struct GlobalPositionerOptions {
...
+  bool initialize_scales_from_geometry = false;
 
   // Flags for which parameters to optimize
   bool optimize_positions = true;
   bool optimize_points = true;
   bool optimize_scales = true;
+  // Fix the first active observation scale when optimizing scales.
+  bool fix_first_scale = true;
+  // Loss multiplier for observations without a focal-length prior.
+  double uncalibrated_observation_weight = 0.5;
```
B1ueber2y: nit: should this sit aside loss_function_scale rather than here in the header?

src/colmap/estimators/global_positioning.cc
B1ueber2y: Thanks for the clarification. Thinking about it, optimize_points may not be needed here either? Lets drop and keep it what initially had.

B1ueber2y (approved):
LGTM. Thanks!

sarlinpe (approved):

PR #4764, opened 2026-09-20
Replace EPNP with TinySolver refinement for absolute pose estimation

ahojnnes:
## Summary

Replace the EPNP non-minimal solver used for local optimization in absolute
pose LO-RANSAC with iterative Levenberg-Marquardt refinement
(`colmap::TinySolver` on the quaternion + translation manifold, autodiff
Jacobian, normalized-plane ray residuals), following the GP4PS self-LO
pattern (`LORANSAC<P3PEstimator, P3PEstimator>`). `EPNPEstimator` is removed.

Additionally, P3P/P4PF models now store `Rigid3d` instead of raw 3x4 matrices,
avoiding conversions at both ends (P3P builds directly from
`poselib::CameraPose` via `ConvertPoseLibPoseToRigid3d`).

Notes for reviewers:

- The refinement is camera-model agnostic (observations enter as rays). For
  pinhole cameras it shares the pixel-error minimizer exactly (verified: the
  in-tree implementation reproduces prototype numbers bit-for-bit).
- LO now triggers at >= 3 inliers (`kMinNumSamples`) instead of 4 with EPNP.
- An analytic-Jacobian variant was prototyped: ~2x faster per `Refine` call
  but no end-to-end difference (LO is negligible vs sampling/scoring), so the
  simpler autodiff version is shipped.

## Benchmark: EPNP vs TinySolver LO (synthetic, paired seeds)

Setup: 200 random scenes per cell, 100 points each, focal uniform in 400-2000,
paired RANSAC seeds, production RANSAC options (4px inlier threshold, 12px for
the sigma=4 cells; min 10 trials; confidence 0.99999; 10k trial cap). The EPNP
column uses the pre-PR `EPNPEstimator` extracted from git history as the LO
estimator; the TinySolver column uses the shipped `P3PEstimator::Refine`
self-LO. Same P3P minimal stage, thresholds, and scoring — only the LO step
differs. Recall requires rotation error < 1 deg and relative translation
error < 5%. Pairs below are EPNP / TinySolver.

| noise | outliers | success [%] | recall [%] | ms/run | med rot err [deg] | med trans err [%] |
|---|---|---|---|---|---|---|
| 0px | 0% | 100 / 100 | 100 / 100 | 0.1 / 0.0 | 0.0000 / 0.0000 | 0.000 / 0.000 |
| 0px | 30% | 100 / 100 | 100 / 100 | 0.2 / 0.1 | 0.0000 / 0.0000 | 0.000 / 0.000 |
| 0px | 50% | 100 / 100 | 100 / 100 | 0.4 / 0.3 | 0.0000 / 0.0000 | 0.000 / 0.000 |
| 1px | 0% | 100 / 100 | 100 / 100 | 0.1 / 0.1 | 0.0201 / 0.0107 | 0.084 / 0.031 |
| 1px | 30% | 100 / 100 | 100 / 100 | 0.2 / 0.1 | 0.0226 / 0.0122 | 0.092 / 0.041 |
| 1px | 50% | 100 / 100 | 100 / 100 | 0.4 / 0.3 | 0.0268 / 0.0148 | 0.108 / 0.048 |
| 4px | 0% | 100 / 100 | 100 / 100 | 0.2 / 0.1 | 0.0819 / 0.0438 | 0.350 / 0.133 |
| 4px | 30% | 100 / 100 | 100 / 100 | 0.2 / 0.1 | 0.0946 / 0.0521 | 0.366 / 0.180 |
| 4px | 50% | 100 / 100 | 100 / 100 | 0.4 / 0.3 | 0.1045 / 0.0632 | 0.399 / 0.219 |

Takeaways:

- Roughly 2x lower median rotation/translation errors in every noisy cell.
- Success/recall tied at 100% (dominated by the P3P minimal stage).
- Runtime tied or slightly better; the solver converges in 0-3 iterations at
  3-12us per `Refine` call.

## Test plan

- `absolute_pose_test`: 6/6 pass (new `P3PRefine` test: LO-RANSAC recovery,
  perturbed-init convergence, underdetermined-model guard).
- `pose_test`: 11/11 pass (end-to-end `EstimateAbsolutePose` on the new path).
- `scripts/format/c++.sh` clean.


src/colmap/estimators/solvers/absolute_pose.cc
```diff
@@ -2,14 +2,82 @@
...
 #include <PoseLib/solvers/p4pf.h>
+#include <ceres/tiny_solver_autodiff_function.h>
 
 namespace colmap {
+namespace {
+
+// The manifold of a cam_from_world transform: rotation on SO(3), with the
+// translation as Euclidean parameters. The ambient parameter layout matches
+// TinyPnPCostFunctor: [qx, qy, qz, qw, tx, ty, tz].
+using CamFromWorldManifold =
```
B1ueber2y: nit: How about Rigid3dManifold?

B1ueber2y (approved):
Thank you! Interesting to see EPNP stepping off the stage.

src/colmap/estimators/solvers/absolute_pose.cc
ahojnnes: Done — renamed to Rigid3dManifold (also updated its use in the stacked P4PF PR).

PR #4765, opened 2026-09-20
Add observation covariance weighting to global positioning

Zador-Pataki:
## Motivation

Allow global positioning to account for per-observation uncertainty through covariance weighting, supporting downstream research that models different uncertainty across observations and directions.

## Changes

- Introduce `ObservationKey = (image_id, point2D_idx)` to identify observations independently of track ordering.
- Accept optional per-observation covariance matrices in C++ and as a Python dictionary. Matrices must be finite, symmetric, positive-definite 3×3 covariances in world coordinates; a supplied map must cover every observation included in the problem.
- Apply covariance weighting through the existing `CovarianceWeightedCostFunctor` for standard and rig BATA residuals.


src/colmap/scene/track.h
```diff
@@ -22,6 +23,9 @@ struct TrackElement {
   inline bool operator!=(const TrackElement& other) const;
 };
 
+// Identifies an image measurement, independently of its track membership.
+using ObservationKey = std::pair<image_t, point2D_t>;
```
B1ueber2y: This seems to be identical as TrackElement in the same file?

B1ueber2y (commented):
I am actually concerned about this PR. 

From my understanding, the covariance you input in vidmap is rank deficient, so this code probably wont work. Maybe we should add an interface in colmap to take left sqrt or the information matrix in the wrapper.

src/colmap/scene/track.h
sarlinpe: TrackElement cannot currently be used as a map or dict key in either C++ or Python, since `std::hash<TrackElement>` isn't defined - consider reusing PairHash:
```c++
template <>
struct hash<colmap::TrackElement> {
  std::size_t operator()(const colmap::TrackElement& t) const noexcept {
    return colmap::PairHash{}(std::make_pair(t.image_id, t.point2D_idx));
  }
};
```

sarlinpe:
To handle a 2D image covariance $\Sigma_{2\text{D}} \in \mathbb{R}^{2\times 2}$ correctly in global positioning:
1. **The caller should only pass the natural $2\times 2$ pixel covariance $\Sigma_{2\text{D}}$** (which is always symmetric and positive-definite).
2. The BATA cost function decomposes the 3D residual into:
   - A **2D transverse bearing error** in the tangent plane $T_{\mathbf{b}}S^2$, whitened by the exact propagated $2\times 2$ covariance $\Sigma_{\text{tangent}}^{-1}$.
   - A **1D longitudinal scale constraint** along the ray $\mathbf{v}$, weighted to anchor $s \approx 1/\text{depth}$ and prevent scale collapse.

#### Step 1: Propagate 2D Pixel Covariance to the Bearing Tangent Space
Let $\mathbf{x} \in \mathbb{R}^2$ be the pixel coordinate and $\Sigma_{\text{pix}} \in \mathbb{R}^{2\times 2}$ its covariance.
1. Unproject to normalized image coordinates 

$$\mathbf{p} = \pi^{-1}(\mathbf{x}) = \begin{pmatrix} x \\ y \\ 1 \end{pmatrix}$$

   $$\Sigma_{\text{norm}} = J_{\pi^{-1}} \Sigma_{\text{pix}} J_{\pi^{-1}}^\top \in \mathbb{R}^{2\times 2}$$

   where for a pinhole camera, $J_{\pi^{-1}} = \mathrm{diag}(1/f_x, 1/f_y)$.

3. Form the unit bearing $\mathbf{b} = \mathbf{p} / \Vert{}\mathbf{p}\Vert{} \in S^2 \subset \mathbb{R}^3$. Choose an orthonormal basis $E = [\mathbf{e}_1, \mathbf{e}_2] \in \mathbb{R}^{3\times 2}$ for the 2D tangent plane orthogonal to $\mathbf{b}$ ($E^\top E = I_2$, $E^\top \mathbf{b} = \mathbf{0}$). For example, via a Householder reflection of $\mathbf{b}$:

$$\mathbf{u} = \frac{\mathbf{b} + \mathrm{sign}(b_z)\mathbf{e}_3}{\Vert{}\mathbf{b} + \mathrm{sign}(b_z)\mathbf{e}_3\Vert{}}, \quad H = I_3 - 2\mathbf{u}\mathbf{u}^\top, \quad E = H_{[:, 1:2]}$$

4. The projection from normalized coordinates to the tangent space has Jacobian:

$$J_{\text{tangent}} = \frac{1}{\Vert{}\mathbf{p}\Vert{}} E^\top \begin{bmatrix} I_2 \\ \mathbf{0}^\top \end{bmatrix} \in \mathbb{R}^{2\times 2}$$

   The $2\times 2$ tangent covariance is:

   $$\Sigma_{\text{tangent}} = J_{\text{tangent}} \Sigma_{\text{norm}} J_{\text{tangent}}^\top \in \mathbb{R}^{2\times 2}$$

   Since $\Sigma_{\text{pix}}$ is positive definite and $J_{\text{tangent}}$ is full rank, $\Sigma_{\text{tangent}}$ is strictly symmetric and positive definite.

#### Step 2: Form the Full $3\times 3$ Whitening Operator
In world coordinates, the unit ray is $\mathbf{v} = R^\top \mathbf{b}$, and the tangent basis is $E_{\text{world}} = R^\top E \in \mathbb{R}^{3\times 2}$.

Let $L_{\text{tangent}} \in \mathbb{R}^{2\times 2}$ be the Cholesky factor of the inverse tangent covariance:

$$L_{\text{tangent}} L_{\text{tangent}}^\top = \Sigma_{\text{tangent}}^{-1}$$

Along the ray $\mathbf{v}$, we retain BATA's scale-defining constraint $1 - s \mathbf{v}^\top(X - c)$ with a scalar weight $w_{\parallel}$. To match the average scale of the transverse whitening, choose:

$$w_{\parallel} = \sqrt{\frac{1}{2} \mathrm{tr}(\Sigma_{\text{tangent}}^{-1})}$$

(Or $w_{\parallel} = 1$ if using normalized dimensionless units).

The complete $3\times 3$ whitening operator $W \in \mathbb{R}^{3\times 3}$ is:

$$W = \begin{bmatrix} L_{\text{tangent}}^\top E_{\text{world}}^\top \\ w_{\parallel} \mathbf{v}^\top \end{bmatrix} \in \mathbb{R}^{3\times 3}$$

Applying $W$ to the 3D BATA residual $\mathbf{e} = \mathbf{v} - s(X - c)$ gives:

$$\mathbf{r}_{\text{weighted}} = W \mathbf{e} = \begin{bmatrix} -s L_{\text{tangent}}^\top E_{\text{world}}^\top (X - c) \\ w_{\parallel} (1 - s \mathbf{v}^\top (X - c)) \end{bmatrix} \in \mathbb{R}^3$$

- **Residuals 0 and 1:** Exactly measure the 2D directional angular error, correctly whitened by the 2D measurement covariance $\Sigma_{\text{tangent}}$.
- **Residual 2:** Enforces the BATA algebraic scale constraint $s \approx 1/\text{depth}$, preventing scale collapse.
- The cost function produces 3 residuals, preserving full compatibility with Ceres' Schur complement solver (`SPARSE_SCHUR`) and variable ordering.

ahojnnes:
Thanks for the detailed write-up — the transverse + scale decomposition makes sense to me.

One concern: as written, the propagation goes through the `z=1` plane (`J_π⁻¹`, then `J_tangent = (1/‖p‖) Eᵀ[I;0]`). That assumes a finite plane point `p = b/b_z`, i.e. the forward hemisphere only. For fisheye rays at ≥90° incidence or spherical/equirectangular back-facing rays, `‖p‖→∞` at 90° and is undefined beyond, so `Σ_tangent` collapses and the whitening blows up.

Would it be better to skip the plane and differentiate the ray directly per model — `J = d(CamRayFromImg)/dx ∈ ℝ³ˣ²`, `J_t = EᵀJ`, `Σ_tangent = J_t Σ_pix J_tᵀ` (i.e. on top of `CameraModelCamRayFromImg`, not `CameraModelCamFromImg`)? Step 2 then stays unchanged since BATA already operates on unit rays.

Zador-Pataki:
Thanks both @sarlinpe @ahojnnes, implemented! could you take another look? To keep the interface simple for now, this uses a single isotropic pixel stddev. I called it experimental_observation_stddev so it’s clear that we can subsequently deprecate it and upgrade to full per-observation covariances.

I moved the uncertainty input from the default global positioner’s arguments to options.experimental_observation_stddev. If unset, global positioning runs with its existing default behavior.

One concern remains: when a stddev is set, the residuals are whitened, so the default Huber scale of 0.1 changes meaning and is far too small. How should we handle this?

src/colmap/estimators/global_positioning.cc
```diff
@@ -21,6 +24,32 @@ Eigen::Vector3d RandVector3d(double low, double high) {
                          RandomUniformReal(low, high));
 }
 
+Eigen::Matrix3d ComputeObservationCovariance(
```
sarlinpe, line 27: Consider explaining in a comment why the code sets the radial precision equal to the mean tangent precision (imbalance of scale of residuals, etc.).

src/colmap/estimators/global_positioning.cc
```diff
@@ -21,6 +24,32 @@ Eigen::Vector3d RandVector3d(double low, double high) {
...
+    const CamRayWithJac& ray,
+    const Eigen::Quaterniond& cam_from_world,
+    const double pixel_stddev) {
+  const Eigen::Matrix2d gram = ray.jacobian.transpose() * ray.jacobian;
+  // The Gram matrix has the two tangent covariance eigenvalues. Match their
+  // mean precision along the bearing to retain BATA's scale constraint.
+  const double radial_variance = 2.0 / gram.inverse().trace();
```
sarlinpe, line 34: For a point that lies near the valid image boundary of a distorted fisheye lens where the projection derivative vanishes or becomes ill-conditioned, radial_variance vanishes to zero. Should we guard against this?

src/colmap/estimators/global_positioning.h
```diff
@@ -46,6 +47,10 @@ struct GlobalPositionerOptions {
   // Scaling factor for the loss function
   double loss_function_scale = 0.1;
 
+  // Isotropic observation uncertainty in pixels. Disabled when unset.
+  // When enabled, loss_function_scale applies to whitened residuals.
```
sarlinpe, line 51: Should we then set loss_function_scale to 1 when experimental_observation_stddev is provided? or what would be an appropriate value? Without whitening, we have a scale of 0.1 rad.

Zador-Pataki: I set it to 1 for now, but i'll update it once we agree. Assuming a 2 px keypoint stddev and a 700 px focal length, the previous 0.1 rad scale corresponds to a whitened loss scale of approximately 35. In practice, I've also found that global positioning needs a larger angular threshold before robustification kicks in. 

Since camera intrinsics are available during global positioning, would it make sense to let users specify the robust scale in radians and convert it internally? This would remove the need for the conditional default and would keep the existing default of 0.1 rad

I implemented a conditional default: 0.1 without observation uncertainty and 1 with it. Explicit values override this. The conditional default feels a bit hacky. Not sure what the best design is. Thoughts?

Zador-Pataki:
@sarlinpe thanks! Mostly resolved with one open question remaining.

src/colmap/estimators/global_positioning.h
sarlinpe: When whitening with covariance, the scale is not in radians, it's in units of stddev. Would it make sense to run a quick benchmarking on ETH3D to find out the optimal scale value when observation uncertainty is enabled?

PR #4769, opened 2026-09-21
Fix TinySolver iteration count

B1ueber2y:
Found while reviewing https://github.com/colmap/colmap/pull/4756


B1ueber2y:
Seems to be a minor bug inherited from the upstream ceres-solver.

sarlinpe (approved):

